PR #44, opened 2019-10-10
If passing an env var value that can be parsed with yaml, pass it through parsed

swist:
Really useful project!

This helps keep parity when you merge multiple configs, one that defines
values as bools in YAML and another one that interpolates these boolean
values from the environment

I have noticed that you have added other `NodeTypes` but haven't used them, so please tell me if this is not how you planned this to happen and I'll update accordingly.

### DIFF
--- a/omegaconf/omegaconf.py
+++ b/omegaconf/omegaconf.py
@@ -14,7 +14,7 @@ from .config import Config
 def register_default_resolvers():
     def env(key):
         try:
-            return os.environ[key]
+            return yaml.safe_load(os.environ[key])
         except KeyError:
             raise KeyError("Environment variable '{}' not found".format(key))
 
--- a/tests/test_interpolation.py
+++ b/tests/test_interpolation.py
@@ -156,6 +156,34 @@ def test_env_interpolation_not_found():
         c.path
 
 
+@pytest.mark.parametrize("value,expected", [
+   ("false", False),
+   ("off", False),
+   ("no", False),
+   ("true", True),
+   ("on", True),
+   ("yes", True),
+   ("10", 10),
+   ("-10", -10),
+   ("10.0", 10.0),
+   ("-10.0", -10.0),
+   ("foo: bar", {'foo': 'bar'}),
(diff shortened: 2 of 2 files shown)


omry:
Hi, thanks for the bug report and the PR.
I am guessing you became aware of OmegaConf through [Hydra](https://cli.dev/), if not do take a look at it as well.

The correct place to fix this is probably to change the built in env resolver to pass use the yaml parser to convert the value to the appropriate type.
same thing happens when handling dotlists (or cli arguments).

Also, can you fix the test to test other types?
you can parameterize it on:
```
@pytest.mark.parametrize("value,expected",[
   ("false", False),
   ("off", False),
   ("10", 10),
   ("10.0", 10.0),
]
```

or something along those lines.

Re: node types:
Those are not used yet except through the API. but will at some point. I have plans to introduce annotation of config values in the config file via some DSL.
```
#@type: str
#@choice: [10, foo, bar]
foo: 10 # will be string anyway, and must be 10,foo or bar.
```
This requires a new yaml parser and my focus right now is on Hydra so its' on hold.
if you want to help with the yaml parser take a look at [Antelope](https://github.com/omry/antelope).

swist:
Updated the PR!

swist:
I'm hoping not many people rely on this doing the `stringly-typed` bools though. In our org's code i've found a couple of comparisons to `x == 'true'` which I'm hoping to remove with this pr

omry:
Awesome, this is much better.
I think breaking backward compatibility here is acceptable.

Curious, what is your org and how did you learn about OmegaConf?

swist:
I work at OakNorth. I've been looking for a config library for the project I'm working on (I maintained something similar to this in my previous job) and came across hydra when idly browsing facebook

omry:
Great, thanks for the info.
Are you using Hydra at OakNorth or just OmegaConf?

swist:
Just omegaconf for now

omry:
Cool.
Do consider Hydra, It is a bigger change but you get much more in return.
If you are already using OmegaConf it should be fairly straight forward to try.

PR #46, opened 2019-10-23
Support passing variable number of arguments to resolvers.

swist:
This PR adds the ability to pass in 0 or more arguments to resolvers:
```
${foo:}               # zero arguments
${foo:a}              # one argument (currently supported)
${foo:a,b,..,n}       # n arguments
```

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -255,6 +255,26 @@ This example creates a resolver that adds 10 the the given value.
     1000
 
 
+Custom resolvers support variadic argument lists in the form of a comma separated list of zero or more values (coming in OmegaConf 1.3.1).
+Whitespaces are stripped from both ends of each value ("foo,bar" is the same as "foo, bar ").
+You can use literal commas and spaces anywhere by escaping (:code:`\,` and :code:`\ `).
+.. doctest::
+
+    >>> OmegaConf.register_resolver("concat", lambda x,y: x+y)
+    >>> c = OmegaConf.create({
+    ...     'key1': '${concat:Hello,World}',
+    ...     'key_trimmed': '${concat:Hello , World}',
+    ...     'escape_whitespace': '${concat:Hello,\ World}',
+    ... })
+    >>> c.key1
+    'HelloWorld'
+    >>> c.key_trimmed
+    'HelloWorld'
+    >>> c.escape_whitespace
+    'Hello World'
+
+
+
 Merging configurations
 ----------------------
 Merging configurations enables the creation of reusable configuration files for each logical component
--- a/omegaconf/config.py
+++ b/omegaconf/config.py
@@ -416,7 +416,9 @@ class Config(object):
         return ret
 
     def _resolve_single(self, value):
-        match_list = list(re.finditer(r"\${(\w+:)?([\w\.%_-]+?)}", value))
+        key_prefix = r"\${(\w+:)?"
+        legal_characters = r"([\w\.%_ \\,-]*?)}"
+        match_list = list(re.finditer(key_prefix+legal_characters, value))
         if len(match_list) == 0:
             return value
 
--- a/tests/test_interpolation.py
+++ b/tests/test_interpolation.py
@@ -233,7 +233,7 @@ def test_resolver_cache_1():
 
 def test_resolver_cache_2():
     """
-    Tests that resolver cache is not shared between different OmegaConf objects 
+    Tests that resolver cache is not shared between different OmegaConf objects
     """
     try:
         OmegaConf.register_resolver("random", lambda _: random.randint(0, 10000000))
@@ -246,6 +246,21 @@ def test_resolver_cache_2():
         assert c1.k != c2.k
         assert c1.k == c1.k
         assert c2.k == c2.k
+    finally:
+        OmegaConf.clear_resolvers()
+
+
+@pytest.mark.parametrize('resolver,name,key,result', [
+    (lambda *args: args, 'arg_list', "${my_resolver:cat, dog}",("cat", "dog")),
+    (lambda *args: args, 'escape_comma',"${my_resolver:cat\, do g}",("cat, do g",)),
+    (lambda *args: args, 'escape_whitespace',"${my_resolver:cat\, do g}",("cat, do g",)),
+    (lambda: 'zero', 'zero_arg',"${my_resolver:}",("zero")),
+])
+def test_resolver_that_allows_a_list_of_arguments(resolver, name, key, result):
+    try:
+        OmegaConf.register_resolver('my_resolver', resolver)
+        c = OmegaConf.create({name: key})
+        assert c[name] == result
     finally:
         OmegaConf.clear_resolvers()
 
--- a/omegaconf/omegaconf.py
+++ b/omegaconf/omegaconf.py
@@ -2,6 +2,7 @@
 import copy
 import io
 import os
+import re
 import sys
 import warnings
 from contextlib import contextmanager
@@ -137,6 +138,19 @@ class OmegaConf:
         target.merge_with(*others[1:])
         return target
 
+
+    @staticmethod
+    def _tokenize_args(string):
+        if string is None or string == '':
+            return []
+        def _unescape_word_boundary(match):
+            if match.start() == 0 or match.end() == len(match.string):
+                return ''
+            return match.group(0)
+        escaped = re.split(r'(?<!\\),', string)
+        escaped = [re.sub(r'(?<!\\) ', _unescape_word_boundary, x) for x in escaped]
+        return [re.sub(r'(\\([ ,]))', lambda x: x.group(2), x) for x in escaped]
+
     @staticmethod
     def register_resolver(name, resolver):
         assert callable(resolver), "resolver must be callable"
@@ -145,7 +159,7 @@ class OmegaConf:
 
         def caching(config, key):
             cache = OmegaConf.get_cache(config)[name]
-            val = cache[key] if key in cache else resolver(key)
+            val = cache[key] if key in cache else resolver(*OmegaConf._tokenize_args(key))
             cache[key] = val
             return val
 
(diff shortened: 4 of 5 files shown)


swist:
BTW when do you plan to publish a new version to pypi? Would be nice to not have to run my fork

omry:
I am planning to make a pass at the open issues first.
It may take a few months as I am still heavily focused on Hydra and also got a bunch of travel coming in the following months.

In the mean time, you can install from master by putting something like this in your setup.py:
(Or probably also instructing pip to install this directly).

```python
        install_requires=[
            "omegaconf@git+ssh://git@github.com/omry/omegaconf.git@master",
        ],
```

omry:
What do you think about extending the resolver to properly accept multiple arguments instead?
Allowing any argument like this will deny the possibility of cleanly doing it in the future.

tests/test_interpolation.py
```diff
@@ -250,6 +250,30 @@ def test_resolver_cache_2():
         OmegaConf.clear_resolvers()
 
 
+def test_resolver_that_allows_any_char():
+    def env_or_default(x):
+        split = x.split(",")
+        if len(split) > 1:
+            return os.getenv(split[0], split[1])
+        return os.getenv(split[0])
```
omry:
```python
    def env_or_default(name, def = None):
        if def is not None:
            return os.getenv(name, def)
        return os.getenv(name)

    c = OmegaConf.create(
      dict(
        default_env_set="${env_or_default:arg1,env_name,env_default}",
      )
    )
```

omry: something like this.
It's more work but more useful.

swist: Looks good, I'll have a crack later this week

swist: Should we then make `,` a special token that denotes an argument list?

omry: I think this is reasonable, but it would be good to also support \, to escape it and pass a literal ,.
another reasonable thing is to strip whitespace around arguments (and escape whitespace with '\ '.

omry: This can also naturally support 0 argument calls:
`${foo:}`

omegaconf/config.py
```diff
@@ -399,32 +399,32 @@ def merge_with(self, *others):
         Config._re_parent(self)
 
     @staticmethod
-    def _resolve_value(root_node, inter_type, inter_key):
+    def _resolve_value(root_node, inter_type, *args):
```
omry: I prefer that we keep the value as a string until we actually call the function.

swist:
Cool, will update later tomorrow, got the tokenizer in the right place, just having some pains dealing with escaping spaces - ideally something like this would work, but no clean solution comes to mind off the top of my head

```
tokenize('cat dog, dog cat') == ['cat dog', 'dog cat']
tokenize('cat dog,\ dog cat') == ['cat dog', ' dog cat']
```

omry:
Great, I will take a closer look later.
Ideally, this would be handled by a proper tokenizer like antlr, but OmegaConf does not (yet) has this dependency.
It would also be pretty involved to set this up cleanly (see http://github.com/omry/antelope for an example).

Re:
> ```
> tokenize('cat dog, dog cat') == ['cat dog', 'dog cat']
> tokenize('cat dog,\ dog cat') == ['cat dog', ' dog cat']
> ```

I think something like this would be a good overall strategy:
1. split by un-escaped commas:
'cat dog,\ dog cat' -> ['cat dog', '\ dog cat']
2. replace '\,' and '\ ' by ',' and ' '.
3. for each element, strip whitespace for both ends.

The first part is a bit tricky, you can achieve that relatively cleanly by replacing '\,' with an illegal character like *, then split by that, and then replace it back.

docs/source/usage.rst
```diff
@@ -255,6 +255,27 @@ This example creates a resolver that adds 10 the the given value.
     1000
 
 
+Custom resolvers also support variadic argument lists. Commas designate the boundaries between arguments.
+Any you can escape them using the :code:`\` character. The strings will get stripped on both ends unles you escape
+the whitespace with :code:`\`
```
omry: I would like to simplify this a bit, what do you think about:
```
Custom resolvers support variadic argument lists in the form of a comma separated list of zero or more values.
Whitespaces are stripped from both ends of each value ("foo,bar" is the same as "foo, bar ").
You can use literal commas and spaces anywhere by escaping ('/,' and '/ ').
```

docs/source/usage.rst
```diff
@@ -255,6 +255,27 @@ This example creates a resolver that adds 10 the the given value.
...
+
+.. doctest::
+
+    >>> OmegaConf.register_resolver("concat", lambda x,y: x+y)
+    >>> c = OmegaConf.create({
+    ...     'key': '${concat:Hello\ ,World}'
+    ...     'no_escape': '${concat:Hello, World}'
+    ...     'escape_whitespace': '${concat:Hello,\ World}'
+    ... })
```
omry: Order from simple to complicated.
```
'key1': '${concat:Hello,World}'
'key_trimmed': '${concat:Hello, World}'
'escape_whitespace': '${concat:Hello,\ World}'
```

omegaconf/config.py
```diff
@@ -416,7 +416,7 @@ def _resolve_value(root_node, inter_type, inter_key):
         return ret
 
     def _resolve_single(self, value):
-        match_list = list(re.finditer(r"\${(\w+:)?([\w\.%_-]+?)}", value))
+        match_list = list(re.finditer(r"\${(\w+:)?(.+?)?}", value))
```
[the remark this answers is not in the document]
swist: Ah that came from the first pass, when I wanted to pass a string and have the resolver handle the delimiter. Allowing any string is useful when you want to give a default connection string or random string - that was the initial usecase for this PR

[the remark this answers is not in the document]
omry: Gotcha.
Can you make this tighter?
as I mentioned, going wide here make future changes.
The only reason we can make space and comma special now without worrying about backward compatibility is that they were simply illegal before so we can be sure no code relying on them being in the config.
by staying as narrow as possible, we keep the option to add additional reserved characters if the future.
for example, we might want to support function like syntax: foo(a,b,c) at some point.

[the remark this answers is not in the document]
swist: My next best guess is something like this: `\${(\w+:)?([\w\.%-._~:\/\\ ?#\[\]\@!$&'()*+,;=]+)?`, 😂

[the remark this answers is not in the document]
swist: I'll go through my usecase and think what is the set that I expect to use in the near future. I think I will be able to get away with alphanumerics and `:`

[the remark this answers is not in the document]
swist: Actually, I think just comma will work for my usecase, completely forgot that I could compose things back with string interpolator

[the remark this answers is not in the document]
omry: Okay, so this is where regex101 can come in handy.
it let's you build int tests for your regex, testing that specific strings are being matched.

this: `\w\.%_-` is matching any alphanum, dot, _, - and %.
we can to add , to it, and also allow spaces.
I would also suggest that you refactor it into a separate variable to make things easier to parse.

Check this up:
https://regex101.com/r/CB8Tvn/1/tests

It has what seems (in a quick test) a valid regex.
you can use the tool to generate python code.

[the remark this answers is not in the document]
omry: version 2 also handles ${foo:}
https://regex101.com/r/CB8Tvn/2/tests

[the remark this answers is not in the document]
swist: Yeah yours is just missing the escape char `\` and we’re good.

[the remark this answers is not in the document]
swist: Cool split it up and added ` \,` to legal list

docs/source/usage.rst
```diff
@@ -255,6 +255,26 @@ This example creates a resolver that adds 10 the the given value.
     1000
 
 
+Custom resolvers support variadic argument lists in the form of a comma separated list of zero or more values.
```
omry: Please add a note saying this is from OmegaConf 1.3.1 (There are similar notes around the docs about deprecated things (which I need to clean up, duh)).

tests/test_interpolation.py
```diff
@@ -250,6 +250,21 @@ def test_resolver_cache_2():
         OmegaConf.clear_resolvers()
 
 
+@pytest.mark.parametrize('resolver,name,key,result', [
+    (lambda *args: args, 'arg_list', "${my_resolver:cat, dog}",("cat", "dog")),
+    (lambda *args: args, 'escape_comma',"${my_resolver:cat\, do g}",("cat, do g",)),
+    (lambda *args: args, 'escape_whitespace',"${my_resolver:cat\, do g}",("cat, do g",)),
+    (lambda: 'zero', 'zero_arg',"${my_resolver:}",("zero")),
+])
```
omry: Not familiar with this syntex. Why do you need it here?

docs/source/usage.rst
swist: Updated

tests/test_interpolation.py
swist: I'm not sure what you mean? Would you prefer I passed in entire config dictionaries as opposed to pairs of `key, string_to_resolve`?

omry: oh, I was confused by the lambda being passed there, I get why you need this now.

omry:
Thanks a lot for this!
By the way, It will be useful to include to add an optional default value to the environment variable resolver by default.

omry:
General feedback:
When I merge, I can squash in the GitHub UI just prior to merging the PR.
For future diffs, it will be easier if you don't force push to allow me to see the incremental difference from the previously reviewed diff(s).

swist:
Cool no problem, I normally default to pull-rebase so that merge commits don't come through and I didn't want to spam history with a bunch of `wip` commits.

PR #49, opened 2019-10-26
implemented and tested __contains__ (key in cfg)

omry:
implemented __contains__ in DictConfig.

A key is in a dict if there is an associated value that is not None and does not result in a MissingMandatoryValue or KeyError exceptions.


omry:
Closes #50

PR #52, opened 2019-11-02
Bugfix/sometimes envvars begin with yaml special chars

swist:
If your secrets begin with characters like :, >, / and you use
the env resolver, it can cause quite a lot of havoc. This is one
pass at a solution for it

### DIFF
--- a/omegaconf/omegaconf.py
+++ b/omegaconf/omegaconf.py
@@ -15,9 +15,13 @@ from .config import Config
 def register_default_resolvers():
     def env(key):
         try:
-            return yaml.safe_load(os.environ[key])
+            value = os.environ[key]
         except KeyError:
             raise KeyError("Environment variable '{}' not found".format(key))
+        try:
+            return yaml.safe_load(value)
+        except yaml.scanner.ScannerError:
+            return value
 
     OmegaConf.register_resolver('env', env)
 
--- a/tests/test_interpolation.py
+++ b/tests/test_interpolation.py
@@ -147,6 +147,22 @@ def test_env_interpolation1():
         del os.environ['foobar']
         OmegaConf.clear_resolvers()
 
+@pytest.mark.parametrize('value', [
+    '>1234',
+    ':1234',
+    '/1234',
+])
+def test_env_string_val_but_invalid_yaml(value):
+    try:
+        os.environ['foobar'] = value
+        c = OmegaConf.create(dict(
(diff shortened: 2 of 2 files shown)


swist:
In my ideal universe, understanding whether something is yaml parseable and comes back as the correct python primitive and handling unparseable yaml scalars as strings should be as transparent as possible.

We could perhaps try to identify if the string is OK to feed to yaml.safe_load before parsing it instead?

```Characters that cannot be used at the start of an unquoted string:

! Tag like !!null
& Anchor like &mapping_for_later_use
* Alias like *mapping_for_later_use
-<space> Block sequence entry
:<space> Block mapping entry
?<space> Explicit mapping key
{, }, [, ] Flow mapping or sequence
, Flow Collection entry seperator
# Comment
|, > Block Scalar
@, '`' (backtick) Reserved characters
", ' Double and single quote```

omry:
We used yaml parsing as a shortcut for getting the primitives.
For command line, this actually allows more - for example you can specify a list in JSON format there:
Foo=[a,b,c]

For env resolver, I think this makes less sense and will probably not work properly.

The proper solution here might be to implement the type detection and conversion to primitive directly.
This can be a new function that is very specific about the conversions it’s making:
10: int
10.0 float
True|FAlse : bool
Optional: “10” : string
Anything else: string.

omry:
Finally, there is always an option for you to implement a custom env revolver (in your own project) that does what you want, although I think it’s good to improve the core env resolver.

swist:
What do you mean by "it probably won't work properly for env" as opposed to CLI args?

omry:
Cli args are merges into the config object. A list in the cli is converted to ListConfig, which has all the properties of OmegaConf.
A resolver returning a list is different. For example, if that list has an interpolation it will not resolve properly.

swist:
Gotcha. I agree that something that actually supports parsing things from env scalars to scalars in omegaconf is probably more useful - something that always assumes that the envvars are yaml scalars perhaps?

omry:
Yeah, I described that logic above.
Yaml scalars can also be multi line strings that are out of scope here probably.

omry:
By the way, I now remember that I initially has yaml parsing in the env resolver and removed it.
Some strings are not good as yaml strings.

omry:
@swist, so you intend to finish the work on this PR?

omry:
Handling this in #79

PR #55, opened 2019-11-05
Enabling black and flake8 linting and formatting

omry:


omry:
@swist FYI: I updated your recent tests a bit because the escaping was actually incorrect (lint error).
(See second diff in stack).
This should not cause any use facing changes.

Issue #61, opened 2019-11-06
Omegaconf.create accept Protocol like Class


tamuhey:
Hi, thank you for this great repo!

I want to instantiate `omegaconf.Config` with protocol like class to tell IDE what member the config have, that is:

```python
class A:
    a: int = 1

cfg: A = omegaconf.OmegaConf.create(A)
print(cfg.a)
```
```
1
```

omry:
Hi @tamuhey,
This seems very specific to some IDEs.
Is there any common standard to what a "protocol like class" is?
Can you achieve the same results by doing something like:
```
class A:
    a: int = 1

cfg = your_create(A)
print(cfgf.a)
```
?

omry:
Can you explain what is the common use pattern here?
What are you getting by using OmegaConf? 
For example, are you merging your config objects with anything else?

tamuhey:
Sorry for not explaining enough.

1. "Protocol like class" means [typing.Protocol](https://docs.python.org/3/library/typing.html#typing.Protocol) like class that provides structural subtyping for type checkes.

2. common use pattern

- The class (`class A` in previous example) provides structural subtyping, and you can use tab completion with IDE (e.g. VSCode)
- The class also provides default value of config in its class variable.

3. > For example, are you merging your config objects with anything else?

Yes, I'm merging config objects, that's why I'm using omegaconf.
For example:

```python
class ConfA:
    a: int = 1


default_config_a = create_config(ConfA)


def fn_a(cfg: ConfA = default_config_a):
    print(cfg.a)


class ConfB:
    b: str = "foo"


default_config_b = create_config(ConfB)


def fn_b(cfg: ConfB = default_config_b):
    print(cfg.b)


default_config = omegaconf.OmegaConf.create(
    {"a": default_config_a, "b": default_config_b, "foo": 10}
)


def fn_manager(cfg=default_config):
    fn_a(cfg.a)
    fn_b(cfg.b)
    print(cfg.foo)


fn_manager()
```
```
1
foo
10
```

omry:
When I asked about merge,I was thinking more about OmegaConf.merge(conf1, conf2, ..) or conf1.merge_with(conf2, ...).

OmegaConf is surprisingly complex for such a small library and I want to very careful adding additional API surface area.
You already asked (more or less) the same question at the Hydra repo, the answer there was that it does not belong in Hydra. the same goes for OmegaConf - which is even a lower level library.

[...]

PR #62, opened 2019-11-06
Omegaconf.create accept Protocol like Class

tamuhey:
from #61


tamuhey:
Close. See the discussion in #61

PR #79, opened 2019-11-11
env resolver no longer parsing yaml strings 

omry:
No longer using yaml parser to decode environment variables as it causes issues for strings that are not valid yaml.


omry:
Supersedes #52

Issue #80, opened 2019-11-12
"a" in cfg is false if cfg.a is None


omry:
```
{'a': None, 'b': '${a}'}
>>> "a" in a
True
>>> "b" in a
False
```

omry:
fixed by #84

PR #81, opened 2019-11-12
`in` operator should only return false if key is actually missing.

swist:
Addresses #80. I'm not sure if I like the solution with a sentinel, but seems to be in line with your philosophy of doing minimum possible changes

Do you see a better way other than special_casing the `select()` behaviour? It seems like is used in more than one way

### DIFF
--- a/tests/test_basic_ops_dict.py
+++ b/tests/test_basic_ops_dict.py
@@ -142,9 +142,9 @@ def test_dict_keys():
 def test_pickle_get_root():
     # Test that get_root() is reconstructed correctly for pickle loaded files.
     with tempfile.TemporaryFile() as fp:
-        c1 = OmegaConf.create(dict(a=dict(a1=1, a2=2,),))
+        c1 = OmegaConf.create(dict(a=dict(a1=1, a2=2)))
 
-        c2 = OmegaConf.create(dict(b=dict(b1="???", b2=4, bb=dict(bb1=3, bb2=4,),),))
+        c2 = OmegaConf.create(dict(b=dict(b1="???", b2=4, bb=dict(bb1=3, bb2=4))))
         c3 = OmegaConf.merge(c1, c2)
 
         import pickle
@@ -193,6 +193,8 @@ def test_dict_pop():
         ({"a": 1, "b": "???", "c": "${b}"}, "c", False),
         ({"a": 1, "b": "${not_found}"}, "b", False),
         ({"a": "${unknown_resolver:bar}"}, "a", True),
+        ({"a": None, "b": "${a}"}, "b", True),
+        ({"a": "cat", "b": "${a}"}, "b", True),
     ],
 )
 def test_in_dict(conf, key, expected):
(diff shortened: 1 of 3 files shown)


tests/test_basic_ops_dict.py
```diff
@@ -142,9 +142,9 @@ def test_dict_keys():
 def test_pickle_get_root():
     # Test that get_root() is reconstructed correctly for pickle loaded files.
     with tempfile.TemporaryFile() as fp:
-        c1 = OmegaConf.create(dict(a=dict(a1=1, a2=2,),))
+        c1 = OmegaConf.create(dict(a=dict(a1=1, a2=2)))
 
-        c2 = OmegaConf.create(dict(b=dict(b1="???", b2=4, bb=dict(bb1=3, bb2=4,),),))
+        c2 = OmegaConf.create(dict(b=dict(b1="???", b2=4, bb=dict(bb1=3, bb2=4))))
```
swist: I don't know why these got caught in `black` crossfire

omry:
Thanks!
I have no such philosophy, I consider the unit tests the guardians against bad changes. 

I just spotted this bug yesterday during a hackathon so I didn't think much about the solution.
I did think it might be tricky, and I thought that the solution might involve splitting select into two functions, maybe:
```
def _dereference_key(key):
   # return key if not an interpolation
   # else return the key pointed to by key, in the test case this would return "a" -> "a" and "b" -> "a"
   # also "x.b" -> "x.a" in a test like {"x": {"a": 10, "b": "${x.a}"}}
```

select can then be implemented as something like:
```
def select(self, key):
  real_key = self._dereference_key(key)
  # current select logic on real_key
```

Then contains can be implemented as:
```
def __contains__(self , key )
   real_key = self._dereference_key(key)
   # contains on real_key
```

Thoughts?

omry:
@swist, I created a chat for Hydra and OmegaConf (see main readme). would be great if you can join.

swist:
I have tried something similar - adding `exists` which behaves similarly to `select` but doesn't return the value, but wasn't happy with that. Perhaps delegating a bit of work that select currently does will work better

omry:
Yeah, this is what I had in mind.
Need to see what the code would actually look like when implemented to really evaluate if this is cleaner than you proposed solution.

omry:
One thing that guides me in general is to be careful exposing new public APIs (those are easy to introduce and hard to remove).
Any reason you went with this and not with my proposed private API?

exists is not a good name here, it sounds like __contains__(self, key).
I think _dereference_key is better here, but as I said it may be tricky to implement cleanly.
Since select can be implemented in terms of _dereference_key() this will also minimize code duplication.

swist:
Didn't really mean to make it a public API, it was just me digging through the reflog so I could show you what a half-way shop to `_dereference_key` would look like
If you think that this is a better direction than sentinels I'll have a look at `_dereference_key` sometime this week

omry:
Oh, gotcha.
Yeah, my gut feeling is that this should result in a cleaner change, despite the clunky name for _dereference_key().
Thanks again.

omry:
FYI:
I plan to release a new version of OmegaConf either this weekend or next week, this is the last bug I want to address before this new version.

Do you think you will be able to finish this PR soon?
If you don't have time I don't mind taking over.

swist:
I was going to look at it today!

tests/test_basic_ops_dict.py
omry: Interesting, black does not fix this for me.
On the other hand, it's also fine with the version without the extra commas.

swist:
fixed in #84

PR #86, opened 2019-11-23
Structured configs

### LINKED ISSUE #87: Construction of OmegaConf from Structured Configs
The associated PR (#86 ) is adding support for initializing OmegaConf from class files and objects.
Details below

1. Config classes and objects (dataclass or attrs classes) can be converted to OmegaConf object.

2. The following value types are supported:
 - with a default value assigned to it.
 - Optional fields (can accept None)
 - assigned omegaconf.MISSING, Equivalent for '???' in the yaml files (Mandatory missing).
   The semantics here is that this value is not yet set. it must be set prior to read access.
 - Interpolation using omegaconf.II() function :
   II("foo") is equivalent to ${foo} in the YAML file.
   Examples:
   - `foo=II("bar")`  : Foo inherits the type and value of bar at runtime.
   - `foo=II("http://${host}:${port}/")`  : foo is always a string. actual value is determined by host and port
   - `foo=II("env:USER")` : foo is evaluated on access, the type and value is determined by the function (env in this case)

Example for bool types (The same is supported for int, str, float and enum):
```python
@dataclass
class BoolConfig:
    with_default: bool = True
    null_default: Optional[bool] = None
    mandatory_missing: bool = MISSING
    interpolation: bool = II("with_default")
```

3. the typing information in them is available and acted on at runtime (composition and command line overrides are at runtime).

4. type annotations can be used to get static type checking.

Currently the following is supported for both dataclasses and attr classes.
Examples below are @dataclasses. attr classes are similar.
```python
@dataclass
class Database:
    name: str = "demo_app_db"
    # A an untyped list. will accept any type that can be represented by OmegaConf
    tables: List = field(default_factory=lambda: ["table1", "table2"])
    port: int = 3306

    # A list of integers, will reject anything that cannot be converted to int at runtime
    admin_ports: List[int] = field(default_factory=lambda: [8080, 1080])

    # A dict of string -> int, will reject any value that cannot be converted to int at runtime
    error_levels: Dict[str, int] = field(
        default_factory=lambda: {"info": 0, "error": 10}
    )

```

This class, and objects of this class can be used to construct an OmegaConf object:
```python
# Create from a class
conf1 = OmegaConf.create(Database)

# Create from an object
conf2 = OmegaConf.create(Database(port=3307))
```

Python type annotation can then be used on the config object to static type checking:
```python
def foo(cfg : Database):
   cfg.port = 10 # passes static type check
   cfg.port = "nope" # fails static type check
```
5. Composition and overrides are often happening at runtime (from files, or command line flags).
To support that, there is also a runtime validation and conversion.
The following will succeed at runtime:
```python
conf.merge_with_dotlist(['port=30']) # success
conf.merge_with_dotlist(['port=nope']) # runtime error
```
6. OmegaConf objects created from Structured configs will be set to struct mode, this means that any access to a field not in the struct will result in a runtime error:

```python
conf.pork = 80 # fail
```

7. Typed containers are validated at runtime:
```python
conf: Database = OmegaConf.create(Database)
conf.admin_ports[0] = 999  # okay
assert conf.admin_ports[0] == 999

conf.admin_ports[0] = "1000"  # also ok!
assert conf.admin_ports[0] == 1000

    conf.admin_ports[0] = "fail"

    conf.error_levels["info"] = "fail"
```
8. Nested Object/Classes are supported, as well as Enums.
```python

class Protocol(Enum):
    HTTP = 0
    HTTPS = 1


@dataclass
class ServerConfig:
    # Nested configs are by value.
    # This will be expanded to default values for Database
    db1: Database = Database()

    # This will be expanded to the values in the created Database instance
    db2: Database = Database(tables=["table3", "table4"])

    host: str = "localhost"
    port: int = 80
    website: str = MISSING
    protocol: Protocol = MISSING
    debug: bool = False
```
9. Enums can be assigned by Enum, by name string or by value (the later should be used at runtime only).

```python
    conf: ServerConfig = OmegaConf.create(ServerConfig(protocol=Protocol.HTTP))
    assert conf.protocol == Protocol.HTTP

    # The HTTPS string is converted to Protocol.HTTPS per the enum name
    conf.protocol = "HTTPS"
    assert conf.protocol == Protocol.HTTPS

    # The value 0 is converted to Protocol.HTTP per the enum value
    conf.protocol = 0
    assert conf.protocol == Protocol.HTTP

    # Enum names are care sensitive
        conf.protocol = "https"
 
    for value in (True, False, "1", "http", 1.0):
            conf.protocol = value
```

10. Frozen state is preserved (and is recursive):
```python
    @dataclass(frozen=True)
    class FrozenClass:
        x: int = 10
        list: List = field(default_factory=lambda: [1, 2, 3])

    conf = OmegaConf.create(FrozenClass)
    with pytest.raises(ReadonlyConfigError):
        conf.x = 20

    # Read-only flag is recursive
    with pytest.raises(ReadonlyConfigError):
        conf.list[0] = 20
```

### LINKED ISSUE #88: Drop support for Python 2.7
Python 2.7 is reaching end of life in 1/1/2020

1. ~~Provide an error when running OmegaConf 1.5 with Python 2.7, pointing users to pin 1.4 if they wish to use Python 2.7~~
2. ~~Remove 2.7 support from master, update docs, unit tests (Additional advice [here](https://python3statement.org/practicalities/)~~
3. ~~Clean up 2.7 specific and supporting dependencies~~
4. Leverage Python 3.x features: Type annotations, Mypy/Pyre. clean up Python 2.7 support

### LINKED ISSUE #99: Wrong exception on indirect access to missing value
```yaml
foo:
  bar: ???
  foo: ${bar}
```

```python
cfg = OmegaConf.create(..)
cfg.foo # should throw MissingMandatoryValue but throws KeyError
```

### LINKED ISSUE #101: Create API to test if an object is a config object
```python
OmegaConf.is_list(cfg)
OmegaConf.is_dict(cfg)
OmegaConf.is_config(cfg)
```

### LINKED ISSUE #102: Create an API to test if a key is MISSING
```
OmegaConf.is_missing(cfg, key)
```

### LINKED ISSUE #103: Rename omegaconf.Config to omegaconf.Container
Config is too generic, and users should use OmegaConf.is_config(obj) now anyway.

omry:
Closes #87 
Closes #101
Closes #102 
Closes #103
Closes #88 
Closes #99

### DIFF
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -9,7 +9,7 @@ from ._utils import (
     is_structured_config,
     _re_parent,
 )
-from .config import Config
+from .container import Container
 from .errors import (
     ReadonlyConfigError,
     MissingMandatoryValue,
@@ -22,7 +22,7 @@ from .node import Node
 from .nodes import ValueNode
 
 
-class DictConfig(Config):
+class DictConfig(Container):
     def __init__(self, content, parent: Optional[Node] = None, element_type=Any):
         super().__init__(element_type=element_type, parent=parent)
 
@@ -81,7 +81,7 @@ class DictConfig(Config):
         self._validate_access(key)
         self._validate_type(key, value)
 
-        if isinstance(value, Config):
+        if isinstance(value, Container):
             value = copy.deepcopy(value)
             value._set_parent(self)
 
@@ -229,9 +229,9 @@ class DictConfig(Config):
 
     def __eq__(self, other):
         if isinstance(other, dict):
-            return Config._dict_conf_eq(self, DictConfig(other))
+            return Container._dict_conf_eq(self, DictConfig(other))
         if isinstance(other, DictConfig):
-            return Config._dict_conf_eq(self, other)
+            return Container._dict_conf_eq(self, other)
         return NotImplemented
 
     def __ne__(self, other):
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -1,63 +1,96 @@
-from .config import Config
 import copy
+from enum import Enum
+
+from typing import Any, Optional
+
+from ._utils import (
+    get_structured_config_data,
+    is_structured_config_frozen,
+    is_structured_config,
+    _re_parent,
+)
+from .config import Config
 from .errors import (
     ReadonlyConfigError,
     MissingMandatoryValue,
     UnsupportedInterpolationType,
+    UnsupportedValueType,
+    UnsupportedKeyType,
+    ValidationError,
 )
-from .nodes import BaseNode, UntypedNode
+from .node import Node
+from .nodes import ValueNode
 
 
 class DictConfig(Config):
-    def __init__(self, content, parent=None):
-        super(DictConfig, self).__init__()
-        assert isinstance(content, dict)
-        self.__dict__["content"] = {}
-        self.__dict__["parent"] = parent
-        for k, v in content.items():
-            self.__setitem__(k, v)
+    def __init__(self, content, parent: Optional[Node] = None, element_type=Any):
+        super().__init__(element_type=element_type, parent=parent)
 
-    def __deepcopy__(self, memodict={}):
+        self.__dict__["content"] = {}
+        self.__dict__["_type"] = None
+        if is_structured_config(content):
+            d = get_structured_config_data(content)
+            for k, v in d.items():
+                self.__setitem__(k, v)
+
+            if is_structured_config_frozen(content):
+                self._set_flag("readonly", True)
+
+            if isinstance(content, type):
+                self.__dict__["_type"] = content
+            else:
+                self.__dict__["_type"] = type(content)
+
+        else:
+            for k, v in content.items():
+                self.__setitem__(k, v)
+
+    def __deepcopy__(self, memo={}):
         res = DictConfig({})
-        self._deepcopy_impl(res)
+        res.__dict__["content"] = copy.deepcopy(self.__dict__["content"], memo=memo)
+        res.__dict__["flags"] = copy.deepcopy(self.__dict__["flags"], memo=memo)
+        res.__dict__["_element_type"] = copy.deepcopy(
+            self.__dict__["_element_type"], memo=memo
+        )
+
+        res.__dict__["_type"] = copy.deepcopy(self.__dict__["_type"], memo=memo)
+        _re_parent(res)
         return res
 
     def __copy__(self):
-        res = DictConfig({})
+        res = DictConfig(content={}, element_type=self.__dict__["_element_type"])
         res.__dict__["content"] = copy.copy(self.__dict__["content"])
-        res.__dict__["parent"] = self.__dict__["parent"]
+        res.__dict__["_type"] = self.__dict__["_type"]
+        res.__dict__["flags"] = copy.copy(self.__dict__["flags"])
+        _re_parent(res)
         return res
 
     def copy(self):
         return copy.copy(self)
 
     def __setitem__(self, key, value):
-        assert isinstance(key, str)
+        if isinstance(key, Enum):
+            key = key.name
 
-        value = self._prepare_value_to_add(key, value)
+        if not isinstance(key, str):
+            raise UnsupportedKeyType(f"Key type is not str ({type(key).__name__})")
 
-        if key not in self.content and self._get_flag("struct") is True:
-            raise KeyError(
-                "Accessing unknown key in a struct : {}".format(self.get_full_key(key))
+        if self._get_flag("readonly"):
+            raise ReadonlyConfigError(self.get_full_key(key))
+
+        self._validate_access(key)
+        self._validate_type(key, value)
+
+        if isinstance(value, Config):
+            value = copy.deepcopy(value)
+            value._set_parent(self)
+
+        try:
+            self._set_item_impl(key, value)
+        except UnsupportedValueType:
+            raise UnsupportedValueType(
+                f"key {self.get_full_key(key)}: {type(value).__name__} is not a supported type"
             )
-
-        input_config_or_node = isinstance(value, (BaseNode, Config))
-        if key in self:
-            # BaseNode or Config, assign as is
-            if input_config_or_node:
-                self.__dict__["content"][key] = value
-            else:
-                # primitive input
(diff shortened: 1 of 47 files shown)


omegaconf/dictconfig.py
```diff
@@ -42,7 +42,7 @@ def __setitem__(self, key, value):
             )
 
         input_config_or_node = isinstance(value, (BaseNode, Config))
-        if key in self:
+        if self.get_node(key):
```
swist: Does this resolve the `Node` or resolve node's value? If the latter then it's not what you want

omry: the node itself.

omry: The node. iirc the problem is that if the value is ??? this returns false now and it was breaking the semantics I was going for with the new logic.

omegaconf/utils.py
```diff
@@ -0,0 +1,184 @@
...
+            if value == attr.NOTHING:
+                if is_nested:
+                    value = type_
+                else:
+                    raise ValueError(
+                        "Missing default value for {}, to indicate default must be populated later use '???'".format(
+                            name
+                        )
+                    )
+        if is_nested and value in (None, "???"):
```
tamuhey: I think hard coded `"???"` is undesirable. For example, `omegaconf.UNK` like `attr.NOTHING` or  `dataclasses.MISSING` is preferable.

tamuhey:
Using `typing.Optional` fails:
```python
@dataclass
class Foo:
    a: Optional[int] = 1


OmegaConf.create(Foo)
```
```
AttributeError                            Traceback (most recent call last)
<ipython-input-59-71ae7d57826e> in <module>
      4 
      5 
----> 6 OmegaConf.create(Foo)

~/work/omegaconf/omegaconf/omegaconf.py in create(obj, parent)
     37 
     38         if is_structured_config(obj):
---> 39             d = get_structured_config_data(obj)
     40             ret = DictConfig(d, parent)
     41             OmegaConf.set_struct(ret, True)

~/work/omegaconf/omegaconf/utils.py in get_structured_config_data(obj)
    135 def get_structured_config_data(obj):
    136     if is_dataclass(obj):
--> 137         return get_dataclass_data(obj)
    138     if is_attr_class(obj):
    139         return get_attr_data(obj)

~/work/omegaconf/omegaconf/utils.py in get_dataclass_data(obj)
    107             dict,
    108         ) and not is_dataclass(type_):
--> 109             value = node_wrap(type_, value)
    110 
    111         d[name] = value

~/work/omegaconf/omegaconf/utils.py in node_wrap(type_, value)
     32         node = StringNode(value)
     33     else:
---> 34         raise ValueError("Unexpected object type : {}".format(type_.__name__))
     35     return node
     36 

/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/typing.py in __getattr__(self, attr)
    700         if '__origin__' in self.__dict__ and not _is_dunder(attr):
    701             return getattr(self.__origin__, attr)
--> 702         raise AttributeError(attr)
    703 
    704     def __setattr__(self, attr, val):

AttributeError: __name__
```

omegaconf/utils.py
```diff
@@ -0,0 +1,184 @@
...
+    from enum import Enum
+    from .nodes import EnumNode
+
+
+def node_wrap(type_, value):
+    from typing import Any
+
+    if type_ == Any:
+        node = UntypedNode(value)
+    elif isinstance(type_, type) and issubclass(type_, Enum):
```
tamuhey: Is the first statement of this line `isinstance(type_, type)` necessary?

tamuhey:
You can resolve `Optional` type as follows:
```python
def resolve_optional(type_):
    if getattr(type_, "__origin__", None) is Union:
        args = type_.__args__
        if len(args) == 2 and args[1] is type(None):
            return args[0]
    return type_

assert resolve_optional(Optional[int]) is int
assert resolve_optional(int) is int
```

omegaconf/utils.py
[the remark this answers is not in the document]
omry: > Using `typing.Optional` fails:

Fixed and pushed to this PR, thanks.

> Is the first statement of this line isinstance(type_, type) necessary?

I thought it was, there was some error about issubclass only working on classes.
tests are passing now without it so I am removing it.

> I think hard coded "???" is undesirable. For example, omegaconf.NOTHING like attr.NOTHING 
or dataclasses.MISSING is preferable.

I agree.
I should also be able to get rid of the casts for "???" if that thing is of type Any.
Will give it a shot.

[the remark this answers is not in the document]
tamuhey: > Fixed and pushed to this PR, thanks.

Raises another error

```python
@dataclass
class Foo:
    a: Optional[int] = 1


OmegaConf.create(Foo)
```
```
ValueError                                Traceback (most recent call last)
<ipython-input-257-71ae7d57826e> in <module>
      4 
      5 
----> 6 OmegaConf.create(Foo)

~/work/omegaconf/omegaconf/omegaconf.py in create(obj, parent)
     38         if is_structured_config(obj):
     39             d = get_structured_config_data(obj)
---> 40             ret = DictConfig(d, parent)
     41             OmegaConf.set_struct(ret, True)
     42             return ret

~/work/omegaconf/omegaconf/dictconfig.py in __init__(self, content, parent)
     16         self.__dict__["parent"] = parent
     17         for k, v in content.items():
---> 18             self.__setitem__(k, v)
     19 
     20     def __deepcopy__(self, memodict={}):

~/work/omegaconf/omegaconf/dictconfig.py in __setitem__(self, key, value)
     35         assert isinstance(key, str)
     36 
---> 37         value = self._prepare_value_to_add(key, value)
     38 
     39         if key not in self.content and self._get_flag("struct") is True:

~/work/omegaconf/omegaconf/config.py in _prepare_value_to_add(self, key, value)
    523             raise ValueError(
    524                 "key {}: {} is not a primitive type".format(
--> 525                     full_key, type(value).__name__
    526                 )
    527             )

ValueError: key a: IntegerNode is not a primitive type
```

The code should be involved in test.

omry:
Progress:
support for Optional[T], List[T] and Dict[str, T] added.
T is limited to supported simple types (str, int, float, bool and enum)

omry:
This is functionally complete as far as I can tell.
I would appreciate a review pass on the docs and on the tests (especially changed tests)

PR #90, opened 2019-12-01
Fix bug in FloatNode.__eq__

### LINKED ISSUE #89: Bug in FloatNode.__eq__
```python
from omegaconf.nodes import FloatNode

x = FloatNode()
x == 1
```

raises

```
TypeError                                 Traceback (most recent call last)
<ipython-input-163-39f43fc4ed44> in <module>
      2 
      3 x = FloatNode()
----> 4 x == 1

~/work/omegaconf/omegaconf/nodes.py in __eq__(self, other)
    115             other_val = other
    116 
--> 117         return self.val == other_val or (math.isnan(self.val) and math.isnan(other_val))
    118 
    119 

TypeError: must be real number, not NoneType
```

due to the following line:

https://github.com/omry/omegaconf/blob/7f521275c71565b5bb51d736699e3a11d51bd9cf/omegaconf/nodes.py#L88

`math.isnan` accepts `typing.SupportsFloat` but not `NoneType`.

tamuhey:
from #89 

Before this PR, the following code raises
```python
from omegaconf.nodes import FloatNode

x = FloatNode()
x == 1
```

After this PR, not raise.

### DIFF
--- a/omegaconf/nodes.py
+++ b/omegaconf/nodes.py
@@ -85,7 +85,11 @@ class FloatNode(BaseNode):
         else:
             other_val = other
 
-        return self.val == other_val or (math.isnan(self.val) and math.isnan(other_val))
+        if self.val == other_val:
+            return True
+        if not hasattr(self.val, "__float__") or not hasattr(other_val, "__float__"):
+            return False
+        return math.isnan(self.val) and math.isnan(other_val)
 
 
 class BooleanNode(BaseNode):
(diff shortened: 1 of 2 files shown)


omry:
Thanks!
I have a few requests, but let's park it until I land #87, It has a lot of changes to nodes.py.

omry:
I fixed this as a part of #86 while improving test coverage.
Test for the patch, closing.

PR #96, opened 2019-12-11
Fix disable automatic conversion of date strings in yaml decoding

### LINKED ISSUE #95: Disable automatic conversion of date strings in from_dotlist
### Issue
When instantiating OmegaConf from `from_dotlist`, yaml loader tries to convert date-formatted strings into `date` type, therefore raises error when validating the type of values.

```
 OmegaConf.from_dotlist(["my_date=2019-11-11"])

>>>

.../site-packages/omegaconf/config.py in _prepare_value_to_add(self, key, value)
    448         if not Config.is_primitive_type(value):
    449             full_key = self.get_full_key(key)
--> 450             raise ValueError("key {}: {} is not a primitive type".format(full_key, type(value).__name__))
    451
    452         if self._get_flag('readonly'):

ValueError: key my_date: date is not a primitive type
```
This error also blocks the overriding of configuration values by using cli arguments in hydra.

### Potential workaround
Could be solved by removing implicit timestamp resolvers in default `yaml.SafeLoader`.
(Related stackoverflow: https://stackoverflow.com/questions/34667108/ignore-dates-and-times-while-parsing-yaml)

shuki-hpcnt:
Fixes #95 by removing timestamp implicit resolver in yaml loader. 
```
In [3]:  OmegaConf.from_dotlist(["my_date=2019-11-11"])
Out[3]: {'my_date': '2019-11-11'}
```

### DIFF
--- a/news/95.bugfix
+++ b/news/95.bugfix
@@ -1,0 +1,1 @@
+Disable automatic conversion of date strings in `from_dotlist`
--- a/news/95.bugfix
+++ b/news/95.bugfix
@@ -1,1 +1,1 @@
-Disable automatic conversion of date strings in `from_dotlist`
+Disable automatic conversion of date strings in yaml decoding
--- a/omegaconf/config.py
+++ b/omegaconf/config.py
@@ -39,6 +39,14 @@ def get_yaml_loader():
         ),
         list(u"-+0123456789."),
     )
+    loader.yaml_implicit_resolvers = {
+        key: [
+            (tag, regexp)
+            for tag, regexp in resolvers
+            if tag != u"tag:yaml.org,2002:timestamp"
+        ]
+        for key, resolvers in loader.yaml_implicit_resolvers.items()
+    }
     return loader
 
 
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -50,7 +50,15 @@ def test_cli_passing():
 
 
 @pytest.mark.parametrize(
-    "input_, expected", [(["a=1", "b.c=2"], dict(a=1, b=dict(c=2)))]
+    "input_, expected",
+    [
+        # simple
+        (["a=1", "b.c=2"], dict(a=1, b=dict(c=2))),
+        # string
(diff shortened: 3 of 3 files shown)


omry:
Thanks, can you add a unit test?

omry:
And a news fragment file in news?
Something like:
news/Issue-num.bugfix:
Short description of bug fix

shuki-hpcnt:
Should I add new fragment under version `1.4.0`?

omry:
Add the news fragment in the dir, not in the news file. 
I have an upcoming major change coming to OmegaConf, which is potentially going to break compatibility in some ways. I will try to include your patch in a maintenance branch for 1.4 but it would be awesome if you can do some early testing for the new version (2.0.0rc1) when I release it.

omry:
Changes are at #86.
Changes there are massive but well tested.
I am planning a few other incompatible fixed before 2.0.0.

omry:
Please check the issue #87 for details about that that huge PR is about. It’s adds some great new stuff.

shuki-hpcnt:
I've added simple test and news as requested.
Also checked the new PR and issue for Omegaconf 2.0. New features look nice! Looking forward to use it :)

omry:
So actually your diff disables date implicit conversion inOmegaConf period.
Not just for from_dotlist.

news/95.bugfix
```diff
@@ -0,0 +1 @@
+Disable automatic conversion of date strings in `from_dotlist`
```
omry: Can you change it to be not about dotlist?
This change is effecting all of OmrgaConf YAML decoding.

shuki-hpcnt: I only checked `config.py` and thought that `merge_with_dotlist` is the only part that uses yaml loader, but you are indeed right! I'll fix that.

omry:
Alright, thanks for this.
I created a new 1.4_branch for bug fixes of the existing release line and I will patch it with your changes.

omry:
I released 1.4.1rc1
You should be able to install it with `pip install --upgrade --pre omegaconf`.

PR #110, opened 2019-12-29
add unicode support for pretty.

arisliang:


omegaconf/container.py
```diff
@@ -296,7 +296,7 @@ def pretty(self, resolve=False):
         :return: A string containing the yaml representation.
         """
         container = OmegaConf.to_container(self, resolve=resolve, enum_to_str=True)
-        return yaml.dump(container, default_flow_style=False)
+        return yaml.dump(container, default_flow_style=False, allow_unicode=allow_unicode)
```
omry: I think it's safe to just allow unicode.
no need to extend the interface for this.

omry:
Also, please add a news snippet file to news.
Something like:

*news/110.bugfix*
Fixed pretty to handle strings with unicode characters correctly

omry:
Oh, and of course - a unit test verifying this remains fixed.
(It should reproduce the problem you are fixing here, and should fail if your fix is not in place).

Thanks!

omegaconf/container.py
arisliang: Removed the interface extension.

arisliang:
`news/110.bugfix` - added news
unit test:

omry:
Can you try to rebase on master? there still seems to be merge errors.

arisliang:
There seems to too many conflicts when rebasing. I try re-create the PR

PR #155, opened 2020-02-21
Compact key support

### LINKED ISSUE #152: Enhancement: Compact key support
## Compact keys
This will probably come in OmegaConf 2.1 as 2.0 is closed to new features at this time.

Given:
```yaml
hydra:
  sweeper:
    params:
      budget: 100
      num_workers: 10
```

Allow the same config to be represented by:

```yaml
hydra.sweeper.params:
  budget: 100
  num_workers: 10
```
## Compatibility considerations
This changes the semantic meaning of a key with a dot, from a 
```yaml
foo.bar: 10
```
to:
```yaml
foo:
  bar: 10
```

People using `.` in their keys now should be aware of it and ideally stop doing it (in 1.4 or 2.0).
OmegaConf 2.0 will add a deprecation warning that can be disabled by setting the environment variable `OC_DISABLE_DOT_ACCESS_WARNING` to the value "1".

To address this, there will be multiple opt out mechanisms (fully specification tbd).
some ideas:

- [x] Allow keys with dot by escaping the `.` : `foo\.bar`
- [ ] Possibly allow file level opt out via a special header
- [ ] Possibly allow API level opt out when loading the file or when constructing a DictConfig object.

omry:
TODO:

- [ ] documentation
- [ ] compact pretty print() by default. flag to turn it off.
- [ ] convenience function to convert to standard yaml
- [ ] ability to disable compact keys while creating a config object (including loading).
- [ ] ability to opt file based on a file header (`#omegaconf:compact_keys=false` for example)

closes #152

### DIFF
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -257,6 +257,8 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
             self._format_and_raise(key=key, value=value, cause=e)
 
     def __set_impl(self, key: Union[str, Enum], value: Any) -> None:
+        from omegaconf import ListConfig
+
         if isinstance(key, str):
             sub_keys = key.split(".")
             cur = self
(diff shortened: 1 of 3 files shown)


omry:
Abandoning this, see #332.

PR #158, opened 2020-02-28
Add support for pathlib.Path

alexvicegrab:
Addresses issue #159

Test for pathlib.Path added.

### DIFF
--- a/omegaconf/omegaconf.py
+++ b/omegaconf/omegaconf.py
@@ -181,7 +181,7 @@ class OmegaConf:
                     )
 
     @staticmethod
-    def load(file_: Union[str, IO[bytes]]) -> Union[DictConfig, ListConfig]:
+    def load(file_: Union[str, pathlib.Path, IO[bytes]]) -> Union[DictConfig, ListConfig]:
         from ._utils import get_yaml_loader
 
         if isinstance(file_, (str, pathlib.Path)):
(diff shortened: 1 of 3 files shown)


tests/test_serialization.py
```diff
@@ -37,6 +38,20 @@ def save_load_from_filename(conf: Container, resolve: bool, expected: Any) -> No
         os.unlink(fp.name)
 
 
+def save_load_from_pathlib_path(conf: Container, resolve: bool, expected: Any) -> None:
```
omry: instead of duplicating the other function please parametrize it with pytest.mark.parametrize()

omegaconf/omegaconf.py
```diff
@@ -183,7 +184,7 @@ def create(  # noqa F811
     def load(file_: Union[str, IO[bytes]]) -> Union[DictConfig, ListConfig]:
```
omry: Please update the typing information on save and load to reflect this change.

tests/test_serialization.py
alexvicegrab: Thanks @omry.

I was able to implement the other suggestions, and while I would like to keep things DRY, I'm not quite sure how to parametrise using the original save_load_from_filename. Could you please point me in the right direction?

alexvicegrab: Made an implementation. Hopefully this is reasonably elegant :)

omegaconf/omegaconf.py
alexvicegrab: Done

tests/test_serialization.py
omry: It's good, and trickier than usual because you can't construct the object outside due to the temp file.
good job :).

omry:
Thanks for the fix!

omry:
whoops, failed lint.
unclear to me why CI didn't run for your PR.

alexvicegrab:
Oops, strange. I ran only pytest. Will run nox also next time.

Thanks for the quick responses and help!

omry:
cheers and thanks for the PR!

Issue #170, opened 2020-03-22
Set default value when environment variable is missing


kfollesdal:
It would be nice to be able to set a default value in yaml file for
environment variable interpolation when a variable is missing. 

E.g syntax can be `${env:ENVIROMENT_VARIABLE,default_value}`

`a: ${env:VALUE_A,1}`

if `VALUE_A `do not exist give `1`

omry:
Thanks, I have been wanting to do this.
The motivation for allowing multiple arguments for resolvers was exactly this use case, but I never got to add this.

PR #171, opened 2020-03-22
Optional default value for env resolver.

kfollesdal:
Proposal for a solution for #170

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -305,6 +305,12 @@ Input yaml file:
     >>> print(conf.user.home)
     /home/omry
 
+It is possible to set a default value if the environment variable is missing. 
+In the following example, we set 'omry' as the default value if the environment
+ variable USER is not defined.
+
+.. include:: env_default_interpolation.yaml
+   :code: yaml
 
 Custom interpolations
 ^^^^^^^^^^^^^^^^^^^^^
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -316,7 +316,7 @@ variable 'DB_PASSWORD' is not defined.
     ... })
     >>> print(cfg.database.password)
     12345
-    >>> cfg = cfg.copy()  # clear resolver catch
+    >>> OmegaConf.clear_cache(cfg) # clear resolver cache
     >>> os.environ["DB_PASSWORD"] = 'secret'
     >>> print(cfg.database.password)
     'secret'
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -306,11 +306,20 @@ Input yaml file:
     /home/omry
 
 It is possible to set a default value if the environment variable is missing. 
-In the following example, we set 'omry' as the default value if the environment
- variable USER is not defined.
+In the following example, we set '1234' as the default value if the environment
+variable 'DB_PASSWORD' is not defined.
 
-.. include:: env_default_interpolation.yaml
-   :code: yaml
+.. doctest::
+
+    >>> cfg = OmegaConf.create({
+    ...       'database': {'password': '${env:DB_PASSWORD,12345}'}
+    ... })
+    >>> print(cfg.database.password)
+    12345
+    >>> cfg = cfg.copy()  # clear resolver catch
+    >>> os.environ["DB_PASSWORD"] = 'secret'
+    >>> print(cfg.database.password)
+    'secret'
 
 Custom interpolations
 ^^^^^^^^^^^^^^^^^^^^^
--- a/omegaconf/omegaconf.py
+++ b/omegaconf/omegaconf.py
@@ -97,7 +97,7 @@ def register_default_resolvers() -> None:
         try:
             return decode_primitive(os.environ[key])
         except KeyError:
-            if default:
+            if default is not None:
                 return decode_primitive(default)
             else:
                 raise KeyError("Environment variable '{}' not found".format(key))
--- a/omegaconf/omegaconf.py
+++ b/omegaconf/omegaconf.py
@@ -93,11 +93,14 @@ Resolver = Union[Resolver0, Resolver1, Resolver2, Resolver3]
 
 
 def register_default_resolvers() -> None:
-    def env(key: str) -> Any:
+    def env(key: str, default: str = None) -> Any:
         try:
             return decode_primitive(os.environ[key])
         except KeyError:
-            raise KeyError("Environment variable '{}' not found".format(key))
+            if default:
+                return decode_primitive(default)
+            else:
+                raise KeyError("Environment variable '{}' not found".format(key))
 
     OmegaConf.register_resolver("env", env)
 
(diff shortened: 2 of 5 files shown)


omegaconf/omegaconf.py
```diff
@@ -93,11 +93,14 @@ def __call__(self, __x1: str, __x2: str, __x3: str) -> Any:
 
 
 def register_default_resolvers() -> None:
-    def env(key: str) -> Any:
+    def env(key: str, default: str = None) -> Any:
         try:
             return decode_primitive(os.environ[key])
         except KeyError:
-            raise KeyError("Environment variable '{}' not found".format(key))
+            if default:
```
omry:
```suggestion
            if default is not None:
```

omry:
Oh, and also update the documentation to reflect that.

kfollesdal:
I have then added tests and documentation.

docs/source/usage.rst
```diff
@@ -305,6 +305,12 @@ Input yaml file:
     >>> print(conf.user.home)
     /home/omry
 
+It is possible to set a default value if the environment variable is missing. 
+In the following example, we set 'omry' as the default value if the environment
+ variable USER is not defined.
```
omry: Since the USER environment variable is always defined, it's not the best example.

omry: 1. USER is not the best example because it is define by default. use a different env name for this example.
2. Show a code example here, this is the actual documentation, it's more important than Jupyter notebook.
You can just make it all inline, no need for an additional file:
```python
cfg = OmegaConf.create({
   "database": {
      password: "${env:DB_PASSWORD,12345}"
   }
})

print(cfg.database.password)
12345

os.environ["DB_PASSWORD"] = secret
print(cfg.database.password)
secret
```

tests/test_interpolation.py
```diff
@@ -137,6 +137,16 @@ def test_env_interpolation_not_found() -> None:
...
 
+def test_env_default_interpolation_missing_env() -> None:
+    if os.getenv("foobar") is not None:
+        del os.environ["foobar"]
+    c = OmegaConf.create(dict(path="/test/${env:foobar,abc}"))
+    assert c.path == "/test/abc"
+
+def test_env_default_interpolation_env_exist() -> None:
+    os.environ["foobar"] = "1234"
+    c = OmegaConf.create(dict(path="/test/${env:foobar,abc}"))
```
omry: I am slowly switching to {} style for dicts. please change the new tests to use that instead of dict().

docs/source/usage.rst
kfollesdal: Example added. Had to put in `cfg = cfg.copy()` to clear resolver catch.

tests/test_interpolation.py
kfollesdal: Done.

docs/source/usage.rst
```diff
@@ -305,6 +305,21 @@ Input yaml file:
...
+variable 'DB_PASSWORD' is not defined.
+
+.. doctest::
+
+    >>> cfg = OmegaConf.create({
+    ...       'database': {'password': '${env:DB_PASSWORD,12345}'}
+    ... })
+    >>> print(cfg.database.password)
+    12345
+    >>> cfg = cfg.copy()  # clear resolver catch
```
omry: That's a hack and the fact it works is a bug, use set_cache:

```python
OmegaConf.set_cache(cfg, {}) # clear resolver cache
```

kfollesdal: Thanks, was looking after a method like that. But get an error, with the following code: 

```
import os

cfg = OmegaConf.create({
      'database': {'password': '${env:DB_PASSWORD,12345}'}
})
print(cfg.database.password)

OmegaConf.set_cache(cfg, {}) # clear resolver cache
os.environ["DB_PASSWORD"] = 'secret'
print(cfg.database.password)
```

```
  File "test.py", line 11, in <module>
    print(cfg.database.password)
  File "C:\Users\kf9762\Desktop\tmp\omegaconf\omegaconf\dictconfig.py", line 182, in __getattr__
    return self.get(key=key, default_value=None)
  File "C:\Users\kf9762\Desktop\tmp\omegaconf\omegaconf\dictconfig.py", line 197, in get
    return self._resolve_with_default(
  File "C:\Users\kf9762\Desktop\tmp\omegaconf\omegaconf\basecontainer.py", line 65, in _resolve_with_default
    value = self._resolve_single(value) if isinstance(value, str) else value
  File "C:\Users\kf9762\Desktop\tmp\omegaconf\omegaconf\basecontainer.py", line 412, in _resolve_single
    return BaseContainer._resolve_value(root, match.group(1), match.group(2))
  File "C:\Users\kf9762\Desktop\tmp\omegaconf\omegaconf\basecontainer.py", line 394, in _resolve_value
    value = resolver(root_node, inter_key)
  File "C:\Users\kf9762\Desktop\tmp\omegaconf\omegaconf\omegaconf.py", line 278, in caching
    cache = OmegaConf.get_cache(config)[name]
KeyError: 'env'
```

kfollesdal: Found solution. Have to use `OmegaConf.set_cache(cfg, defaultdict(dict, {}))`

kfollesdal: Propose to make an own clear catch method #175

kfollesdal: Done

docs/source/usage.rst
```diff
@@ -305,6 +305,21 @@ Input yaml file:
     >>> print(conf.user.home)
     /home/omry
 
+It is possible to set a default value if the environment variable is missing. 
+In the following example, we set '1234' as the default value if the environment
+variable 'DB_PASSWORD' is not defined.
```
omry:
```suggestion
You can specify a default value to use in case the environment variable is not defined.
The following example sets `12345` as the the default value for the `DB_PASSWORD` environment variable.
```

tests/test_interpolation.py
```diff
@@ -282,6 +292,19 @@ def test_copy_cache() -> None:
...
+    cfg = OmegaConf.create({
+          'database': {'password': '${env:DB_PASSWORD,12345}'}
+    })
+    fresh = cfg.__dict__
+    assert cfg.database.password == 12345
+    OmegaConf.clear_cache(cfg)
+    assert cfg.__dict__ == fresh
+    os.environ['DB_PASSWORD'] = 'abcdef'
+    assert cfg.database.password == 'abcdef'
+    del os.environ['DB_PASSWORD']
```
omry: see the other tests testing the cache, in particular `test_resolver_cache_1` and `test_resolver_cache_2`.
They are much more compact.

kfollesdal: Updated with a more compact test.

omry:
awesome, thanks!

I would like to learn more about your use case at some point, can you join the Hydra chat (https://hydra-framework.zulipchat.com/) ?

omry:
interesting that this failes the CI, somehow it didn't trigger.
next time do run tests locally too.

omry:
to get a full CI run, run `nox` locally.
you can see the session names with `nox -l` and run a specific session with `nox -s NAME`.

PR #257, opened 2020-05-26
Report errors when loading YAML files with duplicate keys

jgehring:
This adds a custom constructor to the YAML loader created in
get_yaml_loader() so that duplicate keys in the YAML file result in a
PyYAML ConstructorError.

Duplicate keys are not allowed as per the YAML spec but are silently
ignored by PyYAML. This is a potential source of hard-to-detect errors
due to misconfigurations. It has been reported several times in the
PyYAML issues, e.g. https://github.com/yaml/pyyaml/issues/165. The
specific fix here was proposed at https://gist.github.com/pypt/94d747fe5180851196eb.

### DIFF
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -41,8 +41,10 @@ def isint(s: str) -> bool:
 def get_yaml_loader() -> Any:
     # Custom constructor that checks for duplicate keys
     # (from https://gist.github.com/pypt/94d747fe5180851196eb)
-    def no_duplicates_constructor(loader, node, deep=False):
-        mapping = {}
+    def no_duplicates_constructor(
+        loader: yaml.Loader, node: yaml.Node, deep: bool = False
+    ) -> Any:
+        mapping: Dict[str, Any] = {}
         for key_node, value_node in node.value:
             key = loader.construct_object(key_node, deep=deep)
             value = loader.construct_object(value_node, deep=deep)
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -39,6 +39,23 @@ def isint(s: str) -> bool:
 
 
 def get_yaml_loader() -> Any:
+    # Custom constructor that checks for duplicate keys
+    # (from https://gist.github.com/pypt/94d747fe5180851196eb)
+    def no_duplicates_constructor(loader, node, deep=False):
+        mapping = {}
+        for key_node, value_node in node.value:
+            key = loader.construct_object(key_node, deep=deep)
+            value = loader.construct_object(value_node, deep=deep)
+            if key in mapping:
+                raise yaml.constructor.ConstructorError(
+                    "while constructing a mapping",
+                    node.start_mark,
+                    f"found duplicate key {key}",
+                    key_node.start_mark,
+                )
+            mapping[key] = value
+        return loader.construct_mapping(node, deep)
+
     loader = yaml.SafeLoader
     loader.add_implicit_resolver(
         "tag:yaml.org,2002:float",
@@ -62,6 +79,9 @@ def get_yaml_loader() -> Any:
         ]
         for key, resolvers in loader.yaml_implicit_resolvers.items()
     }
+    loader.add_constructor(
+        yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG, no_duplicates_constructor
+    )
     return loader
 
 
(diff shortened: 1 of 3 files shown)


tests/test_serialization.py
```diff
@@ -105,3 +105,26 @@ def test_pickle_list() -> None:
...
         c1 = pickle.load(fp)
         assert c == c1
+
+
+def test_load_duplicate_keys_top() -> None:
+    from yaml.constructor import ConstructorError
+
+    try:
+        with tempfile.NamedTemporaryFile(delete=False) as fp:
+            fp.write("a:\n  b: 1\na:\n  b: 2\n".encode("utf-8"))
```
omry, line 115: Switch tests to use """ strings to to make them easier to understand.
"""
```suggestion
            content = """
a:
  b: 1
a:
  b: 2
"""
            fp.write(content.encode("utf-8"))
```

jgehring:
I guess "bugfix" is the most appropriate label for this PR, but I'm not totally sure.

omegaconf/_utils.py
```diff
@@ -39,6 +39,25 @@ def isint(s: str) -> bool:
...
+        for key_node, value_node in node.value:
+            key = loader.construct_object(key_node, deep=deep)
+            value = loader.construct_object(value_node, deep=deep)
+            if key in mapping:
+                raise yaml.constructor.ConstructorError(
+                    "while constructing a mapping",
+                    node.start_mark,
+                    f"found duplicate key {key}",
+                    key_node.start_mark,
+                )
```
omry, line 57: To avoid breaking compatibility, I think I prefer a warning instead.
What do you think about something like this?
```
warnings.warn(category=UserWarning, message=...)
```
Also, the warning message is very technical, we can make it clearer.
Maybe by inverting the order: "Duplicate key found while constructing a mapping" (along with embedding anything helpful in there).

tests/test_serialization.py
```diff
@@ -105,3 +105,39 @@ def test_pickle_list() -> None:
...
+        with tempfile.NamedTemporaryFile(delete=False) as fp:
+            content = """
+a:
+  b: 1
+a:
+  b: 2
+"""
+            fp.write(content.encode("utf-8"))
+        with pytest.raises(ConstructorError):
+            OmegaConf.load(fp.name)
```
omry, line 123: this becomes:
```
with pytest.warns(...)
```

omegaconf/_utils.py
jgehring: The format here is specific to `ConstructionError`. The final error reported to the user becomes:
```
yaml.constructor.ConstructorError: while constructing a mapping
  in "/var/folders/37/wpf0jmtx63b0spk4ps1pfbhr0000gn/T/tmp6tqvg9hq", line 3, column 3
found duplicate key b
  in "/var/folders/37/wpf0jmtx63b0spk4ps1pfbhr0000gn/T/tmp6tqvg9hq", line 5, column 3
```

jgehring: I'd argue for making this an error as it seems to be a violation of the YAML spec. A warning can be missed, and in particular when used from Hydra (where it's normal to override specific configuration keys from e.g. the command-line) it seems like it's really easy to introduce misconfigs this way. My specific problem was something like this:

```!yaml
foo:
    bar: 1
    baz: 2
    items:
        - bla
        - blubb

# other items...

# intention: change foo.baz only
foo:
    baz: 3
```

With the default behavior `yaml.load()`, you'll end up with `{'foo': {'baz': 3}}`.

omry: > 
> 
> The format here is specific to `ConstructionError`. The final error reported to the user becomes:
> 
> ```
> yaml.constructor.ConstructorError: while constructing a mapping
>   in "/var/folders/37/wpf0jmtx63b0spk4ps1pfbhr0000gn/T/tmp6tqvg9hq", line 3, column 3
> found duplicate key b
>   in "/var/folders/37/wpf0jmtx63b0spk4ps1pfbhr0000gn/T/tmp6tqvg9hq", line 5, column 3
> ```

What does it look like if you construct from a string?

omry:
Alright, error it is!

omegaconf/_utils.py
jgehring: In that case you'll also get excerpts from the actual string:
```
yaml.constructor.ConstructorError: while constructing a mapping
  in "<unicode string>", line 2, column 1:
    a:
    ^
found duplicate key a
  in "<unicode string>", line 4, column 1:
    a:
    ^
```

PR #277, opened 2020-06-21
[bug] fix #248 List config negative indices

### LINKED ISSUE #248: [bug] ListConfig doesn't support negative indices in basic list operations
```
a = OmegaConf.create({"a": [1, 2, 3, 4, 5, 6, 7, 8]})
b = [1, 2, 3, 4, 5, 6, 7, 8]
print(b[1:-1]) # Works fine
print(a.a[1:-1]) # Throws error
print(b[None:None:-1]) # Works fine
print(a.a[None:None:-1])  # Throws error
```

pereman2:


.pre-commit-config.yaml
```diff
@@ -14,7 +14,7 @@ repos:
     rev: 3.7.9
     hooks:
       - id: flake8
-        additional_dependencies: [-e, git+git://github.com/pycqa/pyflakes.git@1911c20#egg=pyflakes]
+        additional_dependencies: [-e, 'git+git://github.com/pycqa/pyflakes.git@1911c20#egg=pyflakes']
```
omry, line 17: yaml can handle all sorts of crazy strings. did it actually cause a problem?

tests/test_basic_ops_list.py
```diff
@@ -475,14 +475,25 @@ def test_set_with_invalid_key() -> None:
...
+        (OmegaConf.create([1, 2, 3]), slice(None, -1, None), [1, 2]),
+        (OmegaConf.create([1, 2, 3]), slice(None, None, 1), [1, 2, 3]),
+        (OmegaConf.create([1, 2, 3]), slice(None, None, -1), [3, 2, 1]),
+        (OmegaConf.create([1, 2, 3]), slice(1, None, -2), [2]),
+        (OmegaConf.create([1, 2, 3]), slice(None, 1, -2), [3]),
+        (OmegaConf.create([1, 2, 3]), slice(1, 3, -1), []),
+        (OmegaConf.create([1, 2, 3]), slice(3, 1, -1), [3]),
     ],
 )
 def test_getitem(lst: Any, idx: Any, expected: Any) -> None:
```
omry, line 491: Please create a new test for it.
It should take a tandard list ([1,2,3]) and test that the behavior of all the slicing variants is consistent between the standard list and ListConfig.

tests/test_basic_ops_list.py
```diff
@@ -475,14 +475,25 @@ def test_set_with_invalid_key() -> None:
...
     if isinstance(expected, type):
         with pytest.raises(expected):
             lst.__getitem__(idx)
     else:
-        lst.__getitem__(idx) == expected
+        assert lst.__getitem__(idx) == expected
```
omry, line 496: nice catch

.pre-commit-config.yaml
pereman2: Whenever a I committed without quotes it would give me a syntax error pointing to the colons. I added them and so it would work.

tests/test_basic_ops_list.py
pereman2: Do you mean dividing it into another test with the test cases already used ?

.pre-commit-config.yaml
omry: what would give you the syntax error?

tests/test_basic_ops_list.py
omry: I basically want to ensure getitem behavior is consistent for your new implementation with the behavior of the standard list.

find a way to do it while still testing what is already there :)

.pre-commit-config.yaml
pereman2: Executing without quotes this gives an error.
`git commit -m "try without quotes"`

An error has occurred: InvalidConfigError: 
==> File .pre-commit-config.yaml
=====> while scanning a plain scalar
  in "<unicode string>", line 17, column 39
found unexpected ':'
  in "<unicode string>", line 17, column 46

omry: pre-commit seems to be installing the new environment every time I run it which seems like another bug.
I haven't used it in a while because it became too much pain.

You can upgrade the dev dependncies with:
```
pip install -r requirements/dev.txt  --upgrade
```

pereman2: It keeps going anyways. Maybe because of the OS(macos)?

tests/test_basic_ops_list.py
```diff
@@ -482,7 +482,30 @@ def test_getitem(lst: Any, idx: Any, expected: Any) -> None:
...
+        (slice(None, 1, -2)),
+        (slice(1, 3, -1)),
+        (slice(3, 1, -1)),
+    ],
+)
+def test_getitem_slice(sli: slice) -> None:
+    lst = [1, 2, 3]
+    olst = OmegaConf.create([1, 2, 3])
+    expected = lst[sli.start : sli.stop : sli.step]
+    assert olst.__getitem__(sli) == expected
```
omry, line 508: this looks great. I just want to get to the bottom of the quote thing.
probably indicating of some issue on your end.

.pre-commit-config.yaml
omry: can you jump into the Hydra/OmegaConf chat?
https://hydra-framework.zulipchat.com/

omry:
Thanks!

PR #281, opened 2020-07-01
Fix #279 bug optional tuple.

### LINKED ISSUE #279: Bugs on Optional/Tuple type: TypeError: issubclass() arg1 must be a class
Example:

```python
from typing import Optional, Tuple

@dataclass
class A:
    foo: Optional[Tuple[int]] = None
    bar: Tuple[int] = (1, 2, 3)

OmegaConf.structured(A)
```

Stacktrace:
```
.../lib/python3.7/site-packages/omegaconf/omegaconf.py in _node_wrap(type_, parent, is_optional, value, key)
    638     elif type_ == Any or type_ is None:
    639         node = AnyNode(value=value, key=key, parent=parent, is_optional=is_optional)
--> 640     elif issubclass(type_, Enum):
    641         node = EnumNode(
    642             enum_type=type_,

TypeError: issubclass() arg 1 must be a class
```

This happens when an argument to a typing type, e.g. `Tuple`.

pereman2:
Tuple weren't checked when creating from a dataclass. 
Added additional tests so it covers optional lists and optional tuples.

### DIFF
--- a/news/279.bugfix
+++ b/news/279.bugfix
@@ -1,1 +1,1 @@
-Tuples weren't checked when creating a List container.
+Optional[Tuple] types are now supported within DictConfigs.
--- a/news/279.bugfix
+++ b/news/279.bugfix
@@ -1,0 +1,1 @@
+Tuples weren't checked when creating a List container.
--- a/news/279.bugfix
+++ b/news/279.bugfix
@@ -1,1 +1,1 @@
-Optional[Tuple] types are now supported within DictConfigs.
+Optional[Tuple] types are now supported as type annotation in Structured Configs.
--- a/tests/structured_conf/data/attr_classes.py
+++ b/tests/structured_conf/data/attr_classes.py
@@ -257,6 +257,12 @@ class FrozenClass:
 
 
 @attr.s(auto_attribs=True)
+class ContainsFrozen:
+    x: int = 10
+    frozen: FrozenClass = FrozenClass()
+
+
+@attr.s(auto_attribs=True)
 class WithTypedList:
     list: List[int] = [1, 2, 3]
 
--- a/tests/structured_conf/data/attr_classes.py
+++ b/tests/structured_conf/data/attr_classes.py
@@ -304,6 +304,15 @@ class ListExamples:
 
 
 @attr.s(auto_attribs=True)
+class TupleExamples:
+    any: Tuple[Any, Any] = (1, "foo")
+    ints: Tuple[int, int] = (1, 2)
+    strings: Tuple[str, str] = ("foo", "bar")
+    booleans: Tuple[bool, bool] = (True, False)
+    colors: Tuple[Color, Color] = (Color.RED, Color.GREEN)
+
+
+@attr.s(auto_attribs=True)
 class DictExamples:
     any: Dict[str, Any] = {"a": 1, "b": "foo"}
     ints: Dict[str, int] = {"a": 10, "b": 20}
--- a/tests/structured_conf/data/attr_classes.py
+++ b/tests/structured_conf/data/attr_classes.py
@@ -222,6 +222,20 @@ class StringOptional:
 
 
 @attr.s(auto_attribs=True)
+class ListOptional:
+    with_default: Optional[List[int]] = [1, 2, 3]
+    as_none: Optional[List[int]] = None
+    not_optional: List[int] = [1, 2, 3]
+
+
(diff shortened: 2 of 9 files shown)


tests/structured_conf/data/attr_classes.py
```diff
@@ -221,6 +221,20 @@ class StringOptional:
...
 
+@attr.s(auto_attribs=True)
+class ListOptional:
+    with_default: Optional[List[int]] = [1, 2, 3]
+    as_none: Optional[List[int]] = None
+    not_optional: List[int] = [1, 2, 3]
+
+
+@attr.s(auto_attribs=True)
+class TupleOptional:
```
omry, line 232: Since Tuple was not supported before, also add non optional tests for Tuple.

Stonesjtu:
Any idea on how to make a tuple using yaml?

omry:
No such thing.
test it with Structured Configs.
there there is a [big test suite](https://github.com/omry/omegaconf/tree/master/tests/structured_conf), you added data for it but the tests.

the tests inside for an example.

tests/structured_conf/data/attr_classes.py
pereman2: > I think you missed adding an actual test that interacts with your new ListOptional and TupleOptional classes.

I added those two classes on 'test_optional'

tests/structured_conf/test_structured_config.py
```diff
@@ -380,6 +380,8 @@ def test_interpolation(self, class_type: str) -> Any:
             "IntegerOptional",
             "FloatOptional",
             "StringOptional",
+            "ListOptional",
+            "TupleOptional",
```
pereman2, line 384: These two are now tested in here.

tests/structured_conf/test_structured_config.py
```diff
@@ -535,6 +537,52 @@ def test_any(name: str) -> None:
             Color.BLUE,
         ]
 
+    def test_tuple_examples(self, class_type: str) -> None:
```
pereman2, line 540: This new test could be merged with test_list_examples. Do you prefer it this way so its separated?

tests/structured_conf/data/attr_classes.py
```diff
@@ -303,6 +303,15 @@ class ListExamples:
     colors: List[Color] = [Color.RED, Color.GREEN]
 
 
+@attr.s(auto_attribs=True)
```
omry, line 306: nice.

tests/structured_conf/test_structured_config.py
omry: If it's a near duplicate try to merge them. (it should be because underneeth Tuple and List are both converted to the same ListConfig.

pereman2: @omry done

tests/structured_conf/test_structured_config.py
```diff
@@ -493,95 +493,57 @@ def test_typed_list_value_error(self, class_type: str) -> None:
 
     def test_list_examples(self, class_type: str) -> None:
         module: Any = import_module(class_type)
-        conf = OmegaConf.structured(module.ListExamples)
+        conf0 = OmegaConf.structured(module.ListExamples)
+        conf1 = OmegaConf.structured(module.TupleExamples)
```
omry, line 497: try to use `@pytest.mark.parametrize()` on this function and pass in the different class name as well as the class type.

omry:
Great.
Since the actual change to the test is now reflects the evolution of your changes, can you just undo the changes test changes and redo it with a minimal change? (Although your changes are fine, it makes it harder to review that no unintentional changes snuck in).

omry:
Good.
Please update the structured configs documentation to reflect this and add a news fragment in news.

news/279.bugfix
```diff
@@ -0,0 +1 @@
+Tuples weren't checked when creating a List container.
```
omry, line 1: A good news fragment will tell the user what bug was fixed , or what feature was added.
In this case, it should say something about Optional[Tuple] now being supported in DictConfig.

news/279.bugfix
```diff
@@ -0,0 +1 @@
+Optional[Tuple] types are now supported within DictConfigs.
```
omry, line 1: Optional[Tuple[...]] types are now a supported annotation type when defining Structured Configs.

omry:
Thanks!

PR #285, opened 2020-07-08
DictConfig copy minor refactor

pereman2:
Since it is a shallow copy its contents should't be copied, instead we can add the same attributes back.


omry:
1. what is the context of this change?
2. is this solving a bug? which one?
3. where is the news fragment?
4. where are the tests?

pereman2:
I've change my mind, what i did messed up the parents of the copy. I'll close the PR

PR #299, opened 2020-07-14
Fix #296 Float/Boolean/Int represented as string, pretty.

### LINKED ISSUE #296: Float as string transforms into float type when creating config.
How to reproduce:
`OmegaConf.create(OmegaConf.create("abc: '10e2'").pretty())`

Output:
`abc: 1000.0`

pereman2:
When representing Float/Boolean/Int as string and then printing with pretty causes the string to be printed without quotes which can cause problems.
Added tests in both DictConfig and ListConfig.
Added news file.
Closes #296 .

### DIFF
--- a/news/296.bugfix
+++ b/news/296.bugfix
@@ -1,0 +1,1 @@
+bool, ints and floats represented as string are now formatted with surrounding quotes on pretty()
--- a/news/296.bugfix
+++ b/news/296.bugfix
@@ -1,1 +1,1 @@
-bool, ints and floats represented as string are now formatted with surrounding quotes on pretty()
+strings containing valid ints and floats represented are converted to quoted strings instead of the primitives in pretty()
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -57,7 +57,7 @@ YAML_BOOL_TYPES = [
 ]
 
 
-class DumperWrapper(yaml.Dumper):  # type: ignore
+class OmegaConfDumper(yaml.Dumper):  # type: ignore
     pass
 
 
@@ -101,10 +101,10 @@ def get_yaml_loader() -> Any:
             mapping[key] = value
         return loader.construct_mapping(node, deep)
 
-    class SafeLoaderWrapper(yaml.SafeLoader):  # type: ignore
+    class OmegaConfLoader(yaml.SafeLoader):  # type: ignore
         pass
 
-    loader = SafeLoaderWrapper
+    loader = OmegaConfLoader
     loader.add_implicit_resolver(
         "tag:yaml.org,2002:float",
         re.compile(
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -58,27 +58,20 @@ YAML_BOOL_TYPES = [
 
 
 class OmegaConfDumper(yaml.Dumper):  # type: ignore
-    pass
+    def str_representer(dumper: yaml.Dumper, data: str) -> yaml.ScalarNode:
+        with_quotes = yaml_is_bool(data) or is_int(data) or is_float(data)
+        return dumper.represent_scalar(
+            yaml.resolver.BaseResolver.DEFAULT_SCALAR_TAG,
+            data,
+            style=("'" if with_quotes else None),
+        )
 
 
-def isbool(b: str) -> bool:
+OmegaConfDumper.add_representer(str, OmegaConfDumper.str_representer)
+
+
+def yaml_is_bool(b: str) -> bool:
     return b in YAML_BOOL_TYPES
-
-
-def isfloat(f: str) -> bool:
-    try:
-        float(f)
-        return True
-    except ValueError:
-        return False
-
-
-def isint(s: str) -> bool:
-    try:
-        int(s)
-        return True
-    except ValueError:
-        return False
 
 
 def get_yaml_loader() -> Any:
@@ -361,25 +354,28 @@ def get_value_kind(value: Any, return_match_list: bool = False) -> Any:
         return ret(ValueKind.STR_INTERPOLATION)
 
 
+def is_bool(st: str) -> bool:
+    st = str.lower(st)
+    return st == "true" or st == "false"
+
+
+def is_float(st: str) -> bool:
+    try:
+        float(st)
+        return True
+    except ValueError:
+        return False
+
+
+def is_int(st: str) -> bool:
+    try:
+        int(st)
+        return True
+    except ValueError:
+        return False
+
+
 def decode_primitive(s: str) -> Any:
-    def is_bool(st: str) -> bool:
-        st = str.lower(st)
-        return st == "true" or st == "false"
-
-    def is_float(st: str) -> bool:
-        try:
-            float(st)
-            return True
-        except ValueError:
-            return False
-
-    def is_int(st: str) -> bool:
-        try:
-            int(st)
-            return True
-        except ValueError:
-            return False
-
     if is_bool(s):
         return str.lower(s) == "true"
 
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -30,6 +30,7 @@ except ImportError:  # pragma: no cover
     attr = None  # type: ignore # pragma: no cover
 
 
+# source: https://yaml.org/type/bool.html
 YAML_BOOL_TYPES = [
     "y",
     "Y",
@@ -54,6 +55,10 @@ YAML_BOOL_TYPES = [
     "Off",
     "OFF",
 ]
+
+
+class DumperWrapper(yaml.Dumper):  # type: ignore
+    pass
 
 
 def isbool(b: str) -> bool:
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -58,6 +58,8 @@ YAML_BOOL_TYPES = [
 
 
 class OmegaConfDumper(yaml.Dumper):  # type: ignore
+    str_representer_added = False
+
     def str_representer(dumper: yaml.Dumper, data: str) -> yaml.ScalarNode:
         with_quotes = yaml_is_bool(data) or is_int(data) or is_float(data)
         return dumper.represent_scalar(
@@ -67,7 +69,11 @@ class OmegaConfDumper(yaml.Dumper):  # type: ignore
         )
 
 
-OmegaConfDumper.add_representer(str, OmegaConfDumper.str_representer)
+def get_omega_conf_dumper() -> Type[OmegaConfDumper]:
+    if not OmegaConfDumper.str_representer_added:
+        OmegaConfDumper.add_representer(str, OmegaConfDumper.str_representer)
+        OmegaConfDumper.str_representer_added = True
+    return OmegaConfDumper
 
 
 def yaml_is_bool(b: str) -> bool:
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -30,6 +30,44 @@ except ImportError:  # pragma: no cover
     attr = None  # type: ignore # pragma: no cover
 
 
+YAML_BOOL_TYPES = [
+    "y",
+    "Y",
+    "yes",
+    "Yes",
+    "YES",
+    "n",
+    "N",
+    "no",
+    "No",
+    "NO",
+    "true",
+    "True",
+    "TRUE",
+    "false",
+    "False",
+    "FALSE",
+    "on",
+    "On",
+    "ON",
+    "off",
+    "Off",
+    "OFF",
+]
+
+
+def isbool(b: str) -> bool:
+    return b in YAML_BOOL_TYPES
+
+
+def isfloat(f: str) -> bool:
+    try:
+        float(f)
+        return True
+    except ValueError:
+        return False
+
+
 def isint(s: str) -> bool:
     try:
         int(s)
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -60,6 +60,7 @@ YAML_BOOL_TYPES = [
 class OmegaConfDumper(yaml.Dumper):  # type: ignore
     str_representer_added = False
 
+    @staticmethod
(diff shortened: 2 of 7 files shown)


omry:
1. Test on multiple cases for boolean values (TRUE, True, true. There are a handful of other boolean constants in yaml)
2. Config this problem even exists for booleans
3. pyyaml dump uses single quote for strings, do that too.
4. How do you differentiate between 1.0 and "1.0" in the config? both will parse to float. I think your current solution will convert the real float to a string as well. (add a test for this).

pereman2:
> 1. Test on multiple cases for boolean values (TRUE, True, true. There are a handful of other boolean constants in yaml)

Done

> 2. Config this problem even exists for booleans

I tried it with booleans and it worked the same as floats

> 3. pyyaml dump uses single quote for strings, do that too.

Done.

> 4. How do you differentiate between 1.0 and "1.0" in the config? both will parse to float. I think your current solution will convert the real float to a string as well. (add a test for this).

pyyaml provides a method to change how a certain datatype is represented, in this case str, is the only one changed thus, a 1.0 will work as usual.

omegaconf/basecontainer.py
```diff
@@ -224,9 +224,37 @@ def pretty(self, resolve: bool = False, sort_keys: bool = False) -> str:
...
+                return is_float
+
+            with_quotes = is_bool(data) or is_int(data) or is_float(data)
+            return dumper.represent_scalar(
+                yaml.resolver.BaseResolver.DEFAULT_SCALAR_TAG,
+                data,
+                style=("single-quoted" if with_quotes else None),
+            )
+
+        dumper.add_representer(str, str_representer)
```
pereman2, line 274: This is the method that changes the representer of a certain datatype.

omegaconf/basecontainer.py
```diff
@@ -228,7 +228,30 @@ def pretty(self, resolve: bool = False, sort_keys: bool = False) -> str:
 
         def str_representer(dumper: yaml.Dumper, data: str) -> yaml.ScalarNode:
             def is_bool(data: str) -> bool:
-                return data in ["True", "False"]
+                return data in [
```
omry, line 231: Looks like this is from https://yaml.org/type/bool.html ?

1. Keep the same order.
2. Add a link to it in a comment.

tests/test_basic_ops_list.py
```diff
@@ -50,11 +50,56 @@ def test_pretty_list_unicode() -> None:
...
+        "FALSE",
+        "on",
+        "On",
+        "ON",
+        "off",
+        "Off",
+        "OFF",
+    ],
+)
+def test_pretty_string_boolean(yaml_bool: str) -> None:
```
omry, line 89: refactror your tests, you should have have that list more than once in the tests.

tests/test_basic_ops_list.py
```diff
@@ -50,11 +50,56 @@ def test_pretty_list_unicode() -> None:
...
+        yaml_bool
+    )
+    assert c.pretty() == expected
+
+
+def test_pretty_string_int() -> None:
+    c = OmegaConf.create(["1", 1])
+    expected = """- '1'
+- 1
 """
```
omry: easier to understand it with \n.
something lke:
assert c.pretty() == "'- 1'\n- 1\n"

tests/test_basic_ops_list.py
```diff
@@ -50,11 +50,56 @@ def test_pretty_list_unicode() -> None:
...
+        "off",
+        "Off",
+        "OFF",
+    ],
+)
+def test_pretty_string_boolean(yaml_bool: str) -> None:
+    c = OmegaConf.create([yaml_bool, 1])
+    expected = """- '{0}'
+- 1
+""".format(
```
omry, line 93: fstrings.

omegaconf/basecontainer.py
```diff
@@ -245,7 +268,7 @@ def is_float(data: str) -> bool:
             return dumper.represent_scalar(
                 yaml.resolver.BaseResolver.DEFAULT_SCALAR_TAG,
                 data,
-                style=("single-quoted" if with_quotes else None),
+                style=("'" if with_quotes else None),
```
omry, line 271: What is with_quotes? where is it coming from?

omegaconf/basecontainer.py
```diff
@@ -228,7 +228,30 @@ def pretty(self, resolve: bool = False, sort_keys: bool = False) -> str:
...
+                    "off",
+                    "Off",
+                    "OFF",
+                    "y",
+                    "Y",
+                    "n",
+                    "N",
+                ]
 
             def is_int(data: str) -> bool:
```
omry, line 256: is -10 an int?
```
"-1".isdecimal()
False
```

omegaconf/basecontainer.py
omry: move this logic somewhere else. _utils?

omegaconf/basecontainer.py
omry: "-1".isdecimal()
False

news/296.bugfix
```diff
@@ -0,0 +1 @@
+bool, ints and floats represented as string are now formatted with surrounding quotes on pretty()
```
omry, line 1:
```suggestion
strings containing valid ints and floats represented are converted to quoted strings instead of the primitives in pretty()
```

omegaconf/basecontainer.py
```diff
@@ -224,9 +224,60 @@ def pretty(self, resolve: bool = False, sort_keys: bool = False) -> str:
...
+            def is_int(data: str) -> bool:
+                return data.isdecimal()
+
+            def is_float(data: str) -> bool:
+                is_float = True
+                try:
+                    float(data)
+                except ValueError:
+                    is_float = False
+                return is_float
```
omry, line 265:
```suggestion
            def is_float(data: str) -> bool:
                is_float = True
                try:
                    float(data)
                    return True
                except ValueError:
                    return False
```

tests/test_basic_ops_list.py
```diff
@@ -50,6 +50,60 @@ def test_pretty_list_unicode() -> None:
...
+        yaml_bool
+    )
+    assert c.pretty() == expected
+
+
+def test_pretty_string_int() -> None:
+    c = OmegaConf.create(["1", 1])
+    expected = """- '1'
+- 1
+"""
```
omry, line 103: string with newlines like above.

omegaconf/basecontainer.py
pereman2: Since bool, float and int are the ones causing the problem we are quoting them if they are strings. If a string is not one of them its not necessary to quote.

omegaconf/basecontainer.py
pereman2: True mb

omegaconf/basecontainer.py
omry: I understand, my question is about the usage of the variable `with_quotes`. where is it defined?

omegaconf/_utils.py
```diff
@@ -30,6 +30,44 @@
     attr = None  # type: ignore # pragma: no cover
 
 
+YAML_BOOL_TYPES = [
```
omry, line 33: Add a comment with a link to the source.

omegaconf/basecontainer.py
omry: Ah, I missed the definition (it was folded).

omegaconf/basecontainer.py
omry: This method is changing the actual pyyaml dumper.
You already fixed on bug caused by doing something similar :)

define your own dumper (in _utils) and use it here with

```python
def dump(data, stream=None, Dumper=Dumper, **kwds):
```

omegaconf/basecontainer.py
```diff
@@ -225,9 +226,11 @@ def pretty(self, resolve: bool = False, sort_keys: bool = False) -> str:
         :param sort_keys: If True, will print dict keys in sorted order. default False.
         :return: A string containing the yaml representation.
         """
-        from omegaconf import OmegaConf
 
-        dumper = yaml.Dumper
+        return self._dump(resolve=resolve, sort_keys=sort_keys)
+
+    def _dump(self, stream=None, Dumper=DumperWrapper, **kwds) -> str:  # type: ignore
```
omry, line 232: 1. I just meant you should call dump passing a Dumper, is this the function _dump adding anything?
2. I don't like DumperWrapper, maybe OmegaConfDumper? (same for the other Wrapper, by the way).

omegaconf/basecontainer.py
```diff
@@ -237,13 +240,16 @@ def str_representer(dumper: yaml.Dumper, data: str) -> yaml.ScalarNode:
                 style=("'" if with_quotes else None),
             )
 
-        dumper.add_representer(str, str_representer)
-        container = OmegaConf.to_container(self, resolve=resolve, enum_to_str=True)
+        Dumper.add_representer(str, str_representer)
+        container = OmegaConf.to_container(
+            self, resolve=kwds["resolve"], enum_to_str=True
```
omry, line 245: so _dump is just making the rest of the code more complicated.
let's get rid of it, I didn't mean this.

omegaconf/basecontainer.py
pereman2: 1. Ah okay i thought you wanted a separate function.

2. Sure

omegaconf/_utils.py
```diff
@@ -30,6 +30,49 @@
...
+    "Off",
+    "OFF",
+]
+
+
+class OmegaConfDumper(yaml.Dumper):  # type: ignore
+    pass
+
+
+def isbool(b: str) -> bool:
```
omry, line 64: call this function is_yaml_bool

omegaconf/basecontainer.py
```diff
@@ -224,9 +228,23 @@ def pretty(self, resolve: bool = False, sort_keys: bool = False) -> str:
...
 
+        def str_representer(dumper: yaml.Dumper, data: str) -> yaml.ScalarNode:
+            with_quotes = isbool(data) or isint(data) or isfloat(data)
+            return dumper.represent_scalar(
+                yaml.resolver.BaseResolver.DEFAULT_SCALAR_TAG,
+                data,
+                style=("'" if with_quotes else None),
+            )
+
+        Dumper = OmegaConfDumper
```
omry, line 239: no point in assigning to a temp variable Dumper.

omegaconf/basecontainer.py
```diff
@@ -224,9 +228,23 @@ def pretty(self, resolve: bool = False, sort_keys: bool = False) -> str:
...
+        Dumper.add_representer(str, str_representer)
```
omry, line 240: This should happen in the contructor of OmegaConfDumper, not during Dump.

omegaconf/_utils.py
```diff
@@ -58,10 +101,10 @@ def no_duplicates_constructor(
...
 
-    class SafeLoaderWrapper(yaml.SafeLoader):  # type: ignore
+    class OmegaConfLoader(yaml.SafeLoader):  # type: ignore
         pass
 
-    loader = SafeLoaderWrapper
+    loader = OmegaConfLoader
     loader.add_implicit_resolver(
         "tag:yaml.org,2002:float",
         re.compile(
```
omry, line 110: this should happen in the constructor of OmegConfLoader.

omegaconf/_utils.py
omry: I think those functions are already defined in decode_primitive, refactor. (use those instead of what's in that function, except for the bool one which is a yaml_bool).

omegaconf/_utils.py
```diff
@@ -30,12 +30,48 @@
...
+    def str_representer(dumper: yaml.Dumper, data: str) -> yaml.ScalarNode:
+        with_quotes = yaml_is_bool(data) or is_int(data) or is_float(data)
+        return dumper.represent_scalar(
+            yaml.resolver.BaseResolver.DEFAULT_SCALAR_TAG,
+            data,
+            style=("'" if with_quotes else None),
+        )
+
+
+OmegaConfDumper.add_representer(str, OmegaConfDumper.str_representer)
```
omry, line 70: instead of calling this static function, can you just add it directly on the OmegaConfDumper object?

pereman2: In the pyyaml source code representers are added that way since pyyaml requires a Dumper class not a object.

pereman2: maybe i can define a function that creates that class, adds representer and returns the class

omry: If it wants a class, give it a class.
Can you define the class without calling this helper function?
it seems to me like the representers should be a list on the class.
can you add to it directoy?

omegaconf/_utils.py
```diff
@@ -67,7 +69,11 @@ def str_representer(dumper: yaml.Dumper, data: str) -> yaml.ScalarNode:
         )
 
 
-OmegaConfDumper.add_representer(str, OmegaConfDumper.str_representer)
+def get_omega_conf_dumper() -> Type[OmegaConfDumper]:
+    if not OmegaConfDumper.str_representer_added:
+        OmegaConfDumper.add_representer(str, OmegaConfDumper.str_representer)
+        OmegaConfDumper.str_representer_added = True
+    return OmegaConfDumper
```
omry, line 77: You can create a sort of static initializer as described here:
https://softwareengineering.stackexchange.com/questions/380210/why-isnt-there-a-static-initializer-in-python

pereman2: add_representer is a classmethod so it doesn't work, I've tried :/

omry: Sigh. I just commented here: https://github.com/yaml/pyyaml/issues/221#issuecomment-658485897

For now do the best you can.

omegaconf/_utils.py
omry: I am landing this because I am releasing another rc, but please follow up on this one in another diff.

OmegaConfLoader should work like OmegaConfDumper.

omry:
Well done :).

PR #307, opened 2020-07-16
OmegaConf.save() not accepting dataclasses fix

### LINKED ISSUE #287: Relax OmegaConf.save() to accept Any
It should happily accept either a real OmegaConf object or a structured config (something annotated as dataclass).
there should be a runtime check to make sure input is valid.

pereman2:
Closes #287 .
Added test.
Added news.

### DIFF
--- a/omegaconf/omegaconf.py
+++ b/omegaconf/omegaconf.py
@@ -35,6 +35,8 @@ from ._utils import (
     get_dict_key_value_types,
     get_list_element_type,
     get_type_of,
+    is_attr_class,
+    is_dataclass,
     is_dict_annotation,
     is_int,
     is_list_annotation,
@@ -261,7 +263,7 @@ class OmegaConf:
 
     @staticmethod
     def save(
-        config: Container, f: Union[str, pathlib.Path, IO[Any]], resolve: bool = False
+        config: Any, f: Union[str, pathlib.Path, IO[Any]], resolve: bool = False
     ) -> None:
         """
         Save as configuration object to a file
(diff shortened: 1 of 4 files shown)


omegaconf/omegaconf.py
```diff
@@ -261,14 +263,16 @@ def load(file_: Union[str, pathlib.Path, IO[Any]]) -> Union[DictConfig, ListConf
...
+        config: Any, f: Union[str, pathlib.Path, IO[Any]], resolve: bool = False
     ) -> None:
         """
         Save as configuration object to a file
         :param config: omegaconf.Config object (DictConfig or ListConfig).
         :param f: filename or file object
         :param resolve: True to save a resolved config (defaults to False)
         """
+        if is_dataclass(config) or is_attr_class(config):
+            config = OmegaConf.create(config)
```
omry, line 275: I was actually thinking about just accepting DictConfig annotated as something else here, but this works too.

tests/test_serialization.py
```diff
@@ -79,6 +81,45 @@ def test_save_load__from_filename(
...
+@dataclass
+class Dataclass:
+    number: int = 2
+    done: bool = True
+
+
+@attr.s(auto_attribs=True)
+class Attr:
+    number: int = 2
+    done: bool = True
```
omry, line 93: There are some test dataclasses in tests/__init__.py
Use something from there (and add an attr.s one to complete the test).

tests/test_serialization.py
```diff
@@ -79,6 +81,45 @@ def test_save_load__from_filename(
...
+
+
+@pytest.mark.parametrize(
+    "input_,resolve,expected,file_class",
+    [
+        (Dataclass, False, {"number": 2, "done": True}, str),
+        (Attr, False, {"number": 2, "done": True}, str),
```
omry, line 100: for completeness, test instance as well.(Dataclass()).
Also, use better names. Dataclas and Attr are terrible :)

PR #308, opened 2020-07-16
Implement nested interpolations

### LINKED ISSUE #100: [Feature Request] resolve individual resolver arguments
It would be nice to enable recursive interpolation. 

Example : 

```python
import omegaconf

OmegaConf.register_resolver("plus", lambda x,y: float(x)+float(y))

c = OmegaConf.create({
        "i": 1,
        'key1': '${plus:1,2}',
        'key2': '${plus:${i},2}'
 })
print(c.key1)
# 3.0
print(c.key2)
# '${plus:1,2}'
```

But one would hope to get `3.0` out

Note that 2 recursion would be enough to enable the user to create deeper interpolation if they need it (by calling an other key which had already been interpolated).

odelalleau:
This is to fix #100 

Doc and tests are missing, I'm aware of it, I wanted to first get feedback on the implementation itself, in case there'd be any major concern.

The main parsing logic is in `_evaluate_complex()`, it's actually not too complex since we can sequentially iterate on all characters in the string and evaluate interpolations in the order we encounter their end `}`. So the core logic is simple but there are still "details" to be addressed (see concerns below).

This is not a super efficient implementation: (i) there are many string copies in the process, and (ii) the worst case complexity is quadratic in case of deep nesting (Edit: not anymore, see comment below). That being said, I don't expect it to be a real issue in the large majority of practical use cases.

A few potential concerns:
1. Although all tests currently pass, this can actually break backward compatibility since for instance if you had `a: 1` as config, then `${${a}}` used to evaluate to the string`"${1}"`, when now it would crash when trying to resolve this new interpolation. Hopefully people were not relying on this!
2. My current implementation returns a `None` value instead of crashing in case of syntax error. This isn't good, but right now the only two flags to trigger a crash are `throw_on_missing` and `throw_on_resolution_failure`, because there could be no syntax error in the previous code (for instance `${a}_${a` would be resolved into `"1_${a"`). Should I add a `throw_on_parsing_failure` flag?
3. With my current implementation `${a` resolves to `"${a"` while `${a}_${a` triggers a parsing failure. This inconsistency isn't good => I was thinking of using my new interpolation logic whenever it's not a simple interpolation but there is still a `${` in the string to be parsed, which should address this issue. Thoughts?
4. I'm not planning to allow escaping of the special `${` and `}` tokens, although it probably wouldn't be that hard. Let me know if you think this is important.

### DIFF
--- a/docs/notebook/Tutorial.ipynb
+++ b/docs/notebook/Tutorial.ipynb
@@ -731,7 +731,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "You can add additional interpolation types using custom resolvers. This example creates a resolver that adds 10 the the given value."
+    "You can add additional interpolation types using custom resolvers, which take strings as inputs. This example creates a resolver that adds 10 to the given value (note the need to cast it to `int`)."
    ]
   },
   {
@@ -761,6 +761,40 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "## Nested interpolations"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "You can use interpolations within interpolations, for instance to perform an\n",
+    "operation over config variables:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 25,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "3\n"
+     ]
+    }
+   ],
+   "source": [
+    "OmegaConf.register_resolver(\"plus_int\", lambda x, y: int(x) + int(y))\n",
+    "conf = OmegaConf.create({\"a\": 1, \"b\": 2, \"a_plus_b\": \"${plus_int:${a},${b}}\"})\n",
+    "print(conf.a_plus_b)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "# Merging configurations\n",
     "Merging configurations enables the creation of reusable configuration files for each logical component instead of a single config file for each variation of your task.\n",
     "\n",
@@ -780,7 +814,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 25,
+   "execution_count": 26,
    "metadata": {
     "pycharm": {
      "name": "#%%\n"
@@ -807,7 +841,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 26,
+   "execution_count": 27,
    "metadata": {
     "pycharm": {
      "name": "#%%\n"
@@ -831,7 +865,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 27,
+   "execution_count": 28,
    "metadata": {
     "pycharm": {
      "name": "#%%\n"
@@ -884,7 +918,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.8.1"
+   "version": "3.8.3"
   },
   "pycharm": {
    "stem_cell": {
--- a/docs/notebook/Tutorial.ipynb
+++ b/docs/notebook/Tutorial.ipynb
@@ -795,6 +795,40 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "Another typical use case is to to dynamically select a sub-config:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 26,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Default: cfg.plan = plan A\n",
+      "After selecting plan B: cfg.plan = plan B\n"
+     ]
+    }
+   ],
+   "source": [
+    "cfg = OmegaConf.create(\"\"\"\n",
+    "plans:\n",
+    "    A: plan A\n",
+    "    B: plan B\n",
+    "selected_plan: A\n",
+    "plan: ${plans.${selected_plan}}\n",
+    "\"\"\")\n",
+    "print(f\"Default: cfg.plan = {cfg.plan}\")\n",
+    "cfg.selected_plan = \"B\"\n",
+    "print(f\"After selecting plan B: cfg.plan = {cfg.plan}\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "# Merging configurations\n",
     "Merging configurations enables the creation of reusable configuration files for each logical component instead of a single config file for each variation of your task.\n",
     "\n",
@@ -814,7 +848,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 26,
+   "execution_count": 27,
    "metadata": {
     "pycharm": {
      "name": "#%%\n"
@@ -841,7 +875,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 27,
+   "execution_count": 28,
    "metadata": {
     "pycharm": {
      "name": "#%%\n"
@@ -865,7 +899,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 28,
+   "execution_count": 29,
    "metadata": {
     "pycharm": {
      "name": "#%%\n"
--- a/docs/notebook/Tutorial.ipynb
+++ b/docs/notebook/Tutorial.ipynb
@@ -521,7 +521,7 @@
     "\n",
     "OmegaConf support variable interpolation, Interpolations are evaluated lazily on access.\n",
     "\n",
-    "### Config node interpolation"
+    "## Config node interpolation"
    ]
   },
   {
@@ -620,7 +620,41 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### Environment variable interpolation\n",
+    "Interpolations may be nested, enabling more advanced behavior like dynamically selecting a sub-config:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 20,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Default: cfg.plan = plan A\n",
+      "After selecting plan B: cfg.plan = plan B\n"
+     ]
+    }
+   ],
+   "source": [
+    "cfg = OmegaConf.create(\"\"\"\n",
+    "plans:\n",
+    "    A: plan A\n",
+    "    B: plan B\n",
+    "selected_plan: A\n",
+    "plan: ${plans.${selected_plan}}\n",
+    "\"\"\")\n",
+    "print(f\"Default: cfg.plan = {cfg.plan}\")\n",
+    "cfg.selected_plan = \"B\"\n",
+    "print(f\"After selecting plan B: cfg.plan = {cfg.plan}\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Environment variable interpolation\n",
     "\n",
     "Environment variable interpolation is also supported.\n",
     "\n",
@@ -629,7 +663,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 20,
+   "execution_count": 21,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -647,7 +681,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 21,
+   "execution_count": 22,
    "metadata": {
     "pycharm": {
      "name": "#%%\n"
@@ -672,7 +706,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 22,
+   "execution_count": 23,
    "metadata": {
     "pycharm": {
      "name": "#%%\n"
@@ -704,7 +738,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 23,
+   "execution_count": 24,
    "metadata": {},
    "outputs": [
     {
@@ -736,7 +770,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 24,
+   "execution_count": 25,
    "metadata": {
     "pycharm": {
      "name": "#%%\n"
@@ -761,20 +795,12 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Nested interpolations"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "You can use interpolations within interpolations, for instance to perform an\n",
-    "operation over config variables:"
+    "You can take advantage of nested interpolations to perform operations over variables:"
    ]
   },
   {
    "cell_type": "code",
-   "execution_count": 25,
+   "execution_count": 26,
    "metadata": {},
    "outputs": [
     {
@@ -789,40 +815,6 @@
     "OmegaConf.register_resolver(\"plus_int\", lambda x, y: int(x) + int(y))\n",
     "conf = OmegaConf.create({\"a\": 1, \"b\": 2, \"a_plus_b\": \"${plus_int:${a},${b}}\"})\n",
     "print(conf.a_plus_b)"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "Another typical use case is to to dynamically select a sub-config:"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": 26,
-   "metadata": {},
-   "outputs": [
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "Default: cfg.plan = plan A\n",
-      "After selecting plan B: cfg.plan = plan B\n"
-     ]
-    }
-   ],
-   "source": [
-    "cfg = OmegaConf.create(\"\"\"\n",
-    "plans:\n",
-    "    A: plan A\n",
-    "    B: plan B\n",
-    "selected_plan: A\n",
-    "plan: ${plans.${selected_plan}}\n",
-    "\"\"\")\n",
-    "print(f\"Default: cfg.plan = {cfg.plan}\")\n",
-    "cfg.selected_plan = \"B\"\n",
-    "print(f\"After selecting plan B: cfg.plan = {cfg.plan}\")"
    ]
   },
   {
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -369,12 +369,12 @@ Another typical use case is to to dynamically select a sub-config:
 .. doctest::
 
     >>> cfg = OmegaConf.create("""
-        plans:
-            A: plan A
-            B: plan B
-        selected_plan: A
-        plan: ${plans.${selected_plan}}
-        """)
+    ... plans:
+    ...     A: plan A
+    ...     B: plan B
+    ... selected_plan: A
+    ... plan: ${plans.${selected_plan}}
+    ... """)
     >>> print(f"Default: cfg.plan = {cfg.plan}")
     Default: cfg.plan = plan A
     >>> cfg.selected_plan = "B"
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -364,6 +364,23 @@ operation over config variables:
     >>> c.a_plus_b
     3
 
+Another typical use case is to to dynamically select a sub-config:
+
+.. doctest::
+
+    >>> cfg = OmegaConf.create("""
+        plans:
+            A: plan A
+            B: plan B
+        selected_plan: A
+        plan: ${plans.${selected_plan}}
+        """)
+    >>> print(f"Default: cfg.plan = {cfg.plan}")
+    Default: cfg.plan = plan A
+    >>> cfg.selected_plan = "B"
+    >>> print(f"After selecting plan B: cfg.plan = {cfg.plan}")
+    After selecting plan B: cfg.plan = plan B
+
 
 Merging configurations
 ----------------------
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -322,8 +322,8 @@ The following example sets `12345` as the the default value for the `DB_PASSWORD
 
 Custom interpolations
 ^^^^^^^^^^^^^^^^^^^^^
-You can add additional interpolation types using custom resolvers.
-This example creates a resolver that adds 10 the the given value.
+You can add additional interpolation types using custom resolvers, which take strings as inputs.
+This example creates a resolver that adds 10 to the given value (note the need to cast it to `int`).
 
 .. doctest::
 
@@ -351,6 +351,18 @@ You can use literal commas and spaces anywhere by escaping (:code:`\,` and :code
     >>> c.escape_whitespace
     'Hello World'
 
+Nested interpolations
+^^^^^^^^^^^^^^^^^^^^^
+
+You can use interpolations within interpolations, for instance to perform an
+operation over config variables:
+
+.. doctest::
+
+    >>> OmegaConf.register_resolver("plus_int", lambda x, y: int(x) + int(y))
+    >>> c = OmegaConf.create({"a": 1, "b": 2, "a_plus_b": "${plus_int:${a},${b}}"})
+    >>> c.a_plus_b
+    3
 
 
 Merging configurations
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -286,6 +286,23 @@ Example:
     >>> print(type(conf.client.url).__name__)
     str
 
+Interpolations may be nested, enabling more advanced behavior like dynamically selecting a sub-config:
+
+.. doctest::
+
+    >>> cfg = OmegaConf.create("""
+    ... plans:
+    ...     A: plan A
+    ...     B: plan B
+    ... selected_plan: A
+    ... plan: ${plans.${selected_plan}}
+    ... """)
+    >>> print(f"Default: cfg.plan = {cfg.plan}")
+    Default: cfg.plan = plan A
+    >>> cfg.selected_plan = "B"
+    >>> print(f"After selecting plan B: cfg.plan = {cfg.plan}")
+    After selecting plan B: cfg.plan = plan B
+
 
 Environment variable interpolation
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -336,6 +353,7 @@ This example creates a resolver that adds 10 to the given value (note the need t
 Custom resolvers support variadic argument lists in the form of a comma separated list of zero or more values.
 Whitespaces are stripped from both ends of each value ("foo,bar" is the same as "foo, bar ").
 You can use literal commas and spaces anywhere by escaping (:code:`\,` and :code:`\ `).
+
 .. doctest::
 
     >>> OmegaConf.register_resolver("concat", lambda x,y: x+y)
@@ -351,11 +369,7 @@ You can use literal commas and spaces anywhere by escaping (:code:`\,` and :code
     >>> c.escape_whitespace
     'Hello World'
 
-Nested interpolations
-^^^^^^^^^^^^^^^^^^^^^
-
-You can use interpolations within interpolations, for instance to perform an
-operation over config variables:
+You can take advantage of nested interpolations to perform operations over variables:
 
 .. doctest::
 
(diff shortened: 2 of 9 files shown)


odelalleau:
FYI here is a toy script I've been using for basic tests:

```

conf_str = """
a: 1
b: ${a} == 1
c: US
d: ER
e: ${env:${c}${d}}
f:
    US: 2
g: ${f.${c}} == 2
h: ${syntax_error.${a}
i: "{test}_${a}"
j: ${a}_${syntax_error
k: ${a
l: ${b} ${a
"""


def main():
    cfg = OmegaConf.create(conf_str)
    print(cfg.pretty())
    print(f"{cfg.b=}")
    print(f"{cfg.e=}")
    print(f"{cfg.g=}")
    print(f"{cfg.h=}")
    print(f"{cfg.i=}")
    print(f"{cfg.j=}")
    print(f"{cfg.k=}")
    print(f"{cfg.l=}")


if __name__ == "__main__":
    main()
```

and its output:

```
a: 1
b: ${a} == 1
c: US
d: ER
e: ${env:${c}${d}}
f:
  US: 2
g: ${f.${c}} == 2
h: ${syntax_error.${a}
i: '{test}_${a}'
j: ${a}_${syntax_error
k: ${a
l: ${b} ${a

cfg.b='1 == 1'
cfg.e='odelalleau'
cfg.g='2 == 2'
cfg.h=None
cfg.i='{test}_1'
cfg.j=None
cfg.k='${a'
cfg.l=None
```

odelalleau:
Just had a flash that I could actually make the code both simpler & faster (now O(N)) => pushed 218bae71ca1e1e8d35696234ae6c2847cad10a01

omegaconf/_utils.py
```diff
@@ -319,7 +319,7 @@ def get_value_kind(value: Any, return_match_list: bool = False) -> Any:
     MANDATORY_MISSING : "???
```
omry, line 319: fix quotes please

omegaconf/base.py
```diff
@@ -4,10 +4,33 @@
...
 
+@dataclass
+class InterpolationRange:
+    """Used to store start/stop indices of interpolations being evaluated"""
+
+    start: int = -1
+    stop: int = -1
+
+
+class ParseError(Exception):
```
omry, line 25: Put it errors and inherit from OmegaConfBaseException

omegaconf/base.py
```diff
@@ -4,10 +4,33 @@
...
+    """Thrown when unable to parse a complex interpolation"""
+
+    @staticmethod
+    def assert_(condition: Any) -> None:
+        if not condition:
+            raise ParseError()
```
omry, line 31: I don't think this logic belongs in the exception class.

omegaconf/base.py
```diff
@@ -377,28 +400,63 @@ def _resolve_interpolation(
...
                 )
-                # if failed to resolve, return None for the whole thing.
-                if new_val is None:
-                    return None
-                new += orig[last_index : match.start(0)] + str(new_val)
-                last_index = match.end(0)
-
-            new += orig[last_index:]
+            except ParseError:
+                return None
```
omry, line 413: It it fly, if there is a need to suppress it in some scenarios we can add another flag.

omegaconf/base.py
```diff
@@ -377,28 +400,63 @@ def _resolve_interpolation(
...
             return StringNode(value=new, key=key)
         else:
             assert False
 
+    def _evaluate_simple(self, value: str, **kw: Any) -> Optional["Node"]:
+        """Evaluate a simple interpolation"""
+        value_kind, match_list = get_value_kind(value=value, return_match_list=True)
```
omry, line 420: Can you reuse the previous parsing results here instead of re-parse?

omegaconf/base.py
```diff
@@ -377,28 +400,63 @@ def _resolve_interpolation(
...
-                last_index = match.end(0)
-
-            new += orig[last_index:]
+            except ParseError:
+                return None
             return StringNode(value=new, key=key)
         else:
             assert False
 
+    def _evaluate_simple(self, value: str, **kw: Any) -> Optional["Node"]:
```
omry, line 418: why **kw and not explicit parameters?

omegaconf/base.py
```diff
@@ -377,28 +400,63 @@ def _resolve_interpolation(
...
+        value_kind, match_list = get_value_kind(value=value, return_match_list=True)
+        assert value_kind == ValueKind.INTERPOLATION and len(match_list) == 1
+        match = match_list[0]
+        node = self._resolve_simple_interpolation(
+            inter_type=match.group(1), inter_key=match.group(2), **kw,
+        )
+        return node
+
+    def _evaluate_complex(self, value: str, **kw: Any) -> Optional[str]:
+        """Evaluate a complex interpolation (>1, nested, with other strings...)"""
```
omry, line 429: Please put a high level description of what this supports and how it works.
In the next major version I am planning to introduce a proper parser generator which will simplify this greatly.
It will help knowing what this function is supposed to be doing when integrating it.

omry:
Please add proper tests, it's probably easier than doing it using your hack script.
see examples in the existing tests for pytest.mark.parameterize()

omry:
be sure to include tests for custom interpolations containing nested interpolations as well.

omegaconf/base.py
odelalleau: I could try but the problem is that the current parsing results are "incomplete", for instance if the string is `${f.${c}}` then it only matches `${c}`. So there could be two options:
1. Re-use the matches that have been found before: should be relatively easy to implement, but the code would be harder to read => not sure it's worth it
2. Improve the matching logic in `get_value_kind()` to fully cover all matches: this seems tricky, at least if we want to keep the current RE-based approach (https://stackoverflow.com/questions/133601/can-regular-expressions-be-used-to-match-nested-patterns)

omegaconf/base.py
odelalleau: I use this pattern sometime when there are arguments that are not actually used by a function, but need to be provided so that they can be forwarded to another function. It's similar to the typical `__init__(self, arg1, arg2, **kw)` used with `super().__init__(**kw)` in some places. Main benefit is to avoid modifying the code in a lot of places whenever you need to modify some of these arguments. There are drawbacks though and I know some people don't like it, so I'll just change it here (will also be more consistent with the rest of the code).

omegaconf/_utils.py
odelalleau: Fixed in fb08218

omegaconf/base.py
odelalleau: Done in fb08218

omegaconf/base.py
odelalleau: Changed in fb08218

omegaconf/base.py
odelalleau: Changed in fb08218

I still hard to guard the exception within an `if throw_on_resolution_failure:` block as otherwise there would be an infinite recursion on error (because the code preparing the error message also tries to parse the same expression)

omegaconf/base.py
odelalleau: Changed in 038f63c

omegaconf/base.py
odelalleau: Done in ec534f9

odelalleau:
NB: I added 540ba7d which solves an issue in my original implementation that was always returning a String node, even when the evaluated expression was another type.

omegaconf/base.py
```diff
@@ -403,30 +398,65 @@ def _resolve_interpolation(
             try:
                 # NB: `match_list` is ignored here as complex interpolations may be
```
omry, line 399: what is NB ?

omegaconf/base.py
omry: Okay, don't worry about it then.
Once we have a proper grammar this will be a million time easier.
in fact, check this out:
https://github.com/facebookresearch/hydra/blob/master/hydra/grammar/Override.g4#L78-L84

I think it will need to be changed to support complex interpolations from the command line.
In fact, I think it will need to be promoted from a token to a full fledged rule.

omegaconf/base.py
```diff
@@ -403,30 +398,65 @@ def _resolve_interpolation(
...
-        """Evaluate a complex interpolation (>1, nested, with other strings...)"""
+    def _evaluate_complex(
+        self,
+        value: str,
+        key: Any,
+        throw_on_missing: bool,
+        throw_on_resolution_failure: bool,
+    ) -> Optional["Node"]:
+        """
+        Evaluate a complex interpolation (>1, nested, with other strings...).
```
omry, line 442: add a few examples for complex interpolations here.

omegaconf/base.py
```diff
@@ -403,30 +398,65 @@ def _resolve_interpolation(
...
+
+        As an example, consider the string `value` set to "${foo.${bar}}":
+            1. We initialized our result with the original string: "${foo.${bar}}"
+            2. Scanning `value` from left to right, the first interpolation to be
+               closed is "${bar}". We evaluate it: if it resolves to "baz", we update
+               the result string to "${foo.baz}".
+            3. The next interpolation to be closed is now "${foo.baz}". We evaluate it:
+               if it resolves to "abc", we update the result accordingly to "abc".
+            4. We are done scanning `value` => the final result is "abc".
+        """
```
omry, line 457: Awesome description, thanks.

omegaconf/base.py
odelalleau: Nota Bene (https://www.urbandictionary.com/define.php?term=NB)
Will replace with "Note"

omegaconf/base.py
odelalleau: Yes a grammar will make things cleaner and more consistent. At least this PR should still enable some useful use cases in the meantime.

I ran into another issue related to the regex logic in `get_value_kind()`, which prevented me from fixing the inconsistency between `${a` (no error) and `${a}_${a ` (parsing error), which I had pointed out in point 3 above. It'd be a bit long to explain it in details, and annoying to fix, so I decided to give up on it for now since the grammar should fix it one way or another (= either always allowing it or always throwing an error).

odelalleau:
> Please add proper tests, it's probably easier than doing it using your hack script.
> see examples in the existing tests for pytest.mark.parameterize()

Tests added in 5476596

omegaconf/base.py
odelalleau: Done in 088f28a

omry:
undraft when ready for final review.

omegaconf/base.py
odelalleau: Done in 46f6d60

omegaconf/errors.py
```diff
@@ -103,3 +103,9 @@ class ConfigValueError(OmegaConfBaseException, ValueError):
     """
     Thrown from a config object when a regular access would have caused a ValueError.
     """
+
+
+class ParseError(OmegaConfBaseException):
+    """
+    Thrown when unable to parse a complex interpolation.
+    """
```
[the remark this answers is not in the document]
odelalleau: Done in 67231e0

odelalleau:
> undraft when ready for final review.

Should be ready :)

A note about additional commits I haven't mentioned in the above comments:
* 5fa85ef fixed an issue caused by the fact I was assuming all `${..}` segment were interpolations, while currently this is not the case (I guess that might change later with the grammar, but I preferred not to touch the regex in `get_value_kind()` for now)
* 2f99a64 adds documentation

omry:
A tiny coverage issue, but still 99.99% is not 100% :)

docs/notebook/Tutorial.ipynb
```diff
@@ -731,7 +731,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "You can add additional interpolation types using custom resolvers. This example creates a resolver that adds 10 the the given value."
+    "You can add additional interpolation types using custom resolvers, which take strings as inputs. This example creates a resolver that adds 10 to the given value (note the need to cast it to `int`)."
```
omry, line 734: I will improve this when I have a proper grammar. for now this is indeed taking strings which is unfortunate.

docs/notebook/Tutorial.ipynb
```diff
@@ -757,6 +757,40 @@
...
+      "3\n"
+     ]
+    }
+   ],
+   "source": [
+    "OmegaConf.register_resolver(\"plus_int\", lambda x, y: int(x) + int(y))\n",
+    "conf = OmegaConf.create({\"a\": 1, \"b\": 2, \"a_plus_b\": \"${plus_int:${a},${b}}\"})\n",
+    "print(conf.a_plus_b)"
+   ]
+  },
```
omry, line 793: Can you also add an example for a simple nested interpolation?
something along the lines of:
```python
cfg = OmegaConf.create({
  "plans":{
    "A": "plan A",
    "B": "plan B",
  },
  "selected_plan": "A"
  "plan": "${plans.${selected_plan}}"
})

cfg.plan
>> "Plan A"
cfg.selected_plan = "B"
cfg.plan
>> "Plan B"
```

docs/source/usage.rst
```diff
@@ -351,6 +351,18 @@ You can use literal commas and spaces anywhere by escaping (:code:`\,` and :code
     >>> c.escape_whitespace
     'Hello World'
 
+Nested interpolations
```
omry, line 354: added the example for simple nested interpolation here as well.

omegaconf/base.py
omry: Please explain the problem more and how a user would hit this.
I like this change but I want to make sure it does not opening up significant issues we are not ready to deal with.
It will be MANY months before OmegaConf 2.1.

omegaconf/base.py
```diff
@@ -439,15 +449,22 @@ def _evaluate_complex(
         throw_on_resolution_failure: bool,
     ) -> Optional["Node"]:
         """
-        Evaluate a complex interpolation (>1, nested, with other strings...).
+        Evaluate a complex interpolation.
+
+        A complex interpolation is more elaborate than "${a}" or "${a:b,c}", e.g.:
+            "I really like ${liked}" (concatenating a string)
+            "${subject} {verb} ${object}" (concatenating multiple strings)
```
omry, line 456: those two are referred to everywhere else as string interpolations. This is actually user facing terminology:

see [this](https://docs.google.com/presentation/d/e/2PACX-1vT_UIV7hCnquIbLUm4NnkUpXvPEh33IKiUEvPRF850WKA8opOlZOszjKdZ3tPmf8u7hGNP6HpqS-NT5/pub?start=false&loop=false&delayms=3000&slide=id.g84632f636b_5_148).

You are adding support for a new dimension, which is nested interpolations.
this is orthogonal to simple (`${foo.bar}` or string (`${foo}_bar`)) interpolations. both can be nested or not.

omegaconf/base.py
```diff
@@ -523,10 +540,10 @@ def _re_parent(self) -> None:
 
 def _parse_assert(condition: Any, msg: str = "") -> None:
```
omry, line 541: this method does not parse anything, certainly not an assert ;)

maybe `_cond_parse_error(condition: Any, msg: str = "")`?

tests/test_interpolation.py
```diff
@@ -347,11 +348,99 @@ def test_incremental_dict_with_interpolation() -> None:
...
+        for key, expected in expected_dict.items():
+            assert OmegaConf.select(c, key) == expected
+    finally:
+        OmegaConf.clear_resolvers()
+
+
+@pytest.mark.parametrize(  # type: ignore
+    "cfg,key",
+    [
+        ({"a": "PATH", "b": "${env:${a}"}, "b"),
```
omry, line 433: Why is this an error?

omegaconf/base.py
odelalleau: With both the current master and this PR, you can use:
```
a: 1
b: ${a
```
and it will run just fine, with `conf.b` being equal to `"${a"`.

However, with this PR, if you do:
```
a: 1
b: x_${a
```
then trying to access `conf.b` will trigger a parsing error. This is because by adding `x_` as prefix, it's not a simple interpolation anymore, so it goes through my new parsing code that ensures that all opening braces are properly closed.

Ideally I would like the first case to also trigger an error. I don't think anyone should be relying on the fact that right now it "works". But I don't see an easy way to make it happen.

So this is not really adding a problem, it's more that I'm unable to fix an existing problem :)

docs/notebook/Tutorial.ipynb
odelalleau: Cool, I just added this comment because it's easy to do something like `${plus:1,2}` and end up with `"12"` instead of `3` :)

docs/notebook/Tutorial.ipynb
odelalleau: Good idea! Done in a15623d

docs/source/usage.rst
odelalleau: Also in a15623d

omegaconf/base.py
odelalleau: Updated docstring in 3170517

Note that `_evaluate_complex()` is now also used for string interpolations (that's why I used "complex" instead of "nested").

omegaconf/base.py
omry: This is fine. It's good that you are noticing these differences but don't try too hard to preserve 100% compatibility.
some small differences around error cases are fine.

omegaconf/base.py
odelalleau: Hmm, your comment about simple interpolation being potentially nested makes me think I need to to a renaming pass over the code though, because `_resolve_simple_interpolation()` doesn't support nesting. I had interpreted "simple" as "a single `${..}` block" until now. This will have to wait until next week.

tests/test_interpolation.py
odelalleau: Oops, renamed in 94474ae
I intended it to be understood as "parsing_assert" but yeah, that was a bad name :)

omegaconf/base.py
odelalleau: Oops, renamed in 94474ae
I intended it to be understood as "parsing_assert" but yeah, that was a bad name :)

tests/test_interpolation.py
odelalleau: Missing closing brace (I'm glad you missed it, it means such an error is useful :) )

omry: sneaky sneaky error.

omry:
almost there. You have some ci issues. You can run it locally with nox.

omegaconf/base.py
omry: Yeah, maybe take a break and come back for a consistency pass later.

odelalleau:
> almost there. You have some ci issues. You can run it locally with nox.

Should be fixed now (2b473fe) -- taking note to run nox locally next time :)

omry:
Did a full review and added a few more comments.

This is looking great, the only thing that makes me pause here is that generally I do not add new features for minor versions.
This should be in OmegaConf 2.1, so this entire thing would probably go into a new branch for 2.1.
we can have a 2.1.0 pre release package to make it easier to install.

This reminds me, please add a news fragment (see [this](https://hydra.cc/docs/next/development/contributing/#news-entries) this for context).

docs/notebook/Tutorial.ipynb
```diff
@@ -757,6 +757,74 @@
     "print(conf.key)"
    ]
   },
+  {
```
omry, line 760: Looking at this diff is making it hard to know the structure is right.
The structure that makes sense to me:

1. Variable interpolation section contains a mention and an example of nesting interpolations.
2. Custom interpolation mentions that nested interpolations are supported, providing no example or a very short one.

docs/source/usage.rst
```diff
@@ -351,6 +351,35 @@ You can use literal commas and spaces anywhere by escaping (:code:`\,` and :code
...
+^^^^^^^^^^^^^^^^^^^^^
```
omry, line 355: same feedback as the one for the notebook:
Nested interpolations applies to all interpolations, they should be documented before custom interpolations.
I am a bit torn about mentioning nested interpolations in custom interpolations once we have a general description.

Also use the term "nested interpolations" or "interpolations can be nested".

docs/notebook/Tutorial.ipynb
odelalleau: Question related to this, currently the section structure in the notebook is like this:
```
# Variable interpolation
### Config node interpolation
### Environment variable interpolation
## Custom interpolations
## Nested interpolations
```
I assume that the first two sub-sections should use `##` instead, am I correct?

odelalleau:
> This reminds me, please add a news fragment (see [this](https://hydra.cc/docs/next/development/contributing/#news-entries) this for context).

Added news in edbb09b

omegaconf/_utils.py
```diff
@@ -316,10 +316,10 @@ def get_value_kind(value: Any, return_match_list: bool = False) -> Any:
     """
     Determine the kind of a value
     Examples:
-    MANDATORY_MISSING : "???
+    MANDATORY_MISSING : "???"
     VALUE : "10", "20", True,
     INTERPOLATION: "${foo}", "${foo.bar}"
```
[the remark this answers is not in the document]
odelalleau: I'm afraid not: this is related to the naming issue discussed above (the meaning of "simple" interpolation), which will take a bit of time to fix properly. Currently `get_value_kind()` returns `STR_INTERPOLATION` for a nested simple interpolation, and this is required by my implementation.

omegaconf/_utils.py
```diff
@@ -709,6 +709,17 @@ def type_str(t: Any) -> str:
         return ret
 
 
+def update_string(to_s: str, start: int, stop: int, from_s: str) -> Optional[str]:
```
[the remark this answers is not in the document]
odelalleau: Changed in a97ddfd

tests/test_interpolation.py
```diff
@@ -347,11 +348,100 @@ def test_incremental_dict_with_interpolation() -> None:
...
+    "cfg,expected_dict",
+    [
+        (
+            """
+            a: 1
+            b: a
+            c: ${${b}}
+            """,
+            {"c": 1},  # basic nesting
+        ),
```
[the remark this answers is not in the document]
odelalleau: Done in 7ed9492

tests/test_utils.py
```diff
@@ -485,3 +485,21 @@ def test_is_list_annotation(type_: Any, expected: Any) -> Any:
...
+    [
+        (-1, 1),
+        (1, -1),
+        (2, 1),
+        (len("hello world") + 1, 100),
+        (0, len("hello world") + 1),
+    ],
+)
+def test_update_string_error(start: int, stop: int) -> None:
+    assert _utils.update_string("hello world", start, stop, "i") is None
```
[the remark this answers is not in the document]
odelalleau: Done in a97ddfd

docs/notebook/Tutorial.ipynb
odelalleau: Changed in 6339fcd

docs/source/usage.rst
odelalleau: Changed in 6339fcd

For now I kept the example in custom interpolations as personally I think it's a typical use case that's worth highlighting.

odelalleau:
Just FYI nox session `coverage-3.6` passed on Github but failed on my Macbook. Seems to be related to the `id=` parameter of `pytest.param()`. Not sure if I should just revert that change (?)

docs/notebook/Tutorial.ipynb
omry: Probably, you can see what it looks like in the local browser by building the docs and opening the index locally.
Please do that do to make sure it all looks good (there are some details on how to do it at the bottom of [CONTRIBUTING.md](https://github.com/omry/omegaconf/blob/master/CONTRIBUTING.md).

docs/source/usage.rst
```diff
@@ -286,6 +286,23 @@ Example:
...
+    ...     A: plan A
+    ...     B: plan B
+    ... selected_plan: A
+    ... plan: ${plans.${selected_plan}}
+    ... """)
+    >>> print(f"Default: cfg.plan = {cfg.plan}")
+    Default: cfg.plan = plan A
+    >>> cfg.selected_plan = "B"
+    >>> print(f"After selecting plan B: cfg.plan = {cfg.plan}")
+    After selecting plan B: cfg.plan = plan B
```
omry, line 304:
```suggestion
    >>> print(cfg.plan) # default plan
    plan A
    >>> cfg.selected_plan = "B"
    >>> print(cfg.plan) # new plan
    plan B
```

omegaconf/_utils.py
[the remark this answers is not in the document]
omry: Please add a comment calling it out. This is a subtle but important details for this function.

omegaconf/base.py
```diff
@@ -5,7 +5,20 @@
...
+    UnsupportedInterpolationType,
+)
+
+
+@dataclass
+class InterpolationRange:
+    """Used to store start/stop indices of interpolations being evaluated"""
+
+    start: int = -1
+    stop: int = -1
```
omry, line 21: could it make more sense to use Optional[int] instead of -1 here?

tests/test_interpolation.py
```diff
@@ -347,11 +348,106 @@ def test_incremental_dict_with_interpolation() -> None:
...
+    "cfg,expected_dict",
+    [
+        pytest.param(
+            """
+            a: 1
+            b: a
+            c: ${${b}}
+            """,
+            {"c": 1},
+            id="basic nesting",
```
omry, line 366: Please avoid special characters in ids, no spaces, no ().
(While this works, it confuses other tools like Pycharm).

tests/test_interpolation.py
```diff
@@ -347,11 +348,106 @@ def test_incremental_dict_with_interpolation() -> None:
...
+            assert OmegaConf.select(c, key) == expected
+    finally:
+        OmegaConf.clear_resolvers()
+
+
+@pytest.mark.parametrize(  # type: ignore
+    "cfg,key",
+    [
+        # All these examples have non-matching braces.
+        ({"a": "PATH", "b": "${env:${a}"}, "b"),
```
omry, line 440: can also add pytest.param with ids here.

odelalleau:
> Since it passes in CI it's okay.
> I did not see this before, can you show me the failure?

So I was wrong, this has nothing to do with adding `id=`, these errors actually occur in the `master` branch for me. Here's a prototypical example:

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/test_base_config.py:480: in <lambda>
    @pytest.mark.parametrize("copy_method", [lambda x: copy.copy(x), lambda x: x.copy()])
omegaconf/listconfig.py:310: in copy
    return copy.copy(self)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

x = [1, 2, '${0}']

    def copy(x):
        """Shallow copy operation on arbitrary Python objects.

        See the module's __doc__ string for more info.
        """

        cls = type(x)

        copier = _copy_dispatch.get(cls)
        if copier:
            return copier(x)

        try:
            issc = issubclass(cls, type)
        except TypeError: # cls is not a class
            issc = False
        if issc:
            # treat it as a regular class:
            return _copy_immutable(x)

        copier = getattr(cls, "__copy__", None)
        if copier:
>           return copier(x)
E           TypeError: __copy__() takes 1 positional argument but 2 were given
```

I haven't really tried to figure out the root cause. Just did a quick test with a conda Python 3.6 env and I can't repro. But it systematically fails with nox...

omegaconf/_utils.py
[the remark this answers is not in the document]
odelalleau: Comment added in e818f78

docs/notebook/Tutorial.ipynb
odelalleau: Checked, structure looks good to me.

I realized though that one example was too wide, at least on my computer (had to scroll to the right to see the full line of code). I fixed that in 5500100

docs/source/usage.rst
odelalleau: Took your change (d50df1a)

omegaconf/base.py
odelalleau: Done in b3bedcb (actually making `start` a mandatory `int` to simplify type checks)

tests/test_interpolation.py
odelalleau: Done in f25f17f

tests/test_interpolation.py
odelalleau: Done in f25f17f

omegaconf/base.py
odelalleau: Done in b3bedcb (actually making start a mandatory int to simplify type checks)

omry:
> > Since it passes in CI it's okay.
> > I did not see this before, can you show me the failure?
> 
> So I was wrong, this has nothing to do with adding `id=`, these errors actually occur in the `master` branch for me. Here's a prototypical example:
> ...
> 
> I haven't really tried to figure out the root cause. Just did a quick test with a conda Python 3.6 env and I can't repro. But it systematically fails with nox...

Can you file an issue with a minimal repro? it's a bit hard to parse here.

odelalleau:
> Can you file an issue with a minimal repro? it's a bit hard to parse here.

See #314 and associated PR #315

odelalleau:
Just a note that I rebased on top of master to be sure everything still worked fine

omry:
Awesome work, I am ready to land this.
See my final comment (that was stuck in pending for 2 days :) ).

omry:
lending this to be able to test things on your followup issue. Please respond to my comment.

omegaconf/_utils.py
```diff
@@ -316,10 +316,15 @@ def get_value_kind(value: Any, return_match_list: bool = False) -> Any:
...
+    MANDATORY_MISSING : "???"
     VALUE : "10", "20", True,
     INTERPOLATION: "${foo}", "${foo.bar}"
-    STR_INTERPOLATION: "ftp://${host}/path"
+    STR_INTERPOLATION: "ftp://${host}/path", "${foo.${bar}}"
+
+    Note in particular that in the current implementation, "${foo.${bar}}" is
+    identified as a string interpolation (`STR_INTERPOLATION`) while it is
+    actually a (nested) simple interpolation. This discrepancy w.r.t naming
+    conventions will be addressed in a future update.
```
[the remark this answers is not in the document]
odelalleau: Yes, please assign #319 to me
Thanks for landing :)

PR #312, opened 2020-07-19
Fix changing List type with a different type

### LINKED ISSUE #300: With a List[int] config option (structured config), a single integer is still parsed as an integer
This is a copy of https://github.com/facebookresearch/hydra/issues/758 moved to OmegaConf as per @omry 's request.

The following code runs fine while it should raise an error since `fudge` is supposed to be a list of integers, not an integer.

```python
from typing import List

@dataclass
class Config:
    fudge: List[int] = field(default_factory=lambda: [])

cfg.fudge = 1
```

### LINKED ISSUE #326: Runtime error when creating a config from a dataclass using a Set.
```python
from typing import Set
@dataclass
class Foo:
    s : Set[str]
OmegaConf.structured(Foo(s = {"abc"}))
```

Results in `TypeError: issubclass() arg 1 must be a class`.

1. Set usage error should be specific to set.
2. Unexpected types should be rejected with a proper error and not a runtime error.

pereman2:
Closes #300 and closes #326 .
Added tests on changing list and changed test_node_wrap. I changed anynode's ref_type default value to Any so it would work(will need to change).

When changing a List, Tuple value it would treat it as it didn't have a target_node and therefore would wrap a new node instead of setting value.


omegaconf/basecontainer.py
```diff
@@ -435,7 +436,11 @@ def assign(value_key: Any, value_to_assign: Any) -> None:
             elif input_config:
                 assign(key, value)
             else:
-                self.__dict__["_content"][key] = wrap(key, value)
+                if change_node:
```
omry, line 439: can you use the should_set_value for this instead of introducing another variable that changes the control flow of this complex function?

Try to extend it to cover this case too.

tests/__init__.py
```diff
@@ -114,3 +114,9 @@ class A:
 @dataclass
 class B:
     x: A = MISSING
+
+
+@dataclass
+class ListD:
+    fudge: List[int] = field(default_factory=lambda: [])
+    tup: Tuple[int, int] = field(default_factory=lambda: (1, 1))
```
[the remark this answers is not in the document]
pereman2: I forgot to change these names, ups.

omegaconf/basecontainer.py
```diff
@@ -374,10 +374,11 @@ def _set_item_impl(self, key: Any, value: Any) -> None:
...
         special_value = value is None or value == "???"
-        should_set_value = (
-            target_node_ref is not None
-            and isinstance(target_node_ref, Container)
-            and special_value
+        should_set_value = target_node_ref is not None and (
+            (
+                isinstance(target_node_ref, Container)
+                and (special_value or target_node_ref._metadata.ref_type)
+            )
```
omry, line 381: okay, this is simpler.
Can you articulate the logic behind using the ref_type of the node as an indication of the need to set vs wrap?

pereman2: Since listconfigs didn't use ref_type before, and should_set_value was only True when the target was a container and the value was 'special', checking if the container has a ref_type means that it should be changed instead of wrapping a new node.

omry: The fact ListConfig did not use ref_type before should not be a factor here.
Can this be simplified? reading your explanation here does not give me a solid "ahhh, I get it" feeling.

pereman2:
it works like this but i want to simplify this

pereman2:
I will clean later when I have a better solution.

pereman2:
hey, nox is failing testing the jupyter notebook. 
Is this something new `ERROR  - pytest.PytestDeprecationWarning: Direct construction of IPyNbFile `?

omegaconf/base.py
```diff
@@ -531,6 +531,9 @@ def _re_parent(self) -> None:
                         if isinstance(item, Container):
                             item._re_parent()
 
+    def _has_ref_type(self) -> bool:
+        return self._metadata.ref_type not in (None, Any)
```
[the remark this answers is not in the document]
pereman2: When changing the value of a key with primitive container if the said key doesn't have a ref_type assigned (None) it should be set instead of wrapping a new node.
In Any's case, if a dictconfig has an Any ref_type when setting its value with a different type of dict, dictconfig or structured_config it fails so, we need to wrap a new node instead.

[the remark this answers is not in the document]
omry: Please explain with examples. I am not following.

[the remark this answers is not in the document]
pereman2: Let's say that a container has a ref_type if only it isn't None
```
# foo with none ref_type
cfg = OmegaConf.create({"foo": DictConfig(content={})})
cfg.foo = [1,2] # foo doesn't have a ref_type so it wraps
# foo with any ref_type
ref_type = Any 
cfg = OmegaConf.create({"foo": DictConfig(ref_type=ref_type, content={})})
cfg.foo = [1,2] # here it should wrap but instead, it sets value resulting in an error
```
If I'm understanding this correctly a key that has an Any ref_type, should be able to wrap new nodes independently  of their type. So that's why I'm including Any there, so the second case won't give an error.

pereman2:
> Please reduce the number of constants representing Any reftype to one (Either Any or None, be sure to be consistent across DictConfig and ListConfig)

I'm working on it, a lot of test are failing because of the test now.

omry:
Okay, no rush. just wanted to make sure the PR status reflects our discussion.

omegaconf/nodes.py
```diff
@@ -143,8 +143,8 @@ def __init__(
         super().__init__(
             parent=parent,
             value=value,
-            metadata=Metadata(
-                ref_type=None, object_type=None, key=key, optional=is_optional
+            metadata=Metadata(  # ref_type should be Optional[Type[Any]]
```
omry, line 146: What is this comment telling the reader?

tests/test_basic_ops_dict.py
```diff
@@ -560,6 +561,28 @@ def test_set_with_invalid_key() -> None:
...
+    cfg = OmegaConf.create({"foo": [1, 2]})
+    cfg.foo = ListClass()
+
+
+@pytest.mark.parametrize("value", [1, True, "str", 3.1415, ["foo", True, 1.2]])  # type: ignore
+def test_set_list_different_type(value: Any) -> None:
+    cfg = OmegaConf.structured(ListClass)
+    with pytest.raises(ValidationError):
+        cfg.list = value
+        cfg.tuple = value
```
omry, line 574: which one of the two assignments is raiding the validation error?

omry, line 574: Those tests should be moved to the structured configs tests,.

tests/test_nodes.py
```diff
@@ -359,30 +359,41 @@ class DummyEnum(Enum):
...
-        (int, 42, IntegerNode),
-        (float, 3.1415, FloatNode),
-        (bool, True, BooleanNode),
-        (str, "foo", StringNode),
+        (Any, Any, 10, AnyNode),
+        (DummyEnum, DummyEnum, DummyEnum.FOO, EnumNode),
+        (int, int, 42, IntegerNode),
+        (float, float, 3.1415, FloatNode),
+        (bool, bool, True, BooleanNode),
+        (str, str, "foo", StringNode),
```
omry, line 369: Do you support an IntegerNode with Any refrtype?

omegaconf/nodes.py
pereman2: oh, I should have checked that comment, I think that i was having an error when type checking. I'll look into it.

tests/test_basic_ops_dict.py
pereman2: Both, I'll remove the second one.

tests/test_nodes.py
pereman2: No.

tests/test_basic_ops_dict.py
omry: nope, only the first one :)
the second one never gets a chance to run.

tests/structured_conf/test_structured_config.py
```diff
@@ -691,6 +691,29 @@ def test_promote_to_object(self, class_type: str) -> None:
         assert OmegaConf.get_type(conf) == module.BoolConfig
         assert conf.with_default is False
 
+    def test_set_key_with_with_dataclass(self, class_type: str) -> None:
+        module: Any = import_module(class_type)
+        cfg = OmegaConf.create({"foo": [1, 2]})
+        cfg.foo = module.ListClass()
```
omry, line 697: Look at test_assignment_to_nested_structured_config, I think it's testing the same thing.

tests/structured_conf/test_structured_config.py
```diff
@@ -691,6 +691,29 @@ def test_promote_to_object(self, class_type: str) -> None:
...
+        module: Any = import_module(class_type)
+        cfg = OmegaConf.structured(module.ListClass)
+        value = [1, 2, 3]
+        cfg.list = value
+        cfg.tuple = value
+        assert cfg.list == value
+        assert cfg.tuple == value
+
+    @pytest.mark.parametrize("value", [1, True, "str", 3.1415, ["foo", True, 1.2]])  # type: ignore
+    def test_set_list_different_type(self, class_type: str, value: Any) -> None:
```
omry, line 709:
```suggestion
    def test_assign_wrong_type_to_list(self, class_type: str, value: Any) -> None:
```

omry, line 709: Add a similar test for dictionary.

tests/structured_conf/test_structured_config.py
pereman2: Ok.

pereman2: It's not, here ListClass is assigned in an Any ref_type key, in test_assignment_to_nested_structured_config Nested is assigned in a Nested ref_type key.
Anyways, I added that test because there was a part of basecontainer that wasn't covered and, this test completed the coverage.

omry: If the code coverage said it was not covered than it was not covered.

omegaconf/dictconfig.py
```diff
@@ -186,12 +188,19 @@ def is_typed(c: Any) -> bool:
         if is_dict(value_type) and is_dict(target_type):
             return
 
+        def is_not_valid_type(value_type: Any, target_type: Any) -> bool:
+            return (
+                type_is_primitive_list(value_type)
+                or is_primitive_type(value_type)
+                or not issubclass(value_type, target_type)
+            )
+
```
omry, line 197: I don't understand the logic here.
why is a list not valid? why is a primitive not valid?

pereman2: Checking if issubclass(value_type, target_type) fails every time primitive types are passed.

omry: Can you explain the additional check for is_primitive_list?
won't issubclass catch that? what case is this handling?

pereman2:
```
@dataclass
class ListClass:
    list: List[int] = field(default_factory=lambda: []) 
    tuple: Tuple[int, int] = field(default_factory=lambda: (1, 1)) 
    dict: Dict[str, Any] = field(default_factory=lambda: {"foo": "bar"})


cfg = OmegaConf.structured(ListClass)
cfg.dict =  1 # fails
cfg.dict = [1, True] # fails
```
Both assignments end up with 
```
  File "/Users/peristocles/contributions/omegaconf/omegaconf/dictconfig.py", line 302, in __setattr__
    self.__set_impl(key, value)
  File "/Users/peristocles/contributions/omegaconf/omegaconf/dictconfig.py", line 286, in __set_impl
    self._set_item_impl(key, value)
  File "/Users/peristocles/contributions/omegaconf/omegaconf/basecontainer.py", line 376, in _set_item_impl
    self._validate_set(key, value)
  File "/Users/peristocles/contributions/omegaconf/omegaconf/dictconfig.py", line 137, in _validate_set
    self._validate_set_merge_impl(key, value, is_assign=True)
  File "/Users/peristocles/contributions/omegaconf/omegaconf/dictconfig.py", line 203, in _validate_set_merge_impl
  File "/Users/peristocles/anaconda3/envs/omegaconf/lib/python3.6/typing.py", line 1148, in __subclasscheck__
    raise TypeError("Parameterized generics cannot be used with class "
TypeError: Parameterized generics cannot be used with class or instance checks
```

omegaconf/dictconfig.py
```diff
@@ -186,12 +188,19 @@ def is_typed(c: Any) -> bool:
         if is_dict(value_type) and is_dict(target_type):
             return
```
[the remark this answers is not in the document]
pereman2: What do you mean?

omegaconf/dictconfig.py
omry: I see, but why are you checking specifically list and not dict?

I just ran into something related in Hydra:
```
# Allows a list to be assigned to an int:
python examples/tutorials/structured_configs/1_minimal/my_app.py 'port=[1,2]'
Host: localhost, port: [1, 2]

# But correctly rejects a float:
omry@Coronadev:~/dev/hydra$ python examples/tutorials/structured_configs/1_minimal/my_app.py port=1.0
Error merging override port=1.0
Value '1.0' could not be converted to Integer
        full_key: port
        reference_type=Optional[MySQLConfig]
        object_type=MySQLConfig
```

Can you confirm that this works correctly after this fix?

pereman2: > I see, but why are you checking specifically list and not dict?

Dict is checked before, lists raise the same error as primitive types.

> I just ran into something related in Hydra:
> 
> ```
> # Allows a list to be assigned to an int:
> python examples/tutorials/structured_configs/1_minimal/my_app.py 'port=[1,2]'
> Host: localhost, port: [1, 2]
> 
> # But correctly rejects a float:
> omry@Coronadev:~/dev/hydra$ python examples/tutorials/structured_configs/1_minimal/my_app.py port=1.0
> Error merging override port=1.0
> Value '1.0' could not be converted to Integer
>         full_key: port
>         reference_type=Optional[MySQLConfig]
>         object_type=MySQLConfig
> ```
> 
> Can you confirm that this works correctly after this fix?

It isn't working I'll check what is happening. This fix only targeted containers.

omry: > Dict is checked before, lists raise the same error as primitive types.
Okay, maybe check list before as well? the code now is pretty hard to parse.

> It isn't working I'll check what is happening. This fix only targeted containers.
Probably better to file a different bug.

pereman2: > Okay, maybe check list before as well? the code now is pretty hard to parse.

I've been looking to simplify it and, when issubclass fails is because it doesn't support subscripted generics like `List[Any]` or  `Dict[str, int]`

There is a post about how to  [check if a type is a subscripted generic](https://stackoverflow.com/questions/49171189/whats-the-correct-way-to-check-if-an-object-is-a-typing-generic) which results in adding this logic removing previous changes I did but, I would mean adding that chuck of code in _utils.

```
if is_generic(target_type) and not is_base_generic(target_type):
            return
```

pereman2: I've tried it and it solves the problem.

omry: There are already functions in utils to check if something is a Dict or a List (is_dict_annotation, is_list_annotation), as well as functions to check if something is a primitive dict or list.

pereman2: I know but they don't check what is needed. For example for `List[Any]` and `List` in is_list_annotation would return true, but I need only to check if is like `List[Any], 
Dict[str, int]`...

omry: Do we ever use List without  the type (List[T] )?

omegaconf/_utils.py
```diff
@@ -724,3 +722,11 @@ def _ensure_container(target: Any) -> Any:
...
     assert OmegaConf.is_config(target)
     return target
+
+
+def is_generic_list(type_: Any) -> bool:
+    return is_list_annotation(type_) and get_list_element_type(type_) is not None
+
+
+def is_generic_dict(type_: Any) -> bool:
+    return is_dict_annotation(type_) and len(get_dict_key_value_types(type_)) > 0
```
omry, line 732: add comments describing what you mean by generic list and generic dict.

tests/test_utils.py
```diff
@@ -457,7 +457,7 @@ def test_is_list_annotation(type_: Any, expected: Any) -> Any:
         ),
         pytest.param(
             DictConfig(ref_type=Any, content=ConcretePlugin),
-            Any,
+            Optional[Dict[Any, Any]],
             id="DictConfig[ConcretePlugin]_Any_reftype",
```
omry, line 461: Why did you change this?

omry:
don't forget to add a news fragment.

tests/test_utils.py
pereman2: I didn't, I guess it was a fix in pre-commit.

omegaconf/dictconfig.py
pereman2: No. But I guess it could happen

tests/test_utils.py
omry: hmm, if you didn't change it it should not be in this PR.
make sure that you are rebased and that the PR contains only code that you changed.

omry: I didn't understand why you meant.
No,  pre-commit would not do that.
This is changing a test, that test is passing. It seems that this change is required due to a change in your logic.

Please figure out why this had to change.

pereman2: Oh no sorry, I forgot I changed it I changed _utils.get_ref_type so it supports the changes to Dictconfig and listconfig so now Dictconfig with Any ref_type would return Optional[Dict[Any, Any]] instead of Any.

omry: This change does not make sense to me:
Any means it can take any variable (including string or int).
Dict[Any, Any] means it can take any variable as the key and any variable in the value, but still must be an int.
This is actually not sensical for DictConfig because the key doesn't even support Any in general (it supports strings and enums).

Is Dict[Any, Any] a special sentinal value meaning Any? why not use Any then?

pereman2: Then should this be Any too ? `pytest.param(DictConfig(content={}), Optional[Dict[Any, Any]], id="DictConfig") `.

omry: This seems like it should be a Dict.
It also seems like it should be rejected because the dict key type does not support Any.

tests/test_utils.py
```diff
@@ -418,7 +418,7 @@ def test_is_list_annotation(type_: Any, expected: Any) -> Any:
         pytest.param(True, Optional[bool], id="bool"),
         pytest.param("bar", Optional[str], id="str"),
         pytest.param(None, type(None), id="NoneType"),
-        pytest.param({}, Optional[Dict[Any, Any]], id="dict"),
+        pytest.param({}, Optional[Dict[str, Any]], id="dict"),
```
omry, line 421: I think Optional[Dict[Union[str, Enum], Any]] is closer to the truth.

omegaconf/_utils.py
```diff
@@ -553,17 +553,21 @@ def none_as_any(t: Optional[Type[Any]]) -> Union[Type[Any], Any]:
...
         if isinstance(obj, Node):
             ref_type = obj._metadata.ref_type
-        if not is_structured_config(ref_type):
-            kt = none_as_any(obj._metadata.key_type)
-            vt = none_as_any(obj._metadata.element_type)
+        is_optional = obj._is_optional()
+        kt = none_as_any(obj._metadata.key_type)
+        vt = none_as_any(obj._metadata.element_type)
+        if ref_type is Any and kt is Any and vt is Any and obj._value() != "???":
+            pass
```
omry, line 560: 1. This is easier to read:
```suggestion
            ref_type = Any
```

2. Missing Container has their own ref type.

```python
ref_type(DictConfig(content="???", element_type=str, key_type=int)) == Optional[Dict[str, int]]
```
Make sure there is a test for this (and for missing ListConfig).

3. Also containers that are none. Use OmegaConf.is_missing() and OmegaConf.is_none() to test that instead of comparing to '???' and None directly.

pereman2: > Also containers that are none. Use OmegaConf.is_missing() and OmegaConf.is_none() to test that instead of comparing to '???' and None directly.
Well should it  be obj._is_missing() instead because obj is already the node to check?

omry: You can use either.

omegaconf/_utils.py
```diff
@@ -556,18 +556,18 @@ def none_as_any(t: Optional[Type[Any]]) -> Union[Type[Any], Any]:
         is_optional = obj._is_optional()
         kt = none_as_any(obj._metadata.key_type)
         vt = none_as_any(obj._metadata.element_type)
-        if ref_type is Any and kt is Any and vt is Any and obj._value() != "???":
-            pass
+        if ref_type is Any and kt is Any and vt is Any and not obj._is_missing():
+            ref_type = Any  # type: ignore
```
omry, line 560: what about  Container._is_none() ?
it's another case where you should not use ref_type Any (and add a test to cover it).

tests/test_utils.py
```diff
@@ -460,6 +460,11 @@ def test_is_list_annotation(type_: Any, expected: Any) -> Any:
             Any,
             id="DictConfig[ConcretePlugin]_Any_reftype",
         ),
+        pytest.param(
+            DictConfig(content="???", element_type=int, key_type=str),
+            Optional[Dict[str, int]],
+            id="DictConfig[str, int]_missing",
+        ),
```
omry, line 467: cover missign and none for DictConfig and ListConfig.

omegaconf/dictconfig.py
```diff
@@ -186,6 +188,8 @@ def is_typed(c: Any) -> bool:
...
 
+        if is_generic_list(target_type) or is_generic_dict(target_type):
+            return
```
omry, line 192: Can I assign a List[str] to a Dict[str, str] now?

omegaconf/listconfig.py
```diff
@@ -487,7 +488,12 @@ def _set_value(self, value: Any) -> None:
         ):
             self.__dict__["_content"] = value
         else:
-            assert is_primitive_list(value) or isinstance(value, ListConfig)
+            if not (is_primitive_list(value) or isinstance(value, ListConfig)):
+                msg = (
+                    f"Invalid value assigned : {value} is not a "
```
omry, line 493: add a test for this new error in test_errors.

omegaconf/dictconfig.py
pereman2: No you can't.

omry:
Ran into this one which could be related to this fix.
Can you check to see if it's catching it?

```python
@dataclass
class Module:
    name: str = MISSING
    classes: List[str] = MISSING


@dataclass
class Config:
    modules: List[Module] = MISSING

print(OmegaConf.merge(Config, {"modules": [{"should_error": "foo"}]}))
```

tests/test_errors.py
```diff
@@ -812,6 +812,18 @@ def finalize(self, cfg: Any) -> None:
         ),
         id="list:create_not_optional:_set_value(None)",
     ),
+    pytest.param(
+        Expected(
+            create=lambda: ListConfig(content=[1, 2]),
+            op=lambda cfg: cfg._set_value(True),
+            exception_type=ValidationError,
+            object_type=None,
+            msg="Invalid value assigned : True is not a subclass of ListConfig or list",
```
omry, line 821: Do we always show the object content in those kind of errors?
it could get crowded for large objects. should this show the type instead?

pereman2:
> Ran into this one which could be related to this fix.
> Can you check to see if it's catching it?
> 
> ```python
> @dataclass
> class Module:
>     name: str = MISSING
>     classes: List[str] = MISSING
> 
> 
> @dataclass
> class Config:
>     modules: List[Module] = MISSING
> 
> print(OmegaConf.merge(Config, {"modules": [{"should_error": "foo"}]}))
> ```
sure

pereman2:
It's not catching it.

omry:
> It's not catching it.

Please file a different issue for that one.

tests/test_errors.py
omry: ping

pereman2: I don't know about that one. I guess we can show the type.

omry: Okay, follow the same pattern here:
```
msg="Invalid type assigned : int is not a subclass of FoobarParams. value: 20",
```

Value last makes it easier to read in case the value is big.

PR #315, opened 2020-07-20
Fix nox session coverage-3.6 with Python <= 3.6.3

### LINKED ISSUE #314: nox session coverage-3.6 fails with python <= 3.6.3
Repro steps:

```
conda create -n py363-tmp python=3.6.3 --yes
conda activate py363-tmp
pytest tests/test_base_config.py::TestCopy
```

Observe multiple errors like:
```
FAILED tests/test_base_config.py::TestCopy::test_copy[src0-<lambda>0] - TypeError: __copy__() takes 1 positional argument but 2 were given
```

The error is related to this item in Python 3.6.4 changelog (https://docs.python.org/3.6/whatsnew/changelog.html#python-3-6-4-final):
```
bpo-28556: Two minor fixes for typing module: allow shallow copying instances of generic classes, [...]
```

I ran into this because on my MacBook `nox` is re-using `/opt/homebrew/opt/python36/bin/python3.6` which is 3.6.2.

I couldn't find a way to force `nox` to use Python >= 3.6.4 so I hacked around a PR to just skip these tests when this happens (will submit it shortly).

odelalleau:
Not sure it's the best way to fix #314 but at least it avoids adding noise when testing locally :)

### DIFF
--- a/tests/test_base_config.py
+++ b/tests/test_base_config.py
@@ -523,6 +523,17 @@ class TestCopy:
         assert id(cfg[0]) == id(cp[0])
 
 
+# This test is to ensure 100% coverage with Python <= 3.6.3, even when the
+# above tests are skipped.
+@pytest.mark.skipif(  # type: ignore
+    sys.version_info >= (3, 6, 4), reason="Copy is tested in `TestCopy`"
+)
+def test_enforce_copy_coverage() -> None:
+    cfg = OmegaConf.create([])
+    with raises(TypeError):
+        cfg.copy()
+
+
 def test_not_implemented() -> None:
     with raises(NotImplementedError):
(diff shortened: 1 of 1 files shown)


odelalleau:
Thanks, that looks better, though after skipping these tests I ran into a coverage issue, so I had to add another "dummy" test for Python <= 3.6.3 to ensure proper coverage. Should be all good now.

omry:
Great, thanks.

omry:
Looking at the changes made by this diff out of context I had no idea why it was needed and when I commented out locally of course everything was fine.

I wonder if there is a better way to solve your local problem.
maybe install a supported version of Python 3.6 or remove it from your PATH somehow?

PR #316, opened 2020-07-21
Allow bracketed lists as arguments to resolvers

### LINKED ISSUE #318: Brackets are not allowed in interpolations
Currently an interpolation of the form
```
a: ${my_resolver:[0\,1\,2]}
```
is not recognized as an interpolation, and thus is not resolved.

This will mostly become a significant issue once nested interpolations are added (#308), since one may want to do something like:
```
parallel_job_ids: [0, 3, 9]
hydra:
  launcher:
    params:
      cpus_per_task: ${get_cpus_per_task:${parallel_job_ids}}
```
where `get_cpus_per_task()` is a custom resolver that computes the number of CPUs to be reserved on the cluster, from the list of parallel jobs to run.

This is the motivation for #316

odelalleau:
This fixes #318 

This is needed if we want to be able to apply a resolver on a config variable which is a list (using nested interpolations).

Note that doing so is a bit cumbersome though, as we need to do something like:
```
my_list: [0, 1, 2]
my_function_of_list: ${my_function:${escape_commas:${my_list}},other_arg}
```
where `escape_commas` is a resolver that escapes the commas in the string representation of a list.

It may not look great but at least it works, enabling use cases that were not possible before (to the best of my knowledge).

### DIFF
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -327,7 +327,7 @@ def get_value_kind(value: Any, return_match_list: bool = False) -> Any:
     """
 
     key_prefix = r"\${(\w+:)?"
-    legal_characters = r"([\w\.%_ \\/:,-]*?)}"
+    legal_characters = r"([\w\.%_ \\/:,-\[\]]*?)}"
     match_list: Optional[List[Match[str]]] = None
 
     def ret(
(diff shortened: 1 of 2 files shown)


omry:
Please explain the motivation in more details. I don't understand what this is doing.

omry:
Also, before sending pull requests, *especially* PRs that are adding new features please open an issue and allow for some discussion.

odelalleau:
Closing (superseded by #321)

PR #320, opened 2020-07-22
Update CONTRIBUTING.md

pereman2:


CONTRIBUTING.md
```diff
@@ -45,6 +45,16 @@ Imports are sorted using isort, use `isort .` to sort all imports prior to pushi
...
 
+### Modifying Jupyter notebook
+
+In order to change the Jupyter notebook you first need to open it with `jupyter notebook`.
+Change the cell you want and then, execute it so the expected output is shown. 
+Note that the output after you execute the cell is saved as expected ouput for further 
+testing.
+
+In case that that the `in[number]` of cells aren't in order you should reset the kernel with 
+`00` and then execute all the cells.
```
omry, line 56: what is "reset the kernel with `00`" ?

pereman2: It restarts the jupyter notebook. I can change it to how to do it without a shortcut:
"In case that that the `in[number]` of cells aren't in order you should go to the kernel in the toolbar and restart it."

omry: Ok, no need to document Jupyter notebook shortcuts.

PR #335, opened 2020-08-20
Fix bug that allowed containers to be set as a value in ValueNodes

### LINKED ISSUE #324: ValueNodes are not validating correctly.
Nodes are being assigned wrong types. The following code shows that a list is assigned to a int, bool and str type which shouldn't happen.
```
@dataclass
class A:
    a: int = 5 
    b: bool = True
    c: str = "var"


cfg = OmegaConf.structured(A)
# Should raise error
cfg.a = [1, 2]
cfg.b = [1, 2]
cfg.c = [1, 2]
```

pereman2:
Closes #324 .
I added tests but I think maybe I should remove the one in test_structured_config since is essentially the same as test_errors. 
The variable should_set_value is getting pretty big but this fixed it.


omegaconf/basecontainer.py
```diff
@@ -397,6 +391,17 @@ def _set_item_impl(self, key: Any, value: Any) -> None:
...
             target_node = isinstance(target_node_ref, ValueNode)
+        should_set_value = target_node_ref is not None and (
+            (
+                # Target node is a container and has an special value or has an
+                # ref_type assigned
+                isinstance(target_node_ref, Container)
+                and (special_value or target_node_ref._has_ref_type())
+            )
+            # if target_node is a valuenode and is not an anynode then
+            # it should set value
```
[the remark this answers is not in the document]
pereman2: Because AnyNode could be anything from primitive to container?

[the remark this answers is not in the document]
omry: You are not supposed to put containers in AnyNode, containers are ALWAYS converted to DictConfig or ListConfig.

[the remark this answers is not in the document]
pereman2: Then if you make a config `cfg = OmegaConf.create({"foo": 4})` and `cfg.foo = []` it should wrap a new ListConfig. But if instead the config is:
```
@dataclass
class C:
    foo: Any = 4
cfg= OmegaConf.structured(C)
cfg.foo = []
```
it should give an error?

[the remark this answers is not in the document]
omry: No, it should do the same in both cases.
Why would the second one give an error?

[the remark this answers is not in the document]
pereman2: Because AnyNode doesn't allow containers.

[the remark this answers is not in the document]
omry: As I said, that container (dict or list) should be wrapped in a DictConfig or a ListConfig and REPLACE the AnyNode.
It can do that because the ref_type allows it.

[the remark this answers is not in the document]
pereman2: Then an AnyNode will never set its value it will always wrap a new node even if it's a primitive type. ref_type=Any is the default so it should never be a problem wrapping always.

omegaconf/basecontainer.py
```diff
@@ -391,16 +397,18 @@ def _set_item_impl(self, key: Any, value: Any) -> None:
 
         elif isinstance(self.__dict__["_content"], list):
             target_node = isinstance(target_node_ref, ValueNode)
+        # We use set_value if:
+        # 1. Target node is a container and the value is MISSING or None
+        # 2. Target node is a container and has an explicit ref_type
+        # 3. Target node is a valuenode. If it's an AnyNode then if the
+        #    value is a primitive type it should set.
```
omry, line 404: These are not change requests:

for 1: reading this I am wondering why we don't always use set_value if the is_missing() or is_none()

for 2: it only makes sense to me that we use set_value on a container if the incoming value is a compatible. (dict, missing or None)

for 3: I am having a hard time parsing this one.

omegaconf/basecontainer.py
```diff
@@ -391,16 +397,18 @@ def _set_item_impl(self, key: Any, value: Any) -> None:
...
             (
-                # Target node is a container and has an special value or has an
-                # ref_type assigned
                 isinstance(target_node_ref, Container)
                 and (special_value or target_node_ref._has_ref_type())
             )
-            # if target_node is a valuenode and is not an anynode then
-            # it should set value
             or (target_node and not isinstance(target_node_ref, AnyNode))
+            or (isinstance(target_node_ref, AnyNode) and is_primitive_type(value))
```
omry, line 411: This got me worried.
You completely changed the logic without changing any test.
This means you are just not covering this condition in a test, and I have no way to understand what you are trying to do here and if I agree with it.

Add tests that are covering this branch.

omegaconf/basecontainer.py
pereman2: > for 3: I am having a hard time parsing this one.

If the target  is a NodeValue then it should set his value. Furthermore if it's an AnyNode it should wrap when the input is a container and set when the input is an compatible type(primitive type).

omry: much better.

omegaconf/basecontainer.py
```diff
@@ -412,6 +412,7 @@ def _set_item_impl(self, key: Any, value: Any) -> None:
         )
 
         def wrap(key: Any, val: Any) -> Node:
+            print(key, val)
```
omry, line 417: Do not commit this.
I also recommend that you use a debugging instead of printing to debug things.

tests/test_basic_ops_dict.py
```diff
@@ -530,6 +530,28 @@ def test_set_with_invalid_key() -> None:
...
+
+def test_set_valuenode() -> None:
+    cfg = OmegaConf.structured(User)
+    a_before = cfg._get_node("age")
+    cfg.age = 12
+    assert id(cfg._get_node("age")) == id(a_before)
+    with pytest.raises(ValidationError):
+        cfg.age = []
+
+
```
omry, line 554: Nice :)

omegaconf/basecontainer.py
pereman2: oops

omegaconf/basecontainer.py
omry: so, the new tests you added, are they failing with the previous version of this code?

omegaconf/nodes.py
```diff
@@ -177,6 +177,12 @@ def __init__(
         )
 
     def validate_and_convert(self, value: Any) -> Optional[str]:
+        from omegaconf import OmegaConf
+
+        if OmegaConf.is_config(value) or is_primitive_container(value):
+            raise ValidationError(
+                "Value '$VALUE' is not a valid string (type $VALUE_TYPE)"
+            )
```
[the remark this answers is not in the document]
omry: did you address this? re-reviewing the entire diff looking for things like this is time consuming.
if you did not address this please do, if you did please clarify it here.

omegaconf/basecontainer.py
omry: still waiting for an answer on for this one.

tests/test_nodes.py
```diff
@@ -85,6 +85,22 @@ def test_valid_inputs(type_: type, input_: Any, output_: Any) -> None:
         (IntegerNode, "-abc"),
         (BooleanNode, "Nope"),
         (BooleanNode, "Yup"),
+        (StringNode, [1, 2]),
```
omry, line 88: I want to see invalid assignment to AnyNode here as well (dict, list, DictConfig and ListConfig, also illegal values like objects that are not Structured Configs).

omegaconf/basecontainer.py
pereman2: > so, the new tests you added, are they failing with the previous version of this code?

Sorry I didn't see this comment.

They do fail under previous changes because it wraps always instead of setting value when it should.

omegaconf/nodes.py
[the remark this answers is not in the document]
pereman2: It rejects them with the current implementation, line149: `if not is_primitive_type(value):`

tests/test_nodes.py
pereman2: Okay

omry: By assignment I mean set_value(), AnyNode can be replaced by a container.

pereman2: I'll extend the test_set_anynode with more values then.

tests/test_nodes.py
```diff
@@ -74,6 +74,10 @@ def test_valid_inputs(type_: type, input_: Any, output_: Any) -> None:
     assert str(node) == str(output_)
 
 
+class Person:
```
omry, line 77: There is already a class we can use for this, look for "IllegalType"

tests/test_nodes.py
```diff
@@ -101,15 +105,26 @@ def test_valid_inputs(type_: type, input_: Any, output_: Any) -> None:
...
 )
 def test_invalid_inputs(type_: type, input_: Any) -> None:
     empty_node = type_()
-    with pytest.raises(ValidationError):
-        empty_node._set_value(input_)
 
-    with pytest.raises(ValidationError):
-        type_(input_)
+    if type_ == AnyNode:
+        with pytest.raises(UnsupportedValueType):
```
omry, line 119: didn't we say we can make UnsupportedValueType a subclass of ValidationError?

pereman2: Ah I thought you meant we could make them on 2.1

pereman2: I misunderstood you, I'll change it.

pereman2:
I rebased incorrectly i'll re-commit.

tests/test_basic_ops_dict.py
```diff
@@ -539,17 +539,39 @@ def test_set_with_invalid_key() -> None:
         cfg[1] = "a"  # type: ignore
 
 
-def test_set_anynode() -> None:
+@pytest.mark.parametrize("value", [1, 3.14, True, None, Enum1.FOO])  # type: ignore
```
omry, line 542: these tests should probably be in test_nodes.py

pereman2: Well this one tests if the node is not replaced. That cannot be done in test_nodes.

pereman2: But yeah, I missed adding theses test_cases in test_valid_inputs in test_nodes

omry: It's okay. Please move the tests to test_nodes.

tests/test_errors.py
```diff
@@ -305,6 +305,18 @@ def finalize(self, cfg: Any) -> None:
         ),
         id="DictConfig[str,Color]:setitem_bad_value",
     ),
+    pytest.param(
+        Expected(
+            create=lambda: OmegaConf.structured(User),
+            op=lambda cfg: cfg.__setitem__("name", [1, 2]),
```
omry, line 311: Why not `cfg.name = [1,2]` ?

tests/test_errors.py
```diff
@@ -305,6 +305,18 @@ def finalize(self, cfg: Any) -> None:
...
+            exception_type=ValidationError,
+            msg="Value '[1, 2]' is not a valid string (type list)",
```
omry, line 313: This is a better error:
```
Cannot convert 'list' to string : '[1, 2]'
```

tests/test_nodes.py
omry: remove `class Person` and use `IllegalType` instead.

tests/test_nodes.py
omry: You don't need to special-case AnyNode now that the Exception is a ValidationError.

tests/test_errors.py
pereman2: Lambda doesn't allow that, and every other test uses setitem

omry: oh, right. forgot about that shit.

PR #339, opened 2020-08-26
make is_float return False for float(st) == float("inf")

my-tien:
use case:

`key: /path/to/data-${env:identifier}`

where identifier might be a combination of 6 chars and digits, e.g.
`identifier=2e1354`

Currently the variable interpolation will create this:
`key: /path/to/data-inf`

This commit causes `is_float` to return False here, so that the identifier is inserted as string:
`key: /path/to/data-2e1354`

### DIFF
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -368,8 +368,7 @@ def is_bool(st: str) -> bool:
 
 def is_float(st: str) -> bool:
     try:
-        float(st)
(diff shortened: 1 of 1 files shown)


omry:
This is the wrong fix for this problem. inf is a float. you are just hacking around the actual issue.
In your issue, provide full information such as what version of OmegaConf you are using and minimal repro.

PR #340, opened 2020-08-26
Change black version in .pre-commit-config.yaml.

pereman2:
Change version and reformat with the new changes. 
I changed to 20.8b1 version in black because as mentioned in https://github.com/psf/black/issues/420:
> Notably it gives the illusion that the latest version is being used but in reality it uses the version at install time.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ repos:
       - id: isort
 
   - repo: https://github.com/psf/black
-    rev: stable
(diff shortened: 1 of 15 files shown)


.pre-commit-config.yaml
```diff
@@ -5,10 +5,10 @@ repos:
       - id: isort
 
   - repo: https://github.com/psf/black
-    rev: stable
+    rev: 20.8b1
```
omry, line 8: nasty.

PR #352, opened 2020-08-30
Fix quoting of keys.

pereman2:
For some reason 'yes' as key gets quoted as opposed to every other yaml_bool_type. I don't know which prefix would be better instead of '\_'.


pereman2:
I didn't find a thing to change the behaviour one keys in specific so I had to override the represent_mapping method and change it with what I needed.

pereman2:
bool variables that their length is >1 like 'yes' will always print with quotes but instead the ones with length of 1 like 'y' could be changed.

omry:
Okay, I closed the original issue as invalid. no need to hack around it.

PR #366, opened 2020-09-09
Properly crash when trying to access child of a value node

### LINKED ISSUE #364: [Bug] Using ${foo.bar} returns ${foo} if `foo` is not a container
```
cfg = OmegaConf.create(dict(foo=0, bug="${foo.bar}"))
print(cfg.bug)
```
The above code prints "0" while I would expect it to crash.

What puzzles me is that this behavior is actually tested right now:
```
def test_indirect_interpolation2() -> None:
    d = {
        "a": {"aa": 10},
        "b": "${a.aa}",
        "c": "${b.aa}",
    }

    cfg = OmegaConf.create(d)
    assert cfg.c == 10
```
This test doesn't make sense to me: accessing `cfg.c` should crash since `b` is equal to 10 so `b.aa` should be invalid. Am I missing something?

odelalleau:
Fix #364

NB: I didn't understand the motivation for the two lines I removed -- but tests still passed after removing them so I guess it should be ok :)

### DIFF
--- a/news/364.bugfix
+++ b/news/364.bugfix
@@ -1,0 +1,1 @@
+Properly crash when accessing `${foo.bar}` if `foo` is a value node (instead of silently returning `${foo}`)
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -252,7 +252,6 @@ class Container(Node):
         :param key:
         :return:
         """
-        from . import ValueNode
         from .omegaconf import _select_one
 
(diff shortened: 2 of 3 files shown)


omry:
Clearly there was motivation :)
I will take a look to see why it was.

omry (approved):
Yup, looks like the sole purpose of this was to ensure the bad test passes :).

PR #395, opened 2020-09-25
Fix bug assigning structured classes of different element_type in dictconfigs #386

### LINKED ISSUE #386: Structured configs with Dict annotations allowing forbidden objects.
**Describe the bug**
DictConfig with primitive type as element type is allowing other structured configs as a valid element. 

**To Reproduce**
```python
@dataclass
class A:
   pass
@dataclass
class C:
    foo: Dict[str, int] = field(default_factory=lambda: {"a": A()})

cfg = OmegaConf.structured(C) # Doesn't raise an error
```

**Expected behaviour**

It should raise an error when assigning structured configs to a different element type in DictConfigs.

**context**
- [x] OmegaConf version: 2.0.2
- [x] Python version: 3.6, 3.7 and 3.8
- [x] Operating system : mac

pereman2:
When a ValueNode was being assigned to a DictConfig it was never check if it was a valid type(if DictConfig had an element_type). Moreover looking at the bug I replaced the test for Str2IntWithStrField because it should raise an error. 
Closes #386 .

### DIFF
--- a/news/386.bugfix
+++ b/news/386.bugfix
@@ -1,0 +1,1 @@
+Fix bug that allowed Structured classes to be assigned to DictConfig with different element type.
--- a/news/386.bugfix
+++ b/news/386.bugfix
@@ -1,1 +1,1 @@
-Fix bug that allowed Structured classes to be assigned to DictConfig with different element type.
+Fix bug that allowed instances of Structured Configs to be assigned to DictConfig with different element type.
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -435,7 +435,11 @@ class BaseContainer(Container, ABC):
                 target = self._get_node(key)
                 if target is None:
                     if is_structured_config(val):
-                        ref_type = OmegaConf.get_type(val)
+                        element_type = self._metadata.element_type
+                        if element_type is Any:
+                            ref_type = OmegaConf.get_type(val)
+                        else:
+                            ref_type = element_type
                 else:
                     is_optional = target._is_optional()
                     ref_type = target._metadata.ref_type
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -147,7 +147,25 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
         self._validate_set_merge_impl(key, value, is_assign=False)
 
     def _validate_set(self, key: Any, value: Any) -> None:
+        self._validate_set_value_node(key, value)
         self._validate_set_merge_impl(key, value, is_assign=True)
+
+    def _validate_set_value_node(self, key: Any, value: Any) -> None:
+        from omegaconf.omegaconf import _maybe_wrap
+
+        element_type = self._metadata.element_type
+        dummy_value = copy.deepcopy(value)
+
+        if isinstance(dummy_value, ValueNode) and element_type is not Any:
+            optional = dummy_value._metadata.optional
+            dummy_parent = DictConfig(content={})
+            _maybe_wrap(
+                ref_type=element_type,
+                key=key,
+                value=dummy_value._value(),
+                is_optional=optional,
+                parent=dummy_parent,
+            )
 
     def _validate_set_merge_impl(self, key: Any, value: Any, is_assign: bool) -> None:
         from omegaconf import OmegaConf
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -153,7 +153,6 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
         from omegaconf import OmegaConf
 
         vk = get_value_kind(value)
-
         if vk in (ValueKind.INTERPOLATION, ValueKind.STR_INTERPOLATION):
             return
 
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -147,30 +147,13 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
         self._validate_set_merge_impl(key, value, is_assign=False)
 
     def _validate_set(self, key: Any, value: Any) -> None:
-        self._validate_set_value_node(key, value)
         self._validate_set_merge_impl(key, value, is_assign=True)
-
-    def _validate_set_value_node(self, key: Any, value: Any) -> None:
-        from omegaconf.omegaconf import _maybe_wrap
-
-        element_type = self._metadata.element_type
-        dummy_value = copy.deepcopy(value)
-
-        if isinstance(dummy_value, ValueNode) and element_type is not Any:
-            optional = dummy_value._metadata.optional
-            dummy_parent = DictConfig(content={})
-            _maybe_wrap(
-                ref_type=element_type,
-                key=key,
-                value=dummy_value._value(),
-                is_optional=optional,
-                parent=dummy_parent,
-            )
 
     def _validate_set_merge_impl(self, key: Any, value: Any, is_assign: bool) -> None:
         from omegaconf import OmegaConf
 
         vk = get_value_kind(value)
+
         if vk in (ValueKind.INTERPOLATION, ValueKind.STR_INTERPOLATION):
             return
 
@@ -192,6 +175,10 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
                     )
 
         if value == "???":
+            return
+
+        if is_assign and isinstance(value, ValueNode) and self._has_element_type():
+            self._check_assign_with_wrap_value(key, value)
             return
 
         target: Optional[Node]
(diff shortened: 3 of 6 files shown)


omegaconf/dictconfig.py
```diff
@@ -147,8 +147,26 @@ def _validate_merge(self, key: Any, value: Any) -> None:
         self._validate_set_merge_impl(key, value, is_assign=False)
 
     def _validate_set(self, key: Any, value: Any) -> None:
+        self._validate_set_value_node(key, value)
         self._validate_set_merge_impl(key, value, is_assign=True)
 
+    def _validate_set_value_node(self, key: Any, value: Any) -> None:
```
omry, line 153: why is this a new function and not in _validate_set_merge_impl? (which is supposed to validate both set and assign based on the is_assign flag)

tests/structured_conf/test_structured_config.py
```diff
@@ -897,15 +907,8 @@ def test_str2str_with_field(self, class_type: str) -> None:
...
-        cfg = OmegaConf.structured(module.DictSubclass.Str2IntWithStrField())
-        assert cfg.foo == "bar"
-
-        cfg.one = 1
-        assert cfg.one == 1
-
         with pytest.raises(ValidationError):
-            # bad
-            cfg.hello = "world"
+            OmegaConf.structured(module.DictSubclass.Str2IntWithStrField())
```
omry, line 911: You changed the semantics of the test.
it was testing assignment, and now it's testing construction.
Don't do that, if you want to test something new add another test or change the test logic in a way that EXTENDS what it's doing.

Your test changes should reflect the bug you are fixing and should not reduce coverage of other functionality.
also, the bug description is talking about assignment, so I am not sure why this is even a reasonable test for what you are actually testing.

omry (changes requested):

omegaconf/dictconfig.py
pereman2: I find it confusing to stack different types of validation on the same function without splitting into different functions. I think this way it's more readable than hacking my way trough that big _validate_set_merge_impl function.

tests/structured_conf/test_structured_config.py
pereman2: Better this way 6397e1c ?

omegaconf/dictconfig.py
omry: even if I agree with this, now you have:
1. function that does both.
2. another function that does one.

Adding a new function without considering the existing design is not ok.

omry: Please combine it into _validate_set_merge_impl, we can have a followup refactoring to make the code easier to read later.

tests/structured_conf/data/attr_classes.py
```diff
@@ -393,15 +393,15 @@ class Str2User(Dict[str, User]):
     class Str2StrWithField(Dict[str, str]):
         foo: str = "bar"
 
-    @attr.s(auto_attribs=True)
-    class Str2IntWithStrField(Dict[str, int]):
-        foo: str = "bar"
```
omry, line 398: The idea with foo here is that you can have a fields in dataclass extending Dict.

make that field an int to conform with the superclass key type instead of removing it.

tests/structured_conf/test_structured_config.py
```diff
@@ -743,7 +743,17 @@ def test_assign_wrong_type_to_list(self, class_type: str, value: Any) -> None:
...
+        "value",
+        [
+            1,
+            True,
+            "str",
+            3.1415,
+            ["foo", True, 1.2],
+            {"foo": True},
+            User(age=1, name="foo"),
+            {"xd": User(age=1, name="foo")},
```
omry, line 755: what is "xd"?
why not "user"?

tests/structured_conf/test_structured_config.py
```diff
@@ -895,24 +905,18 @@ def test_str2str_with_field(self, class_type: str) -> None:
...
         def test_usr2str(self, class_type: str) -> None:
             module: Any = import_module(class_type)
             with pytest.raises(KeyValidationError):
                 OmegaConf.structured(module.DictSubclass.Error.User2Str())
 
+        def test_str2int_with_field_of_different_type(self, class_type: str) -> None:
+            module: Any = import_module(class_type)
+            cfg = OmegaConf.structured(module.DictSubclass.Error.Str2IntWithStrField())
+            with pytest.raises(ValidationError):
+                cfg.foo = "str"
```
omry, line 918: This tests the assignment of an incorrect type to an existing field (once you add `foo: int = 10` to the class.
```
cfg.foo = "str"
```

This tests creation of a new field with an incorrect type:
```
cfg.foo2 = "str"
```

omry (changes requested):

omegaconf/dictconfig.py
```diff
@@ -243,6 +230,24 @@ def is_typed(c: Any) -> bool:
...
+        dummy_value = copy.deepcopy(value)
+        optional = dummy_value._metadata.optional
+        dummy_parent = DictConfig(content={})
+        _maybe_wrap(
+            ref_type=element_type,
+            key=key,
+            value=dummy_value._value(),
+            is_optional=optional,
+            parent=dummy_parent,
+        )
```
omry, line 246: It looks like you are using maybe_wrap to do the validation and you are not doing anything with the return value.
It seems a wasteful, If this should really get wrapped, it will happen twice.

Also, what's the deal with the dummy parent and you using it's value in the strangest and probably incorrect way?

omegaconf/basecontainer.py
```diff
@@ -435,7 +435,11 @@ def wrap(key: Any, val: Any) -> Node:
                 target = self._get_node(key)
                 if target is None:
                     if is_structured_config(val):
-                        ref_type = OmegaConf.get_type(val)
+                        element_type = self._metadata.element_type
+                        if element_type is Any:
+                            ref_type = OmegaConf.get_type(val)
+                        else:
+                            ref_type = element_type
```
omry, line 442: Can you explain this change? I am not sure what this is doing.

pereman2: Wrapping a value in a dictconfig that has element type with the ref_type of the node as the element type ensures that node is validated correctly. This solved the test case `{"user": User(age=1, name="foo")}` for `test_assign_wrong_type_to_dict`

omegaconf/dictconfig.py
pereman2: What happened with the assignment of `User(age=1, name="foo")` is that when getting its data transformed everything to value nodes so, when you wrapped a value node it didn't validate it since it didn't really wrap a new node so I wrapped a new value with the node's value.

news/386.bugfix
```diff
@@ -0,0 +1 @@
+Fix bug that allowed Structured classes to be assigned to DictConfig with different element type.
```
omry, line 1:
```suggestion
Fix bug that allowed instances of Structured Configs to be assigned to DictConfig with different element type.
```

omegaconf/dictconfig.py
```diff
@@ -153,6 +153,7 @@ def _validate_set_merge_impl(self, key: Any, value: Any, is_assign: bool) -> Non
         from omegaconf import OmegaConf
 
         vk = get_value_kind(value)
+
```
omry, line 156: Unless black added this, remove.
```suggestion
```

omry (changes requested):
Looking better.
Any of this applies to assignment to lists?

```python
@dataclass
class ListConfig:
  list: List[int]

...
cfg = OmegaConf.structured(ListConfig(list=[1])
cfg.append(User(..)) # does this fail?
cfg[0] = User(..) # does this fail?
```

pereman2:
> Any of this applies to assignment to lists?
> 
> ```python
> @dataclass
> class ListConfig:
>   list: List[int]
> 
> ...
> cfg = OmegaConf.structured(ListConfig(list=[1])
> cfg.append(User(..)) # does this fail?
> cfg[0] = User(..) # does this fail?
> ```
They do fail.

**append**

image

**assignment**

image

omegaconf/dictconfig.py
```diff
@@ -225,6 +229,21 @@ def is_typed(c: Any) -> bool:
             )
             raise ValidationError(msg)
 
+    def _check_assign_value_node(self, key: Any, value: Any) -> None:
+        from omegaconf import OmegaConf
+
+        element_type = self._metadata.element_type
+        value_type = OmegaConf.get_type(value)
+        if value_type is not Any and element_type != value_type:
```
omry, line 237: You are checking for type equality but the error message is talking about a subclass.
which one is it?

omegaconf/dictconfig.py
```diff
@@ -176,6 +176,10 @@ def _validate_set_merge_impl(self, key: Any, value: Any, is_assign: bool) -> Non
         if value == "???":
             return
 
+        if is_assign and isinstance(value, ValueNode) and self._has_element_type():
+            self._check_assign_value_node(key, value)
```
omry, line 180: Looks like you modeled your check on it, but you did it wrong and now the error you copied does not match your check.
try to consolidate this logic _validate_set_merge_impl().
along the way, see if you can do the TODO in it and simplify it a bit.

omry (changes requested):

omegaconf/dictconfig.py
omry: Fix this.

tests/structured_conf/data/attr_classes.py
```diff
@@ -393,15 +393,15 @@ class Str2User(Dict[str, User]):
...
-
     class Error:
         @attr.s(auto_attribs=True)
         class User2Str(Dict[User, str]):
             pass
 
+        @attr.s(auto_attribs=True)
+        class Str2IntWithStrField(Dict[str, int]):
+            foo: int = 1
+
```
omry, line 404: Why is this  under Error?
The above one is an error because it's an illegal Structured Config.
There is nothing wrong with the one you moved here.

pereman2: But this one is used to test that gives an error. I'll move it

omry (approved):
Great.
As a followup, you can refactor the validation flow into two different functions.

Issue #396, opened 2020-09-28
Bad select value kind for INTERPOLATION 


fcennecf:
**Describe the bug**
The incorrectly parsed action for default value with specific char.

**To Reproduce**
There is app.yaml with next context and absent an env variable
``` 
config:
  DATABASE_URI: ${env:DATABASE_URI, "postgresql+psycopg2"}
```
**Expected behavior**
cfg = OmegaConf.load("app.yaml")
assert cfg.config.DATABASE_URI == "postgresql+psycopg2"

**Additional context**
- [ ] OmegaConf version: master
- [ ] Python version: 3.8.1
- [ ] Operating system : Linux localhost.localdomain 5.3.7-301.fc31.x86_64 #1 SMP Mon Oct 21 19:18:58 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
- [ ] Please provide a minimal repro

omry:
See my comment on the PR.
cc @odelalleau

PR #397, opened 2020-09-28
Fix 396

fcennecf:
Fix https://github.com/omry/omegaconf/issues/396


omry:
There is already work underway to improve the interpolation support that will cover this and will come in OmegaConf 2.1.

PR #402, opened 2020-10-04
Refactor _validate_set_merge_impl

pereman2:


omegaconf/dictconfig.py
```diff
@@ -176,73 +192,97 @@ def _validate_set_merge_impl(self, key: Any, value: Any, is_assign: bool) -> Non
...
+            valid = True
+        return valid
+
+    def _get_target(self, key: Any) -> Optional[Node]:
+        target: Optional[Node]
+        if key is None:
+            target = self
+        else:
+            target = self._get_node(key)
+        return target
```
omry, line 285: I don't like this.
you added a bunch of tiny functions that are sensible only in when called from the the validation and does not make any sense in any other context.

I would rather have a minimal amount of duplication between the validate_set and validate_merge than to add a page of code that is not useful in any other context.

omegaconf/dictconfig.py
```diff
@@ -176,73 +192,97 @@ def _validate_set_merge_impl(self, key: Any, value: Any, is_assign: bool) -> Non
...
-            msg = (
-                f"Invalid type assigned : {type_str(value_type)} is not a "
-                f"subclass of {type_str(element_type)}. value: {value}"
-            )
-            raise ValidationError(msg)
+    def _is_special_value(self, value: Any) -> bool:
+        # target must be optional by now. no need to check the type of value if None.
+        return value == "???" or value is None
+
+    def _get_assign_target_type(self, target: Optional[Node], value: Any) -> Any:
```
omry, line 252: called once, inline.

omry (changes requested):

omegaconf/dictconfig.py
```diff
@@ -147,17 +147,33 @@ def _validate_get(self, key: Any, value: Any = None) -> None:
...
+        self._validate_non_optional(key, value)
+        if self._is_special_value(value):
+            return
+
+        if is_assign:
+            self._validate_set_assign(key, value)
+        else:
+            self._validate_set_merge(key, value)
 
+    def _is_value_interpolation(self, value: Any) -> bool:
```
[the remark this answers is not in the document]
pereman2: No. It fails.

omegaconf/dictconfig.py
pereman2: I understand what you mean but, wouldn't that make it less readable ?

omegaconf/dictconfig.py
```diff
@@ -147,17 +147,33 @@ def _validate_get(self, key: Any, value: Any = None) -> None:
                 )
 
     def _validate_merge(self, key: Any, value: Any) -> None:
-        self._validate_set_merge_impl(key, value, is_assign=False)
+        self._validate(key, value, is_assign=False)
 
     def _validate_set(self, key: Any, value: Any) -> None:
-        self._validate_set_merge_impl(key, value, is_assign=True)
+        self._validate(key, value, is_assign=True)
```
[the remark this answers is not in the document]
pereman2: I'm changing it.

omegaconf/_utils.py
```diff
@@ -764,3 +764,13 @@ def is_generic_dict(type_: Any) -> bool:
...
+
+
+def is_generic_type(type_: Any) -> bool:
+    return is_generic_dict(type_) or is_generic_list(type_)
+
+
+def is_typed_dictconfig(c: Any) -> bool:
+    from omegaconf.dictconfig import DictConfig
+
+    return isinstance(c, DictConfig) and c._metadata.ref_type not in (Any, dict)
```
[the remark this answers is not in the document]
pereman2: > * is ref_type ever `dict`?

Yes. For example this test case in `test_merge` covers it:

```
pytest.param(
             [{}, ConfWithMissingDict],
             {"dict": "???"},
             id="merge_missing_dict_into_missing_dict",
         ),
```

omegaconf/dictconfig.py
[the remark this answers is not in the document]
omry: how?

omegaconf/dictconfig.py
omry: I may seem that adding many small functions enhance readability but it's not always a clear cut.
If I have to jump around the code a lot to understand why it's doing it's reducing readability.
If something is just being used once, I would rather have the code where it's being used.
You can use local boolean variables to enhance readability.
You can even use local functions if you feel something is best described in a function. This way you do not contaminate the module with utility functions that are not a part of the API.

omegaconf/_utils.py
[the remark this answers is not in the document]
pereman2: > * this function is checking ref_type, not object_type. from reading it's name I would assume it's checking the object type.

In DictConfigs obj_type is always `None` so I thought that having a ref_type would also mean that a DictConfig is "typed".

omegaconf/dictconfig.py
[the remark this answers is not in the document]
pereman2: OmegaConf.is_interpolation checks if a node in a Container is interpolation. Here we are checking if the value(not the target value) is an interpolation.

omegaconf/dictconfig.py
```diff
@@ -146,41 +146,15 @@ def _validate_get(self, key: Any, value: Any = None) -> None:
...
-                        cause=ValidationError("field '$FULL_KEY' is not Optional"),
-                    )
-
-        if value == "???":
-            return
-
-        if is_assign and isinstance(value, ValueNode) and self._has_element_type():
-            self._check_assign_value_node(key, value)
+        self._validate_non_optional(key, value)
+        # target must be optional by now. no need to check the type of value if None.
```
omry, line 156: I find this comment odd, because you are doing that exactly in the following line :)

omry:
Retry DeepCode

omegaconf/dictconfig.py
pereman2: True, I was just moving around the comment with the "linked" code to not lose it. I'll remove it.

omegaconf/dictconfig.py
```diff
@@ -189,60 +162,108 @@ def _validate_set_merge_impl(self, key: Any, value: Any, is_assign: bool) -> Non
         else:
             target = self._get_node(key)
 
-        if target is None:
-            return
+        def is_value_node(value: Any) -> bool:
+            valid = False
+            if isinstance(value, ValueNode) and self._has_element_type():
+                valid = True
+            return valid
```
omry, line 169: You can replace this function with a single line where you use it:
```
is_value_node = isinstance(value, ValueNode) and self._has_element_type()
```

omegaconf/dictconfig.py
```diff
@@ -189,60 +162,108 @@ def _validate_set_merge_impl(self, key: Any, value: Any, is_assign: bool) -> Non
...
 
-        def is_typed(c: Any) -> bool:
-            return isinstance(c, DictConfig) and c._metadata.ref_type not in (Any, dict)
+        def is_valid_target(target: Optional[Node]) -> bool:
+            valid = True
+            if target is None or not is_typed_dictconfig(target):
+                valid = False
+            return valid
```
omry, line 175: You can replace this function with a single line where you use it:
```
is_valid_target = target is None or not is_typed_dictconfig(target)
```

omegaconf/dictconfig.py
omry: This also reads wrong, if self has or has no element has no bearing on the value being a value node.

omegaconf/dictconfig.py
```diff
@@ -189,60 +162,108 @@ def _validate_set_merge_impl(self, key: Any, value: Any, is_assign: bool) -> Non
...
 
-        # target must be optional by now. no need to check the type of value if None.
-        if value is None:
-            return
+        def get_assign_target_type(target: Optional[Node], value: Any) -> Any:
+            if is_value_node(value):
+                type_ = self._metadata.element_type
+            else:
+                type_ = target._metadata.ref_type  # type: ignore
+            return type_
```
omry, line 185: target_type = self._metadata.element_type if is_value_node(target) else target._metadata.ref_type

omegaconf/_utils.py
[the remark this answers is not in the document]
omry: That's actually wrong. DictConfig object type is not None if it's a Structured Config.

omry (changes requested):
I feel like this PR is taking a lot of energy.

omegaconf/_utils.py
```diff
@@ -764,3 +764,7 @@ def is_generic_dict(type_: Any) -> bool:
     :return: bool
     """
     return is_dict_annotation(type_) and len(get_dict_key_value_types(type_)) > 0
+
+
+def is_generic_type(type_: Any) -> bool:
```
omry, line 769: Maybe a better name is `is_generic_container()`?

omegaconf/dictconfig.py
```diff
@@ -146,103 +145,103 @@ def _validate_get(self, key: Any, value: Any = None) -> None:
...
-        else:
-            target = self._get_node(key)
+        assign_value_node = isinstance(value, ValueNode) and self._has_element_type()
+        target_has_ref_type = isinstance(
+            target, DictConfig
+        ) and target._metadata.ref_type not in (Any, dict)
+        is_invalid_target = target is None or not target_has_ref_type
 
-        if target is None:
+        if is_invalid_target and not assign_value_node:
```
omry, line 166: This reads wrong, why would we want to return on an invalid target? (whatever that means)?

omegaconf/dictconfig.py
```diff
@@ -146,103 +145,103 @@ def _validate_get(self, key: Any, value: Any = None) -> None:
...
-        def is_typed(c: Any) -> bool:
-            return isinstance(c, DictConfig) and c._metadata.ref_type not in (Any, dict)
+        target_type = (
+            self._metadata.element_type if assign_value_node else target._metadata.ref_type  # type: ignore
+        )
+        value_type = OmegaConf.get_type(value)
 
-        if not is_typed(target):
+        if is_generic_type(target_type):
             return
```
omry, line 175: This also reads really weird to me.
Why is ValueNode assignment getting special treatment, and what is that treatment exactly.
Also, under what circumstances is the element_type a generic type?
Is it a case like this one?
```
foo : Dict[str, List[int]]
```
If so, it seems completely illegal to assign a value node to it, so I don't understand why you return here.

omry (commented):
This is easier to read, but still hard to understand. I left a few questions.

omegaconf/dictconfig.py
pereman2: > This also reads really weird to me.
> Why is ValueNode assignment getting special treatment, and what is that treatment exactly.

ValueNode is getting special treatment because when assigning a structured config into a DictConfig it first converts structured config's data to Nodes as opposed to every other instance(raw values). 

> Also, under what circumstances is the element_type a generic type?
> Is it a case like this one?
> 
> ```
> foo : Dict[str, List[int]]
> ```
> 
> If so, it seems completely illegal to assign a value node to it, so I don't understand why you return here.

We return when is a generic container because it can cause errors like `omegaconf.errors.ConfigTypeError: Subscripted generics cannot be used with class and instance checks` on 
```python
validation_error = (
            target_type is not None
            and value_type is not None
        )
```

omry: > ValueNode is getting special treatment because when assigning a structured config into a DictConfig it first converts structured config's data to Nodes as opposed to every other instance(raw values).

Read this sentence again.
You explain that value node is getting special treatment because Structured Configs assignment is this and that.
Maybe it's Structured Config assignment that is getting special treatment?

> We return when is a generic container because it can cause errors like omegaconf.errors.ConfigTypeError: Subscripted generics cannot be used with class and instance checks on ...

So instead of fixing the logic here, why can't you fix the check to handle Subscripted generics?

```python
validation_error = (
            target_type is not None
            and not is_generic_type(target_type)
            and value_type is not None
        )
```

Also, does that mean that we can assign the value `1` to a subscripted generic?

pereman2: > Read this sentence again.
> You explain that value node is getting special treatment because Structured Configs assignment is this and that.
> Maybe it's Structured Config assignment that is getting special treatment?

Then how would you check for example:
```python
@dataclass
class DictClass:

@dataclass
class User:
    age: int = 4
    name: str = "foo"

cfg.foo = User() # should raise error because of name
```
It assigns a User which is a structured config but, since it doesn't have a ref_type to check it has to be checked when setting its values age and name. Am I missing something?

Structured Config is getting special treatment indirectly checking nodes. 
> > We return when is a generic container because it can cause errors like omegaconf.errors.ConfigTypeError: Subscripted generics cannot be used with class and instance checks on ...
> 
> So instead of fixing the logic here, why can't you fix the check to handle Subscripted generics?
> 
> ```python
> validation_error = (
>             target_type is not None
>             and not is_generic_type(target_type)
>             and value_type is not None
>             and not issubclass(value_type, target_type)
>         )
> ```
> 

I can move it to validation_error but isn't it logically the same? 

> Also, does that mean that we can assign the value `1` to a subscripted generic?

You can't, because 1 is being wrapped with a Dict annotation ref_type. It is tested on `test_assign_wrong_type_to_dict`

omry:
This exchange is a great example of why it's important to insist that the code makes sense.
As far as I can tell just by looking at the behavior below, you just found  bug here.

```python
@dataclass
class DictClass:

@dataclass
class User:
    age: int = 4
    name: str = "foo"

cfg.foo = User() # should raise error because of name
```

playing with this, I see several problems with the current behavior here:
1: assigning 10 to foo throws but modifies it anyway:
```
In [25]: cfg = OmegaConf.structured(DictClass)

In [26]: cfg.foo
Out[26]: {'a': 4}

In [27]: cfg.foo = 10
ValidationError: 
        full_key: foo
        reference_type=Optional[DictClass]
        object_type=DictClass
In [28]: cfg.foo
Out[28]: {}
```
1.1 the validation error has no message for some reason.

2. The behavior in your example does not make sense to me:

```python
In [31]: cfg = OmegaConf.structured(DictClass)
In [32]: cfg.foo = User()
ValidationError: Invalid type assigned : str is not a subclass of int. value: foo
        full_key: foo.name
        reference_type=Dict[str, int]
        object_type=dict
```
What does the error has to do with my action?
I expect it to fail because I assign User to Dict, not because of anything to do with name.

So this seems like a few bugs different bugs in this area.
Can you file issues for 1 and 2?
I think we should clear them before we proceed with this refactoring.
2 is potentially more disruptive, let's see what fixing it means for other tests when you try to fix it.

pereman2:
> This exchange is a great example of why it's important to insist that the code makes sense.
> As far as I can tell just by looking at the behavior below, you just found bug here.
> 
> ```python
> @dataclass
> class DictClass:
>     foo: Dict[str, int] = field(default_factory=lambda: {"a": 4})
> 
> @dataclass
> class User:
>     age: int = 4
>     name: str = "foo"
> 
> cfg = OmegaConf.structured(DictClass)
> cfg.foo = User() # should raise error because of name
> ```
> 
> playing with this, I see several problems with the current behavior here:
> 1: assigning 10 to foo throws but modifies it anyway:

It is creating the dict and then adding the other values(because tries to assign user data). I think that's why.

> ```
> In [25]: cfg = OmegaConf.structured(DictClass)
> 
> In [26]: cfg.foo
> Out[26]: {'a': 4}
> 
> In [27]: cfg.foo = 10
> ValidationError: 
>         full_key: foo
>         reference_type=Optional[DictClass]
>         object_type=DictClass
> In [28]: cfg.foo
> Out[28]: {}
> ```
> 
> 1.1 the validation error has no message for some reason.
> 
> 2. The behavior in your example does not make sense to me:
> 
> ```python
> In [31]: cfg = OmegaConf.structured(DictClass)
> In [32]: cfg.foo = User()
> ValidationError: Invalid type assigned : str is not a subclass of int. value: foo
>         full_key: foo.name
>         reference_type=Dict[str, int]
>         object_type=dict
> ```
> 
> What does the error has to do with my action?
> I expect it to fail because I assign User to Dict, not because of anything to do with name.

Then does that mean that a structured config cannot be assigned to a Dict[str, int] even though User has str keys and int values?

> So this seems like a few bugs different bugs in this area.
> Can you file issues for 1 and 2?
> I think we should clear them before we proceed with this refactoring.
> 2 is potentially more disruptive, let's see what fixing it means for other tests when you try to fix it.

Sure.

omry:
> Then does that mean that a structured config cannot be assigned to a Dict[str, int] even though User has str keys and int values?

I think so, the assignment is illegal under mypy.
merge should be legal, but assignment is a stretch.

omry:
Is there a test covering the behavior in 2?

omry:
patching 2.0 branch with this:
```diff
-        if is_generic_list(target_type) or is_generic_dict(target_type):
-            return
+        if is_assign:
+            if is_generic_dict(target_type):
+                if not is_generic_dict(value_type):
+                    raise ValidationError(
+                        f"Cannot assign {type_str(value_type)} to {type_str(target_type)}"
+                    )
+
+            if is_generic_list(target_type):
+                if not is_generic_list(value_type):
+                    raise ValidationError(
+                        f"Cannot assign {type_str(value_type)} to {type_str(target_type)}"
+                    )
+
         # TODO: simplify this flow. (if assign validate assign else validate merge)
```
All the tests are passing, but that example is failing:
```
$ python 1.py 
  File "1.py", line 15, in <module>
    cfg.foo = User()
  File "/home/omry/dev/omegaconf/omegaconf/dictconfig.py", line 289, in __setattr__
    self._format_and_raise(key=key, value=value, cause=e)
  File "/home/omry/dev/omegaconf/omegaconf/base.py", line 95, in _format_and_raise
    format_and_raise(
  File "/home/omry/dev/omegaconf/omegaconf/_utils.py", line 675, in format_and_raise
    _raise(ex, cause)
  File "/home/omry/dev/omegaconf/omegaconf/_utils.py", line 591, in _raise
    raise ex  # set end OC_CAUSE=1 for full backtrace
omegaconf.errors.ValidationError: Cannot assign User to Dict[str, int]
        full_key: foo
        reference_type=Optional[DictClass]
        object_type=DictClass
```

Need to add some tests covering this and also ensuring things makes sense when merging of course, but this can be the base for a fix.

pereman2:
Did you patch this ?

omry:
No. this was just an example solution I hacked in a few minutes. go ahead and integrate.

pereman2:
I messed the rebase, I'm fixing it.

pereman2:
Hey, whenever you can check this. After fixing the last 2 bugs it looks like this.

omry:
Yes, I am trying to rest this weekend.
Will take a look sometime next week.

omry (approved):
I like it.

PR #408, opened 2020-10-16
Fix #407 Subscripted Dict and List unable to be pickled on python 3.6.

### LINKED ISSUE #407: Unable to pickle typing.Dict[kt, vt] and List[vt] on python 3.6
**Describe the bug**
Pickling generic types `Dict[kt, vt]` and `List[vt]`on python 3.6 produces an error.

**To Reproduce**
```python
@dataclass
class Pick:
    union: Dict[str, int] = field(default_factory=lambda: {})

with tempfile.TemporaryFile() as fp:
    c = OmegaConf.structured(Pick)
    pickle.dump(c, fp)
    fp.flush()
    fp.seek(0)
    c1 = pickle.load(fp)
```

**Expected behavior**
It should be able to pickle.

**Additional context**
- [x] Python version: 3.6

pereman2:
Closes #407 .

Now when pickling structured configs ref_types like `Dict[kt, vt]` and `List[vt]` are transformed to `Dict` and `List` respectively and then reconstructed when loading.

The reason behind this is for ensuring that the config doesn't get modified after pickle.
```python
dict_ = copy.copy(self.__dict__)
dict_["_metadata"] = copy.copy(dict_["_metadata"])
```

### DIFF
--- a/news/407.bugfix
+++ b/news/407.bugfix
@@ -1,0 +1,1 @@
+Fixes bug where subscripted Dict and List couldn't be pickled on python 3.6.
--- a/news/407.bugfix
+++ b/news/407.bugfix
@@ -1,1 +1,1 @@
-Fixes bug where subscripted Dict and List couldn't be pickled on python 3.6.
+Fix Structured Configs with subscripted Dict and List fields failed pickling on python 3.6.
--- a/news/407.bugfix
+++ b/news/407.bugfix
@@ -1,1 +1,1 @@
-Fix Structured Configs with subscripted Dict and List fields failed pickling on python 3.6.
+Fix pickling of Structured Configs with fields annotated as Dict[KT, VT] or List[T] on Python 3.6.
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -90,17 +90,17 @@ class BaseContainer(Container, ABC):
 
     # Support pickle
     def __getstate__(self) -> Dict[str, Any]:
-        dict_ = copy.copy(self.__dict__)
-        dict_["_metadata"] = copy.copy(dict_["_metadata"])
+        dict_copy = copy.copy(self.__dict__)
+        dict_copy["_metadata"] = copy.copy(dict_copy["_metadata"])
         ref_type = self._metadata.ref_type
         if is_container_annotation(ref_type):
             if is_dict_annotation(ref_type):
-                dict_["_metadata"].ref_type = Dict
+                dict_copy["_metadata"].ref_type = Dict
             elif is_list_annotation(ref_type):
-                dict_["_metadata"].ref_type = List
+                dict_copy["_metadata"].ref_type = List
             else:
                 assert False
-        return dict_
+        return dict_copy
 
     # Support pickle
     def __setstate__(self, d: Dict[str, Any]) -> None:
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -89,10 +89,30 @@ class BaseContainer(Container, ABC):
 
     # Support pickle
     def __getstate__(self) -> Dict[str, Any]:
-        return self.__dict__
+        from omegaconf._utils import is_dict_annotation
+
+        dict_ = copy.copy(self.__dict__)
+        dict_["_metadata"] = copy.copy(dict_["_metadata"])
+        if self._has_ref_type():
+            if is_dict_annotation(self._metadata.ref_type):
+                dict_["_metadata"].ref_type = Dict
+            else:
+                dict_["_metadata"].ref_type = List
+        return dict_
 
     # Support pickle
     def __setstate__(self, d: Dict[str, Any]) -> None:
+        from omegaconf import DictConfig
+        from omegaconf._utils import is_generic_dict
+
+        if isinstance(self, DictConfig):
+            key_type = d["_metadata"].key_type
+        element_type = d["_metadata"].element_type
+        if is_generic_dict(d["_metadata"].ref_type):
+            d["_metadata"].ref_type = Dict[key_type, element_type]  # type: ignore
+        else:
+            d["_metadata"].ref_type = List[element_type]  # type: ignore
+
         self.__dict__.update(d)
 
     @abstractmethod
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -89,28 +89,28 @@ class BaseContainer(Container, ABC):
 
     # Support pickle
     def __getstate__(self) -> Dict[str, Any]:
-        from omegaconf._utils import is_dict_annotation
-
         dict_ = copy.copy(self.__dict__)
         dict_["_metadata"] = copy.copy(dict_["_metadata"])
         if self._has_ref_type():
-            if is_dict_annotation(self._metadata.ref_type):
+            ref_type = self._metadata.ref_type
+            if is_dict_annotation(ref_type):
                 dict_["_metadata"].ref_type = Dict
-            else:
+            elif is_list_annotation(ref_type):
                 dict_["_metadata"].ref_type = List
         return dict_
 
     # Support pickle
     def __setstate__(self, d: Dict[str, Any]) -> None:
         from omegaconf import DictConfig
-        from omegaconf._utils import is_generic_dict
+        from omegaconf._utils import is_generic_dict, is_generic_list
 
         if isinstance(self, DictConfig):
             key_type = d["_metadata"].key_type
         element_type = d["_metadata"].element_type
-        if is_generic_dict(d["_metadata"].ref_type):
+        ref_type = d["_metadata"].ref_type
+        if is_generic_dict(ref_type):
             d["_metadata"].ref_type = Dict[key_type, element_type]  # type: ignore
-        else:
+        elif is_generic_list(ref_type):
             d["_metadata"].ref_type = List[element_type]  # type: ignore
 
         self.__dict__.update(d)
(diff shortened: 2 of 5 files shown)


omry:
> The reason behind this is for ensuring that the config doesn't get modified after pickle.
Can you explain what you mean? why is this the reason?

news/407.bugfix
```diff
@@ -0,0 +1 @@
+Fixes bug where subscripted Dict and List couldn't be pickled on python 3.6.
```
omry, line 1: The news is not clear enough.
```suggestion
Fix Structured Configs with subscripted Dict and List fields failed pickling on python 3.6.
```

omegaconf/basecontainer.py
```diff
@@ -89,10 +89,30 @@ def __repr__(self) -> str:
...
-        return self.__dict__
+        from omegaconf._utils import is_dict_annotation
+
+        dict_ = copy.copy(self.__dict__)
+        dict_["_metadata"] = copy.copy(dict_["_metadata"])
+        if self._has_ref_type():
+            if is_dict_annotation(self._metadata.ref_type):
+                dict_["_metadata"].ref_type = Dict
+            else:
+                dict_["_metadata"].ref_type = List
```
omry, line 100: defaulting to list here is bad practice. if we ever have another type this will be broken.
```python
if is_dict...:
  ...
elif is_list...:
  ...
else:
  assert False
```

omegaconf/basecontainer.py
```diff
@@ -89,10 +89,30 @@ def __repr__(self) -> str:
...
+        from omegaconf import DictConfig
+        from omegaconf._utils import is_generic_dict
+
+        if isinstance(self, DictConfig):
+            key_type = d["_metadata"].key_type
+        element_type = d["_metadata"].element_type
+        if is_generic_dict(d["_metadata"].ref_type):
+            d["_metadata"].ref_type = Dict[key_type, element_type]  # type: ignore
+        else:
+            d["_metadata"].ref_type = List[element_type]  # type: ignore
```
omry, line 114: ditto

tests/__init__.py
```diff
@@ -144,3 +144,13 @@ class Module:
...
+
+
+@dataclass
+class SubscriptedList:
+    list: List[int] = field(default_factory=lambda: [1, 2])
+
+
+@dataclass
+class SubscriptedDict:
+    dict: Dict[str, int] = field(default_factory=lambda: {"foo": 4})
```
omry, line 156: Curious, cab you check what happens to dataclasses that extends `Dict[str, int]` for example?

omry (changes requested):
Nice. a few minor requests.

omegaconf/basecontainer.py
pereman2: Okay

tests/__init__.py
pereman2: Pickling this class works just fine because the ref_type stored is `DictSubclass`
```python
@dataclass
class DictSubclass(Dict[str, int]):
    pass
```

It should be `ref_type=Dict[str, int]` since obj_type is `DictSubclass` imo.

pereman2:
> The reason behind this is for ensuring that the config doesn't get modified after pickle.

> Can you explain what you mean? why is this the reason?

Making `__getstate__` modify ref_type to ensure that python 3.6 pickles subscripted Dicts or Lists will modify the config after dumping if it isn't copied. 

So, if you want to dump cfg and then keep working with cfg you don't want pickle to modify it.

omry:
I am just not understanding the sentence:
> The reason behind this is for ensuring that the config doesn't get modified after pickle.

It appears in the function, the motivation there is to ensure pickle works, the implementation changes the metadata.
seems to be confusing the motivation (reason) with the implementaiton.

About the implementation:
Would it make more sense to change the representation of the metadata instead?
we already have ref type, key type and element type. essentially we have everything we need.

get_ref_type() can construct the proper ref type when it's being called.
if anything is using _metadata.ref_type directly they should probably be changed to use the function instead.

pereman2:
> Instead of storing Dict[str, str] as the ref_type, you can store Dict.
> get_ref_type() can construct the proper ref type when it's being called.
> if anything is using _metadata.ref_type directly they should probably be changed to use the function instead.

In the tests to check if it's the correct ref_type I used `get_ref_type()` so, that means if we change access to ref_type to be always using the function it should work. I'll remove the copy lines.

Do I open another issue involving changing every access to ref_type to use the function `get_ref_type`?

omry:
I see it as an alternative solution for this same issue.
If you can implement it you can just push it to this PR.

pereman2:
> I see it as an alternative solution for this same issue.
> If you can implement it you can just push it to this PR.

I'll try it here.

omry:
Added a comment.

On second though, let's do the followup fix in a different diff.
I think we will want this patched in 2.0 as well.
for 2.0 (not 2.1) we should probably go with a minimal fix, which is what you have here.
This fix will be easier to backport to the 2.0 branch.

Once you address my comment I am going to merge it as is, you can have a followup that is cleaning up (changing usage of metadata.ref_type) in 2.1.

pereman2:
Okay, do I open an issue involving the change of access to ref_type?

tests/__init__.py
omry: ref_type for structured configs is determined by the annotation. I think that if there is no annotation it defaults to the dataclass itself.

What are you actually seeing?

news/407.bugfix
```diff
@@ -0,0 +1 @@
+Fix Structured Configs with subscripted Dict and List fields failed pickling on python 3.6.
```
omry, line 1:
```suggestion
Fix pickling of Structured Configs with fields annotated as Dict[KT, VT] or List[T] on Python 3.6.
```

omry (changes requested):

omegaconf/basecontainer.py
```diff
@@ -89,28 +89,28 @@ def __repr__(self) -> str:
...
         dict_ = copy.copy(self.__dict__)
         dict_["_metadata"] = copy.copy(dict_["_metadata"])
         if self._has_ref_type():
-            if is_dict_annotation(self._metadata.ref_type):
+            ref_type = self._metadata.ref_type
+            if is_dict_annotation(ref_type):
                 dict_["_metadata"].ref_type = Dict
-            else:
+            elif is_list_annotation(ref_type):
                 dict_["_metadata"].ref_type = List
```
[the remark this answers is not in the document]
pereman2: The third option the should be a pass since we don't want nothing to be modified in that case.

tests/__init__.py
pereman2: `DictClass(Dict[str, int])` ref_type is DictSubclass instead of `Dict[str, int]`

```python
@dataclass
class DictSubclass(Dict[str, int]):
    pass
from omegaconf._utils import get_ref_type
x = OmegaConf.structured(DictSubclass())
print(x._metadata.ref_type)
print(get_ref_type(x))

<class '__main__.DictSubclass'>
typing.Union[__main__.DictSubclass, NoneType]
``

Comment (omry) [reply tests/__init__.py:156]: This seems correct to me.

File: omegaconf/basecontainer.py [hunk 7]
```diff
@@ -89,10 +89,30 @@ def __repr__(self) -> str:
...
     def __getstate__(self) -> Dict[str, Any]:
-        return self.__dict__
+        dict_ = copy.copy(self.__dict__)
+        dict_["_metadata"] = copy.copy(dict_["_metadata"])
+        if self._has_ref_type():
+            ref_type = self._metadata.ref_type
+            if is_dict_annotation(ref_type):
+                dict_["_metadata"].ref_type = Dict
+            elif is_list_annotation(ref_type):
+                dict_["_metadata"].ref_type = List
```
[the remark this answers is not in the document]
omry: Can you make this change?

omegaconf/basecontainer.py
```diff
@@ -89,10 +89,30 @@ def __repr__(self) -> str:
...
+
+        if isinstance(self, DictConfig):
+            key_type = d["_metadata"].key_type
+        element_type = d["_metadata"].element_type
+        ref_type = d["_metadata"].ref_type
+        if is_generic_dict(ref_type):
+            d["_metadata"].ref_type = Dict[key_type, element_type]  # type: ignore
+        elif is_generic_list(ref_type):
+            d["_metadata"].ref_type = List[element_type]  # type: ignore
+
```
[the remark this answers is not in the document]
omry: and here too?

omry (changes requested):

tests/__init__.py
omry: Which is exaclty what I was expecting.

omegaconf/basecontainer.py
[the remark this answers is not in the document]
pereman2: The third option the should be a pass since we don't want anything to be modified in that case.

omegaconf/basecontainer.py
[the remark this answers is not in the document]
pereman2: same comment as above

omegaconf/basecontainer.py
```diff
@@ -91,12 +92,14 @@ def __repr__(self) -> str:
     def __getstate__(self) -> Dict[str, Any]:
         dict_ = copy.copy(self.__dict__)
```
omry, line 93: for clarity, can you rename this to `dict_copy` ?

omry (approved):
one nit about variable name, otherwise ready to merge.

omegaconf/basecontainer.py
pereman2: @omry done.

PR #416, opened 2020-10-22
Fix Structured Config assignment to Dict annotation raising wrong error.

### LINKED ISSUE #410: Assignment of Structured Config to Subscripted Dict raising wrong error.
**Describe the bug**
Assigning a structured config into a Subscripted Dict error message is not representing correctly the action.
**To Reproduce**

```python
@dataclass
class DictClass:

@dataclass
class User:
    age: int = 4
    name: str = "foo"
cfg.foo = User()
ValidationError: Invalid type assigned : str is not a subclass of int. value: foo
        full_key: foo.name
        reference_type=Dict[str, int]
        object_type=dict
```
**Expected behaviour**
Structured config assignment to subscripted dicts should't be allowed and should have an descriptive error message.

**Additional context**
- [x] Python version: 3.8

pereman2:
Close #410 .
With this fix check_assign_value_node is no longer needed.
Added a test in test_error to check the error message and, added test cases for test_assign_wrong_type_to_list and test_assign_wrong_type_to_dict for completion.


pereman2 (commented):
I've noticed that it keeps modifying the dict1 value in the test cases `{"foo": True}` and `{"user": User(age=1, name="foo")}` (test_assign_wrong_type_to_dict) because when setting dict1 value to one of these it first creates a empty dict and then starts assigning its values if it passes validate_set.
```python
class DictSubclass:
    dict1: Dict[str, int] = ...
cfg.OmegaConf.structured(DictSubclass)
cfg.dict1 = {"foo": True} 
# 1. {"foo": True} valid assignment so it creates {}
# 2. tries to assign True to {} but it's not a valid value_type so, it raises error
# 3. {} stays as new dict1 value
```

So if validate_set encounters an invalid assignment it should be able to recover somehow with a cache for example.

omegaconf/dictconfig.py
```diff
@@ -597,9 +584,9 @@ def _set_value(self, value: Any, flags: Optional[Dict[str, bool]] = None) -> Non
             elif isinstance(value, dict):
                 for k, v in value.items():
                     self.__setitem__(k, v)
-            else:
+            else:  # pragma: no cover
```
omry, line 587: Why is this not covered?

pereman2: No reason, I just moved this comment from the sentence below there because it made more sense.

omry: this comment says that test coverage is not considered this line as needed to be covered.
it's not a comment, it's a directive to the code coverage.
try to remove it and see if we still get 100% coverage.
if not, try to add a test that will make it get to 100%.
if it's impossible, please explain why.

pereman2: It's impossible, if it is a structured config the following code will check it.
```python
if (
     is_assign
     and is_container_annotation(target_type)
     and not is_container_annotation(value_type)
 ):
```
If it isn't a structured config it will wrap a new node anyways.

omry: If this code never runs, why do we need it?

pereman2: We don't. Well to have a default case could be useful sometime but now, no.

omry: ok. I guess we can keep it.
If there is no reason, keep the pragma unchanged.

omegaconf/dictconfig.py
```diff
@@ -179,10 +178,6 @@ def _validate_set_merge_impl(self, key: Any, value: Any, is_assign: bool) -> Non
         if value == "???":
             return
 
-        if is_assign and isinstance(value, ValueNode) and self._has_element_type():
-            self._check_assign_value_node(key, value)
-            return
-
```
omry, line 185: While I like removing unneeded functions, it looks like this function is checking something different (and also producing a different error).
can you explain this change more?

pereman2: Validate set for structured config before was checked with its value nodes now, if we are assigning a structured config to a subscripted dict it will raise an error so, this will never be needed.

omegaconf/dictconfig.py
pereman2: Okay

omegaconf/dictconfig.py
omry: Gotcha.

omry (approved):
Looking good, I see that you also removed the no cover, meaning this case is now covered by tests.

PR #417, opened 2020-10-23
Fix assignment of primitive type to Subscripted Dict with wrong error message.

### LINKED ISSUE #409: Assignment of primitive type to subscripted Dict modifying the value.
**Describe the bug**
Assigning an int into a Subscripted Dict raises an error without a descriptive message also, it modifies the value after it to an empty dict.
**To Reproduce**

```python
@dataclass
class DictClass:

@dataclass
class User:
    age: int = 4
    name: str = "foo"

cfg.foo = 10
```
Output 
```
ValidationError: 
        full_key: foo
        reference_type=Optional[DictClass]
        object_type=DictClass
```
After the error cfg.foo is modified:
`print(cfg.foo) # prints '{}'`

**Expected behaviour**
`cfg.foo = 1` should raise a descriptive error message.

**Additional context**
- [x] Python version: 3.8

pereman2:
Close #409 .


pereman2:
This type of assignment modifies the value anyways as with #416 . I think catching the exception here and then restoring to it's previous state must work. Do I change this on another diff? It matters for both this PR and #416 .
```python
elif isinstance(value, dict):
    for k, v in value.items():
          self.__setitem__(k, v)
```

news/409.bugfix
```diff
@@ -0,0 +1 @@
+Assigning a primitive type to a Subscripted Dict now raises a descriptive message.
```
omry, line 1: This is two different bugs.
1. Bad error message
2. Mutating the container on exception

I would say that the second one is actually much worse, but you seems to ignore it here.

omegaconf/dictconfig.py
```diff
@@ -574,6 +574,7 @@ def _set_value(self, value: Any, flags: Optional[Dict[str, bool]] = None) -> Non
             self.__dict__["_content"] = "???"
             self._metadata.object_type = None
         else:
+            previous_state = copy.copy(self.__dict__["_content"])
```
omry, line 577: Why do you need to copy it? it's expensive and I am not seeing the value in that.

omegaconf/dictconfig.py
```diff
@@ -596,10 +597,15 @@ def _set_value(self, value: Any, flags: Optional[Dict[str, bool]] = None) -> Non
 
             elif isinstance(value, dict):
                 for k, v in value.items():
-                    self.__setitem__(k, v)
+                    try:
+                        self.__setitem__(k, v)
+                    except ValidationError as e:
+                        self.__dict__["_content"] = previous_state
+                        raise e
+
```
omry, line 605: You can do that on any exception.

omry (changes requested):

omegaconf/dictconfig.py
pereman2: Content is changed to {}, that means if it isn't copied it will still be modified.

news/409.bugfix
pereman2: Do you want another diff for this ? Or do I change this news fragment?

omry: No need for another diff, but create two issues and have two news fragments in your diff.

omegaconf/dictconfig.py
omry: copying the entire subtree on assignment is expensive.
content is ASSIGNED {}, if you keep a reference to the previous one is it changing? is this something you have actually seen or are you just thinking it will change?

omry: ah, I got it.
I think this is probably okay because it's a shallow copy, but please do some benchmarking of the initialization of OmegaConf from a fairly big dictionary to ensure this does not introduce a regression speed. we should over time improve performance or at least not make it slower.

pereman2: With a dict of 1M key-value pair.

```python
@dataclass
class DictClass:

cols = [str(i) for i in range(1000000)]
data = [i for i in range(1000000)]
dict_ = dict(zip(cols, data))

t1 = datetime.now()
cfg.foo = dict_
t2 = datetime.now()
cfg.foo = {"a": 1}
t3 = datetime.now()

print("Execution time: ", (t2-t1))
print("Assigning new value(it'll shallow copy content) time: ", t3-t2)
```
**Removing the shallow copy**
image

**With the shallow copy**
image
As you can see there isn't a significative difference between both implementations.

omry: Check a dictionary with some nesting.
How about 10 levels, each one with two splits (should be about ~1000 leaf nodes).

```yaml
tree:
  left:
    left: 1
    right: 1
  right:
    left: 1
    right: 1
```

omry, line 577: 1. Add a comment saying this is shallow copy.
2. better variable names are `content_backup` or `prev_content`.

pereman2: With and without it causes segmentation fault.

omry: also, take an average of many iterations.

omry (changes requested):
see comments

omegaconf/dictconfig.py
pereman2: 1024 leaf nodes
**Without copy**
image
**With copy**
image
It looks like it doesn't change anything. I can do further statistics if you prefer.

omry: No, it sounds good. can you paste your script here or in a GitHub gist?

pereman2: https://gist.github.com/pereman2/1b9069bf68a2df2a8023dcf34fd4147d

omry: Your benchmark is checking assignment, my concern is about initialization of configs like that one.

pereman2: I'll test creating.

tests/structured_conf/test_structured_config.py
```diff
@@ -787,6 +787,7 @@ def test_assign_wrong_type_to_dict(self, class_type: str, value: Any) -> None:
         cfg = OmegaConf.structured(module.ConfigWithDict2)
         with pytest.raises(ValidationError):
             cfg.dict1 = value
+        cfg == OmegaConf.structured(module.ConfigWithDict2)
```
pereman2, line 790: I've just noticed that the content_backup was wrong and I didn't see this stupid line of code. I'll check it tomorrow got a test test first in the morning.

omry: oops.

pereman2: I fixed it, the exception is picked at `__setitem__` so it made sense to restore the backup there. I've made previous_content assigning _content to it instead of copying.

pereman2: check this

omegaconf/dictconfig.py
```diff
@@ -574,7 +575,8 @@ def _set_value(self, value: Any, flags: Optional[Dict[str, bool]] = None) -> Non
             self.__dict__["_content"] = "???"
             self._metadata.object_type = None
         else:
-            previous_state = copy.copy(self.__dict__["_content"])
+            # shallow copy of content
+            self.__dict__["_previous_content"] = self.__dict__["_content"]
```
omry, line 579: I think we should absolutely not go this way.
Set should be atomic, you are introducing a new object state change here.

omry (changes requested):

news/409.bugfix
omry: Can you add a second news fragment?
If you want to use the same issue you can name them 409.bugfix.1 and 409.bugfix.2.

omegaconf/dictconfig.py
```diff
@@ -596,10 +602,24 @@ def _set_value(self, value: Any, flags: Optional[Dict[str, bool]] = None) -> Non
...
-                raise ValidationError(msg=msg)  # pragma: no cover
+                raise ValidationError(msg)  # pragma: no cover
+
+    def _restore_backup(self) -> None:
+        if (
+            "_previous_content" in self.__dict__
+            and self.__dict__["_previous_content"] is not None
+        ):
+            self.__dict__["_content"] = self.__dict__["_previous_content"]
+            del self.__dict__["_previous_content"]
```
omry, line 622: remove this function.

omegaconf/dictconfig.py
```diff
@@ -574,6 +574,8 @@ def _set_value(self, value: Any, flags: Optional[Dict[str, bool]] = None) -> Non
             self.__dict__["_content"] = "???"
             self._metadata.object_type = None
         else:
+            # shallow copy of content
```
omry, line 577: instead of sprinkling this function with error handling, you can do something like:
```python
def _set_value(self, value: Any, flags: Optional[Dict[str, bool]] = None) -> None:
  try:
     # copy backup
     self._set_value_impl(...)
  except:
    # restore backup
    raise

def _set_value_impl(self, value: Any, flags: Optional[Dict[str, bool]] = None) -> None:
   ... # same logic as the current function before your changes.
```

omry (changes requested):
See comments.
Also, you have CI failures.

omegaconf/dictconfig.py
pereman2: Makes more sense.

omry (approved):
This is much better, thanks.

PR #424, opened 2020-10-28
Use get_type_hints() to resolve string annotations

### LINKED ISSUE #303: [Bug] Crash when using postponed evaluation of annotations
# 🐛 Bug

Structured configs seem not to work with the upcoming postponed evaluation of annotations (https://www.python.org/dev/peps/pep-0563/) which will be the default in Python 3.10.

Obviously not really a problem yet but flagging it now as a FYI.

## To reproduce

** Minimal Code/Config snippet to reproduce **

```
from __future__ import annotations

@dataclass
class Config:
    a: int = 0

```

** Stack trace/error message **
```
  File "test2.py", line 10, in <module>
  File "/private/home/odelalleau/.conda/envs/py37-hydra/lib/python3.7/site-packages/omegaconf/omegaconf.py", line 134, in structured
    return OmegaConf.create(obj, parent)
  File "/private/home/odelalleau/.conda/envs/py37-hydra/lib/python3.7/site-packages/omegaconf/omegaconf.py", line 171, in create
    return OmegaConf._create_impl(obj=obj, parent=parent)
  File "/private/home/odelalleau/.conda/envs/py37-hydra/lib/python3.7/site-packages/omegaconf/omegaconf.py", line 220, in _create_impl
    element_type=element_type,
  File "/private/home/odelalleau/.conda/envs/py37-hydra/lib/python3.7/site-packages/omegaconf/dictconfig.py", line 74, in __init__
    self._set_value(content)
  File "/private/home/odelalleau/.conda/envs/py37-hydra/lib/python3.7/site-packages/omegaconf/dictconfig.py", line 546, in _set_value
    data = get_structured_config_data(value)
  File "/private/home/odelalleau/.conda/envs/py37-hydra/lib/python3.7/site-packages/omegaconf/_utils.py", line 258, in get_structured_config_data
    return get_dataclass_data(obj)
  File "/private/home/odelalleau/.conda/envs/py37-hydra/lib/python3.7/site-packages/omegaconf/_utils.py", line 202, in get_dataclass_data
    ref_type=type_, is_optional=is_optional, key=name, value=value, parent=None,
  File "/private/home/odelalleau/.conda/envs/py37-hydra/lib/python3.7/site-packages/omegaconf/omegaconf.py", line 681, in _maybe_wrap
    key=key,
  File "/private/home/odelalleau/.conda/envs/py37-hydra/lib/python3.7/site-packages/omegaconf/omegaconf.py", line 641, in _node_wrap
    elif issubclass(type_, Enum):
TypeError: issubclass() arg 1 must be a class
```

thomkeh:
Closes #303 

When using `from __future__ import annotations` (which will become default behavior in Python 3.10), type annotations are stored as strings. The standard library provides `typing.get_type_hints()` to convert these strings to types.

For the tests, I simulated this behavior by wrapping the type annotations in quotes:

```python
@dataclass
class Foo:
    a: "int"
```
This has the same effect and also works on Python 3.6. (The future import was introduced in Python 3.7.)

I was a bit unsure about the tests. Let me know if I can improve them.

### DIFF
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -4,7 +4,7 @@ import re
 import string
 import sys
 from enum import Enum
-from typing import Any, Dict, List, Match, Optional, Tuple, Type, Union
+from typing import Any, Dict, List, Match, Optional, Tuple, Type, Union, get_type_hints
 
 import yaml
 
@@ -225,9 +225,10 @@ def get_dataclass_data(
     flags = {"allow_objects": allow_objects} if allow_objects is not None else {}
     dummy_parent = OmegaConf.create({}, flags=flags)
     d = {}
+    resolved_hints = get_type_hints(get_type_of(obj))
     for field in dataclasses.fields(obj):
         name = field.name
-        is_optional, type_ = _resolve_optional(field.type)
+        is_optional, type_ = _resolve_optional(resolved_hints[field.name])
         type_ = _resolve_forward(type_, obj.__module__)
 
         if hasattr(obj, name):
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -224,6 +224,18 @@ def test_is_dataclass(mocker: Any) -> None:
     assert not _utils.is_dataclass(10)
 
 
+def test_resolve_types() -> None:
+    # simulate `from __future__ import annotations` by using strings for the types
+    @dataclass
+    class Foo:
+        a: "int" = "2"  # type: ignore
+        b: "float" = "0.0"  # type: ignore
+        c: "str" = 2  # type: ignore
+
+    assert _utils.get_structured_config_data(Foo) == {"a": 2, "b": 0.0, "c": "2"}
+    assert _utils.get_structured_config_data(Foo()) == {"a": 2, "b": 0.0, "c": "2"}
+
+
 def test_is_attr_class(mocker: Any) -> None:
     @attr.s
     class Foo:


thomkeh:
I don't think the pre-commit.ci failure is related?

omry:
Looks good.
1. Please add a news fragment in the news directory (similar to what is described [here](https://hydra.cc/docs/development/documentation#news-entries)).
2. pre-commit.ci error is unrelated, don't worry about it.
3. Can you point me to some docs saying using forward (quoted) types is simulating the issue?

thomkeh:
1. Done.
2. 👍 
3. Quoted types might not be 100% identical to postponed annotations. This is what I found:

The [relevant PEP](https://www.python.org/dev/peps/pep-0563/) says:

> This PEP proposes changing function annotations and variable annotations so that they are no longer evaluated at function definition time. Instead, they are preserved in `__annotations__` in **string form**.

and 

> This PEP is meant to solve the problem of forward references in type annotations.

So, postponed annotations certainly remove the need for quoted forward references. [The PEP also makes clear](https://www.python.org/dev/peps/pep-0563/#resolving-type-hints-at-runtime) that using `get_type_hints()` solves the problem:

> For code that uses type hints, the `typing.get_type_hints(obj, globalns=None, localns=None)` function correctly evaluates expressions back from its string form. Note that all valid code currently using `__annotations__` should already be doing that since a type annotation can be expressed as a string literal.

Additionally, [the mypy documentation](https://mypy.readthedocs.io/en/stable/common_issues.html) states:

> If you are running Python 3.7+ you can use `from __future__ import annotations` as a (nicer) alternative to string quotes, read more in PEP 563.

If this doesn't convince you, I would need to add a test that's only run on Python 3.7+ (where the future import is available). Tell me what you think!

omry:
I am not sure exactly how `from __future__ import annotations` works to enable future compatibility.
does it have side effects or do you need to use it directly for the import to have the effect?

One possibility is to have testing that depends on the Python version, but in the presence of side effects from that import (as I suspect is the case), I am not sure that this is a good idea.

Related:
What would be the implications of `from __future__ import annotations` on the execution path of OmegaConf import?
Will there be behavior changing side effects beyond the scope of OmegaConf itself (that may cause issues to user's code?)

thomkeh:
In the [announcement of Python 3.7](https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations) (where the future import was introduced), it says:

> Instead of compiling code which executes expressions in annotations at their definition time, the compiler stores the annotation in a string form equivalent to the AST of the expression in question. If needed, annotations can be resolved at runtime using `typing.get_type_hints()`. In the common case where this is not required, the annotations are cheaper to store (since short strings are interned by the interpreter) and make startup time faster.

So this does seem to affect some deeper level of the compiler.

---

> Will there be behavior changing side effects beyond the scope of OmegaConf itself (that may cause issues to user's code?)

I don't think so. In my experience, it seems to work on a per-module level. Meaning that it doesn't matter if you import a future activated module from one that is not activated, or vice-versa. I think this is also how the `from __future__ import print_function` worked in Python 2 (which affects the parser).

If you have two modules **a.py** and **b.py** and they look like this:

**a.py**
```python
def f(a: unknown_name):
    pass
```
**b.py**
```python
from __future__ import annotations
from a import f
```
and you run **b.py**, it throws an error:
```
  File "b.py", line 2, in <module>
    from a import f
  File "/Users/tk324/dev/python/a.py", line 3, in <module>
    def f(a: unknown_name):
NameError: name 'unknown_name' is not defined
```
but if you change **a.py** to be
```python
from __future__ import annotations
def f(a: unknown_name):
    pass
```
it works. So the effect of the future import seems to be locally restricted.

I really tried finding more detailed documentation for this, but wasn't much I could find.

In general though, I think OmegaConf should use `typing.get_type_hints()` in any case. Even if it doesn't solve #303 .

omry:
Cool, thanks for testing.
This opens up the option of creating a specific test file module to test the behavior with future annotations directly instead of (or in addition to) simulating it the way you did.

Can you add such a test? you can condition it con Python 3.7+ before the import.

thomkeh:
I changed the test. Is this what you had in mind? A major difficulty is that `from __future__ import annotations` has to be the first code in a module, otherwise it throws a syntax error.

news/303.bugfix
```diff
@@ -0,0 +1 @@
+Fix type error that occurs when using postponed evaluation of annotations
```
omry, line 1:
```suggestion
Support for Python 3.10 postponed annotation evaluation
```

Also, this is a feature, not a bugfix.
rename accordingly.

omry (changes requested):
Awesome.
One nit with the news file.

thomkeh:
Done!

PR #434, opened 2020-11-08
Fix #433 bug where ListConfig is modified after assignment of wrong value.

### LINKED ISSUE #433: Assignment of invalid value to ListConfig modifies the value.
**Describe the bug**
Assigning a list with invalid values to a ListConfig keeps the content modified after raising the error.
**To Reproduce**
```python
    @dataclass
    class ListClass:
        list: List[int] = field(default_factory=lambda: [1])
    cfg = OmegaConf.structured(ListClass)
    try:
        cfg.list = [1, 2, "foo"]
    except:
        pass
    print(cfg.list) # output [1, 2]

```

**Expected behavior**
cfg.list should return to be [1] 
**Additional context**
- [x] Python version: 3.8

pereman2:
Closes #433 .

### DIFF
--- a/news/433.bugfix
+++ b/news/433.bugfix
@@ -1,0 +1,1 @@
+Fix assignment of an invalid value to a ListConfig to raise an exception without modifying the config object
--- a/news/433.bugfix
+++ b/news/433.bugfix
@@ -1,1 +1,1 @@
-Fix assignment of an invalid value to a ListConfig to raise an exception without modifying the config object
+Fix bug where assignment of an invalid value to a ListConfig raised an exception but left the object modified.
--- a/omegaconf/listconfig.py
+++ b/omegaconf/listconfig.py
@@ -504,6 +504,16 @@ class ListConfig(BaseContainer, MutableSequence[Any]):
         return False
 
     def _set_value(self, value: Any, flags: Optional[Dict[str, bool]] = None) -> None:
+        try:
+            previous_content = self.__dict__["_content"]
+            self._set_value_impl(value, flags)
+        except Exception as e:
+            self.__dict__["_content"] = previous_content
+            raise e
+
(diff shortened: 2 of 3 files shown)


news/433.bugfix
```diff
@@ -0,0 +1 @@
+Fix assignment of an invalid value to a ListConfig to raise an exception without modifying the config object
```
omry, line 1: Describe the bug this is fixing, not the fix.
```suggestion
Fix bug where assignment of an invalid value to a ListConfig raised an exception but left the object modified.
```

omegaconf/listconfig.py
```diff
@@ -504,6 +504,16 @@ def __contains__(self, item: Any) -> bool:
         return False
 
     def _set_value(self, value: Any, flags: Optional[Dict[str, bool]] = None) -> None:
+        try:
+            previous_content = self.__dict__["_content"]
```
omry, line 508: interesting that there is no need for a shallow copy here (unlike the DictConfig). how come?

omry (changes requested):
Nice. a couple off comments.

omegaconf/listconfig.py
pereman2: `self.__dict__["_content"] = {}`allocates a new pointer thus making `previous_content` able to be assigned the last content state without needing a copy. I forgot to ask you if I should copy this behaviour in dictconfig in this same diff.

omry: Not sure I follow what behavior.
maybe separate PR to clean DictConfig's shallow copy?

omry (approved):

omegaconf/listconfig.py
pereman2: Okay

Issue #427, opened 2020-10-29
Nested Containers in structured config type annotation


gibiansky:
**Describe the bug**
Using Dict[str, List[str]] or other similar complex value types fails.

**To Reproduce**
```python
from dataclasses import dataclass
from typing import Dict, Tuple, List

from omegaconf import OmegaConf, MISSING

@dataclass
class Config:
    # If I change List[str] to the following, same error:
    #   Tuple[str, str]
    #   Tuple[str, ...]
    #   Dict[str, str]
    data: Dict[str, List[str]] = MISSING

# omegaconf.errors.ValidationError: Unsupported value type : typing.List[str]
OmegaConf.structured(Config)
```

**Expected behavior**
It should work.

Instead you get Validation error (on 2.0) and the following on master:
```
Traceback (most recent call last):
  File "test.py", line 15, in <module>
    OmegaConf.structured(Config)
  File "/Users/gibiansky/env/lib/python3.7/site-packages/omegaconf/omegaconf.py", line 145, in structured
    return OmegaConf.create(obj, parent, flags)
  File "/Users/gibiansky/env/lib/python3.7/site-packages/omegaconf/omegaconf.py", line 201, in create
    flags=flags,
  File "/Users/gibiansky/env/lib/python3.7/site-packages/omegaconf/omegaconf.py", line 254, in _create_impl
    flags=flags,
  File "/Users/gibiansky/env/lib/python3.7/site-packages/omegaconf/dictconfig.py", line 105, in __init__
    format_and_raise(node=None, key=None, value=None, cause=ex, msg=str(ex))
  File "/Users/gibiansky/env/lib/python3.7/site-packages/omegaconf/dictconfig.py", line 89, in __init__
    self._set_value(content, flags=flags)
  File "/Users/gibiansky/env/lib/python3.7/site-packages/omegaconf/dictconfig.py", line 552, in _set_value
    raise e
  File "/Users/gibiansky/env/lib/python3.7/site-packages/omegaconf/dictconfig.py", line 549, in _set_value
    self._set_value_impl(value, flags)
  File "/Users/gibiansky/env/lib/python3.7/site-packages/omegaconf/dictconfig.py", line 580, in _set_value_impl
    allow_objects=self._get_flag("allow_objects"),
  File "/Users/gibiansky/env/lib/python3.7/site-packages/omegaconf/_utils.py", line 312, in get_structured_config_data
    return get_dataclass_data(obj, allow_objects=allow_objects)
[...]

PR #443, opened 2020-11-26
Support for containers as element_type

### LINKED ISSUE #436: Tuple in a List (and probably Dict) does not work
```python
@dataclass
class A:
  a: Union[str, float]
  b: List[Tuple[int, int, int]]
```

Trying to use those types leads to errors like:
Unsupported value type : typing.Tuple[int, int, int]

While Union may be debatable, something like Tuple in the lists should be relatively easy to support?

pereman2:
I don't know if this should close #436 and #427 since both mention Tuple which still isn't implemented as a container. 

Maybe I should move merge tests to test_merge? 

With respect to the `should_assign`, I think it's necessary because if `input_config` it's true and element_type is a container it assigns instead of wrapping.


omegaconf/basecontainer.py
```diff
@@ -348,7 +348,8 @@ def expand(node: Container) -> None:
                     with open_dict(dest):
                         dest[key] = src._get_node(key)
                 else:
-                    dest[key] = src._get_node(key)
+                    src_node = src._get_node(key)
+                    dest[key] = src_node
```
omry, line 352: undo this no-op change.

omegaconf/basecontainer.py
```diff
@@ -433,6 +434,8 @@ def _set_item_impl(self, key: Any, value: Any) -> None:
 
         from .nodes import AnyNode, ValueNode
 
+        if isinstance(value, AnyNode):
+            value = value._value()
```
omry, line 438: Please explain why you need it, it seems out of place.
The following block is differentiating between Nodes (isintance(value, Node)) and Containers (the else branch).
Your change seems to be causing AnyNodes to be treated as containers. I don't understand why it's needed.

omegaconf/listconfig.py
```diff
@@ -102,17 +103,32 @@ def _validate_set(self, key: Any, value: Any) -> None:
...
-                    f"subclass of {type_str(target_type)}. value: {value}"
-                )
-                raise ValidationError(msg)
+                raise_validation_error(value_type, target_type, value)
+        if (
+            is_structured_config(value_type)
+            and not is_structured_config(target_type)
+            and target_type not in (Any, None)
+        ):
+            raise_validation_error(value_type, target_type, value)
```
omry, line 131: this block is confused. can you reorganized it to be clearer? (adding a comment about which failure case each condition is detecting will also help).

tests/structured_conf/test_structured_config.py
```diff
@@ -1051,3 +1051,216 @@ def test_update_userlist(
...
+        "obj,cfg2",
+        [
+            (
+                "ContainerInList",
+                OmegaConf.create(
+                    {
+                        "list_with_list": [[4, "invalid_value"]],
+                        "list_with_dict": [{"foo": 4}],
+                    }
+                ),
```
omry, line 1174: Instead of passing and DictConfig pass the dict itself and wrap it in the test function.

omry (changes requested):

omegaconf/basecontainer.py
pereman2: oops

omegaconf/basecontainer.py
pereman2: I don't like this too, I'm going to look for a better way to deal with this. 
I made this hacky fix in order to deal with merge with a Config like `OmegaConf.create({"a": 4})` since `4` will be AnyNode and it cause to not check the value correctly.

tests/structured_conf/test_structured_config.py
```diff
@@ -1103,164 +1103,125 @@ def test_container_as_element_type_invalid(
...
     )
     def test_container_as_element_type_merge(
-        self, class_type: str, obj: str, cfg2: Any, expected: Any
+        self, class_type: str, obj: str, obj2: Any, expected: Any
     ) -> None:
         module: Any = import_module(class_type)
         class_ = getattr(module, obj)
         cfg = OmegaConf.merge(class_)
+        cfg2 = OmegaConf.create(obj2)
         res = OmegaConf.merge(cfg, cfg2)
```
omry: In fact, you can merge the dict directly so you don't even need to create here :)

omegaconf/basecontainer.py
```diff
@@ -490,6 +492,8 @@ def wrap(key: Any, val: Any) -> Node:
                 else:
                     is_optional = target._is_optional()
                     ref_type = target._metadata.ref_type
+            if input_config:
+                val = val._value()
             return _maybe_wrap(
                 ref_type=ref_type,
                 key=key,
```
omry, line 499: If input is a config why do you unwrap and wrap it?

omegaconf/dictconfig.py
```diff
@@ -158,17 +158,28 @@ def _validate_set(self, key: Any, value: Any) -> None:
...
+        )
+        value_type = OmegaConf.get_type(value)
+
+        is_valid_target = not target_has_ref_type and not (
+            (
+                is_structured_config(value_type)
+                and self._metadata.element_type is not Any
+            )
+            or (isinstance(value, AnyNode) and target_type is not Any)
+        )
```
omry, line 178: too much. add some temporary boolean variables to make this clearer.

omegaconf/listconfig.py
```diff
@@ -102,17 +103,26 @@ def _validate_set(self, key: Any, value: Any) -> None:
...
+
+            raise ValidationError(msg)
+
+        if (
+            is_structured_config(value_type)
+            or is_structured_config(target_type)
+            or isinstance(value, ValueNode)
+        ):
+            if value_type is not None and not issubclass(value_type, target_type):
+                raise_validation_error(value_type, target_type, value)
```
omry, line 125: This is bad form:

```python
if A:
  if B:
    ...
```

you can do:

```python
if A and B:
  ...
```

omegaconf/listconfig.py
```diff
@@ -234,10 +244,16 @@ def append(self, item: Any) -> None:
...
+                value = item._value()
+
             node = _maybe_wrap(
                 ref_type=self.__dict__["_metadata"].element_type,
                 key=index,
-                value=item,
+                value=value,
                 is_optional=OmegaConf.is_optional(item),
                 parent=self,
             )
```
omry, line 259: What is this doing?
It looks like if you get a DictConfig or ListConfig you use the content directly and wrap it again.
What is this achieving?

omry (changes requested):

omegaconf/basecontainer.py
pereman2: That's because maybe wrap doesn't really wrap nodes and we need to check if the container value(input_config) has the correct types given a ref_type.  

If we remove this the following test case fails(merging complex dicts):
```
(
   "ComplexDict", # Dict[str, Dict[str, List[int]]]
   {
        "dict": {"foo2": {"var2": ["invalid", 4]}},
   },
),
# merge with ComplexDict
```

tests/structured_conf/test_structured_config.py
pereman2: hmm, I didn't notice it was transformed into dictconfig when merging, good to know.

omegaconf/listconfig.py
pereman2: I thought in this case it would be clearer this way.

omegaconf/listconfig.py
pereman2: As I said above(and other PRs), if you merge two containers you have to deal with maybe_wrap which won't wrap nodes. 

We want to type check whether the contents of the value if we expect the value to be a container, with certain element_type and key_type. The solution I found to work here and other parts of the code(like input_config above) is to unwrap to check again.

omegaconf/listconfig.py
```diff
@@ -101,18 +102,31 @@ def _validate_set(self, key: Any, value: Any) -> None:
...
+
+        def raise_validation_error(
+            value_type: Any, target_type: Any, value: Any
+        ) -> None:
+            msg = (
+                f"Invalid type assigned : {type_str(value_type)} is not a "
+                f"subclass of {type_str(target_type)}. value: {value}"
+            )
+
+            raise ValidationError(msg)
```
omry, line 118: Adding this function here, which is only called once - is making it even harder to understand what you are actually changing here.
If it's only called once, what is the point of extracting it into a local function?

omegaconf/listconfig.py
```diff
@@ -101,18 +102,31 @@ def _validate_set(self, key: Any, value: Any) -> None:
...
+        if (
+            (
+                is_structured_config(value_type)
+                or is_structured_config(target_type)
+                or isinstance(value, ValueNode)
+            )
+            and value_type is not None
+            and not issubclass(value_type, target_type)
+        ):
+            raise_validation_error(value_type, target_type, value)
```
omry, line 129: Can you explain what new cases this is dealing with?
it's significantly more complicated than before and looking at it I suspect it's actually too broad.

omegaconf/listconfig.py
omry: A possibly cleaner solution is to treat the case of a DictConfig assigned onto a DictConfig by validating that the key type and the element types of the src are compatible with those in the target.

Maybe we can implement a function like:
```
is_legal_assignment(dest_type, src_type) -> bool
```

To handle that.
That function can be used to check if the type of a value can be assigned to a specific target type, but can also be used to check if `Dict[str, int]` can be assigned to `Dict[str, Any]` (yes) or to `Dict[int, int]` (no).

There is no need to re-wrap the DictConfig. we can know that all of keys and values are conforming to the specified key type and element type.

omegaconf/listconfig.py
```diff
@@ -234,10 +248,16 @@ def append(self, item: Any) -> None:
             index = len(self)
             self._validate_set(key=index, value=item)
 
+            element_type = self.__dict__["_metadata"].element_type
+            value = item
```
omry, line 252: What is this assignment achieves?

omegaconf/dictconfig.py
```diff
@@ -158,17 +158,28 @@ def _validate_set(self, key: Any, value: Any) -> None:
...
+        target_has_ref_type = isinstance(target, DictConfig) and target_type not in (
+            Any,
+            dict,
+        )
+        value_type = OmegaConf.get_type(value)
+
+        input_container = (
+            is_structured_config(value_type) and self._metadata.element_type is not Any
+        )
```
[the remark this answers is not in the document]
pereman2: Oops, it should be `input_structured_config`. If value_type is a structured and we have an element_type it should do `issubclass(value_type, target_type)` which didn't happen because this case was considered a "valid_value"

omegaconf/listconfig.py
pereman2: Looking back at it doesn't make sense. I think at one point I reused this function(while testing) but now not. I'll remove it.

omegaconf/listconfig.py
pereman2: Well if we have a structured config the way to check is with `issubclass` so that's one case. 

The case with ValueNodes is because the same reason with other PRs where we receive a node instead of a primitive value and, this is where we can check.

omegaconf/listconfig.py
pereman2: Do you mean this function should check the type of every value in DictConfig?

omegaconf/listconfig.py
omry: Sounds good.

omegaconf/listconfig.py
omry: Trying to parse the logic here:
```
        if (
            (
                is_structured_config(value_type)
                or is_structured_config(target_type)
                or isinstance(value, ValueNode)
            )
```

```
If:
  input is structured config (e.g an instance of or a dataclass)
  or element type is a structured config
  or value we are assigning a value node
```
This is already lumping multiple unrealted things together.

```
            and value_type is not None
```

and the value type is not a subclass of the element type.
=> fail.


To me a better way to express this sounds like:
```
if value is a structured config:
  if element type is a structured config and value type is not a subclass
    => fail
```

I don't know what the input being a value node has to do with it, shouldn't the check be performed on the unboxed input?

omegaconf/listconfig.py
omry: Please clear this assignment if it does not do anything useful.

omegaconf/listconfig.py
omry: I mean that a DictConfig can be assumed to be legal.
There was validation performed on it's input when it was created.
When you are assigning a DictConfig, you should not need to recursively re-validate it.

You can look at the key type and element types:
if they are matching (or more specific), the assignment is legal.

omegaconf/listconfig.py
pereman2: > Trying to parse the logic here:
> 
> ```
>         if (
>             (
>                 is_structured_config(value_type)
>                 or is_structured_config(target_type)
>                 or isinstance(value, ValueNode)
>             )
> ```
> 
> ```
> If:
>   input is structured config (e.g an instance of or a dataclass)
>   or element type is a structured config
>   or value we are assigning a value node
> ```
> 
> This is already lumping multiple unrealted things together.
> 
> ```
>             and value_type is not None
>             and not issubclass(value_type, target_type)
> ```
> 
> and the value type is not a subclass of the element type.
> => fail.
> 
> To me a better way to express this sounds like:
> 
> ```
> if value is a structured config:
>   if element type is a structured config and value type is not a subclass
>     => fail
> ```

I agree.
> 
> I don't know what the input being a value node has to do with it, shouldn't the check be performed on the unboxed input?

ValueNodes are a bit tricky because of implicit conversion. I think this process could be much more simplified with :
```
if value is not a subscripted container type and not subclass..
```

But since value nodes implicitly convert values we cannot. Therefore, we arrive to a point where in the following case 
```python
@attr.s(auto_attribs=True)
class ComplexDict:
    dict: Dict[str, Dict[str, List[int]]] = {"foo": {"var": [1, 2]}}

obj = {
          "dict": {"foo2": {"var2": ["invalid", 4]}},
      }
c = OmegaConf.structured(ComplexDict)
OmegaConf.merge(c, obj) # no error
```
This doesn't raise an error because "invalid" is an StringNode which cannot be checked its value without defining an explicit check in `_validate_set` and, if we check with the unboxed value(str literal) it can cause regression problems since, for example, the int value `4` can be StringNode too(`issubclass(int, str)` will fail but this is ok in omegaconf).

I don't know if I explained this correctly, I hope this one clarifies my concerns.

tests/structured_conf/test_structured_config.py
```diff
@@ -1070,3 +1070,176 @@ def test_merge_missing_list_promotes_target_type(self, class_type: str) -> None:
...
+                    "list": [[{"foo": User()}]],
+                },
+            ),
+            (
+                "ComplexDict",
+                {
+                    "dict": {"foo2": {"var2": ["invalid", 4]}},
+                },
+            ),
+        ],
```
omry, line 1235: try to remove trialing comma from lines like:
```
 "list": [[{"foo": "invalid"}]],
```

I will likely result in a more compact formatting.

omegaconf/listconfig.py
```diff
@@ -234,6 +243,11 @@ def append(self, item: Any) -> None:
             index = len(self)
             self._validate_set(key=index, value=item)
 
+            element_type = self.__dict__["_metadata"].element_type
+            expect_container = is_container_annotation(element_type)
+            if expect_container and isinstance(item, BaseContainer):
+                item = item._value()
+
```
omry, line 250: I am  not seeing new tests for deepcopy/copy that are testing with containers as element types.
Can you add some?

omegaconf/listconfig.py
```diff
@@ -101,18 +102,26 @@ def _validate_set(self, key: Any, value: Any) -> None:
...
+            )
+            and value_type is not None
+            and not issubclass(value_type, target_type)
+        ):
+            msg = (
+                f"Invalid type assigned : {type_str(value_type)} is not a "
+                f"subclass of {type_str(target_type)}. value: {value}"
+            )
+
+            raise ValidationError(msg)
```
omry, line 125: This still looks too complicated.
can you try to make it more readable by refacotring it a bit? (maybe extract local boolean values and use them in conditions like in DictConfig).

omry (changes requested):

omegaconf/listconfig.py
```diff
@@ -101,18 +102,25 @@ def _validate_set(self, key: Any, value: Any) -> None:
...
+        )
+        if (
+            needs_type_validation
+            and value_type is not None
+            and not issubclass(value_type, target_type)
+        ):
+            msg = (
+                f"Invalid type assigned : {type_str(value_type)} is not a "
+                f"subclass of {type_str(target_type)}. value: {value}"
+            )
```
pereman2, line 121: I extracted this variable. Do you think I should leave a message explaining why we need `needs_type_validation` ?

omegaconf/listconfig.py
```diff
@@ -252,12 +255,27 @@ def __setitem__(self, index: Union[int, slice], value: Any) -> None:
...
+            ):
+                item_ref_type = item._metadata.ref_type
+                if is_container_annotation(item_ref_type) and not is_legal_assignment(
+                    element_type, item_ref_type
+                ):
+                    _raise_invalid_assignment(element_type, item_ref_type, item)
+                else:
+                    # regular dict without annotation
+                    item = item._value()
+
```
pereman2, line 278: I added the function `is_legal_assignment` in utils.py to not re-wrap as you mentioned. Even though we check with `is_legal_assignment`, it's necessary a default case where we re-wrap in case it's a regular dict. For example:

The following snippet where `d`is a DictConfig without a Dict annotation, if there wasn't a default option where it re-wraps the DictConfig, it wouldn't be possible to find if it's a legal assignment.

```python3
@dataclass
class ContainerInList:
    list_with_list: List[List[int]] = field(default_factory=lambda: [[0, 1], [2, 3]])
    list_with_dict: List[Dict[str, int]] = field(
        default_factory=lambda: [{"foo": 1, "foo2": 2}]
    )
d = {"list_with_list": [[4, 5]], "list_with_dict": [{"foo": User()}]}
cfg = OmegaConf.structured(ContainerInList)
OmegaConf.merge(cfg, d) # no error
```

Re-wrapping wouldn't be necessary if somehow we built `ref_type`, `element_type`, `key_type` on runtime for DictConfigs like `d`. I don't know if this is impossible to do.

pereman2: Moreover, I added a test case example in 0b9cbfe which is the only one that covers this example. I'll wait for your review and then, complete with more test cases like the one in 0b9cbfe.

omegaconf/dictconfig.py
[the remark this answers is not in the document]
omry: self._metadata.element_type can also be a primitive type. (int, str etc).
I think you should reorganize the logic here a bit.

omegaconf/listconfig.py
```diff
@@ -101,18 +101,20 @@ def _validate_set(self, key: Any, value: Any) -> None:
...
-                and value_type is not None
-                and not issubclass(value_type, target_type)
-            ):
-                msg = (
-                    f"Invalid type assigned : {type_str(value_type)} is not a "
-                    f"subclass of {type_str(target_type)}. value: {value}"
-                )
-                raise ValidationError(msg)
+
+        needs_type_validation = is_structured_config(value_type) or isinstance(
```
omry, line 109: what if the input is a primitive, say `int`. does it mean that it does not need type validation?

omegaconf/basecontainer.py
```diff
@@ -747,6 +751,23 @@ def prepand(full_key: str, parent_type: Any, cur_type: Any, key: Any) -> str:
 
         return full_key
 
+    # If item has a Container annotation like Dict[str, int] it validates against the target element_type
+    # if it doesn't have an annotation it returns the raw value
+    def _resolve_container_assignment(self, item: Any) -> Any:
```
omry, line 756: This function is doing too much, which also leads to it having an opaque name.
It's doing some assignment validation (and raising an exception), and unwrappingg the object in some cases.

Assignment validation is done in validate_set. It's not clear to me why you want to perform a partial validation from various places.
Unwrapping can happen as a part of assignment. (_set_value() on DictConfig and ListConfig).

omegaconf/dictconfig.py
```diff
@@ -199,7 +211,7 @@ def _validate_set(self, key: Any, value: Any) -> None:
             and not issubclass(value_type, target_type)
         )
         if validation_error:
-            self._raise_invalid_value(value, value_type, target_type)
+            _raise_invalid_assignment(target_type, value_type, value)
```
omry, line 214: Is this a fix to https://github.com/facebookresearch/hydra/issues/1322 by any chance?

omegaconf/listconfig.py
```diff
@@ -252,12 +254,15 @@ def __setitem__(self, index: Union[int, slice], value: Any) -> None:
...
             from omegaconf.omegaconf import OmegaConf, _maybe_wrap
 
             index = len(self)
             self._validate_set(key=index, value=item)
 
+            item = self._resolve_container_assignment(item)
+
             node = _maybe_wrap(
                 ref_type=self.__dict__["_metadata"].element_type,
                 key=index,
```
omry, line 268: see, here for example the responsibility of _resolve_container_assignment() can be split between validate_set() above and _maybe_wrap() below.

tests/__init__.py
```diff
@@ -192,3 +192,21 @@ class InterpolationList:
...
+        default_factory=lambda: {"foo": {"var": 1}}
+    )
+
+
+@dataclass
+class ContainerInList:
+    list_with_list: List[List[int]] = field(default_factory=lambda: [[0, 1], [2, 3]])
+    list_with_dict: List[Dict[str, int]] = field(
+        default_factory=lambda: [{"foo": 1, "foo2": 2}]
+    )
```
omry, line 212: Can you also cover cases where the type is a structured config?

```
Dict[str, List[User]] 
Dict[str, Dict[str, User]] 
List[List[User]]
List[Dict[str, User]]
```

tests/structured_conf/test_structured_config.py
```diff
@@ -1144,3 +1144,191 @@ def test_merge_missing_list_promotes_target_type(self, class_type: str) -> None:
...
+        self, class_type: str, obj: str, obj2: Any, expected: Any
+    ) -> None:
+        module: Any = import_module(class_type)
+        class_ = getattr(module, obj)
+        cfg = OmegaConf.merge(class_)
+        res = OmegaConf.merge(cfg, obj2)
+        assert res == expected
+
+    @pytest.mark.parametrize(
+        "obj,obj2",
```
omry, line 1239: the first field here is a class name (class_type), not an object. looks like this mistake is repeated in all the new tests here.

tests/structured_conf/test_structured_config.py
```diff
@@ -1144,3 +1144,191 @@ def test_merge_missing_list_promotes_target_type(self, class_type: str) -> None:
...
+            ("ContainerInDict", {"users": {"default": 1}}, "dict_with_dict"),
+            ("ContainerInDict", {"users": MISSING}, "dict_with_dict"),
+            ("ContainerInDict", {"users": [1, 2]}, "dict_with_list"),
+            ("ContainerInDict", {"users": MISSING}, "dict_with_list"),
+            ("ComplexList", [[{"number": 1}]], "list"),
+            ("ComplexDict", {"users": {"data": [1, 2]}}, "dict"),
+        ],
+    )
+    def test_container_as_element_type_valid(
+        self, class_type: str, obj: str, value: Any, node: str
```
omry, line 1164: rename node to key

tests/structured_conf/test_structured_config.py
```diff
@@ -1144,3 +1144,191 @@ def test_merge_missing_list_promotes_target_type(self, class_type: str) -> None:
...
+            ("ContainerInDict", {"users": [1, User()]}, "dict_with_list"),
+            ("ComplexList", [[{"number": True}]], "list"),
+            ("ComplexList", [[{"number": User()}]], "list"),
+            ("ComplexDict", {"users": {"data": True}}, "dict"),
+            ("ComplexDict", {"users": {"data": User()}}, "dict"),
+        ],
+    )
+    def test_container_as_element_type_invalid(
+        self, class_type: str, obj: str, value: Any, node: str
+    ) -> None:
```
omry, line 1191: rename node to key

tests/structured_conf/test_structured_config.py
```diff
@@ -1144,3 +1144,191 @@ def test_merge_missing_list_promotes_target_type(self, class_type: str) -> None:
         c3 = OmegaConf.merge(c1, c2)
         with pytest.raises(ValidationError):
             c3.missing.append("xx")
+
+    @pytest.mark.parametrize(
```
omry, line 1148: This is a ton of new tests, good job.
Move them to a dedicated file though.

`test_structured_config_container_element.py`?

tests/test_basic_ops_dict.py
```diff
@@ -811,3 +812,20 @@ def test_self_assign_list_value_with_ref_type(c: Any) -> None:
...
+
+
+@pytest.mark.parametrize(
+    "node, value",
+    [
+        pytest.param("dict_with_dict", {"a": 2}, id="copy_dict_with_dict"),
+        pytest.param("dict_with_list", [3, 4], id="copy_dict_with_list"),
+    ],
+)
+def test_copy_container_in_dict(node: str, value: Any) -> None:
```
omry, line 824: node -> key

tests/test_basic_ops_list.py
```diff
@@ -623,3 +623,20 @@ def test_shallow_copy_none() -> None:
     c._set_value([1])
     assert c[0] == 1
     assert cfg._is_none()
+
+
+@pytest.mark.parametrize(
+    "node, value",
```
omry, line 629: node -> key

omry (changes requested):

omegaconf/basecontainer.py
pereman2: I'm going to tackle this between today and tomorrow.

omegaconf/dictconfig.py
pereman2: No, in dictconfig.py:629 
```python
else:  # pragma: no cover
    msg = f"Unsupported value type : {value}"
    raise ValidationError(msg)
```
should print type_str(type(value)). Should I fix it in this PR?

omry: I am not sure it's that simple. we can do it in a different pr (need a clean standalone repro first).

tests/structured_conf/data/attr_classes.py
```diff
@@ -548,6 +549,18 @@ class ContainerInList:
...
 
 
+@attr.s(auto_attribs=True)
+class ContainerInDictWithUser:
+    dict_with_list: Dict[str, List[User2]] = {"foo": [User2()]}
+    dict_with_dict: Dict[str, Dict[str, User2]] = {"foo": {"var": User2()}}
+
+
+@attr.s(auto_attribs=True)
+class ContainerInListWithUser:
```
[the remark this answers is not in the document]
pereman2: Does `ContainerWithUserInDict` look good for you ?

pereman2:
I've added more test cases in fa2c0a2 which are failing. This was after reviewing coverage output.

tests/structured_conf/data/attr_classes.py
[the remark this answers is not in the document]
omry: DictWithNestedUserContainer?

omegaconf/_utils.py
```diff
@@ -740,3 +746,73 @@ def is_generic_dict(type_: Any) -> bool:
...
+        is_legal = is_legal_list_assignment(dest_type, src_type)
+    elif is_dict_annotation(dest_type) and is_dict_annotation(src_type):
+        is_legal = is_legal_dict_assignment(dest_type, src_type)
+    return is_legal
+
+
+def is_legal_dict_assignment(dest_type: Any, src_type: Any) -> bool:
+    key_values_pair_dest = get_dict_key_value_types(dest_type)
+    key_values_pair_src = get_dict_key_value_types(src_type)
+    return key_values_pair_dest == key_values_pair_src
```
omry, line 804: Isn't this just `dest_type == src_type`?

omry (commented):
I have an overall feeling that this is more complicated than it should be, and that things gets more complicated as I am providing more feedback.

Maybe this is too much for a single PR and we should break it down vertically (support Dict, List and maybe Tuple), each with a standalone PR.
what do you think?

Jasha10:
Closing this as #890 implemented support for containers as `element_type`.

PR #321, opened 2020-07-25
[OmegaConf 2.1] Grammar for parsing of interpolations

odelalleau:
This is heavily WIP. Just putting it out there for visibility and to centralize discussions.

PR #444, opened 2020-11-26
More explicit error when trying to access the child of non-container node

odelalleau:
When using ${a.b}, if `a` is not a container node, the previous code was failing with an obscure AssertionError.

This commit was originally the first commit in the stack in #321 but was unrelated to other changes in that PR, so it makes more sense to create a different PR.

I don't think it's worth adding an issue and a news item for this (since it seems pretty minor) but let me know if I should.

Note: the motivation for adding `test_assertion_error()` is that otherwise coverage would drop below 100%.

### DIFF
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -307,7 +307,11 @@ class Container(Node):
                     throw_on_resolution_failure=throw_on_resolution_failure,
                 )
 
-            assert ret is None or isinstance(ret, Container)
+            if ret is not None and not isinstance(ret, Container):
+                raise ConfigKeyError(
+                    f"Error trying to access {key}: node `{'.'.join(split[0:i + 1])}` "
+                    f"is not a container and thus cannot contain `{split[i + 1]}``"
+                )
             root = ret
 
         if root is None:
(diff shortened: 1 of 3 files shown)


omry:
curious about test_assertion_error.
what line in the code is not covered unless you have that?

omry (approved):
regardless, this looks fine.
I will merge after you answer my question.

odelalleau:
> curious about test_assertion_error.
> what line in the code is not covered unless you have that?

This line: https://github.com/omry/omegaconf/blob/800e8a7ad30db9400b8332d556520a3ecee56b77/omegaconf/_utils.py#L606

omry:
Okay, let's keep that - it may come in handy next time there is some random assertion.

omry:
Generally, the contact of OmegaConf.select is that it returns None if the field is not found.
This fix is causing inconsistent behavior:
```python
In [14]: cfg
Out[14]: {'a': 10, 'b': None}

In [15]: print(OmegaConf.select(cfg, "foo.bar"))
None

In [16]: print(OmegaConf.select(cfg, "b.bar"))
...
ConfigKeyError: Error trying to access b.bar: node `b` is not a container and thus cannot contain `bar`
    full_key: b.bar
    object_type=dict
```

PR #452, opened 2020-12-10
Fix bug #450 where DictConfig's copy didn't properly copy primitive types.

### LINKED ISSUE #450: Shallow copy not working correctly for DictConfig
**Describe the bug**
It seems that DictConfig objects cannot be copied in python. I am not sure if this is expected behaviour or not.

**To Reproduce**
```
a = OmegaConf.create({"a": 1, "b": 2})
b = a.copy()
b["a"] = 42
print(b)
print(a)
```
Outputs:
```
{'a': 42, 'b': 2}
{'a': 42, 'b': 2}
```

### Simple fix
A simple fix would be to first cast the object to a dict:
```
a = OmegaConf.create({"a": 1, "b": 2})
b = dict(a).copy()
b["a"] = 42
print(b)
print(a)
```
Outputs:
```
{'a': 42, 'b': 2}
{'a': 1, 'b': 2}
```

This however only works for first level attributes, nested DictConfigs won't be casted to dict and will therefore be modified inplace.
For instance:
```
a = OmegaConf.create({"a": {"aa": 1}, "b": 2})
b = dict(a).copy()
b["a"]["aa"] = 42
print(b)
print(a)
```
Outputs:
```
{'a': {'aa': 42}, 'b': 2}
{'a': {'aa': 42}, 'b': 2}
```

This caused a downstream bug in fairseq: https://github.com/pytorch/fairseq/issues/3011

pereman2:
Closes #450 .

When copying a dictconfig it copied its `_content` which is a dict, that means that ValueNodes is kept as before(shallow copy of a dict).

### DIFF
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -120,10 +120,12 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
             res.__dict__[k] = copy.copy(v)
         res._re_parent()
         content = self.__dict__["_content"]
-        if content is not MISSING and content is not None:
+        if content != MISSING and content is not None:
+            print(content)
             for k, v in content.items():
                 if isinstance(v, ValueNode):
                     res.__dict__["_content"][k] = copy.copy(v)
+
         return res
 
     def copy(self) -> "DictConfig":
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -113,13 +113,17 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
         return res
 
     def __copy__(self) -> "DictConfig":
+        from .omegaconf import MISSING
+
         res = DictConfig(content=None)
         for k, v in self.__dict__.items():
             res.__dict__[k] = copy.copy(v)
         res._re_parent()
-        for k, v in self.__dict__["_content"].items():
-            if isinstance(v, ValueNode):
-                res.__dict__["_content"][k] = copy.copy(v)
+        content = self.__dict__["_content"]
+        if content is not MISSING and content is not None:
+            for k, v in content.items():
+                if isinstance(v, ValueNode):
+                    res.__dict__["_content"][k] = copy.copy(v)
         return res
 
     def copy(self) -> "DictConfig":
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -117,6 +117,9 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
         for k, v in self.__dict__.items():
             res.__dict__[k] = copy.copy(v)
         res._re_parent()
+        for k, v in self.__dict__["_content"].items():
+            if isinstance(v, ValueNode):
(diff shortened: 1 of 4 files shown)


omegaconf/dictconfig.py
```diff
@@ -117,6 +117,9 @@ def __copy__(self) -> "DictConfig":
         for k, v in self.__dict__.items():
             res.__dict__[k] = copy.copy(v)
         res._re_parent()
+        for k, v in self.__dict__["_content"].items():
+            if isinstance(v, ValueNode):
+                res.__dict__["_content"][k] = copy.copy(v)
```
omry, line 122: content can be None or MISSING, I think it will blow up here.

omry (commented):
Thanks.
See my comment.

omegaconf/dictconfig.py
pereman2: In this case copying Foo doesn't raise and error neither with None. Since we only copy.copy ValueNode.
```python
@dataclass
class Foo:
    a: Dict[str, int] = MISSING
    b: Any = 2
```

omry: I think you misunderstood me.
`__dict__["_content"]` itself can be None or '???'. In that case it will not have an items() method.

pereman2: But, `self.__dict__["_content"]` can only be none or missing if it's specified in a structured config, isn't it?

omry: Yes. but the fact that you were not able to reproduce the problem does not mean it's not there.

if _content is None,  the line `for k, v in self.__dict__["_content"].items():` will fail, right?

omry: Were you able to reproduce it?
It would be good to add a test that tests serializing such DictConfigs.

pereman2: Yes, done in 8fee3099f5c958004d106ddd3d5d8cd14f8d2248

pereman2: Do I add test for both `OmegaConf.save()` and pickle?

omry: save is converting to yaml, right?
does it even hit that code path?

omegaconf/dictconfig.py
```diff
@@ -113,10 +113,19 @@ def __deepcopy__(self, memo: Dict[int, Any] = {}) -> "DictConfig":
...
     def __copy__(self) -> "DictConfig":
+        from .omegaconf import MISSING
+
         res = DictConfig(content=None)
         for k, v in self.__dict__.items():
             res.__dict__[k] = copy.copy(v)
         res._re_parent()
+        content = self.__dict__["_content"]
+        if content != MISSING and content is not None:
+            print(content)
```
omry, line 124: remove the print.

omegaconf/dictconfig.py
pereman2: I think pickle doesn't hit that part too.

pereman2: It doesn't copy, got it. Done in 4881890

omry (approved):
Great, thanks!

PR #454, opened 2020-12-11
Dict int key type

Jasha10:
A preliminary stab at implementing int key-type for DictConfig.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,9 @@
+## 2.0.6 (2020-12-22)
+
+### Features
+
+- OmegaConf now supports `int` for dictionary key types ([#149](https://github.com/omry/omegaconf/issues/149)) 
+
 ## 2.0.5 (2020-11-11)
 
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,9 +1,3 @@
-## 2.0.6 (2020-12-22)
-
-### Features
-
-- OmegaConf now supports `int` for dictionary key types ([#149](https://github.com/omry/omegaconf/issues/149)) 
-
 ## 2.0.5 (2020-11-11)
 
 
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -491,7 +491,7 @@ def valid_value_annotation_type(type_: Any) -> bool:
 
 
 def _valid_dict_key_annotation_type(type_: Any) -> bool:
-    return type_ is None or type_ is Any or issubclass(type_, (str, Enum))
+    return type_ is None or type_ is Any or issubclass(type_, (str, int, Enum))
 
 
 def is_primitive_type(type_: Any) -> bool:
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -491,7 +491,9 @@ def valid_value_annotation_type(type_: Any) -> bool:
 
 
 def _valid_dict_key_annotation_type(type_: Any) -> bool:
-    return type_ is None or type_ is Any or issubclass(type_, (str, int, Enum))
+    from omegaconf import DictKeyType
+
+    return type_ is None or type_ is Any or issubclass(type_, DictKeyType.__args__)  # type: ignore
 
 
 def is_primitive_type(type_: Any) -> bool:
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -24,12 +24,10 @@ from ._utils import (
     is_primitive_type,
     is_structured_config,
 )
-from .base import Container, ContainerMetadata, Node
+from .base import Container, ContainerMetadata, DictKeyType, Node
 from .errors import MissingMandatoryValue, ReadonlyConfigError, ValidationError
 
 DEFAULT_VALUE_MARKER: Any = str("__DEFAULT_VALUE_MARKER__")
-
-DictKeyType = Union[str, int, Enum]
 
 
 class BaseContainer(Container, ABC):
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -29,6 +29,8 @@ from .errors import MissingMandatoryValue, ReadonlyConfigError, ValidationError
 
 DEFAULT_VALUE_MARKER: Any = str("__DEFAULT_VALUE_MARKER__")
 
+DictKeyType = Union[str, int, Enum]
+
 
 class BaseContainer(Container, ABC):
     # static
@@ -187,7 +189,7 @@ class BaseContainer(Container, ABC):
         resolve: bool,
         enum_to_str: bool = False,
         exclude_structured_configs: bool = False,
-    ) -> Union[None, Any, str, Dict[str, Any], List[Any]]:
+    ) -> Union[None, Any, str, Dict[DictKeyType, Any], List[Any]]:
         from .dictconfig import DictConfig
         from .listconfig import ListConfig
 
@@ -528,7 +530,10 @@ class BaseContainer(Container, ABC):
 
     @staticmethod
     def _item_eq(
-        c1: Container, k1: Union[str, int], c2: Container, k2: Union[str, int]
+        c1: Container,
+        k1: Union[DictKeyType, int],
+        c2: Container,
+        k2: Union[DictKeyType, int],
     ) -> bool:
         v1 = c1._get_node(k1)
         v2 = c2._get_node(k2)
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -46,6 +46,8 @@ from .errors import (
 )
 from .nodes import EnumNode, ValueNode
 
+DictKeyType = Union[str, int, Enum]
+
 
 class DictConfig(BaseContainer, MutableMapping[str, Any]):
 
@@ -245,14 +247,14 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
         )
         raise ValidationError(msg)
 
-    def _validate_and_normalize_key(self, key: Any) -> Union[str, int, Enum]:
+    def _validate_and_normalize_key(self, key: Any) -> DictKeyType:
         return self._s_validate_and_normalize_key(self._metadata.key_type, key)
 
     def _s_validate_and_normalize_key(
         self, key_type: Any, key: Any
-    ) -> Union[str, int, Enum]:
+    ) -> DictKeyType:
         if key_type is Any:
-            for t in (str, int, Enum):
+            for t in DictKeyType.__args__:  # type: ignore
                 try:
                     return self._s_validate_and_normalize_key(key_type=t, key=key)
                 except KeyValidationError:
@@ -285,7 +287,7 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
         else:
             assert False, f"Unsupported key type {key_type}"
 
-    def __setitem__(self, key: Union[str, int, Enum], value: Any) -> None:
+    def __setitem__(self, key: DictKeyType, value: Any) -> None:
         try:
             self.__set_impl(key=key, value=value)
         except AttributeError as e:
@@ -295,7 +297,7 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
         except Exception as e:
             self._format_and_raise(key=key, value=value, cause=e)
 
-    def __set_impl(self, key: Union[str, int, Enum], value: Any) -> None:
+    def __set_impl(self, key: DictKeyType, value: Any) -> None:
         key = self._validate_and_normalize_key(key)
         self._set_item_impl(key, value)
 
@@ -338,7 +340,7 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
         except Exception as e:
             self._format_and_raise(key=key, value=None, cause=e)
 
-    def __getitem__(self, key: Union[str, int, Enum]) -> Any:
+    def __getitem__(self, key: DictKeyType) -> Any:
         """
         Allow map style access
         :param key:
@@ -354,7 +356,7 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
         except Exception as e:
             self._format_and_raise(key=key, value=None, cause=e)
 
-    def __delitem__(self, key: Union[str, int, Enum]) -> None:
+    def __delitem__(self, key: DictKeyType) -> None:
         if self._get_flag("readonly"):
             self._format_and_raise(
                 key=key,
@@ -383,14 +385,14 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
         del self.__dict__["_content"][key]
 
     def get(
-        self, key: Union[str, int, Enum], default_value: Any = DEFAULT_VALUE_MARKER
+        self, key: DictKeyType, default_value: Any = DEFAULT_VALUE_MARKER
     ) -> Any:
         try:
             return self._get_impl(key=key, default_value=default_value)
         except Exception as e:
             self._format_and_raise(key=key, value=None, cause=e)
 
-    def _get_impl(self, key: Union[str, int, Enum], default_value: Any) -> Any:
+    def _get_impl(self, key: DictKeyType, default_value: Any) -> Any:
         try:
             node = self._get_node(key=key)
         except ConfigAttributeError:
@@ -403,7 +405,7 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
         )
 
     def _get_node(
-        self, key: Union[str, int, Enum], validate_access: bool = True
+        self, key: DictKeyType, validate_access: bool = True
     ) -> Optional[Node]:
         try:
             key = self._validate_and_normalize_key(key)
@@ -421,7 +423,7 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
         return value
 
     def pop(
-        self, key: Union[str, int, Enum], default: Any = DEFAULT_VALUE_MARKER
+        self, key: DictKeyType, default: Any = DEFAULT_VALUE_MARKER
     ) -> Any:
         try:
             if self._get_flag("readonly"):
@@ -494,7 +496,7 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
     def items(self) -> AbstractSet[Tuple[str, Any]]:
         return self.items_ex(resolve=True, keys=None)
 
-    def setdefault(self, key: Union[str, int, Enum], default: Any = None) -> Any:
+    def setdefault(self, key: DictKeyType, default: Any = None) -> Any:
         if key in self:
             ret = self.__getitem__(key)
         else:
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -245,14 +245,14 @@ class DictConfig(BaseContainer, MutableMapping[str, Any]):
         )
         raise ValidationError(msg)
 
-    def _validate_and_normalize_key(self, key: Any) -> Union[str, Enum]:
+    def _validate_and_normalize_key(self, key: Any) -> Union[str, int, Enum]:
         return self._s_validate_and_normalize_key(self._metadata.key_type, key)
 
     def _s_validate_and_normalize_key(
         self, key_type: Any, key: Any
-    ) -> Union[str, Enum]:
+    ) -> Union[str, int, Enum]:
         if key_type is Any:
-            for t in (str, Enum):
+            for t in (str, int, Enum):
                 try:
                     return self._s_validate_and_normalize_key(key_type=t, key=key)
                 except KeyValidationError:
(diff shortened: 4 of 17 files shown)


omry:
> Before moving further with test coverage, here is a design question:
will we want to support other primitive key types besides int?
Currently (before the above PR), DictConfig keys can be one of
[Any, str, Enum, None]
and DictConfig values can be one of
[Any, str, float, bool, int, None, StructuredConfig].

We will likely want to support float and bool as well (all primitive types).
yaml has no issues with them and if there aren't any special issues I think we can support them.

As for the test you commented out, you can delete it but add tests covering the usage.
You can look at enum key tests, which is the latest addition of a new supported key type for inspiration.

omegaconf/dictconfig.py
```diff
@@ -331,7 +338,7 @@ def __getattr__(self, key: str) -> Any:
         except Exception as e:
             self._format_and_raise(key=key, value=None, cause=e)
 
-    def __getitem__(self, key: Union[str, Enum]) -> Any:
+    def __getitem__(self, key: Union[str, int, Enum]) -> Any:
```
omry, line 341: We should probably create an alias for the union instead of repeating the whole thing many times.

tests/test_basic_ops_dict.py
```diff
@@ -84,6 +84,12 @@ def test_subscript_get() -> None:
     assert "b" == c["a"]
 
 
+def test_subscript_get_int_key() -> None:
+    c = OmegaConf.create("1: b")
+    assert isinstance(c, DictConfig)
+    assert "b" == c[1]
```
omry, line 90: Create from dictionary instead of from string.
```suggestion
def test_subscript_get_int_key() -> None:
    c = OmegaConf.create({1: b})
    assert isinstance(c, DictConfig)
    assert "b" == c[1]
```

omry (changes requested):
Thanks for working on it!
Overall looks pretty good, but I think I will want more tests.
once things are settled down, you can also update the docs and finally you can create a news fragment for this change.

Per your question, it could be a good opportunity to add book and float as well to complete the spectrum of keys to all supported primitives.
This can also be done in a followup PR if you prefer (but if we go that way I will want this completed with tests and documentation updates before we land it).

omegaconf/_utils.py
```diff
@@ -491,7 +491,7 @@ def valid_value_annotation_type(type_: Any) -> bool:
 
 
 def _valid_dict_key_annotation_type(type_: Any) -> bool:
-    return type_ is None or type_ is Any or issubclass(type_, (str, Enum))
+    return type_ is None or type_ is Any or issubclass(type_, (str, int, Enum))
```
odelalleau, line 494: I think a `DictKeyType = Union[str, int, Enum]` should be defined somewhere so that whenever it's changed we only need to modify it in one place.

Here this would translate into writing `issubclass(type_, DictKeyType.__args__)`

Edit: just saw that @omry had the same comment, still leaving this one to show how to obtain the underlying types.

tests/structured_conf/data/attr_classes.py
```diff
@@ -299,10 +299,10 @@ class WithTypedDict:
...
 
 
-@attr.s(auto_attribs=True)
-class ErrorDictIntKey:
-    # invalid dict key, must be str
-    dict: Dict[int, str] = {10: "foo", 20: "bar"}
+# @attr.s(auto_attribs=True)
+# class ErrorDictIntKey:
+#     # invalid dict key, must be str
+#     dict: Dict[int, str] = {10: "foo", 20: "bar"}
```
odelalleau, line 305: Instead of commenting it out, replace it by another type that is not supported. I suggest something like `dict: Dict[object, str] = {object(): "foo", object(): "bar"}` because I don't expect that arbitrary objects will be supported any time soon.

tests/structured_conf/test_structured_config.py
```diff
@@ -546,11 +546,11 @@ def test_merge_dict_with_correct_type(self, class_type: str) -> None:
...
-    def test_typed_dict_key_error(self, class_type: str) -> None:
-        module: Any = import_module(class_type)
-        input_ = module.ErrorDictIntKey
-        with pytest.raises(KeyValidationError):
-            OmegaConf.structured(input_)
+    # def test_typed_dict_key_error(self, class_type: str) -> None:
+    #     module: Any = import_module(class_type)
+    #     input_ = module.ErrorDictIntKey
+    #     with pytest.raises(KeyValidationError):
+    #         OmegaConf.structured(input_)
```
odelalleau, line 553: Uncomment

tests/test_basic_ops_dict.py
```diff
@@ -542,21 +548,21 @@ def test_masked_copy_is_deep() -> None:
...
 
 
-def test_get_with_invalid_key() -> None:
-    cfg = OmegaConf.create()
-    with pytest.raises(KeyValidationError):
-        cfg[1]  # type: ignore
+# def test_get_with_invalid_key() -> None:
+#     cfg = OmegaConf.create()
+#     with pytest.raises(KeyValidationError):
+#         cfg[1]  # type: ignore
```
odelalleau, line 565: Uncomment, replace by another invalid key like `object()`

odelalleau (changes requested):

omry:
A surface area that should be tested is Structured Configs.

A few things to add tests for:
```
@dataclass
class DictWithIntKeys:
  dict : Dict[int, str]

@dataclass
class ExtendDictWithIntKeys(Dict[int, str]:
  ...
```

Anything else I am not thinking about?

Jasha10 (commented):
A related issue with type checking of `OmegaConf.create`:
The following also gives mypy issues:
`a = OmegaConf.create({Enum1.FOO: 123})`
For simplicity, we could consider changing the signature of the `OmegaConf.create` overload to something like `obj: Union[Dict[Any, Any], None] = None`.

tests/test_basic_ops_dict.py
```diff
@@ -84,6 +84,12 @@ def test_subscript_get() -> None:
     assert "b" == c["a"]
 
 
+def test_subscript_get_int_key() -> None:
+    c = OmegaConf.create({1: b})
```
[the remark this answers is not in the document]
omry: Up, looks like it.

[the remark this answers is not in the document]
Jasha10: An alternative to using
    `obj: Union[Dict[Any, Any], None] = None,`
in the signature of `OmegaConf.create` would be to use
    `obj: Union[DictType, None] = None,`
where the alias `DictType = Union[Dict[Str, Any], Dict[int, Any], Dict[Enum, Any]]` is defined in `dictconfig.py` right next to the definition of the alias `DictKeyType = Union[Str, int, Enum]`.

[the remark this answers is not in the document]
omry: Can this be expressed simply as:
```
Dict[DictKeyType, Any]
```
?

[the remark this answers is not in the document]
Jasha10: Or should we use

    obj: Union[Dict[DictKeyType, Any], None] = None,

?
Technically `Union[Dict[DictKeyType, Any], None]` is a supertype of `Union[Dict[Str, Any], Dict[int, Any], Dict[Enum, Any]]`.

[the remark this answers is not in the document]
omry: The Optional variant seems cleaner:

 `obj: Optional[Dict[DictKeyType, Any]] = None,`

omry:
re-request review when ready.

Jasha10:
@omry I've added tests for structured config and general use of Dict[int, ...].
I'm not sure exactly what you had in mind with your comment above about ExtendDictWithIntKeys:
```
@dataclass
class ExtendDictWithIntKeys(Dict[int, str]:
  ...
```
Could you elaborate?

omegaconf/dictconfig.py
```diff
@@ -47,13 +48,13 @@
 from .nodes import EnumNode, ValueNode
 
 
-class DictConfig(BaseContainer, MutableMapping[str, Any]):
+class DictConfig(BaseContainer, MutableMapping[DictKeyType, Any]):
```
Jasha10, line 51: The following is giving me issues with mypy:
```
cfg = OmegaConf.create()
upd = dict(a=1)
cfg.update(upd)
```
The mypy error is:
```
tmp.py:4: error: Argument 1 to "update" of "MutableMapping" has incompatible type "Dict[str, int]"; expected "Mapping[Union[str, int, Enum], Any]"
```
This makes me think that

    class DictConfig(BaseContainer, MutableMapping[DictKeyType, Any]): ...


should be changed to

    class DictConfig(BaseContainer, MutableMapping[Any, Any]): ...

Jasha10: Made the change in commit https://github.com/omry/omegaconf/pull/454/commits/3b193b601a43302a2c5d6107ef2df0afa744992f

odelalleau: Unfortunately I don't really see a better option. I wonder if this is actually a bug in mypy?
(another "fix" would be to explicitly declare `upd: Dict[DictKeyType, Any] = dict(a=1)` but I think your solution is better)
@omry what do you think?

omegaconf/omegaconf.py
```diff
@@ -183,7 +183,7 @@ def create(
     @staticmethod
     @overload
     def create(
-        obj: Union[Dict[str, Any], None] = None,
+        obj: Optional[Dict[Any, Any]] = None,
```
odelalleau, line 186: Why did you end up using `Any` here instead of `DictKeyType`?

tests/test_basic_ops_dict.py
```diff
@@ -72,12 +72,27 @@ def test_mandatory_value() -> None:
         c.a
 
 
+def test_mandatory_value_int_key() -> None:
```
odelalleau, line 75: I think this test (as well as other tests added in this commit) would better be created by parameterizing the existing test operating on strings (here: `test_mandatory_value()`). This way, if we want to support float / bool keys as well for instance, it will be much more straightforward than to copy again the same set of tests.

omegaconf/dictconfig.py
```diff
@@ -494,9 +498,9 @@ def setdefault(self, key: Union[str, Enum], default: Any = None) -> Any:
         return ret
 
     def items_ex(
-        self, resolve: bool = True, keys: Optional[List[str]] = None
-    ) -> AbstractSet[Tuple[str, Any]]:
-        items: List[Tuple[str, Any]] = []
+        self, resolve: bool = True, keys: Optional[Sequence[DictKeyType]] = None
+    ) -> AbstractSet[Tuple[DictKeyType, Any]]:
+        items: List[Tuple[DictKeyType, Any]] = []
```
odelalleau, line 503: Leaving this comment here but it applies to other parts of the code as well:

I see multiple places where the previous code was using only `str` but not `Enum`. Do you know if it's just that the previous code was incorrect, or was there a reason not to have enums? In the latter case, would adding `Enum` back (through `DictKeyType`) be an issue? (I personally have no idea what's the role of Enums here...)

odelalleau (changes requested):

omegaconf/omegaconf.py
odelalleau: Ah, never mind, just saw the answer: https://github.com/omry/omegaconf/pull/454#pullrequestreview-550948711

omegaconf/dictconfig.py
omry: I am okay with making DictConfig extend MutableMapping[Any, Any].

omry:
> @omry I've added tests for structured config and general use of Dict[int, ...].
> I'm not sure exactly what you had in mind with the comment above about ExtendDictWithIntKeys:
> 
> ```
> @dataclass
> class ExtendDictWithIntKeys(Dict[int, str]:
>   ...
> ```
> 
> Could you elaborate?

Extending dicts has a special semantics in OmegaConf Structured Configs.
Look for DictWithFields in the [Structured Configs doc](https://omegaconf.readthedocs.io/en/latest/structured_config.html).
Basically it allows the assignment of fields that are not a part o the dataclass.
In this case, I would expect ExtendDictWithIntKeys to accept int key but reject incompatible keys:

```
cfg : ExtendDictWithIntKeys = OmegaConf.structured(ExtendDictWithIntKeys)
cfg[10] = "ten" # okay
cfg[10.0] = "float" # fail
cfg["10"] = "float" # fail
```

Let's stick to what mypy would accept  (I think mypy should also complain about the same assignments).

omegaconf/dictconfig.py
omry: Enum keys are were added in OmegaConf 2.0 and I am not surprised there are some annotation gaps for it.

omegaconf/dictconfig.py
odelalleau: > I am okay with making DictConfig extend MutableMapping[Any, Any].

Perfect, resolving then

omegaconf/dictconfig.py
odelalleau: > Enum keys are were added in OmegaConf 2.0 and I am not surprised there are some annotation gaps for it.

Ok thanks, resolving then

odelalleau (changes requested):
I just noticed a doc item that should be updated: "Keys must be strings or enums" (https://github.com/omry/omegaconf/blame/master/docs/source/structured_config.rst#L260)

tests/test_basic_ops_dict.py
Jasha10: Done in commit https://github.com/omry/omegaconf/pull/454/commits/35c796a68ec4e8580fa2914ca2ca59f077184941

odelalleau: Nice, thanks! I didn't expect all those `if isinstance(key, str)` cases (since `cfg[key]` works for strings too) but it doesn't hurt to test them I guess :) You could have also used a helper function to switch between the two ways to access keys without having to repeat these checks everywhere -- though your tests actually cover more access cases this way, so not asking you to do that.

tests/test_basic_ops_dict.py
```diff
@@ -66,28 +66,69 @@ def test_getattr_dict() -> None:
...
-def test_subscript_set() -> None:
-    c = OmegaConf.create()
-    c["a"] = "b"
-    assert {"a": "b"} == c
+@pytest.mark.parametrize(
+    "key",
+    ["a", 1],
+)
+class TestDictKeyTypes:
+    def test_mandatory_value(self, key: Any) -> None:
```
odelalleau, line 74: Shouldn't all `key: Any` actually be `key: DictKeyType`?

odelalleau (changes requested):
Just a minor remark on type annotations -- after this I think it should be all good for me!

tests/test_basic_ops_dict.py
Jasha10: Oh yes, absolutely!

odelalleau: Just a little import sorting issue left to fix ;)

pereman2 (dismissed):

odelalleau (approved):
All my remarks have been resolved, thanks!

Jasha10:
@omry I think this is ready for your signoff :)

omry:
Thanks for the ping, I will take a look in a few days!

omegaconf/dictconfig.py
```diff
@@ -264,6 +263,13 @@ def _s_validate_and_normalize_key(
                     f"Key $KEY ($KEY_TYPE) is incompatible with ({key_type.__name__})"
                 )
 
+            return key
+        elif key_type == int:
+            if not isinstance(key, int):
+                raise KeyValidationError(
+                    f"Key $KEY ($KEY_TYPE) is incompatible with ({key_type.__name__})"
+                )
+
```
omry, line 272: Add a test for this error in `test_errors.py`.

omry: Something like this (line 430 in test_errors.py):
```
    pytest.param(
        Expected(
            create=lambda: DictConfig({}, key_type=int),
            op=lambda cfg: cfg.get("foo"),
            exception_type=KeyValidationError,
            msg=dedent(
                """\
                Key foo (str) is incompatible with (int)
                \tfull_key: foo
                \treference_type=Optional[Dict[int, Any]]
                \tobject_type=dict"""
            ),
            key="foo",
            full_key="foo",
        ),
        id="dict[int,Any]:mistyped_key",
    ),
```

omegaconf/basecontainer.py
```diff
@@ -29,6 +29,8 @@
 
 DEFAULT_VALUE_MARKER: Any = str("__DEFAULT_VALUE_MARKER__")
 
+DictKeyType = Union[str, int, Enum]
+
```
omry, line 33: This should probably be defined in base.py

omry (changes requested):
Overall looks pretty good. see my comments.

omegaconf/basecontainer.py
```diff
@@ -528,7 +530,10 @@ def assign(value_key: Any, value_to_assign: Any) -> None:
 
     @staticmethod
     def _item_eq(
-        c1: Container, k1: Union[str, int], c2: Container, k2: Union[str, int]
+        c1: Container,
+        k1: Union[DictKeyType, int],
```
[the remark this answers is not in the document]
odelalleau: Just a note that I had thought about suggesting that as well, then I also thought that keeping the union is semantically clearer (we want either a dict key or an int -- that some day might get written `ListindexType`). Python simplifies it anyway so there's no redundancy in practice.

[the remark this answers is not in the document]
omry: Yeah, I was just scratching my head looking at the `Union[DictKeyType, int]`.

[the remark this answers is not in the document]
Jasha10: A related example: BaseContainer._get_full_key has signature
`def _get_full_key(self, key: Union[str, Enum, int, slice, None]) -> str:`
whose union type could be simplified to either `Union[DictKeyType, int, slice, None]` or `Union[DictKeyType, slice, None]`, depending on which style we choose (either explicitly specifying int for semantic clarity, or excluding it because it's redundant with the `int` in DictKeyType).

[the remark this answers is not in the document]
omry: I guess we can also go with something like:
`Union[DicyKeyType, ListKeyType]` where
```
ListKeyType = Union[int, slice]
```

omry:
Let me know what you guys think of that suggestion.
In any case, I am ready to merge this. Good job @Jasha10!

omegaconf/basecontainer.py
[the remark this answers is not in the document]
odelalleau: > ListKeyType = Union[int, slice]

Not sure we want the slice in there: maybe this specific function would work with slices, but the `get` / `pop` / `insert` methods of `ListConfig` wouldn't. And if a `ListKeyType` is defined, it should probably be used in these methods too.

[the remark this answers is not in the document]
Jasha10: Although this is not a documented feature, `ListConfig.get(slice)` and `ListConfig.pop(slice)` both work as expected.
```
>>> cfg = OmegaConf.create([0,1,2,3,4])
>>> cfg
[0, 1, 2, 3, 4]
>>> cfg.get(slice(1,3))
[1, 2]
>>> cfg.pop(slice(1,3))
[1, 2]
>>> cfg
[0, 3, 4]
```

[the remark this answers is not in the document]
Jasha10: I guess it's not clear how `list_config.insert(slice, ...)` would be defined. So `ListConfig.insert` has stricter typing requirements than `ListConfig.__getindex__`.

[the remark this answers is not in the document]
Jasha10: I guess the confusion is that `slice` represents a set of indices, not a single index, and that methods such as `__getitem__` or `pop` are conventionally overloaded to accept either a single index or a set of indices.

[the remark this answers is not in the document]
omry: Yeah, I realized slice is not going to play well there after making the suggestion.
I think for now we can park this discussion and keep things as is.

> Although this is not a documented feature, ListConfig.get(slice) and ListConfig.pop(slice) both work as expected.
I think this is accidental :), I doubt it's a tested feature.

omry:
Can you add a news fragment?

NEWS.md
```diff
@@ -1,3 +1,9 @@
+## 2.0.6 (2020-12-22)
+
+### Features
+
+- OmegaConf now supports `int` for dictionary key types ([#149](https://github.com/omry/omegaconf/issues/149)) 
+
```
omry, line 6: sorry for not being more specific.
Do not edit the news file.

OmegaConf is using towncrier to generate this file. You can read about the general approach in this Hydra page: https://hydra.cc/docs/development/documentation#news-entries

You should create a specific file under the news directory.

Jasha10: Ok, got it.

omry (approved):
Awesome work, thanks!

Issue #267, opened 2020-06-13
Evaluate performance with big yaml files


omry:
Performance is not great when there are very large configs.
There could be some low hanging fruits that can help improve it.
specifically, evaluate:
- initialization time
- access time
- modification time

PR #463, opened 2020-12-17
Merging a missing value onto an existing value no longer change the target value

### LINKED ISSUE #462: Behavior change: Merging a MISSING node should not change target
```python
$ cat 1.py 

cfg1 = OmegaConf.create({"a" : 1, "b" : "???"})
cfg2 = OmegaConf.create({"a" : "???", "b" : 2})
print(OmegaConf.merge(cfg1, cfg2))

 
omry@Coronadev:~/dev/omegaconf$ python 1.py 
{'a': '???', 'b': 2}
```

Desired behavior is for missing fields being merged onto non-missing fields to not change them.
Desired output:
```
{'a': 1, 'b': 2}
```

omry:
Closes #462


omegaconf/basecontainer.py
```diff
@@ -330,14 +330,24 @@ def expand(node: Container) -> None:
...
+                        assert isinstance(src_node, ValueNode)
                         try:
-                            dest_node._set_value(src_value)
+                            if isinstance(dest_node, AnyNode):
+                                node: ValueNode = copy.copy(src_node)
+                                if node._is_missing():
+                                    node._set_value(dest_node._value())
+                                dest.__setitem__(key, node)
+                            else:
+                                dest_node._set_value(src_value)
```
odelalleau: Isn't there a risk to assign MISSING here?

odelalleau: Never mind, this issue was fixed in latest commit

odelalleau:
Another question is what happens when trying to merge `${foo}` where `foo` is missing?

omegaconf/basecontainer.py
```diff
@@ -360,6 +370,43 @@ def expand(node: Container) -> None:
...
+    @staticmethod
+    def _list_merge(dest: Any, src: Any) -> None:
+        from omegaconf import DictConfig, ListConfig, OmegaConf
+
+        assert isinstance(dest, ListConfig)
+        assert isinstance(src, ListConfig)
+
+        if src._is_missing():
+            # do not change dest if src is MISSING.
+            return
```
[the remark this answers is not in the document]
omry: If you can come up with a test that does not work as one would expect let me know. :)

omegaconf/basecontainer.py
```diff
@@ -330,14 +330,24 @@ def expand(node: Container) -> None:
...
+                            dest.__setitem__(key, src_value)
                     else:
                         assert isinstance(dest_node, ValueNode)
+                        assert isinstance(src_node, ValueNode)
                         try:
-                            dest_node._set_value(src_value)
+                            if isinstance(dest_node, AnyNode):
+                                node: ValueNode = copy.copy(src_node)
+                                if node._is_missing():
+                                    node._set_value(dest_node._value())
```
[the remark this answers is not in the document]
omry: good call.
Performance is actually important.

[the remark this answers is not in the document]
omry: deepcopy dominates the initialization time of non trivial configs.

If you have some cycles : #267

omegaconf/basecontainer.py
```diff
@@ -330,14 +330,24 @@ def expand(node: Container) -> None:
                         dest._validate_merge(key=key, value=src_value)
                         dest_node._merge_with(src_value)
                     else:
-                        dest.__setitem__(key, src_value)
+                        if src_value != MISSING:
+                            dest.__setitem__(key, src_value)
                 else:
                     if isinstance(src_value, BaseContainer):
-                        dest.__setitem__(key, src_value)
+                        if src_value != MISSING:
```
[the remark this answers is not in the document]
odelalleau: > Yes.

I'm not entirely convinced because in your example `a` is the container, but `a != MISSING`. Anyway, resolving since you removed that check.

omry:
> Another question is what happens when trying to merge `${foo}` where `foo` is missing?

Not sure I understand what case you are thinking about, maybe try and tell me if you find a hole?

I am going to merge this. things seems to work in Hydra land with this version after switching default `_self_` to tailing.

omegaconf/basecontainer.py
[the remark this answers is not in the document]
odelalleau: So actually, after looking at it more closely, the conversion I had in mind isn't done for dicts either. Essentially, I was wondering what should happen in these scenarios:
```
1. OmegaConf.merge(OmegaConf.create({"a": 0, "b": 1}), DictConfig(MISSING, element_type=str))
2. OmegaConf.merge(OmegaConf.create([0, 1]), ListConfig(MISSING, element_type=str))
```
I thought that in (1) the key types would be converted to strings, but in (2) the list elements wouldn't. However even in (1) there is no such conversion.

Given the direction of these changes I would expect these conversions to happen. Thoughts?

Note: interestingly, if I swap the inputs to merge, in (1) there is no conversion while in (2) there is...

omegaconf/basecontainer.py
```diff
@@ -301,11 +298,15 @@ def expand(node: Container) -> None:
                 else:
                     node._set_value(type_)
 
-        if dest._is_interpolation() or dest._is_missing():
+        if src._is_missing() and not dest._is_missing():
+            if is_structured_config(get_ref_type(src)):
+                expand(src)
```
odelalleau, line 303: I think this can have unintended side effects as we may end up instantiating a non-empty object in `src` that would erase some fields.
To see it, run `pytest "tests/structured_conf/test_structured_basic.py::TestStructured::test_merge_structured_onto_dict_nested[tests.structured_conf.data.dataclasses]"` after modifying the class `User` in dataclasses.py with `name: str = "hello"`: notice how the name gets overwritten even though `MissingUserField` is supposed to have a missing `User`.

odelalleau (commented):
> Not sure I understand what case you are thinking about, maybe try and tell me if you find a hole?

I think this is a hole:
```python
>>> OmegaConf.merge(OmegaConf.create({"foo": 0, "bar": 1}), OmegaConf.create({"foo": "${bar}", "bar": "???"}))
{'foo': 0, 'bar': 1}
```

omegaconf/basecontainer.py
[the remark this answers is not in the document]
odelalleau: > If you have some cycles : #267

Ask again next year ;)

omegaconf/basecontainer.py
omry: This :
```python
cfg = OmegaConf.merge(
    {"foo": 0, "bar": 1},
    {"foo": "${bar}", "bar": "???"},
)
```

Is as designed because both foo and bar are considered missing.
(By the way, you can merge regular dictionaries directly).

Can you create a minimal example for the second part of your comment?

odelalleau: > Is as designed because both foo and bar are considered missing.

Is there anything that relies on `foo` being missing? Under the semantics of "missing == not set", then `foo` wouldn't be missing: it is set to an interpolation pointing to `bar`.
Even if we want to keep considering `foo` as missing, for the specific case of merge, it would make more sense to me to still merge the interpolation. Otherwise it seems confusing that using, e.g. `"foo_${bar}"` would merge the interpolation, while `${bar}` wouldn't. In addition, not being able to merge an interpolation pointing to a missing value seems limiting (for instance, maybe it's because we want to use the value of this variable in the source config).

> (By the way, you can merge regular dictionaries directly).

Oh! :)

> Can you create a minimal example for the second part of your comment?

I added a test on top of master showing exactly what I was alluding to: https://github.com/odelalleau/omegaconf/commit/ca37fd38f4325e88adb016a66867e104eb081917

omry: > > Is as designed because both foo and bar are considered missing.
> 
> Is there anything that relies on `foo` being missing? Under the semantics of "missing == not set", then `foo` wouldn't be missing: it is set to an interpolation pointing to `bar`.

Many unit tests at the very least.
The interaction between missing, none, interpolations, structured configs, optional etc is tested well.
If you are asking if we can this is something we can break, I will need a good reason before even considering it.

> Even if we want to keep considering `foo` as missing, for the specific case of merge, it would make more sense to me to still merge the interpolation. Otherwise it seems confusing that using, e.g. `"foo_${bar}"` would merge the interpolation, while `${bar}` wouldn't. In addition, not being able to merge an interpolation pointing to a missing value seems limiting (for instance, maybe it's because we want to use the value of this variable in the source config).

Yes, worth looking if it's easy to not use is_missing here but compare the value directly to MISSING.

> 
> > (By the way, you can merge regular dictionaries directly).
> 
> Oh! :)
> 
> > Can you create a minimal example for the second part of your comment?
> 
> I added a test on top of master showing exactly what I was alluding to: [odelalleau@ca37fd3](https://github.com/odelalleau/omegaconf/commit/ca37fd38f4325e88adb016a66867e104eb081917) (right now, this test fails)

Looking at that test, I am not expecting the name to change after this merge.

odelalleau: > If you are asking if we can this is something we can break, I will need a good reason before even considering it.

Followed up on this in your new PR: https://github.com/omry/omegaconf/pull/464#issuecomment-748111885

> > I added a test on top of master showing exactly what I was alluding to: [odelalleau@ca37fd3](https://github.com/odelalleau/omegaconf/commit/ca37fd38f4325e88adb016a66867e104eb081917) (right now, this test fails)
> 
> Looking at that test, I am not expecting the name to change after this merge.

Currently it does, and my guess is it's because of the line I pointed to. A potential fix (not sure if there's a better way) would be to set all fields of `src` to `MISSING` after the expand.

omry: Ouch.
Need to take a closer look at it. switching. can you file a standalone bug to keep track of it?

odelalleau: > Need to take a closer look at it. Can you file a standalone bug to keep track of it?

I created a PR with a fix instead to (hopefully) save time: #465

omegaconf/basecontainer.py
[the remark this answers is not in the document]
odelalleau: @omry just flagging that this may require further looking into -- I'm not entirely sure what is the behavior you would like to see for these

PR #464, opened 2020-12-18
fixed interpolaiton+missing merge

omry:


omegaconf/basecontainer.py
```diff
@@ -339,9 +339,11 @@ def expand(node: Container) -> None:
                     else:
                         assert isinstance(dest_node, ValueNode)
                         assert isinstance(src_node, ValueNode)
+                        # Compare to literal missing, ignoring interpolation
+                        src_node_missing = src_node._value() == "???"
```
odelalleau, line 343: Minor: could re-use `src_value`?

odelalleau (commented):
Looks mostly good but I think something similar is needed for lists. Currently we have:
```python
>>> OmegaConf.merge({"foo": [1, 2, 3], "bar": [4, 5, 6]},
...                 {"foo": "???", "bar": ListConfig(content="${foo}")})```
{'foo': [1, 2, 3], 'bar': [4, 5, 6]}
```

Compare it to the equivalent with dicts (which behaves as expected with this PR):
```python
>>> OmegaConf.merge({"foo": {"a": 1, "b": 2}, "bar": {"c": 3}},
...                 {"foo": "???", "bar": DictConfig(content="${foo}")})
{'foo': {'a': 1, 'b': 2}, 'bar': '${foo}'}
```

odelalleau:
Also, out of curiosity I checked what happened if `_is_missing()` always returned `False` for interpolations: the answer is that although tests that test exactly this are failing, nothing else breaks (on current master branch).
Considering that in 756d7b4 you already changed it so that `${foo}_bar` is *not* missing anymore when `foo` is missing, I'm not sure it would be such a stretch to consider saying that `${foo}` isn't missing either.
(it's not a big deal, we can keep it a special case for merge, but I see it as a way to simplify things a bit more)

omry:
> Also, out of curiosity I checked what happened if `_is_missing()` always returned `False` for interpolations: the answer is that although tests that test exactly this are failing, nothing else breaks (on current master branch).
> Considering that in [756d7b4](https://github.com/omry/omegaconf/commit/756d7b40d861704090f8f7ff9b1299fde88e885e) you already changed it so that `${foo}_bar` is _not_ missing anymore when `foo` is missing, I'm not sure it would be such a stretch to consider saying that `${foo}` isn't missing either.
> (it's not a big deal, we can keep it a special case for merge, but I see it as a way to simplify things a bit more)
> 
> Edit: Hydra tests also pass

I tried too and 155 tests failed.
```diff
omry@Coronadev:~/dev/omegaconf$ git diff
diff --git a/omegaconf/basecontainer.py b/omegaconf/basecontainer.py
index 1ef0398..9f57307 100644
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -605,6 +605,7 @@ class BaseContainer(Container, ABC):
         return self.__dict__["_content"] is None
 
     def _is_missing(self) -> bool:
+        return False
         try:
             self._dereference_node(
                 throw_on_resolution_failure=False, throw_on_missing=True
diff --git a/omegaconf/nodes.py b/omegaconf/nodes.py
index 85ef5d9..99ab979 100644
--- a/omegaconf/nodes.py
+++ b/omegaconf/nodes.py
@@ -96,6 +96,7 @@ class ValueNode(Node):
         return node._metadata.optional
 
     def _is_missing(self) -> bool:
+        return False
         if self._is_interpolation():
             try:
                 node = self._dereference_node(
```

omegaconf/basecontainer.py
omry: yes!

odelalleau:
> I tried too and 155 tests failed.

The `return False` should be gated by `if self._is_interpolation():`

odelalleau:
> If you want I can merge this PR and you can take a stab and finding and finding and fixing more edge cases :)

If you want you can merge this PR and I can create issues for you to fix :P
(if it's a trivial fix I'll do it, but if not it'll be on you)

odelalleau (approved):
Looks good to merge -- I'll let you decide whether or not you may want to consider making all interpolations non missing

omry:
> Looks good to merge -- I'll let you decide whether or not you may want to consider making all interpolations non missing
Still not convinced of your claim. With the following I get 16 failures.

```diff
omry@Coronadev:~/dev/omegaconf$ git diff
diff --git a/omegaconf/basecontainer.py b/omegaconf/basecontainer.py
index da77f33..a0be282 100644
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -603,16 +603,12 @@ class BaseContainer(Container, ABC):
         return self.__dict__["_content"] is None
 
     def _is_missing(self) -> bool:
-        try:
-            self._dereference_node(
-                throw_on_resolution_failure=False, throw_on_missing=True
-            )
+        if self._is_interpolation():
             return False
-        except MissingMandatoryValue:
-            ret = True
-
-        assert isinstance(ret, bool)
-        return ret
+        else:
+            ret = self._value() == "???"
+            assert isinstance(ret, bool)
+            return ret
 
     def _is_optional(self) -> bool:
         return self.__dict__["_metadata"].optional is True
diff --git a/omegaconf/nodes.py b/omegaconf/nodes.py
index 85ef5d9..685ea0b 100644
--- a/omegaconf/nodes.py
+++ b/omegaconf/nodes.py
@@ -97,18 +97,7 @@ class ValueNode(Node):
 
     def _is_missing(self) -> bool:
         if self._is_interpolation():
-            try:
-                node = self._dereference_node(
-                    throw_on_resolution_failure=False, throw_on_missing=False
-                )
-                if node is None:
-                    # resolution failure
-                    return False
-                else:
-                    return node._is_missing()
-            except ConfigKeyError:
-                # Will happen if interpolation is accessing keys that does not exist.
-                return False
+            return False
         else:
             ret = self._value() == "???"
             assert isinstance(ret, bool)
```

omry:
Fixed two additional bugs (one new and one old).
Merging once CI passes.

odelalleau:
> Still not convinced of your claim. With the following I get 16 failures.

16 sounds about right. When I looked into them this morning they were all tests related to testing whether `${foo}` is missing when `foo` is missing (which obviously are expected to fail).

omry:
What do you think about this?
```python
In [3]: cfg = OmegaConf.create({"a": "${b}", "b" : "???"})
In [4]: "b" in cfg
Out[4]: False
In [5]: "a" in cfg
Out[5]: False

```

odelalleau:
> What do you think about this?

I think it would make sense to match the `_is_missing()` behavior: `x in y` if `y._get_node(x)` succeeds and `not y._get_node(x)._is_missing()`

odelalleau:
> > What do you think about this?
> 
> I think it would make sense to match the `_is_missing()` behavior: `x in y` if `y._get_node(x)` succeeds and `not y._get_node(x)._is_missing()`

Thinking about it a bit more, I'm not sure... if we do this, then `for x in y` should also skip over missing keys. This could be confusing / lead to subtle bugs.
What would be the benefits of this change?

omry:
1. This is not a change. This is the current behavior, and it depends on the behavior of is_missing().
2. Iteration will fail if you having missing items:
```python
cfg = OmegaConf.create({"a": "${b}", "b" : "???", "c": 10})
In [16]: [k for k in cfg.keys()]
Out[16]: ['a', 'b', 'c']

In [17]: [(k,v) for k in cfg.items()]
MissingMandatoryValue: Missing mandatory value : b
        full_key: a
        reference_type=Optional[Dict[Union[str, Enum], Any]]
        object_type=dict
```

This is as designed, missing items and actively hostile. You should not have any when you are using your config.

For internal purposes, there is a second iteration function that allow iteration on such configs:
```
In [19]: [(k,v) for k,v in cfg.items_ex(resolve=False)]
Out[19]: [('a', '${b}'), ('b', '???'), ('c', 10)]
```

odelalleau:
> 1. This is not a change. This is the current behavior, and it depends on the behavior of is_missing().
> (...)

Oh, ok! Then I think it may be a bit confusing, but why not. What I find potentially is that the following fails:
```python
cfg = OmegaConf.create({"a": "${b}", "b" : "???", "c": 10})
for k in cfg:
    assert k in cfg
```

omry:
Yes, but it's failing in a very visible way so it's not so bad.
I was asking about the semantics on if for missing values (and for interpolation to missing values) and you seem to agree with it.
This was implemented before OmegaConf.is_missing() and I think it's overall making sense.
We can potentially change the iteration to skip missing keys completely, but any such semantic changes are potentially breaking and we need to a good reason to do them.

odelalleau:
> I was asking about the semantics on if for missing values (and for interpolation to missing values) and you seem to agree with it.

Yes, no objection from me :)
(though I wouldn't mind either `key in dict` to ignore the missingness of the value -- I'm just not going to push for one or the other)

> We can potentially change the iteration to skip missing keys completely, but any such semantic changes are potentially breaking and we need to a good reason to do them.

Probably not worth it (actually seems a bit more dangerous than the current behavior, IMO)

PR #465, opened 2020-12-18
Fix edge case when merging a missing structured config

odelalleau:
As discussed in https://github.com/omry/omegaconf/pull/463#discussion_r545868699

### DIFF
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -305,12 +305,12 @@ class BaseContainer(Container, ABC):
                     node._set_value(type_)
 
         src_ref_type = get_ref_type(src)
+        assert src_ref_type is not None
         if (
             src._is_missing()
             and not dest._is_missing()
             and is_structured_config(src_ref_type)
         ):
-            assert src_ref_type is not None
             # Replace `src` with a prototype of its corresponding structured config
             # whose fields are all missing (to avoid overwriting fields in `dest`).
             src = _create_structured_with_missing_fields(
@@ -387,14 +387,7 @@ class BaseContainer(Container, ABC):
                 else:
                     dest[key] = src._get_node(key)
 
-        if src_type is not None and not is_primitive_dict(src_type):
-            dest._metadata.object_type = src_type
-
-        if dest._metadata.ref_type is Any:
-            new_is_optional, new_ref_type = _resolve_optional(src_ref_type)
-            if new_ref_type is not Any:
-                dest._metadata.ref_type = new_ref_type
-                dest._metadata.optional = new_is_optional
+        _udpate_types(node=dest, ref_type=src_ref_type, object_type=src_type)
 
         # explicit flags on the source config are replacing the flag values in the destination
         flags = src._metadata.flags
@@ -751,3 +744,14 @@ def _is_missing_value(value: Any) -> bool:
     ret = value == "???"
     assert isinstance(ret, bool)
     return ret
+
+
+def _udpate_types(node: Node, ref_type: type, object_type: Optional[type]) -> None:
+    if object_type is not None and not is_primitive_dict(object_type):
+        node._metadata.object_type = object_type
+
+    if node._metadata.ref_type is Any:
+        new_is_optional, new_ref_type = _resolve_optional(ref_type)
+        if new_ref_type is not Any:
+            node._metadata.ref_type = new_ref_type
+            node._metadata.optional = new_is_optional
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -278,7 +278,7 @@ class BaseContainer(Container, ABC):
     @staticmethod
     def _map_merge(dest: "BaseContainer", src: "BaseContainer") -> None:
         """merge src into dest and return a new copy, does not modified input"""
-        from omegaconf import MISSING, AnyNode, DictConfig, OmegaConf, ValueNode
+        from omegaconf import AnyNode, DictConfig, OmegaConf, ValueNode
 
         assert isinstance(dest, DictConfig)
         assert isinstance(src, DictConfig)
@@ -321,11 +321,7 @@ class BaseContainer(Container, ABC):
         for key, src_value in src.items_ex(resolve=False):
             src_node = src._get_node(key, validate_access=False)
             dest_node = dest._get_node(key, validate_access=False)
-
-            if isinstance(src_value, Container):
-                missing_src_value = src_value._value() == MISSING
-            else:
-                missing_src_value = src_value == MISSING
+            missing_src_value = _is_missing_value(src_value)
 
             if (
                 isinstance(dest_node, Container)
@@ -743,3 +739,11 @@ def _create_structured_with_missing_fields(
     cfg._metadata.object_type = object_type
 
     return cfg
+
+
+def _is_missing_value(value: Any) -> bool:
+    if isinstance(value, Container):
+        value = value._value()
+    ret = value == "???"
+    assert isinstance(ret, bool)
+    return ret
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -316,9 +316,21 @@ class BaseContainer(Container, ABC):
         for key, src_value in src.items_ex(resolve=False):
             src_node = src._get_node(key, validate_access=False)
             dest_node = dest._get_node(key, validate_access=False)
-            if isinstance(dest_node, Container) and OmegaConf.is_none(dest, key):
-                if not OmegaConf.is_none(src_value):
-                    expand(dest_node)
+
+            if isinstance(src_value, Container):
+                missing_src_value = (
+                    not src_value._is_interpolation() and src_value._is_missing()
+                )
+            else:
+                missing_src_value = src_value == MISSING
+
+            if (
+                isinstance(dest_node, Container)
+                and OmegaConf.is_none(dest, key)
+                and not missing_src_value
+                and not OmegaConf.is_none(src_value)
+            ):
+                expand(dest_node)
 
             if dest_node is not None:
                 if dest_node._is_interpolation():
@@ -329,7 +341,10 @@ class BaseContainer(Container, ABC):
                         dest[key] = target_node
                         dest_node = dest._get_node(key)
 
-            if is_structured_config(dest._metadata.element_type):
+            if (
+                is_structured_config(dest._metadata.element_type)
+                and not missing_src_value
+            ):
                 dest[key] = DictConfig(content=dest._metadata.element_type, parent=dest)
                 dest_node = dest._get_node(key)
 
@@ -338,9 +353,8 @@ class BaseContainer(Container, ABC):
                     if isinstance(src_value, BaseContainer):
                         dest._validate_merge(key=key, value=src_value)
                         dest_node._merge_with(src_value)
-                    else:
-                        if src_value != MISSING:
-                            dest.__setitem__(key, src_value)
+                    elif not missing_src_value:
+                        dest.__setitem__(key, src_value)
                 else:
                     if isinstance(src_value, BaseContainer):
                         dest.__setitem__(key, src_value)
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -284,8 +284,10 @@ class BaseContainer(Container, ABC):
         assert isinstance(src, DictConfig)
         src_type = src._metadata.object_type
 
-        # if source DictConfig is an interpolation set the DictConfig one to be the same interpolation.
-        if src._is_interpolation():
+        # If source DictConfig is:
+        #  - an interpolation => set the destination DictConfig to be the same interpolation
+        #  - None => set the destination DictConfig to None
+        if src._is_interpolation() or src._is_none():
             dest._set_value(src._value())
             return
 
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -283,12 +283,15 @@ class BaseContainer(Container, ABC):
         assert isinstance(dest, DictConfig)
         assert isinstance(src, DictConfig)
         src_type = src._metadata.object_type
+        src_ref_type = get_ref_type(src)
+        assert src_ref_type is not None
 
         # If source DictConfig is:
         #  - an interpolation => set the destination DictConfig to be the same interpolation
         #  - None => set the destination DictConfig to None
         if src._is_interpolation() or src._is_none():
             dest._set_value(src._value())
+            _udpate_types(node=dest, ref_type=src_ref_type, object_type=src_type)
             return
 
         dest._validate_merge(key=None, value=src)
@@ -304,8 +307,6 @@ class BaseContainer(Container, ABC):
                 else:
                     node._set_value(type_)
 
-        src_ref_type = get_ref_type(src)
-        assert src_ref_type is not None
         if (
             src._is_missing()
             and not dest._is_missing()
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -300,17 +300,20 @@ class BaseContainer(Container, ABC):
                     node._set_value(type_)
 
         src_ref_type = get_ref_type(src)
-        if src._is_missing() and not dest._is_missing():
-            if is_structured_config(src_ref_type):
-                assert src_ref_type is not None
-                # Replace `src` with a prototype of its corresponding structured config
-                # whose fields are all missing (to avoid overwriting fields in `dest`).
-                src_data = get_structured_config_data(src_ref_type)
-                for v in src_data.values():
-                    v._set_value(MISSING)
-                src = DictConfig(src_data)
-                src._metadata.ref_type = src_ref_type
-                src._metadata.object_type = src_type
+        if (
+            src._is_missing()
+            and not dest._is_missing()
+            and is_structured_config(src_ref_type)
+        ):
+            assert src_ref_type is not None
+            # Replace `src` with a prototype of its corresponding structured config
+            # whose fields are all missing (to avoid overwriting fields in `dest`).
+            src_data = get_structured_config_data(src_ref_type)
+            for v in src_data.values():
+                v._set_value(MISSING)
+            src = DictConfig(src_data)
+            src._metadata.ref_type = src_ref_type
+            src._metadata.object_type = src_type
 
         if (dest._is_interpolation() or dest._is_missing()) and not src._is_missing():
             expand(dest)
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -391,8 +391,10 @@ class BaseContainer(Container, ABC):
             dest._metadata.object_type = src_type
 
         if dest._metadata.ref_type is Any:
-            dest._metadata.ref_type = src_ref_type
-            dest._metadata.optional = src._metadata.optional
+            new_is_optional, new_ref_type = _resolve_optional(src_ref_type)
+            if new_ref_type is not Any:
+                dest._metadata.ref_type = new_ref_type
+                dest._metadata.optional = new_is_optional
 
         # explicit flags on the source config are replacing the flag values in the destination
         flags = src._metadata.flags
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -333,14 +333,13 @@ class BaseContainer(Container, ABC):
             ):
                 expand(dest_node)
 
-            if dest_node is not None:
-                if dest_node._is_interpolation():
-                    target_node = dest_node._dereference_node(
-                        throw_on_resolution_failure=False
-                    )
-                    if isinstance(target_node, Container):
-                        dest[key] = target_node
-                        dest_node = dest._get_node(key)
+            if dest_node is not None and dest_node._is_interpolation():
+                target_node = dest_node._dereference_node(
+                    throw_on_resolution_failure=False
+                )
+                if isinstance(target_node, Container):
+                    dest[key] = target_node
+                    dest_node = dest._get_node(key)
 
             if (
                 is_structured_config(dest._metadata.element_type)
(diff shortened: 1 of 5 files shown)


omegaconf/basecontainer.py
```diff
@@ -298,9 +298,15 @@ def expand(node: Container) -> None:
...
-                expand(src)
+        if (
+            src._is_missing()
+            and not dest._is_missing()
+            and is_structured_config(get_ref_type(src))
+        ):
+            expand(src)
+            # Ensure all fields of `src` are missing to avoid overwriting `dest`.
+            for k in src:
+                src[k] = MISSING
```
omry, line 309: Not the cleanest solution but not terrible either.
It would prefer calling get_structured_config_data() to get the dict, fix it, and then use that to create the DictConfig.
See if you can make it work, if it's too hard we can go with this version.

omry (changes requested):
see comment.

omry:
And by the way, thanks for pitching in on this delicate change!

omegaconf/basecontainer.py
odelalleau: I'm not entirely sure I understood what you meant, but I gave it a shot in 7bed03b
Let me know what you think. Personally I find it harder to understand, and I have one concern with my `v._set_value(MISSING)` which I hope isn't overwriting the value of a node that someone might use later (?)

omry: Yes, this more efficient because it doesn't need to build the DictConfig and then delete fields.

> v._set_value(MISSING) which I hope isn't overwriting the value of a node that someone might use later (?)

I think in general we are interested in the types and not the values if the containing field is MISSING.
You can try to not expand at all and see which cases are failing. Using expand there was just convenient, but maybe there is a cleaner way to get the same behavior.

omegaconf/basecontainer.py
```diff
@@ -299,8 +300,15 @@ def expand(node: Container) -> None:
...
-                expand(src)
+            src_ref_type = get_ref_type(src)
+            if is_structured_config(src_ref_type):
+                assert src_ref_type is not None
+                # Replace `src` with a prototype of its corresponding structured config
+                # whose fields are all missing (to avoid overwriting fields in `dest`).
+                src_data = get_structured_config_data(src_ref_type)
+                for v in src_data.values():
+                    v._set_value(MISSING)
+                src = DictConfig(src_ref_type(**src_data), ref_type=src_ref_type)
```
omry, line 311: I think you shouldn't instantiate the dataclass here.
did you try passing `content=src_data` directly?

omegaconf/basecontainer.py
omry: After taking a second look, not exactly what I had in mind. I am not sure my suggestion would work.

odelalleau: > Yes, this more efficient because it doesn't need to build the DictConfig and then delete fields.

I haven't looked closely at `get_structured_config_data()` but since it's returning both keys and values, it seems to me that it's essentially building the same content as a `DictConfig` (except that it's packaged in a regular `dict` rather than a `DictConfig`). Am I missing something?

> > v._set_value(MISSING) which I hope isn't overwriting the value of a node that someone might use later (?)
> 
> I think in general we are interested in the types and not the values if the containing field is MISSING.

Just to clarify my concern: I'm not entirely sure if `v` (a node returned by `get_structured_config_data()`) is a copy that I can safely modify in-place, or may be shared (in other words, would calling again `get_structured_config_data()` return a different `v`, or the same? if it's the same then setting it to `MISSING` is probably unsafe)

> You can try to not expand at all and see which cases are failing. Using expand there was just convenient, but maybe there is a cleaner way to get the same behavior.

This new version isn't actually calling `expand()` anymore (though under the hood it probably does something relatively simliar).

At the end of the day, if `src` is missing but is a structured config type, we still want to make sure that `dest` has all fields for this type. And, for fields that were not already in `dest`, we want them to be missing, but with the correct type. This means creating a node with the proper type and the value set to `MISSING`. Right now we do it in `src`, then merging `src` into `dest`. We could probably do it directly in `dest` but I'm not sure we would actually gain something by it.

I don't really have a better idea right now.

omegaconf/basecontainer.py
odelalleau: Yes that was my first attempt, it failed with:
```
omegaconf.errors.ValidationError: Invalid type assigned : dict is not a subclass of User. value: {'name': '???', 'age': '???'}
```

odelalleau: However, looking into it more closely, all tests pass with:
```python
src = DictConfig(src_data)
```
which worries me a bit because we lost the annotated type of `src` by doing so...

I wonder if we may be missing a `ref_type` change for `dest`, in the same spirit of the `if isinstance(dest_node, AnyNode):` that you added. More precisely, looking for instance at this test:

```python
    def test_merge_structured_onto_dict_nested(self, class_type: str) -> None:
        c1 = OmegaConf.create({"user": {"name": 7}})
        c2 = OmegaConf.merge(c1, module.MissingUserField)
        assert c1 == {"user": {"name": 7}}
        # type of name becomes str
        assert c2 == {"user": {"name": "7", "age": "???"}}
```
Wouldn't we want to also add at the end:
```python
assert c2._get_node("user")._metadata.ref_type == module.User
```
(right now such an assert fails, because the left hand side is `Any`)

odelalleau:
Taking another look at the code, I found a couple more edge cases that weren't properly handled (I think).
See the fix and associated tests in 708ff2b (those two new tests would have failed before this commit)

omegaconf/basecontainer.py
omry: > > Yes, this more efficient because it doesn't need to build the DictConfig and then delete fields.
> 
> I haven't looked closely at `get_structured_config_data()` but since it's returning both keys and values, it seems to me that it's essentially building the same content as a `DictConfig` (except that it's packaged in a regular `dict` rather than a `DictConfig`). Am I missing something?

I think you are correct, but in addition to that, you are constructing an instance (which may blow for anything with nested dataclasses), which in turn will have get_structured_config_data() called on it again, and then construct the final DictConfig object.
each of those steps is expensive.

> > > v._set_value(MISSING) which I hope isn't overwriting the value of a node that someone might use later (?)
> > 
> > 
> > I think in general we are interested in the types and not the values if the containing field is MISSING.
> 
> Just to clarify my concern: I'm not entirely sure if `v` (a node returned by `get_structured_config_data()`) is a copy that I can safely modify in-place, or may be shared (in other words, would calling again `get_structured_config_data()` return a different `v`, or the same? if it's the same then setting it to `MISSING` is probably unsafe)

When you call get_structured_config_data(), the result is your to mangle.
 
> > You can try to not expand at all and see which cases are failing. Using expand there was just convenient, but maybe there is a cleaner way to get the same behavior.
> 
> This new version isn't actually calling `expand()` anymore (though under the hood it probably does something relatively simliar).
> 

Expand is doing the same in a different way.

> At the end of the day, if `src` is missing but is a structured config type, we still want to make sure that `dest` has all fields for this type. And, for fields that were not already in `dest`, we want them to be missing, but with the correct type. This means creating a node with the proper type and the value set to `MISSING`. Right now we do it in `src`, then merging `src` into `dest`. We could probably do it directly in `dest` but I'm not sure we would actually gain something by it.
> 
> I don't really have a better idea right now.

IIRC, the reason expand is there is to support cases where you are merging into a MISSING field that is a dataclass.
we want the to expand the field to that any default values in that field are instantiated into the config and the type information in it is created.

I need to take a second look to see why expand is helping here and really get to the bottom of what we are trying to do in the new user case.

omegaconf/basecontainer.py
omry: Yeah, obviously this is an unusual way to instantiate the DictConfig.
Normally, the DictConfig itself is calling get_structured_config_data(), now we are doing it from the outside.

1. Yes, add the extra check to the test.
2. I think you can fix it by following your `src = DictConfig(src_data)` with a direct assignment of the ref_type and I think also the object type.

omegaconf/basecontainer.py
odelalleau: > > > Yes, this more efficient because it doesn't need to build the DictConfig and then delete fields.
> > 
> > 
> > I haven't looked closely at `get_structured_config_data()` but since it's returning both keys and values, it seems to me that it's essentially building the same content as a `DictConfig` (except that it's packaged in a regular `dict` rather than a `DictConfig`). Am I missing something?
> 
> I think you are correct, but in addition to that, you are constructing an instance (which may blow for anything with nested dataclasses), which in turn will have get_structured_config_data() called on it again, and then construct the final DictConfig object.
> each of those steps is expensive.

But `get_structured_config_data()` is also constructing nested objects as needed -- it seems to me we're only saving the final `DictConfig` creation.

About the rest: ok, thanks!

omry: The object it's constructing is not the final one but an intermediate one that is being converted. you are happy to trace the code and see it if yo don't believe me :).

omegaconf/basecontainer.py
odelalleau: All (hopefully) done in 3deee99

omegaconf/basecontainer.py
odelalleau: > The object it's constructing is not the final one but an intermediate one that is being converted. you are happy to trace the code and see it if yo don't believe me :).

Going to trust you on that :)

odelalleau:
> No, but I made it work with `.structured()` in [e3c544e](https://github.com/omry/omegaconf/commit/e3c544e744887ff8e41945b64cc57b3b6e669df5)

So, actually, I could make it significantly simpler, re-using existing classes => I removed that commit and replaced it with 895f065 instead

omegaconf/basecontainer.py
```diff
@@ -298,19 +299,41 @@ def expand(node: Container) -> None:
...
+        ):
+            assert src_ref_type is not None
+            # Replace `src` with a prototype of its corresponding structured config
+            # whose fields are all missing (to avoid overwriting fields in `dest`).
+            src_data = get_structured_config_data(src_ref_type)
+            for v in src_data.values():
+                v._set_value(MISSING)
+            src = DictConfig(src_data)
+            src._metadata.ref_type = src_ref_type
+            src._metadata.object_type = src_type
```
omry, line 316: Can you pull this to into a helper function?

omegaconf/basecontainer.py
```diff
@@ -298,19 +299,41 @@ def expand(node: Container) -> None:
...
             src_node = src._get_node(key, validate_access=False)
             dest_node = dest._get_node(key, validate_access=False)
-            if isinstance(dest_node, Container) and OmegaConf.is_none(dest, key):
-                if not OmegaConf.is_none(src_value):
-                    expand(dest_node)
+
+            if isinstance(src_value, Container):
+                missing_src_value = src_value._value() == MISSING
+            else:
+                missing_src_value = src_value == MISSING
```
omry, line 328: This too, a small helper will make this complex function a bit more readable.
```python
missing_src_value = missing_src_value(src_value)
```

tests/structured_conf/test_structured_basic.py
```diff
@@ -153,6 +160,10 @@ def test_merge_structured_onto_dict_nested(self, class_type: str) -> None:
         assert c1 == {"user": {"name": 7}}
         # type of name becomes str
         assert c2 == {"user": {"name": "7", "age": "???"}}
+        assert isinstance(c2, DictConfig)
+        c2_user = c2._get_node("user")
+        assert c2_user is not None
+        assert c2_user._metadata.ref_type == module.User
```
omry, line 166:
```
get_ref_type(c2, "user") == module.User
```

tests/structured_conf/test_structured_basic.py
```diff
@@ -161,6 +172,40 @@ def test_merge_structured_onto_dict_nested2(self, class_type: str) -> None:
         assert c1 == {"user": {"name": 7}}
         # type of name remains int
         assert c2 == {"user": {"name": 7, "age": "???"}}
+        assert isinstance(c2, DictConfig)
+        c2_user = c2._get_node("user")
+        assert c2_user is not None
+        assert c2_user._metadata.ref_type == module.User
```
omry, line 178:
```python
assert get_ref_type(c2, "user") == module.User
```

tests/structured_conf/data/attr_classes.py
```diff
@@ -41,11 +41,26 @@ class UserDict:
     dict: Dict[str, User] = MISSING
 
 
+@attr.s(auto_attribs=True)
+class UserWithDefaultName(User):
+    name: str = "bob"
+
+
```
omry, line 48: Instead of creating this class, I think you can just merge with `User(name="bob")` in the test.

tests/structured_conf/data/attr_classes.py
```diff
@@ -41,11 +41,26 @@ class UserDict:
...
 @attr.s(auto_attribs=True)
 class MissingUserField:
     user: User = MISSING
 
 
+@attr.s(auto_attribs=True)
+class MissingUserWithDefaultNameField:
+    user: UserWithDefaultName = MISSING
```
omry, line 56: Ditto:
merge with `User(name=MISSING)`

tests/structured_conf/data/attr_classes.py
```diff
@@ -41,11 +41,26 @@ class UserDict:
...
+
+
+@attr.s(auto_attribs=True)
+class OptionalUserWithDefaultNameField:
+    user: Optional[UserWithDefaultName] = None
```
omry, line 61:
```suggestion
@attr.s(auto_attribs=True)
class OptionalUserWithDefaultNameField:
    user: Optional[User] = User(name="bob")
```

omry (commented):
a few requests that will make this easier to understand and review.

tests/structured_conf/data/attr_classes.py
odelalleau: I don't think I can, could you be more specific regarding which test and how to modify it? Currently I don't merge directly with `UserWithDefaultName`, it's used through `MissingUserWithDefaultNameField`.

The specific case I want to test is when we go into this code block:
```python
        if (
            src._is_missing()
            and not dest._is_missing()
            and is_structured_config(src_ref_type)
        )
```
and when `src_ref_type` is such that if I was calling `OmegaConf.structured(src_ref_type)` then I would get an object with at least one non-missing field.

This is a regression test for the bug that was fixed in d686872, where without this change, the test introduced in this commit would fail because `c2.user.name` would be `bob`.

tests/structured_conf/data/attr_classes.py
odelalleau: Done in 2d39b77 (not exactly your diff, but it was indeed possible to simplify it)

omegaconf/basecontainer.py
odelalleau: Done in 1c4232f

omegaconf/basecontainer.py
odelalleau: Done in 678a186

tests/structured_conf/test_structured_basic.py
odelalleau: Done in 9fabb0e

Although I'm a bit worried because I had to write it:
```python
assert get_ref_type(c2, "user") == Optional[module.User]
```
and this made me realize that if `dest` is optional, then it will remain optional even if `src` isn't. Although this could be changed, I'm not sure it would be a good idea (it's not like for types, where `Any` is a hint that the config was untyped: here we can't necessarilly assume that `optional == True` means that it was set by default to this value)

(I think I mentioned it before, but I'll say it again: given all the edge cases and weird things that might happen, I'm still not convinced that this whole thing is a good idea in the end, even if I kickstarted it)

tests/structured_conf/test_structured_basic.py
odelalleau: Also done in 9fabb0e

tests/structured_conf/data/attr_classes.py
odelalleau: TBD based on the above conversation (https://github.com/omry/omegaconf/pull/465#discussion_r546915451)

tests/structured_conf/test_structured_basic.py
omry: Yes.
I think if we can pull it off it's worth it.
The semantics we are going for is pretty clean:

During merge:
1. missing fields are not overriding the value of non-missing fields (regardless of Structured Configs or not).
2. typed missing fields are replacing the type of untyped fields and are validating it.

I think the simplifying effects from this are worth it even if it's complicated to implement in practice.

omry: > Done in [9fabb0e](https://github.com/omry/omegaconf/commit/9fabb0e3bcaef80d9fd4f859aae005b24ff1739f)
> 
> Although I'm a bit worried because I had to write it:
> 
> ```python
> assert get_ref_type(c2, "user") == Optional[module.User]
> ```
> 
> and this made me realize that if `dest` is optional, then it will remain optional even if `src` isn't. Although this could be changed, I'm not sure it would be a good idea (it's not like for types, where `Any` is a hint that the config was untyped: here we can't necessarilly assume that `optional == True` means that it was set by default to this value)

So see, I think this just found a bug.
By default, fields without a type are optional.
In this case we are replacing the type, so we should also transfer the optional bit to dest.

odelalleau: > In this case we are replacing the type, so we should also transfer the optional bit to dest.

Fix in 786577a: I think it's necessary but not sure if it's sufficient, looking into potential other problems now

omegaconf/basecontainer.py
omry: Looks okay.

tests/structured_conf/test_structured_basic.py
omry: Yeah, I got the same fix while playing with it here:
https://github.com/omry/omegaconf/pull/467/commits/c3edf17b159c147a880d5c71c3d5f6a688ed31a0

tests/structured_conf/data/attr_classes.py
omry: You are correct, this is not working here because we are explicitly interesting in the default from the type.

omry (approved):
overall looks good.
Anything else you want to do here or shall I merge?

omegaconf/basecontainer.py
```diff
@@ -368,6 +389,10 @@ def expand(node: Container) -> None:
         if src_type is not None and not is_primitive_dict(src_type):
             dest._metadata.object_type = src_type
 
+        if dest._metadata.ref_type is Any:
+            dest._metadata.ref_type = src_ref_type
+            dest._metadata.optional = src._metadata.optional
+
```
omry, line 395: I think we should only do this when we are promoting the type of dest to the type of src. (the case of missing overriding non-missing).

omry (changes requested):

odelalleau:
> overall looks good.
> Anything else you want to do here or shall I merge?

I just pushed fb7fd30 which fixes a bug that existed before all the recent changes (do you prefer to have it in a different PR?)

There's other stuff I want to check but I can make another PR if needed, let's get this one merged at least.

tests/test_merge.py
```diff
@@ -201,6 +201,11 @@
             {"user": Group},
             id="merge_into_missing_node",
         ),
+        pytest.param(
+            [{"admin": {"name": "joe", "age": 42}}, Group(admin=None)],
+            {"admin": None},
+            id="merge_none_into_existing_node",
+        ),
```
omry, line 208: In this case, does the type of the result becomes Group?

odelalleau: > In this case, does the type of the result becomes Group?

Almost certainly no, that's what I want to check now actually -- I believe there is a similar issue with interpolations.

omry: This seems to work correctly with normal configs before the fix.
```python
In [4]: OmegaConf.merge({"admin": {"name": "joe", "age": 42}}, {"admin": None})
Out[4]: {'admin': None}
```

Is the fix specific to Structured Configs somehow?

tests/structured_conf/test_structured_basic.py
odelalleau: > Yeah, I got the same fix while playing with it here:
> [c3edf17](https://github.com/omry/omegaconf/commit/c3edf17b159c147a880d5c71c3d5f6a688ed31a0)

Cool, resolving this conversation then

tests/test_merge.py
odelalleau: Kind of, though you'd get the bug if you replaced `None` with `DictConfig(None)` (which usually you wouldn't do manually, this comes from structured configs)

The reason for the different behavior is `None` is not a container, while `DictConfig(None)` is, so the code path isn't the same (there is a `if isinstance(src_value, BaseContainer):`)

omry: Yes, DictConfig(None) is simulating 
```
@dataclass
class Foo:
  x : Optional[Dict[str, Any]] = None
```

omegaconf/basecontainer.py
odelalleau: Changed in 234067894abfd761ff754dfcc827d6fdfc32110d

At the same time this commit fixed an issue where `src_ref_type` could be `Optional[X]`, but we want to get rid of the optional part when assigning `_metadata.ref_type`.

tests/test_merge.py
odelalleau: So, the issue with interpolations / None not updating the types should be fixed by the combination of 4bd4e4e (small refactoring) and 42594c6 (the actual change)

omegaconf/basecontainer.py
```diff
@@ -699,3 +721,38 @@ def prepand(full_key: str, parent_type: Any, cur_type: Any, key: Any) -> str:
...
+def _is_missing_value(value: Any) -> bool:
+    if isinstance(value, Container):
+        value = value._value()
+    ret = value == "???"
+    assert isinstance(ret, bool)
+    return ret
+
+
+def _udpate_types(node: Node, ref_type: type, object_type: Optional[type]) -> None:
```
omry, line 750: _update_types ?

omry (approved):
Great, thanks!

omegaconf/basecontainer.py
odelalleau: Oops, #468

PR #479, opened 2021-01-19
eliminated unnecessary deepcopy calls during construction

### LINKED ISSUE #477: Optimization: reduce number of deepcopies
During assignment of any value to an OmegaConf container, it is deep-copied.
the motivation is primarily address the case a node from an existing config tree is assigned to another config:
```python
    cfg1 = OmegaConf.create({"foo": {"bar": 10}})
    cfg2 = OmegaConf.create({})
    cfg2.foo = cfg1.foo
    assert id(cfg1.foo._get_parent()) == id(cfg1)
    assert id(cfg2.foo._get_parent()) == id(cfg2)
```

Without a deepcopy, the parent of the node cfg1.foo will change and cfg1 will be broken and the asserts will fail (resulting in interpolation and other issues).

Deepcopy is a clean solution, but it's expensive and most cases an overkill.
to improve initialization performance, we can limit deepcopy only to the rare cases where a node that belongs to an existing OmegaConf container is assigned.

This is a potentially breaking change in rare cases but think those are unlikely.

omry:
See #477  for context.

closes #477

### DIFF
--- a/news/477.feature
+++ b/news/477.feature
@@ -1,0 +1,1 @@
+Perf improvement: eliminated unnecessary deepcopy calls during construction
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -232,6 +232,34 @@ def test_create_node_parent_retained_on_assign(node: Any) -> None:
     assert cfg2.zonk._get_parent() is cfg2
 
 
+@pytest.mark.parametrize(
+    "node",
+    [
+        {"a": 0},
+        DictConfig({"a": 0}),
+    ],
+)
+def test_dict_assignment_deepcopy_semantics(node: Any) -> None:
+    cfg = OmegaConf.create()
+    cfg.foo = node
+    node["a"] = 1
+    assert cfg.foo.a == 0
+
+
+@pytest.mark.parametrize(
+    "node",
+    [
+        [1, 2],
+        ListConfig([1, 2]),
+    ],
+)
+def test_list_assignment_deepcopy_semantics(node: Any) -> None:
+    cfg = OmegaConf.create()
+    cfg.foo = node
+    node[1] = 10
+    assert cfg.foo[1] == 2
+
+
 def test_assign_does_not_modify_src_config() -> None:
     cfg1 = OmegaConf.create({"foo": {"bar": 10}})
     cfg2 = OmegaConf.create({})
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -223,6 +223,15 @@ def test_create_node_parent_retained_on_create(node: Any) -> None:
     assert cfg1.foo._get_parent() is cfg1
 
 
+@pytest.mark.parametrize("node", [({"bar": 10}), ([1, 2, 3])])
+def test_create_node_parent_retained_on_assign(node: Any) -> None:
+    cfg1 = OmegaConf.create({"foo": node})
+    cfg2 = OmegaConf.create()
+    cfg2.zonk = cfg1.foo
+    assert cfg1.foo._get_parent() is cfg1
+    assert cfg2.zonk._get_parent() is cfg2
+
+
 def test_assign_does_not_modify_src_config() -> None:
     cfg1 = OmegaConf.create({"foo": {"bar": 10}})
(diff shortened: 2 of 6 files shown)


.pre-commit-config.yaml
```diff
@@ -21,3 +21,5 @@ repos:
     hooks:
       - id: mypy
         args: [--strict]
+        additional_dependencies: ['pytest']
```
odelalleau, line 24: Is this change related to the other changes in the PR? (it's not clear)

news/477.feature
```diff
@@ -0,0 +1 @@
+Perf improvement: eliminated unnecessary deepcopy calls during construction
```
odelalleau, line 1: Not sure if it matters, but this is also a potentially breaking change (if I understand correctly), because of something like:
```python
bar = OmegaConf.create({"zonk": 0})
cfg.foo = bar
bar.zonk = 1
assert cfg.foo.zonk == 0  # would fail now?
```
I didn't test it though so maybe that's not the case?

tests/test_create.py
```diff
@@ -220,6 +220,17 @@ def test_create_node_parent_retained_on_create(node: Any) -> None:
     cfg2 = OmegaConf.create({"zonk": cfg1.foo})
     assert cfg2 == {"zonk": node}
     assert cfg1.foo._get_parent() == cfg1
+    assert id(cfg1.foo._get_parent()) == id(cfg1)
```
odelalleau, line 223: Better use `is` rather than `id`:
```suggestion
    assert cfg1.foo._get_parent() is cfg1
```
(same for other `id`-based comparisons below)

odelalleau: Also I think a test is also needed to check that we don't make a copy anymore. Would the following assert pass?
```python
assert cfg1._get_node("foo") is node
```

odelalleau (changes requested):
A few questions / comments

.pre-commit-config.yaml
odelalleau: Oh right actually this broke a while ago, I ran into issues too and thought it may be a problem with me needing to re-create my env or something, but I never took the time to look at it more closely...
=> resolving

news/477.feature
odelalleau: Ok I'll have to look at the updated code more closely, but as long as you're aware of it that sounds fine for now => resolving

tests/test_create.py
odelalleau: Ok => resolving

odelalleau:
Sorry I'll need to re-review more carefully as the PR has gotten more complex, in particular it's not immediately clear to me what e04e00e6374093e05d961951780cac440c5b360c is fixing (more context would be helpful). This may need to wait 1-2 days due to other priorities on my end.

omry:
> Sorry I'll need to re-review more carefully as the PR has gotten more complex, in particular it's not immediately clear to me what [e04e00e](https://github.com/omry/omegaconf/commit/e04e00e6374093e05d961951780cac440c5b360c) is fixing (more context would be helpful). This may need to wait 1-2 days due to other priorities on my end.

Yeah, it's a bit hard to see because I added the pytest.params there.
There is a new test there (separate function). before this fix this resulted in a runtime error (str does not have items()).
The semantics implemented for copying interpolations there are a bit odd and I want to rethink them.

omry:
I am going to undo the optimization part of this diff as it doesn't make a difference (test changes will stay).

I have an upcoming diff adding a proper benchmark. It shows that create performance is not affected as I suspected by the deepcopy even for large dicts. (tested on dict and DictConfig).

I also found a real perf issue with the flags, adding a flags cache on each node helped a lot (600ms -> 450ms for specific benchmark with 10 levels dict).

By the time you look at this diff it will probably be very different.

odelalleau:
> I am going to undo the optimization part of this diff as it doesn't make a difference (test changes will stay).
> (...)
> By the time you look at this diff it will probably be very different.

Ok in that case I'll mark it as draft, just un-draft it when it's ready for re-review

omry:
See #481

PR #445, opened 2020-11-26
Major changes to interpolations and resolvers

odelalleau:
This is a squashed version of #321. ~~Currently it is built on top of PR #444 so it includes this commit as well (I'm hoping that #444 will be merged quickly enough, but if that's not the case I'll disentangle them).~~ (#444 has been merged)

* Add a grammar to parse interpolations
* Add support for nested interpolations
* Deprecate register_resolver() and introduce new_register_resolver() (that allows resolvers to use non-string arguments, and to decide whether or not to use the cache)
* The `env` resolver now parses environment variables in a way that is consistent with the new interpolation grammar

Fixes #100 #230 #266 #318

PR #484, opened 2021-01-22
Adding support for bool and float as DictConfig key types

### LINKED ISSUE #483: DictConfig: Support float and bool keys
Currently we have `DictKeyType = Union[str, int, Enum]`, i.e. `DictConfig` supports keys of type `str` or `int` or `Enum`.

This feature request is for expansion of `DictKeyType` to include `float` and `bool`:
`DictKeyType = Union[str, int, Enum, float, bool]`.

Additionally, given that `null` is a valid as a key in YAML, we could consider adding support for `None` as a `DictConfig` key, although I suspect that support for `None` may be more difficult than support for `float`/`bool`.

As a starting point for implementation, PR #454 implemented support for `int` key type in `DictConfig`.

### LINKED ISSUE #554: `__delitem__` not as expected for enum key-type DictConfig
Given a config with Enum key-type:
```python
from enum import Enum
from omegaconf import DictConfig

class Enum1(Enum):
    FOO = 1
    BAR = 2

cfg = DictConfig({Enum1.FOO: 'enum key'}, key_type=Enum1)
```
normally it is possible to use either the enum value `Enum1.FOO` or the `str` value `"FOO"` for access:
```python
>>> cfg[Enum1.FOO]
'enum key'
>>> cfg["FOO"]
'enum key'
```

But using string access fails when `__delitem__` is called:
```python
>>> del cfg["FOO"]
  File "/home/jasha10/omegaconf/omegaconf/dictconfig.py", line 405, in __delitem__
    del self.__dict__["_content"][key]
KeyError: 'FOO'
```

The `__delitem__` works as expected when called with an enum value:
```python
>>> del cfg[Enum1.FOO]
>>> assert Enum1.FOO not in cfg
```

### Expected behavior:
The `__delitem__` method should behave the same as `__setitem__` and `__getitem__`: when a `DictConfig` instance has enum-typed keys, either a string or an enum member should be usable to specify what index will be deleted.

Jasha10:
Closes #483
Closes #554

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -51,7 +51,23 @@ From a dictionary
       3: c
     <BLANKLINE>
 
-OmegaConf supports `str`, `int` and Enums as dictionary key types.
+Here is an example of various supported key types:
+
+.. doctest::
+
+    >>> from enum import Enum
+    >>> class Color(Enum):
+    ...   RED = 1
+    ...   BLUE = 2
+    >>> 
+    >>> conf = OmegaConf.create(
+    ...   {"key": "str", 123: "int", True: "bool", 3.14: "float", Color.RED: "Color"}
+    ... )
+    >>> 
+    >>> print(conf)
+    {'key': 'str', 123: 'int', True: 'bool', 3.14: 'float', <Color.RED: 1>: 'Color'}
+
+OmegaConf supports `str`, `int`, `bool`, `float` and Enums as dictionary key types.
 
 From a list
 ^^^^^^^^^^^
--- a/news/483.feature
+++ b/news/483.feature
@@ -1,0 +1,1 @@
+Add DictConfig support for keys of type float and bool
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -19,7 +19,7 @@ from .grammar.gen.OmegaConfGrammarParser import OmegaConfGrammarParser
 from .grammar_parser import parse
 from .grammar_visitor import GrammarVisitor
 
-DictKeyType = Union[str, int, Enum]
+DictKeyType = Union[str, int, Enum, float, bool]
 
 _MARKER_ = object()
 
@@ -179,7 +179,7 @@ class Node(ABC):
         assert False
 
     @abstractmethod
-    def _get_full_key(self, key: Union[str, Enum, int, None]) -> str:
+    def _get_full_key(self, key: Optional[Union[DictKeyType, int]]) -> str:
         ...
 
     def _dereference_node(
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -44,7 +44,7 @@ class BaseContainer(Container, ABC):
 
     def _resolve_with_default(
         self,
-        key: Union[str, int, Enum],
+        key: Union[DictKeyType, int],
         value: Any,
         default_value: Any = DEFAULT_VALUE_MARKER,
     ) -> Any:
@@ -695,11 +695,11 @@ class BaseContainer(Container, ABC):
     def _value(self) -> Any:
         return self.__dict__["_content"]
 
-    def _get_full_key(self, key: Union[str, Enum, int, slice, None]) -> str:
+    def _get_full_key(self, key: Union[DictKeyType, int, slice, None]) -> str:
         from .listconfig import ListConfig
         from .omegaconf import _select_one
 
-        if not isinstance(key, (int, str, Enum, slice, type(None))):
+        if not isinstance(key, (int, str, Enum, float, bool, slice, type(None))):
             return ""
 
         def _slice_to_str(x: slice) -> str:
@@ -713,6 +713,8 @@ class BaseContainer(Container, ABC):
                 key = _slice_to_str(key)
             elif isinstance(key, Enum):
                 key = key.name
+            elif isinstance(key, (int, float, bool)):
+                key = str(key)
 
             if issubclass(parent_type, ListConfig):
                 if full_key != "":
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -275,28 +275,26 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
     def _s_validate_and_normalize_key(self, key_type: Any, key: Any) -> DictKeyType:
         if key_type is Any:
             for t in DictKeyType.__args__:  # type: ignore
-                try:
-                    return self._s_validate_and_normalize_key(key_type=t, key=key)
-                except KeyValidationError:
-                    pass
+                if isinstance(key, t):
+                    return key  # type: ignore
             raise KeyValidationError("Incompatible key type '$KEY_TYPE'")
-        elif key_type == str:
-            if not isinstance(key, str):
+        elif key_type is bool and key in [0, 1]:
+            # Python treats True as 1 and False as 0 when used as dict keys
+            #   assert hash(0) == hash(False)
+            #   assert hash(1) == hash(True)
+            return bool(key)
+        elif key_type in (str, int, float, bool):  # primitive type
+            if not isinstance(key, key_type):
                 raise KeyValidationError(
                     f"Key $KEY ($KEY_TYPE) is incompatible with ({key_type.__name__})"
                 )
 
-            return key
-        elif key_type == int:
-            if not isinstance(key, int):
-                raise KeyValidationError(
-                    f"Key $KEY ($KEY_TYPE) is incompatible with ({key_type.__name__})"
-                )
-
-            return key
+            return key  # type: ignore
         elif issubclass(key_type, Enum):
             try:
-                ret = EnumNode.validate_and_convert_to_enum(key_type, key)
+                ret = EnumNode.validate_and_convert_to_enum(
+                    key_type, key, allow_none=False
+                )
                 assert ret is not None
                 return ret
             except ValidationError:
@@ -377,6 +375,7 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
             self._format_and_raise(key=key, value=None, cause=e)
 
     def __delitem__(self, key: DictKeyType) -> None:
+        key = self._validate_and_normalize_key(key)
         if self._get_flag("readonly"):
             self._format_and_raise(
                 key=key,
@@ -402,7 +401,11 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
                 ),
             )
 
-        del self.__dict__["_content"][key]
+        try:
+            del self.__dict__["_content"][key]
+        except KeyError:
+            msg = "Key not found: '$KEY'"
+            self._format_and_raise(key=key, value=None, cause=ConfigKeyError(msg))
 
     def get(self, key: DictKeyType, default_value: Any = None) -> Any:
         """Return the value for `key` if `key` is in the dictionary, else
--- a/omegaconf/nodes.py
+++ b/omegaconf/nodes.py
@@ -11,7 +11,7 @@ from omegaconf._utils import (
     get_value_kind,
     is_primitive_container,
 )
-from omegaconf.base import Container, Metadata, Node
+from omegaconf.base import Container, DictKeyType, Metadata, Node
 from omegaconf.errors import (
     ConfigKeyError,
     ReadonlyConfigError,
@@ -122,7 +122,7 @@ class ValueNode(Node):
     def _is_interpolation(self) -> bool:
         return _is_interpolation(self._value())
 
-    def _get_full_key(self, key: Union[str, Enum, int, None]) -> str:
+    def _get_full_key(self, key: Optional[Union[DictKeyType, int]]) -> str:
         parent = self._get_parent()
         if parent is None:
             if self._metadata.key is None:
@@ -366,9 +366,9 @@ class EnumNode(ValueNode):  # lgtm [py/missing-equals] : Intentional.
 
     @staticmethod
     def validate_and_convert_to_enum(
-        enum_type: Type[Enum], value: Any
+        enum_type: Type[Enum], value: Any, allow_none: bool = True
     ) -> Optional[Enum]:
-        if value is None:
+        if allow_none and value is None:
             return None
 
         if not isinstance(value, (str, int)) and not isinstance(value, enum_type):
--- a/tests/structured_conf/data/dataclasses.py
+++ b/tests/structured_conf/data/dataclasses.py
@@ -403,10 +403,10 @@ class DictExamples:
         }
     )
     int_keys: Dict[int, str] = field(default_factory=lambda: {1: "one", 2: "two"})
-
-
-@dataclass
-class DictWithEnumKeys:
+    float_keys: Dict[float, str] = field(
+        default_factory=lambda: {1.1: "one", 2.2: "two"}
+    )
+    bool_keys: Dict[bool, str] = field(default_factory=lambda: {True: "T", False: "F"})
     enum_key: Dict[Color, str] = field(
         default_factory=lambda: {Color.RED: "red", Color.GREEN: "green"}
     )
@@ -431,6 +431,14 @@ class DictSubclass:
 
     @dataclass
     class Int2Str(Dict[int, str]):
+        pass
+
+    @dataclass
+    class Float2Str(Dict[float, str]):
+        pass
+
+    @dataclass
+    class Bool2Str(Dict[bool, str]):
         pass
 
     @dataclass
--- a/tests/structured_conf/test_structured_config.py
+++ b/tests/structured_conf/test_structured_config.py
@@ -639,42 +639,47 @@ class TestConfigs:
             Color.BLUE,
         ]
 
-    def test_dict_examples(self, class_type: str) -> None:
+    def test_dict_examples_any(self, class_type: str) -> None:
         module: Any = import_module(class_type)
         conf = OmegaConf.structured(module.DictExamples)
 
-        def test_any(name: str) -> None:
-            conf[name].c = True
-            conf[name].d = Color.RED
-            conf[name].e = 3.1415
-            assert conf[name] == {
-                "a": 1,
-                "b": "foo",
-                "c": True,
-                "d": Color.RED,
-                "e": 3.1415,
-            }
+        dct = conf.any
+        dct.c = True
+        dct.d = Color.RED
+        dct.e = 3.1415
+        assert dct == {"a": 1, "b": "foo", "c": True, "d": Color.RED, "e": 3.1415}
 
-        # any and untyped
-        test_any("any")
+    def test_dict_examples_int(self, class_type: str) -> None:
+        module: Any = import_module(class_type)
+        conf = OmegaConf.structured(module.DictExamples)
+        dct = conf.ints
 
         # test ints
         with pytest.raises(ValidationError):
-            conf.ints.a = "foo"
-        conf.ints.c = 10
-        assert conf.ints == {"a": 10, "b": 20, "c": 10}
+            dct.a = "foo"
+        dct.c = 10
+        assert dct == {"a": 10, "b": 20, "c": 10}
+
+    def test_dict_examples_strings(self, class_type: str) -> None:
+        module: Any = import_module(class_type)
+        conf = OmegaConf.structured(module.DictExamples)
 
         # test strings
         conf.strings.c = Color.BLUE
         assert conf.strings == {"a": "foo", "b": "bar", "c": "Color.BLUE"}
 
-        # tests booleans
+    def test_dict_examples_bool(self, class_type: str) -> None:
+        module: Any = import_module(class_type)
+        conf = OmegaConf.structured(module.DictExamples)
+        dct = conf.booleans
+
+        # test bool
         with pytest.raises(ValidationError):
-            conf.booleans.a = "foo"
-        conf.booleans.c = True
-        conf.booleans.d = "off"
-        conf.booleans.e = 1
-        assert conf.booleans == {
+            dct.a = "foo"
+        dct.c = True
+        dct.d = "off"
+        dct.e = 1
+        assert dct == {
             "a": True,
             "b": False,
             "c": True,
@@ -682,60 +687,110 @@ class TestConfigs:
             "e": True,
         }
 
-        # test colors
-        with pytest.raises(ValidationError):
-            conf.colors.foo = "foo"
-        conf.colors.c = Color.BLUE
-        conf.colors.d = "RED"
-        conf.colors.e = "Color.GREEN"
-        conf.colors.f = 3
-        assert conf.colors == {
-            "red": Color.RED,
-            "green": Color.GREEN,
-            "blue": Color.BLUE,
-            "c": Color.BLUE,
-            "d": Color.RED,
-            "e": Color.GREEN,
-            "f": Color.BLUE,
-        }
+    class TestDictExamples:
+        @pytest.fixture
+        def conf(self, class_type: str) -> DictConfig:
+            module: Any = import_module(class_type)
+            conf: DictConfig = OmegaConf.structured(module.DictExamples)
+            return conf
 
-        # test int_keys
-        with pytest.raises(KeyValidationError):
-            conf.int_keys.foo_key = "foo_value"
-        conf.int_keys[3] = "three"
-        assert conf.int_keys == {
-            1: "one",
-            2: "two",
-            3: "three",
-        }
+        def test_dict_examples_colors(self, conf: DictConfig) -> None:
+            dct = conf.colors
 
-    def test_enum_key(self, class_type: str) -> None:
-        module: Any = import_module(class_type)
-        conf = OmegaConf.structured(module.DictWithEnumKeys)
+            # test colors
+            with pytest.raises(ValidationError):
+                dct.foo = "foo"
+            dct.c = Color.BLUE
+            dct.d = "RED"
+            dct.e = "Color.GREEN"
+            dct.f = 3
+            assert dct == {
+                "red": Color.RED,
+                "green": Color.GREEN,
+                "blue": Color.BLUE,
+                "c": Color.BLUE,
+                "d": Color.RED,
+                "e": Color.GREEN,
+                "f": Color.BLUE,
+            }
 
-        # When an Enum is a dictionary key the name of the Enum is actually used
-        # as the key
-        assert conf.enum_key.RED == "red"
-        assert conf.enum_key["GREEN"] == "green"
-        assert conf.enum_key[Color.GREEN] == "green"
+        def test_dict_examples_str_keys(self, conf: DictConfig) -> None:
+            dct = conf.any
 
-        conf.enum_key["BLUE"] = "Blue too"
-        assert conf.enum_key[Color.BLUE] == "Blue too"
-        with pytest.raises(KeyValidationError):
-            conf.enum_key["error"] = "error"
+            with pytest.raises(KeyValidationError):
+                dct[123] = "bad key type"
+            dct["c"] = "three"
+            assert dct == {
+                "a": 1,
+                "b": "foo",
+                "c": "three",
+            }
+
+        def test_dict_examples_int_keys(self, conf: DictConfig) -> None:
+            dct = conf.int_keys
+
+            # test int keys
+            with pytest.raises(KeyValidationError):
+                dct.foo_key = "foo_value"
+            dct[3] = "three"
+            assert dct == {
+                1: "one",
+                2: "two",
+                3: "three",
+            }
+
+        def test_dict_examples_float_keys(self, conf: DictConfig) -> None:
+            dct = conf.float_keys
+
+            # test float keys
+            with pytest.raises(KeyValidationError):
+                dct.foo_key = "foo_value"
+            dct[3.3] = "three"
+            assert dct == {
+                1.1: "one",
+                2.2: "two",
+                3.3: "three",
+            }
+
+        def test_dict_examples_bool_keys(self, conf: DictConfig) -> None:
+            dct = conf.bool_keys
+
+            # test bool_keys
+            with pytest.raises(KeyValidationError):
+                dct.foo_key = "foo_value"
+            dct[True] = "new value"
+            assert dct == {
+                True: "new value",
+                False: "F",
+            }
+
+        def test_dict_examples_enum_key(self, conf: DictConfig) -> None:
+            dct = conf.enum_key
+
+            # When an Enum is a dictionary key the name of the Enum is actually used
+            # as the key
+            assert dct.RED == "red"
+            assert dct["GREEN"] == "green"
+            assert dct[Color.GREEN] == "green"
+
+            dct["BLUE"] = "Blue too"
+            assert dct[Color.BLUE] == "Blue too"
+            with pytest.raises(KeyValidationError):
+                dct["error"] = "error"
 
     def test_dict_of_objects(self, class_type: str) -> None:
         module: Any = import_module(class_type)
         conf = OmegaConf.structured(module.DictOfObjects)
-        assert conf.users.joe.age == 18
-        assert conf.users.joe.name == "Joe"
+        dct = conf.users
+        assert dct.joe.age == 18
+        assert dct.joe.name == "Joe"
 
-        conf.users.bond = module.User(name="James Bond", age=7)
-        assert conf.users.bond.name == "James Bond"
-        assert conf.users.bond.age == 7
+        dct.bond = module.User(name="James Bond", age=7)
+        assert dct.bond.name == "James Bond"
+        assert dct.bond.age == 7
 
         with pytest.raises(ValidationError):
-            conf.users.fail = "fail"
+            dct.fail = "fail"
 
     def test_list_of_objects(self, class_type: str) -> None:
         module: Any = import_module(class_type)
--- a/tests/test_basic_ops_dict.py
+++ b/tests/test_basic_ops_dict.py
@@ -73,39 +73,51 @@ def test_getattr_dict() -> None:
 
 
 @pytest.mark.parametrize(
-    "key",
-    ["a", 1],
+    "key,match",
+    [
+        pytest.param("a", "a", id="str"),
+        pytest.param(1, "1", id="int"),
+        pytest.param(123.45, "123.45", id="float"),
+        pytest.param(True, "True", id="bool-T"),
+        pytest.param(False, "False", id="bool-F"),
+        pytest.param(Enum1.FOO, "FOO", id="enum"),
+    ],
 )
 class TestDictKeyTypes:
-    def test_mandatory_value(self, key: DictKeyType) -> None:
+    def test_mandatory_value(self, key: DictKeyType, match: str) -> None:
         c = OmegaConf.create({key: "???"})
-        with pytest.raises(MissingMandatoryValue, match=str(key)):
+        with pytest.raises(MissingMandatoryValue, match=match):
             c[key]
         if isinstance(key, str):
-            with pytest.raises(MissingMandatoryValue, match=key):
+            with pytest.raises(MissingMandatoryValue, match=match):
                 getattr(c, key)
 
-    def test_nested_dict_mandatory_value(self, key: DictKeyType) -> None:
+    def test_nested_dict_mandatory_value_inner(
+        self, key: DictKeyType, match: str
+    ) -> None:
         c = OmegaConf.create({"b": {key: "???"}})
-        with pytest.raises(MissingMandatoryValue):
+        with pytest.raises(MissingMandatoryValue, match=match):
             c.b[key]
         if isinstance(key, str):
-            with pytest.raises(MissingMandatoryValue):
+            with pytest.raises(MissingMandatoryValue, match=match):
                 getattr(c.b, key)
 
+    def test_nested_dict_mandatory_value_outer(
+        self, key: DictKeyType, match: str
+    ) -> None:
         c = OmegaConf.create({key: {"b": "???"}})
-        with pytest.raises(MissingMandatoryValue):
+        with pytest.raises(MissingMandatoryValue, match=match):
             c[key].b
         if isinstance(key, str):
-            with pytest.raises(MissingMandatoryValue):
+            with pytest.raises(MissingMandatoryValue, match=match):
                 getattr(c, key).b
 
-    def test_subscript_get(self, key: DictKeyType) -> None:
+    def test_subscript_get(self, key: DictKeyType, match: str) -> None:
         c = OmegaConf.create({key: "b"})
         assert isinstance(c, DictConfig)
         assert "b" == c[key]
 
-    def test_subscript_set(self, key: DictKeyType) -> None:
+    def test_subscript_set(self, key: DictKeyType, match: str) -> None:
         c = OmegaConf.create()
         c[key] = "b"
         assert {key: "b"} == c
@@ -116,6 +128,9 @@ class TestDictKeyTypes:
     [
         ({"a": 10, "b": 11}, "a", {"b": 11}),
         ({1: "a", 2: "b"}, 1, {2: "b"}),
+        ({123.45: "a", 67.89: "b"}, 67.89, {123.45: "a"}),
+        ({True: "a", False: "b"}, False, {True: "a"}),
+        ({Enum1.FOO: "foo", Enum1.BAR: "bar"}, Enum1.FOO, {Enum1.BAR: "bar"}),
     ],
 )
 class TestDelitemKeyTypes:
@@ -124,8 +139,14 @@ class TestDelitemKeyTypes:
         assert c == src
         del c[key]
         assert c == expected
+
+    def test_dict_delitem_KeyError(
+        self, src: Any, key: DictKeyType, expected: Any
+    ) -> None:
+        c = OmegaConf.create(expected)
+        assert c == expected
         with pytest.raises(KeyError):
-            del c["not_found"]
+            del c[key]
 
     def test_dict_struct_delitem(
         self, src: Any, key: DictKeyType, expected: Any
@@ -282,6 +303,7 @@ def test_iterate_dict_with_interpolation() -> None:
 @pytest.mark.parametrize(
     "cfg, key, default_, expected",
     [
+        # string key
         pytest.param({"a": 1, "b": 2}, "a", "__NO_DEFAULT__", 1, id="no_default"),
         pytest.param({"a": 1, "b": 2}, "not_found", None, None, id="none_default"),
         pytest.param(
@@ -311,6 +333,31 @@ def test_iterate_dict_with_interpolation() -> None:
             "default",
             "default",
             id="enum_key_with_default",
+        ),
+        # other key types
+        pytest.param(
+            {123.45: "a", 67.89: "b"},
+            67.89,
+            "__NO_DEFAULT__",
+            "b",
+            id="float_key_no_default",
+        ),
+        pytest.param(
+            {123.45: "a", 67.89: "b"},
+            "not found",
+            None,
+            None,
+            id="float_key_with_default",
+        ),
+        pytest.param(
+            {True: "a", False: "b"},
+            False,
+            "__NO_DEFAULT__",
+            "b",
+            id="bool_key_no_default",
+        ),
+        pytest.param(
+            {True: "a", False: "b"}, "not found", None, None, id="bool_key_with_default"
         ),
     ],
 )
@@ -364,12 +411,23 @@ def test_dict_structured_mode_pop() -> None:
 @pytest.mark.parametrize(
     "cfg, key, expectation",
     [
+        # key not found
         ({"a": 1, "b": 2}, "not_found", pytest.raises(KeyError)),
+        ({1: "a", 2: "b"}, 3, pytest.raises(KeyError)),
+        ({123.45: "a", 67.89: "b"}, 10.11, pytest.raises(KeyError)),
+        ({True: "a"}, False, pytest.raises(KeyError)),
+        ({Enum1.FOO: "bar"}, Enum1.BAR, pytest.raises(KeyError)),
         # Interpolations
         ({"a": "???", "b": 2}, "a", pytest.raises(MissingMandatoryValue)),
+        ({1: "???", 2: "b"}, 1, pytest.raises(MissingMandatoryValue)),
+        ({123.45: "???", 67.89: "b"}, 123.45, pytest.raises(MissingMandatoryValue)),
+        ({True: "???", False: "b"}, True, pytest.raises(MissingMandatoryValue)),
+        (
+            {Enum1.FOO: "???", Enum1.BAR: "bar"},
+            Enum1.FOO,
+            pytest.raises(MissingMandatoryValue),
+        ),
         ({"a": "${b}", "b": "???"}, "a", pytest.raises(MissingMandatoryValue)),
-        # enum key
-        ({Enum1.FOO: "bar"}, Enum1.BAR, pytest.raises(KeyError)),
     ],
 )
 def test_dict_pop_error(cfg: Dict[Any, Any], key: Any, expectation: Any) -> None:
@@ -382,6 +440,7 @@ def test_dict_pop_error(cfg: Dict[Any, Any], key: Any, expectation: Any) -> None
 @pytest.mark.parametrize(
     "conf,key,expected",
     [
+        # str key type
         ({"a": 1, "b": {}}, "a", True),
         ({"a": 1, "b": {}}, "b", True),
         ({"a": 1, "b": {}}, "c", False),
@@ -392,8 +451,10 @@ def test_dict_pop_error(cfg: Dict[Any, Any], key: Any, expectation: Any) -> None
         ({"a": "${unknown_resolver:bar}"}, "a", True),
         ({"a": None, "b": "${a}"}, "b", True),
         ({"a": "cat", "b": "${a}"}, "b", True),
+        # Enum key type
         ({Enum1.FOO: 1, "b": {}}, Enum1.FOO, True),
         ({Enum1.FOO: 1, "b": {}}, "aaa", False),
+        ({Enum1.FOO: 1, "b": {}}, "FOO", False),
         (
             DictConfig(content={Enum1.FOO: "foo"}, key_type=Enum1, element_type=str),
             Enum1.FOO,
@@ -404,10 +465,39 @@ def test_dict_pop_error(cfg: Dict[Any, Any], key: Any, expectation: Any) -> None
             "incompatible_key_type",
             False,
         ),
+        (
+            DictConfig(content={Enum1.FOO: "foo"}, key_type=Enum1, element_type=str),
+            "FOO",
+            True,
+        ),
+        (
+            DictConfig(content={Enum1.FOO: "foo"}, key_type=Enum1, element_type=str),
+            None,
+            False,
+        ),
+        # int key type
         ({1: "a", 2: {}}, 1, True),
         ({1: "a", 2: {}}, 2, True),
         ({1: "a", 2: {}}, 3, False),
         ({1: "a", 2: "???"}, 2, False),
+        ({1: "a", 2: "???"}, None, False),
+        ({1: "a", 2: "???"}, "1", False),
+        (DictConfig({1: "a", 2: "???"}, key_type=int), "1", False),
+        # float key type
+        ({1.1: "a", 2.2: {}}, 1.1, True),
+        ({1.1: "a", 2.2: {}}, "1.1", False),
+        (DictConfig({1.1: "a", 2.2: {}}, key_type=float), "1.1", False),
+        ({1.1: "a", 2.2: {}}, 2.2, True),
+        ({1.1: "a", 2.2: {}}, 3.3, False),
+        ({1.1: "a", 2.2: "???"}, 2.2, False),
+        ({1.1: "a", 2.2: "???"}, None, False),
+        # bool key type
+        ({True: "a", False: {}}, True, True),
+        ({True: "a", False: {}}, False, True),
+        ({True: "a", False: {}}, "no", False),
+        ({True: "a", False: {}}, 1, True),
+        ({True: "a", False: {}}, None, False),
+        ({True: "a", False: "???"}, False, False),
     ],
 )
 def test_in_dict(conf: Any, key: str, expected: Any) -> None:
--- a/tests/test_compare_dictconfig_vs_dict.py
+++ b/tests/test_compare_dictconfig_vs_dict.py
@@ -49,7 +49,7 @@ def python_dict(data: Dict[Any, Any]) -> Dict[Any, Any]:
     return deepcopy(data)
 
 
-@fixture(params=[True, False, None])
+@fixture(params=[None, False, True])
 def struct_mode(request: Any) -> Optional[bool]:
     struct_mode: Optional[bool] = request.param
     return struct_mode
--- a/tests/test_compare_dictconfig_vs_dict.py
+++ b/tests/test_compare_dictconfig_vs_dict.py
@@ -97,28 +97,26 @@ class TestUntypedDictConfig:
         else:
             assert result == cfg[key]
 
-    def test__delitem__(
-        self, python_dict: Any, cfg: DictConfig, key: Any, struct_mode: Optional[bool]
-    ) -> None:
+    @mark.parametrize("struct_mode", [False, None])
+    def test__delitem__(self, python_dict: Any, cfg: DictConfig, key: Any) -> None:
         """Ensure that __delitem__ has same result with python dict as with DictConfig."""
-        if not struct_mode:
-            try:
-                del python_dict[key]
-                assert key not in python_dict
-            except KeyError:
-                with raises(ConfigKeyError):
-                    del cfg[key]
-            else:
+        try:
+            del python_dict[key]
+            assert key not in python_dict
+        except KeyError:
+            with raises(ConfigKeyError):
                 del cfg[key]
-                assert key not in cfg
+        else:
+            del cfg[key]
+            assert key not in cfg
 
+    @mark.parametrize("struct_mode", [True])
     def test__delitem__struct_mode(
-        self, python_dict: Any, cfg: DictConfig, key: Any, struct_mode: Optional[bool]
+        self, python_dict: Any, cfg: DictConfig, key: Any
     ) -> None:
         """Ensure that __delitem__ fails in struct_mode"""
-        if struct_mode:
-            with raises(ConfigTypeError):
-                del cfg[key]
+        with raises(ConfigTypeError):
+            del cfg[key]
 
     def test__contains__(self, python_dict: Any, cfg: Any, key: Any) -> None:
         """Ensure that __contains__ has same result with python dict as with DictConfig."""
@@ -137,59 +135,55 @@ class TestUntypedDictConfig:
         """Ensure that __getitem__ has same result with python dict as with DictConfig."""
         assert python_dict.get(key, "DEFAULT") == cfg.get(key, "DEFAULT")
 
+    @mark.parametrize("struct_mode", [False, None])
     def test_pop(
         self,
         python_dict: Any,
         cfg: DictConfig,
         key: Any,
-        struct_mode: Optional[bool],
     ) -> None:
         """Ensure that pop has same result with python dict as with DictConfig."""
-        if not struct_mode:
-            try:
-                result = python_dict.pop(key)
-            except KeyError:
-                with raises(ConfigKeyError):
-                    cfg.pop(key)
-            else:
-                assert result == cfg.pop(key)
-            assert python_dict.keys() == cfg.keys()
+        try:
+            result = python_dict.pop(key)
+        except KeyError:
+            with raises(ConfigKeyError):
+                cfg.pop(key)
+        else:
+            assert result == cfg.pop(key)
+        assert python_dict.keys() == cfg.keys()
 
+    @mark.parametrize("struct_mode", [True])
     def test_pop_struct_mode(
         self,
         python_dict: Any,
         cfg: DictConfig,
         key: Any,
-        struct_mode: Optional[bool],
     ) -> None:
-        if struct_mode:
-            """Ensure that pop fails in struct mode."""
-            with raises(ConfigTypeError):
-                cfg.pop(key)
+        """Ensure that pop fails in struct mode."""
+        with raises(ConfigTypeError):
+            cfg.pop(key)
 
+    @mark.parametrize("struct_mode", [False, None])
     def test_pop_with_default(
         self,
         python_dict: Any,
         cfg: DictConfig,
         key: Any,
-        struct_mode: Optional[bool],
     ) -> None:
         """Ensure that pop(..., DEFAULT) has same result with python dict as with DictConfig."""
-        if not struct_mode:
-            assert python_dict.pop(key, "DEFAULT") == cfg.pop(key, "DEFAULT")
-            assert python_dict.keys() == cfg.keys()
+        assert python_dict.pop(key, "DEFAULT") == cfg.pop(key, "DEFAULT")
+        assert python_dict.keys() == cfg.keys()
 
+    @mark.parametrize("struct_mode", [True])
     def test_pop_with_default_struct_mode(
         self,
         python_dict: Any,
         cfg: DictConfig,
         key: Any,
-        struct_mode: Optional[bool],
     ) -> None:
         """Ensure that pop(..., DEFAULT) fails in struct mode."""
-        if struct_mode:
-            with raises(ConfigTypeError):
-                cfg.pop(key, "DEFAULT")
+        with raises(ConfigTypeError):
+            cfg.pop(key, "DEFAULT")
 
     def test_keys(self, python_dict: Any, cfg: Any) -> None:
         assert python_dict.keys() == cfg.keys()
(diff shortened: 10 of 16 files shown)


Jasha10:
For the record, I found that implementing support for `None` as a key value was challenging due to special handling of `None` values.

tests/test_basic_ops_dict.py
```diff
@@ -398,10 +403,29 @@ def test_dict_pop_error(cfg: Dict[Any, Any], key: Any, expectation: Any) -> None
             "incompatible_key_type",
             False,
         ),
+        (  # unexpected failure?
+            DictConfig(content={Enum1.FOO: "foo"}, key_type=Enum1, element_type=str),
+            "FOO",
+            False,
+        ),
```
Jasha10, line 410: I was surprised that this test fails.
This was surprising to me because two of the previous tests asserts successfully that
```
assert "FOO" not in OmegaConf.create({Enum1.FOO: 1, "b": {}})
assert "incompatible_key_type" not in OmegaConf.create(DictConfig(content={Enum1.FOO: "foo"}, key_type=Enum1, element_type=str))
```
whereas this test fails when we try
```
assert "FOO" not in OmegaConf.create(DictConfig(content={Enum1.FOO: "foo"}, key_type=Enum1, element_type=str))
```

This means that the string `"FOO"` is being coerced to the `Enum1` type by the `DictConfig.__contains__` method, which seems to contradict the use of the string `"incompatible_key_type"` in the previous test (which suggests that the intended behavior is for `__contains__(self, key)` is to return `False` when `self` has `key_type` that is different from the type of the passed `key`).

tests/test_basic_ops_dict.py
```diff
@@ -398,10 +403,29 @@ def test_dict_pop_error(cfg: Dict[Any, Any], key: Any, expectation: Any) -> None
...
+        ({1.1: "a", 2.2: {}}, "1.1", False),
+        ({1.1: "a", 2.2: {}}, 2.2, True),
+        ({1.1: "a", 2.2: {}}, 3.3, False),
+        ({1.1: "a", 2.2: "???"}, 2.2, False),
+        # bool key type
+        ({True: "a", False: {}}, True, True),
+        ({True: "a", False: {}}, False, True),
+        ({True: "a", False: {}}, "no", False),
+        ({True: "a", False: {}}, 1, False),  # unexpected failure
+        ({True: "a", False: {}}, None, False),  # unexpected AssertionError
```
Jasha10, line 427: I was surprised that these tests fail too. The first test fails because
```
1 in OmegaConf.create({True: "a", False: {}})
```
is `True` (which suggests that `1` is being coerced to `bool`), and the second test fails because `assert ret is not None` triggers `AssertionError` in [`DictConfig._s_validate_and_normalize_key`](https://github.com/7iW/omegaconf/blob/ce97297d6046e97bd22531d90d7357cda2b021fb/omegaconf/dictconfig.py#L305).

Jasha10 (commented):
It seems that some work may be needed to update the `DictConfig.__contains__` method.

tests/test_basic_ops_dict.py
Jasha10:
```
>>> "FOO" in omegaconf.DictConfig({Enum1.FOO: "foo"})
False
>>> "FOO" in omegaconf.DictConfig({Enum1.FOO: "foo"}, key_type=Enum1)
True
```
@omry is this the expected behavior?

omry:
Thanks for working on it, please request review when you are ready (converted to draft).

tests/test_basic_ops_dict.py
Jasha10: It seems that `(1 in {True: "a"}) is True` is standard behavior in python, so the first test result above is as expected.

Jasha10: The second test failure above is resolved in https://github.com/omry/omegaconf/pull/484/commits/61c08fdf713a9811519814da943eaa4b3534148f by changing
```
`assert ret is not None`
```
to
```
if ret is None: raise ValidationError
```

Jasha10:
Hi @omry, I have one design decision to make:

Currently `DictConfig.__contains__` is implemented so that
```
>>> "FOO" in DictConfig({Enum1.FOO: "bar"}, key_type=Enum1)
True
```
i.e. the string `"FOO"` is coerced to `key_type` if `key_type` is a subclass of `Enum`. Do we want the same behavior if a different `key_type` is set (and `key_type is not Any`)?
For example, is coersion from string to `int` or `float` desirable? This would result in:
```
>>> "123" in DictConfig({123: "bar"}, key_type=int)
True
>>> "123.45" in DictConfig({123.45: "bar"}, key_type=float)
True
```

Similarly, `__setitem__`, `__getitem__`, etc coerce the string `"FOO"` to `Enum1.FOO` when `key_type` is `Enum1`. Do we want these methods to coerce the strings `"123"` or `"123.45"` `key_type` is set to `int` or `float`, respectively? The alternative would be to raise e.g. a `KeyValidationError` when `DictConfig({123: "bar"}, key_type=int).__setitem__` is called with a string key.
Do we want `__setitem__`

omry:
Let's try to keep things consistent with standard Python behavior (do not convert keys).
It's worth considering changing the current behavior for enum where it (where "name" in dict matches Enum.name), but this is a breaking change that should be done carefully (and in a sepreate PR).

omegaconf/dictconfig.py
```diff
@@ -287,29 +288,28 @@ def _validate_and_normalize_key(self, key: Any) -> DictKeyType:
...
-
-            return key
-        elif key_type == int:
-            if not isinstance(key, int):
+        elif key_type == bool and (key == 0 or key == 1):
+            # motivated by an edge case in standard python:
+            # >>> assert hash(0) == hash(False)
+            # >>> assert hash(1) == hash(True)
+            # >>> {True: 123, False: 456}[1]
+            # 123
```
Jasha10, line 299: Not clear to me whether this is an implementation detail of CPython or an intended Python language feature.

odelalleau: Language feature, see this StackOverflow thread that asks exactly this: https://stackoverflow.com/questions/2764017/is-false-0-and-true-1-an-implementation-detail-or-is-it-guaranteed-by-the

Jasha10: Oh, very cool, thanks!

Jasha10: What I mean is thank you for the information.

Jasha10: This special handling is motivated by my [conversation](https://github.com/omry/omegaconf/pull/484#issuecomment-765740259) with Omry regarding consistency between OmegaConf and standard python behavior. In particular, because we have
```
>>> {True: 123}[1]
123
```
in standard python, we want to have
```
>>> DictConfig({True: 123}, key_type=bool)[1]
123
```
in OmegaConf.

Jasha10: In all other cases (besides `key_type==bool` and `key in (0,1)`) there is no conversion or coercion of keys, except for some special-cased coersion of `str/int` to `Enum` subclasses.

odelalleau: > What I mean is thank you for the information.

I was confused by this and then realized I accidentally clicked the sad emoji. Sorry, I meant to smile! :) (fixed now!)
 
> in OmegaConf, i.e. if `key_type==bool` then we treat integers `0` and `1` as bool.

Yep makes sense. As a side note I find this Python behavior more annoying than useful but I guess it's how it is...

tests/test_basic_ops_dict.py
```diff
@@ -398,17 +443,162 @@ def test_dict_pop_error(cfg: Dict[Any, Any], key: Any, expectation: Any) -> None
...
+            False,
+        ),
+        # int key type
         ({1: "a", 2: {}}, 1, True),
         ({1: "a", 2: {}}, 2, True),
         ({1: "a", 2: {}}, 3, False),
         ({1: "a", 2: "???"}, 2, False),
+        ({1: "a", 2: "???"}, None, False),
+        ({1: "a", 2: "???"}, "1", False),
+        # (DictConfig({1: "a", 2: "???"}, key_type=int), "1", True),
```
omry, line 463: Why is this commented out?

tests/test_basic_ops_dict.py
```diff
@@ -398,17 +443,162 @@ def test_dict_pop_error(cfg: Dict[Any, Any], key: Any, expectation: Any) -> None
...
+        ({True: "a", False: {}}, None, False),
+        ({True: "a", False: "???"}, False, False),
     ],
 )
 def test_in_dict(conf: Any, key: str, expected: Any) -> None:
     conf = OmegaConf.create(conf)
     assert (key in conf) == expected
 
 
+@pytest.mark.parametrize(
```
omry, line 486: That's a pretty important addition.
1. Please put it in a different file.
2. Please document it a bit, there is a lot going on here.

tests/test_basic_ops_dict.py
```diff
@@ -398,17 +443,162 @@ def test_dict_pop_error(cfg: Dict[Any, Any], key: Any, expectation: Any) -> None
...
+    @pytest.fixture
+    def cfg(self, python_dict: Any) -> DictConfig:
+        return DictConfig(content=python_dict)
+
+    @pytest.fixture
+    def cfg_typed(self, python_dict: Any, cfg_key_type: Any) -> DictConfig:
+        return DictConfig(content=python_dict, key_type=cfg_key_type)
+
+    @pytest.fixture
+    def key_coerced(self, key: Any, cfg_key_type: Any) -> Any:
```
omry, line 514: Is pytest passing other parameters from the function this fixture is used in automatically?!

tests/test_basic_ops_dict.py
```diff
@@ -70,12 +71,13 @@ def test_getattr_dict() -> None:
 
 @pytest.mark.parametrize(
     "key",
-    ["a", 1],
+    ["a", 1, 123.45, True, False, Enum1.FOO],
```
omry, line 74: I think if you remove the trailing comma black will fit it in one line.

tests/test_basic_ops_dict.py
omry: I will probably be simpler to test the dicts with the special enum key in a different class/function instead of making everything more complicated to support them in this test.

omry:
Looking good.
Added line comments.
In addition, the reference_type here is very verbose:
```
OmegaConf.create()[object()]
...
KeyValidationError: Incompatible key type 'object'
        full_key:
        reference_type=Optional[Dict[Union[str, int, Enum, float, bool], Any]]
        object_type=dict
```

What do you think about making it shorter?
```
OmegaConf.create()[object()]

For the default case, I think we should go with one of:
...
Optional[Dict]
Optional[Dict[KT, VT]]
```
Thoughts?

omry (changes requested):

omegaconf/base.py
```diff
@@ -138,7 +138,7 @@ def _format_and_raise(
         assert False
 
     @abstractmethod
-    def _get_full_key(self, key: Union[str, Enum, int, None]) -> str:
+    def _get_full_key(self, key: Union[str, Enum, int, float, bool, None]) -> str:
```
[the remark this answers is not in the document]
Jasha10: I was hesitant to make that change.
The type signature for `BaseContainer._get_full_key` is a little bit different than the above signature for `Node._get_full_key`:
```
    def _get_full_key(
        self, key: Union[str, Enum, int, float, bool, slice, None]
    ) -> str:
```

I'm not super familiar with the internals, but the presence of `slice` in the above signature makes me wonder if `_get_full_key` might be called by some `ListConfig`-related methods too, not just by `DictConfig`...

omegaconf/basecontainer.py
```diff
@@ -49,7 +49,7 @@ def _normalize_ref_type(self) -> None:
 
     def _resolve_with_default(
         self,
-        key: Union[str, int, Enum],
+        key: Union[str, int, Enum, float, bool],
```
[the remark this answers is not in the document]
Jasha10: `_resolve_with_default` gets called by `ListConfig.__getitem__`, which makes me think that something like `Union[DictKeyType, ListIndexType]` would be more correct...

omegaconf/dictconfig.py
```diff
@@ -287,29 +288,28 @@ def _validate_and_normalize_key(self, key: Any) -> DictKeyType:
...
+            return bool(key)
```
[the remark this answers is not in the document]
Jasha10: That works.

omegaconf/basecontainer.py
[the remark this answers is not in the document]
Jasha10: Should I add an alias `ListIndexType = int` to `base.py`?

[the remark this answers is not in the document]
Jasha10: Alternatively, we could just use Union[DictKeyType, int] without making an alias.
For example, we currently have
```
    @staticmethod
    def _item_eq(
        c1: Container,
        k1: Union[DictKeyType, int],
        c2: Container,
        k2: Union[DictKeyType, int],
    ) -> bool:
```
in `basecontainer.py`.

omegaconf/base.py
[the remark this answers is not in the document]
omry: Yup, get_full_key is called to obtain the full key of a node for error reporting purposes.
It's also called for ListConfigs.

slice seems like a mistake though, does the function actually ever receives a slice?
If it does, you can make the signature `Optional[DictKeyType, slice]`.

omegaconf/basecontainer.py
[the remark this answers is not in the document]
omry: int is in DictKeyType now, right?

This doesn't work?
```
    @staticmethod
    def _item_eq(
        c1: Container,
        k1: DictKeyType,
        c2: Container,
        k2: DictKeyType,
    ) -> bool:
```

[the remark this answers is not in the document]
Jasha10: Yes, `int` is in `DictKeyType`. That does work.
The only point was that some uses cases for `_item_eq` or `_resolve_with_default` will be driven by `ListConfig` (not just by `DictConfig`), so one might worry that using the name "DictKeyType" might be misleading because it mentions only "Dict" and not "List".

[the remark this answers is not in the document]
Jasha10: Maybe it would be appropriate to rename `DictKeyType -> BaseKeyType`, or simply `DictKeyType -> KeyType`?

tests/test_basic_ops_dict.py
Jasha10: > Is pytest passing other parameters from the function this fixture is used in automatically?!

Not sure what you mean? When `key_coerced` is used in the `test__setitem__typed` function, the value of `key` that was used to create the fixture value of `key_coerced` is the same as the value of `key` that is passed in to the `test__setitem__typed` function.

Jasha10: > I will probably be simpler to test the dicts with the special enum key in a different class/function instead of making everything more complicated to support them in this test.

Ok, I'll refactor.

tests/test_basic_ops_dict.py
Jasha10: Oops.

Jasha10:
> Looking good.
> Added line comments.
> In addition, the reference_type here is very verbose:
> 
> ```
> OmegaConf.create()[object()]
> ...
> KeyValidationError: Incompatible key type 'object'
>         full_key:
>         reference_type=Optional[Dict[Union[str, int, Enum, float, bool], Any]]
>         object_type=dict
> ```
> 
> What do you think about making it shorter?
> 
> ```
> OmegaConf.create()[object()]
> 
> For the default case, I think we should go with one of:
> ...
> Optional[Dict]
> Optional[Dict[KT, VT]]
> ```
> 
> Thoughts?

Sounds good. As far as picking between `Optional[Dict]` and `Optional[Dict[KT, VT]]`... I'm not sure if I'm qualified because I don't completely understand what `reference_type` is used for.

tests/test_basic_ops_dict.py
omry: > > Is pytest passing other parameters from the function this fixture is used in automatically?!
> 
> Not sure what you mean? When `key_coerced` is used in the `test__setitem__typed` function, the value of `key` that was used to create the fixture value of `key_coerced` is the same as the value of `key` that is passed in to the `test__setitem__typed` function.

So key is coming from the class parametrize, but where is cfg_key_type coming from?

omegaconf/basecontainer.py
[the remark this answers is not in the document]
omry: _item_eq and _resolve_with_default are internal functions. I wouldn't worry too much about them (we can even make them take Any if it simplifies things).

> Maybe it would be appropriate to rename DictKeyType -> BaseKeyType, or simply DictKeyType -> KeyType?

Not sure about it, there is a big difference in the key type between DictConfig and ListConfig.
We wouldn't claim that ListConfig can operate on the full spectrum of types that DictConfig can.

I can see why this function makes you uncomfortable. How about we just unroll the union and use explicit types in cases where we need to support both List and Dict key types?

omry:
> > Thoughts?
> 
> Sounds good. As far as picking between `Optional[Dict]` and `Optional[Dict[KT, VT]]`... I'm not sure if I'm qualified because I don't completely understand what `reference_type` is used for.

reference_type is the equivalent of a field annotation:
```
x : Foo = Bar()
```

Above, Foo is the reference_type and Bar is the object_type.
In the context of error messages, this is just informative.

Jasha10:
I prefer `Optional[Dict]` to `Optional[Dict[KT, VT]]` because it's simpler.

tests/test_basic_ops_dict.py
Jasha10: > So key is coming from the class parametrize, but where is cfg_key_type coming from?

The class `TestDunderMethodsAgainstStandardPythonDict` has two `@parametrize` decorators:
```
@pytest.mark.parametrize(
    "cfg_key_type,data",
    [
        (str, {"a": 10}),
        (int, {1: "a"}),
        (float, {123.45: "a"}),
        (bool, {True: "a"}),
        (Enum1, {Enum1.FOO: "foo"}),
    ],
)
@pytest.mark.parametrize(
    "key",
    ["a", "b", 1, 2, 123.45, 67.89, True, False, Enum1.FOO, Enum1.BAR, "FOO", "BAR"],
)
class TestDunderMethodsAgainstStandardPythonDict:
    ....
```

This means that every possible combination of `cfg_key_type` and `key` is used:
```
  (cfg_key_type, key) == (str, "a"),
  (cfg_key_type, key) == (str, "b"),
  (cfg_key_type, key) == (str, 1),
  (cfg_key_type, key) == (str, 1),
  ...
  (cfg_key_type, key) == (int, "a"),
  (cfg_key_type, key) == (int, "b"),
  (cfg_key_type, key) == (int, 1),
  (cfg_key_type, key) == (int 1),
  ...
  (cfg_key_type, key) == (Enum1, "a"),
  (cfg_key_type, key) == (Enum1, "b"),
  (cfg_key_type, key) == (Enum1, 1),
  (cfg_key_type, key) == (Enum1 1),
  ...
```

omegaconf/basecontainer.py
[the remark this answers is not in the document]
Jasha10: Ok, sounds good :)

omegaconf/dictconfig.py
```diff
@@ -287,29 +288,28 @@ def _validate_and_normalize_key(self, key: Any) -> DictKeyType:
...
                 )
 
-            return key
+            return cast(Union[str, int, bool, float], key)
         elif issubclass(key_type, Enum):
             try:
                 ret = EnumNode.validate_and_convert_to_enum(key_type, key)
-                assert ret is not None
+                if ret is None:
+                    raise ValidationError
```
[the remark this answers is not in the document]
Jasha10: Done, in https://github.com/omry/omegaconf/pull/484/commits/d01d02aec1fffc78b228c4d91b1cf9c7ab34eaa8

omegaconf/base.py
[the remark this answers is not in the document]
Jasha10: Made a change to `base.py:Node._get_full_key` in https://github.com/omry/omegaconf/pull/484/commits/89f29e5f450c2fff3b7b9a3e13c09f18bdaae4af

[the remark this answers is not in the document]
Jasha10: > slice seems like a mistake though, does the function actually ever receives a slice?
> If it does, you can make the signature `Optional[DictKeyType, slice]`.

There are only three test cases where `_get_full_key` receives a slice:
In `test_errors.py`, we have 
```
    pytest.param(
        Expected(
            create=lambda: OmegaConf.create(["???"]),
            op=lambda cfg: cfg.__getitem__(slice(0, 1)),
            exception_type=MissingMandatoryValue,
            msg="Missing mandatory value: [0:1]",
            key=slice(0, 1),
            full_key="[0:1]",
            child_node=lambda cfg: cfg._get_node(slice(0, 1)),
        ),
        id="list:subscript_slice_with_missing",
    ),
```
and in `test_get_full_key.py` we have
```
@pytest.mark.parametrize(
    "cfg, select, key, expected",
    [
        # ... many other test cases ...
        ([1, 2, 3], "", slice(0, 1), "[0:1]"),
        ([1, 2, 3], "", slice(0, 1, 2), "[0:1:2]"),
    ],
)
def test_get_full_key_from_config(
    cfg: Any, select: str, key: Any, expected: Any
) -> None:
    c = OmegaConf.create(cfg)
    _root, _last_key, node = c._select_impl(
        select, throw_on_missing=True, throw_on_resolution_failure=True
    )
    assert node is not None
    assert node._get_full_key(key) == expected
```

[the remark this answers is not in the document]
Jasha10: For now I've made a change to `basecontainer.py:BaseContainer._get_full_key_type` in https://github.com/omry/omegaconf/pull/484/commits/deea49e7ca63eeff773474e07890419073b4a896.

omegaconf/basecontainer.py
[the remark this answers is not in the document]
Jasha10: Resolved in https://github.com/omry/omegaconf/pull/484/commits/cac50b21e6394cbacdf94400406b683fe47ec727

omegaconf/dictconfig.py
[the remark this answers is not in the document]
Jasha10: Done in https://github.com/omry/omegaconf/pull/484/commits/89bf8d0fc9a36fe24bdf9c153d486e61f1cceaed

omegaconf/nodes.py
```diff
@@ -117,7 +117,7 @@ def _is_missing(self) -> bool:
     def _is_interpolation(self) -> bool:
         return _is_interpolation(self._value())
 
-    def _get_full_key(self, key: Union[str, Enum, int, None]) -> str:
+    def _get_full_key(self, key: Union[str, Enum, int, float, bool, None]) -> str:
```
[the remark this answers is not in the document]
Jasha10: Done in https://github.com/omry/omegaconf/pull/484/commits/5dbd9a65197aba4b216470b94d3aa7a5a2b8d015 with unrolled union

tests/structured_conf/test_structured_config.py
```diff
@@ -708,6 +708,25 @@ def test_any(name: str) -> None:
             3: "three",
         }
 
+        # test float_keys
```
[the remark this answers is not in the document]
Jasha10: Done in https://github.com/omry/omegaconf/pull/484/commits/212308a78ebcc3327d3e620fc569696963cc9041

tests/test_basic_ops_dict.py
Jasha10: Fixed in https://github.com/omry/omegaconf/pull/484/commits/85801f4e48f2401e4bfa4d090b4f37fbef38d887

tests/test_basic_ops_dict.py
Jasha10: Done, in https://github.com/omry/omegaconf/pull/484/commits/49590a98dda3da265aaac7109a2440abf6f49c8f

omry:
#505 significantly changes ref_type behavior.
Please check it out and adjust accordingly. Let me know if you have any questions.
By the way: Please join the Hydra and OmegaConf [chat](https://hydra-framework.zulipchat.com/).

omry:
@Jasha10, can you rebase on master and make sure everything still works? looks like there are merge conflicts.

Jasha10:
Yes, I'll rebase and update the branch soon.

Jasha10:
@omry ok I've rebased onto master. This branch should be ready for review.

> In addition, the reference_type here is very verbose:
> 
> ```
> OmegaConf.create()[object()]
> ...
> KeyValidationError: Incompatible key type 'object'
>         full_key:
>         reference_type=Optional[Dict[Union[str, int, Enum, float, bool], Any]]
>         object_type=dict
> ```
> 
> What do you think about making it shorter?

Given the recent PR to change `ref_type`, I assume we no longer need to work on this :)

omry:
> Given the recent PR to change ref_type, I assume we no longer need to work on this :)

Yup!

tests/test_errors.py
```diff
@@ -455,6 +475,38 @@ def finalize(self, cfg: Any) -> None:
...
+        Expected(
+            create=lambda: DictConfig({}, key_type=float),
+            op=lambda cfg: cfg.get("foo"),
+            exception_type=KeyValidationError,
+            msg=dedent(
+                """\
+                Key foo (str) is incompatible with (float)
+                    full_key: foo
+                    object_type=dict"""
+            ),
```
omry, line 488: compare the message only. other components of the exceptions can be compared directly.
```suggestion
            msg="Key foo (str) is incompatible with (float)"
```

tests/test_errors.py
```diff
@@ -455,6 +475,38 @@ def finalize(self, cfg: Any) -> None:
...
+        Expected(
+            create=lambda: DictConfig({}, key_type=bool),
+            op=lambda cfg: cfg.get("foo"),
+            exception_type=KeyValidationError,
+            msg=dedent(
+                """\
+                Key foo (str) is incompatible with (bool)
+                    full_key: foo
+                    object_type=dict"""
+            ),
```
omry, line 504: here too.

tests/test_compare_dictconfig_vs_dict.py
```diff
@@ -0,0 +1,211 @@
+"""
+This file compares dunder methods of DictConfig (e.g. __getitem__ and
+__setitem__) with dunder methods of standard python's dict.
```
omry, line 3: comment is outdated.

tests/test_compare_dictconfig_vs_dict.py
```diff
@@ -0,0 +1,211 @@
...
+)
+def key(request: Any) -> Any:
+    """A key to test indexing into DictConfig."""
+    return request.param
+
+
+@pytest.fixture
+def python_dict(data: Dict[Any, Any]) -> Dict[Any, Any]:
+    """Just a standard python dictionary, to be used in comparison with DictConfig."""
+    return data.copy()
```
omry, line 44: deepcopy is safer.

tests/test_compare_dictconfig_vs_dict.py
```diff
@@ -0,0 +1,211 @@
+"""
```
omry, line 1: What do you think about [this style](https://github.com/omry/omegaconf/commit/d53960c30faa420015457d3730a881b8a92aca2f#diff-8ce1623c935afb10aeeecdcb7f6306610f5e6039448c224b1e10faeab33c4267R63-R65) for comparing?

Seems a bit cleaner to me:

```python
@pytest.mark.parametrize("cfg", [{}, OmegaConf.create({}, flags={"struct": True})])
def test_struct_dict_get(cfg: Any) -> None:
    assert cfg.get("z") is None
```

tests/test_compare_dictconfig_vs_dict.py
```diff
@@ -0,0 +1,211 @@
...
+        False,
+        Enum1.FOO,
+        Enum1.BAR,
+        "FOO",
+        "BAR",
+    ]
+)
+def key(request: Any) -> Any:
+    """A key to test indexing into DictConfig."""
+    return request.param
```
omry, line 38: Not familiar with this style of fixtures. What does it do?

omry (changes requested):

tests/test_compare_dictconfig_vs_dict.py
Jasha10: It is very similar to `@pytest.parametrize`.
When a test function uses the `key` fixture, the test will be parametrized so that each value in the `params` list gets used as a `key`.

omry: I see.
So this is an alternative form to :
```python
@pytest.mark.parameterize("a", [...])
class TestFoo:
  def test_bar(a):
    ...

  def test_rab(a):
    ...
```
?

Jasha10: Yes, exactly.
One possible benefit is that you can use one fixture as input to another fixture, so you could for example write
```python
@pytest.fixture(params=["key_a","key_b"])
def key(request: Any) -> str:
    return request.param

@pytest.fixture(params=["lock1", "lock2"])
def key_in_lock(key, request: Any) -> str:
    lock = request.param
    return lock + "(" + key + ")"

def test_key_in_lock(key_in_lock):
    print(key_in_lock)
```
which outputs
```
lock1(key_a)
lock1(key_b)
lock2(key_a)
lock2(key_b)
```

tests/test_compare_dictconfig_vs_dict.py
Jasha10: Made the change in https://github.com/omry/omegaconf/pull/484/commits/79893d4305910a0312c7b3e99846ef4e4c1a7fdf

omegaconf/dictconfig.py
```diff
@@ -270,35 +270,36 @@ def _validate_and_normalize_key(self, key: Any) -> DictKeyType:
...
                 raise KeyValidationError(
                     f"Key '$KEY' is incompatible with the enum type '{key_type.__name__}', valid: [{valid}]"
                 )
+            else:
+                if ret is None:
+                    valid = ", ".join([x for x in key_type.__members__.keys()])
+                    raise KeyValidationError(
+                        f"Key '$KEY' is incompatible with the enum type '{key_type.__name__}', valid: [{valid}], "
+                    )
+                return ret
```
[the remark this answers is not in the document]
Jasha10: Made the change in https://github.com/omry/omegaconf/pull/484/commits/6d4a1b439e39b9bd078513233e27a05399345f72

tests/test_basic_ops_dict.py
```diff
@@ -67,14 +67,12 @@ def test_getattr_dict() -> None:
...
-@pytest.mark.parametrize(
-    "key",
-    ["a", 1],
-)
+@pytest.mark.parametrize("key", ["a", 1, 123.45, True, False, Enum1.FOO])
 class TestDictKeyTypes:
     def test_mandatory_value(self, key: DictKeyType) -> None:
         c = OmegaConf.create({key: "???"})
-        with pytest.raises(MissingMandatoryValue, match=str(key)):
+        match = key.name if hasattr(key, "name") else str(key)  # type: ignore
```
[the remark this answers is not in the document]
Jasha10: Done in https://github.com/omry/omegaconf/pull/484/commits/4d9710f4e412987c3d8aa47108acf8b1c21721cf

tests/structured_conf/test_structured_config.py
```diff
@@ -682,6 +691,10 @@ def test_any(name: str) -> None:
             "e": True,
         }
 
+    def test_dict_examples_colors(self, class_type: str) -> None:
```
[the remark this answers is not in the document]
Jasha10: Refactored in https://github.com/omry/omegaconf/pull/484/commits/fa0c6627cff09912337bf71dcc75cef2786a3320

tests/test_basic_ops_dict.py
```diff
@@ -111,6 +109,9 @@ def test_subscript_set(self, key: DictKeyType) -> None:
     [
         ({"a": 10, "b": 11}, "a", {"b": 11}),
         ({1: "a", 2: "b"}, 1, {2: "b"}),
+        ({123.45: "a", 67.89: "b"}, 67.89, {123.45: "a"}),
+        ({True: "a", False: "b"}, False, {True: "a"}),
+        ({Enum1.FOO: "foo", Enum1.BAR: "bar"}, Enum1.FOO, {Enum1.BAR: "bar"}),
```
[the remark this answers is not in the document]
Jasha10: Ok, fixed in https://github.com/omry/omegaconf/pull/484/commits/8fa3a498401438846774cad3efd1dc4d4af87ce2 and https://github.com/omry/omegaconf/pull/484/commits/fcecb2769ffeb710ee94453eb7a650c834ad5f29

tests/test_errors.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/484/commits/d9efe4478dbef91c69d59f2721b4e758babd3238

tests/test_compare_dictconfig_vs_dict.py
Jasha10: It is?

tests/test_compare_dictconfig_vs_dict.py
Jasha10: Are you saying to use `DictConfig.get` instead of `DictConfig.__getitem__`?

Jasha10: Ooh, or do you mean creating the `dict` object and the `DictConfig` object next to eachother in the `parametrize` decorator?

Jasha10: Hmm... One issue with that is: some of the tests might modify the `parametrize` data, so other tests using the same data could be affected.

I think that using a fixture to instantiate or deepcopy the data helps to make sure that the test cases are isolated from eachother.

tests/test_compare_dictconfig_vs_dict.py
omry: I started to experiment with this in the structured_ir PR.

tests/test_compare_dictconfig_vs_dict.py
omry: > Hmm... One issue with that is: some of the tests might modify the parametrize data, so other tests using the same data could be affected.

Yeah, either deepcopy the data in the test or have the fixture provide a new copy every time.

tests/test_compare_dictconfig_vs_dict.py
omry: It's testing more than dunder methods now, isn't it?

omegaconf/dictconfig.py
```diff
@@ -270,35 +270,33 @@ def _validate_and_normalize_key(self, key: Any) -> DictKeyType:
...
 
-            return key
+            return key  # type: ignore
         elif issubclass(key_type, Enum):
             try:
                 ret = EnumNode.validate_and_convert_to_enum(key_type, key)
-                assert ret is not None
-                return ret
+                if ret is None:
+                    raise ValidationError
```
omry, line 292: Okay, raising the Validation exception only to catch it a moment later still this still looks a bit weird.

how about:
```python
ret = EnumNode.validate_and_convert_to_enum(key_type, key, alow_none=False)
```

Which will trigger the validation error in validate_and_convert_to_enum?
(Or alternatively, just change it to reject None and make sure other users are not calling it with None).

omegaconf/dictconfig.py
```diff
@@ -270,35 +270,33 @@ def _validate_and_normalize_key(self, key: Any) -> DictKeyType:
...
             except ValidationError:
                 valid = ", ".join([x for x in key_type.__members__.keys()])
                 raise KeyValidationError(
                     f"Key '$KEY' is incompatible with the enum type '{key_type.__name__}', valid: [{valid}]"
                 )
+            else:
+                return ret
```
omry, line 299: I don't like this.
can you just put the return in the try?

tests/test_errors.py
```diff
@@ -552,10 +588,10 @@ def finalize(self, cfg: Any) -> None:
     pytest.param(
         Expected(
             create=lambda: OmegaConf.structured(SubscriptedDict),
-            op=lambda cfg: cfg.__setitem__("dict", 1),
+            op=lambda cfg: cfg.__setitem__("dict_str", 1),
```
omry, line 591: why change that?

tests/test_compare_dictconfig_vs_dict.py
Jasha10: Nope. Just `__getitem__`, `__setitem__`, `__eq__` and `__contains__`.

I just realized when typing this that it might be good to add some tests for `__delitem__` too...

omegaconf/dictconfig.py
Jasha10: Sure. Done in https://github.com/omry/omegaconf/pull/484/commits/2fec127369cb14249758de30f1d5f1b0764108ba

tests/test_errors.py
Jasha10: Previously in the `tests/__init__.py` file we had:
```python
@dataclass
class SubscriptedDict:
    dict: Dict[str, int] = field(default_factory=lambda: {"foo": 4})
```

I changed this `SubscriptedDict` class to look like this:
```python
@dataclass
class SubscriptedDict:
    dict_str: Dict[str, int] = field(default_factory=lambda: {"foo": 4})
    dict_enum: Dict[Color, int] = field(default_factory=lambda: {Color.RED: 4})
    dict_int: Dict[int, int] = field(default_factory=lambda: {123: 4})
    dict_float: Dict[float, int] = field(default_factory=lambda: {123.45: 4})
    dict_bool: Dict[bool, int] = field(default_factory=lambda: {True: 4, False: 5})
```
So the `dict` attribute got renamed to `dict_str`.

tests/test_compare_dictconfig_vs_dict.py
Jasha10: :) Nice!

omegaconf/dictconfig.py
Jasha10: It seems that rejecting None is not an option: If I add `assert value is not None` to the `validate_and_convert_to_enum` method, lots of errors come up in the tests.

Jasha10: I've added an `allow_none: bool = True` keyword argument to the `validate_and_convert_to_enum` method (https://github.com/omry/omegaconf/pull/484/commits/adeadd52223f4df488c458b151fee8a602bcccec).

tests/test_compare_dictconfig_vs_dict.py
omry: and maybe pop and get() :).

How are you accounting for the different behavior of __getitem__ if an item is not found?
This is one of my early mistakes, it returns None in that case.
At some point I would like to transition toward a standard behavior. Maybe after 2.1.

Jasha10: This is the `__getitem__` pattern that I'm currently using in the `test_compare_dictconfig_vs_dict.py` file:
```python
        try:
            result = python_dict[key]
        except KeyError:
            assert cfg[key] is None
        else:
            assert result == cfg[key]
```

omry: Okay, once my #516 lands simplify this.
Can you also test see how things are behaving with and without struct mode? should probably be tested for all functions here).

Jasha10: > and maybe pop and get() :).

Ok, will do!

> Okay, once my #516 lands simplify this.
Can you also test see how things are behaving with and without struct mode? should probably be tested for all functions here).

Yup.

Jasha10: > Okay, once my #516 lands simplify this.
Unfortunately the no simpler now than it was before:
```python
        try:
            result = python_dict[key]
        except KeyError:
            with pytest.raises(ConfigKeyError):
                cfg[key]
        else:
            assert result == cfg[key]
```

Jasha10: We now have tests for all of the following:
```
    __contains__
    __delitem__
    __eq__
    __getitem__
    __setitem__
    get
    pop
    keys
    values
    items
```

Jasha10: I've added conditional branches that deal with different behavior for structured vs non-structured `DictConfig`.

tests/test_omegaconf.py
```diff
@@ -91,7 +91,7 @@ def test_is_missing(
 ) -> None:
     cfg = OmegaConf.create(cfg)
     with expectation:
-        cfg.get(key)
+        cfg[key]
```
Jasha10, line 94: In commit https://github.com/omry/omegaconf/pull/484/commits/0cec9053ab3d630cea6816fbd1b7d6d8278e2416 I made a change to `DictConfig.get`, so that its behavior is consistent with standard `dict.get`:
by default, `cfg.get(key)` should return `None` if `key` is not found.

This change required the above modification to the `test_omegaconf.py` file, using `__getitem__` for testing instead of `get`.

tests/test_compare_dictconfig_vs_dict.py
```diff
@@ -0,0 +1,544 @@
...
+    @pytest.mark.parametrize("use_default", [True, False])
+    def test_pop(
+        self,
+        python_dict: Any,
+        cfg: DictConfig,
+        key: Any,
+        use_default: bool,
+        struct_mode: bool,
+    ) -> None:
+        """Ensure that __getitem__ has same result with python dict as with DictConfig."""
```
omry, line 152: Try to keep tests small.
I think two tests here are better:
`test_pop` and `test_pop_with_default`.

tests/test_compare_dictconfig_vs_dict.py
```diff
@@ -0,0 +1,544 @@
...
+        if struct_mode:
+            with pytest.raises(ConfigTypeError):
+                cfg.pop(key)
```
omry, line 155: If there are behavior changes, it's best to use a different test method, e.g. `test_pop_dict_config_struct_mode`.

tests/test_compare_dictconfig_vs_dict.py
```diff
@@ -0,0 +1,544 @@
...
+            except KeyError:
+                with pytest.raises(ConfigKeyError):
+                    del cfg[key]
+            else:
+                del cfg[key]
+                assert key not in cfg
+
+    def test__contains__(self, python_dict: Any, cfg: Any, key: Any) -> None:
+        """Ensure that __contains__ has same result with python dict as with DictConfig."""
+        assert (key in python_dict) == (key in cfg)
```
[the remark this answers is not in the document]
omry: This test file has a very large surface area. It could be a good idea to followup with a second cleanup pass in a different PR.

tests/test_errors.py
```diff
@@ -170,6 +169,28 @@ def finalize(self, cfg: Any) -> None:
         ),
         id="dict:pop_invalid_nested",
     ),
+    pytest.param(
+        Expected(
+            create=lambda: OmegaConf.create({"foo": "bar"}),
+            op=lambda cfg: cfg.__delitem__("nevermind"),
```
omry, line 175:
```suggestion
            create=lambda: OmegaConf.create({"foo": "bar"}),
            op=lambda cfg: cfg.__delitem__("not_found"),
```

tests/test_omegaconf.py
omry: Gotcha.
This makes sense and is exactly why I wanted to have more disciplined testing for consistency with plain dict.

Under what circumstances the behavior is changing?
should we include a news fragment for it?

omegaconf/dictconfig.py
```diff
@@ -401,12 +400,20 @@ def __delitem__(self, key: DictKeyType) -> None:
...
-        del self.__dict__["_content"][key]
+        try:
+            del self.__dict__["_content"][key]
+        except KeyError:
+            full = self._get_full_key(key=key)
+            if full != key:
+                msg = f"Key not found: '{key}' (path: '{full}')"
+            else:
+                msg = f"Key not found: '{key}'"
+            self._format_and_raise(key=key, value=None, cause=ConfigKeyError(msg))
```
omry, line 411: The full key will be contained in the exception anyway.
Also, you can retype the exception with tyep_override (I think):
```suggestion
        except KeyError as e:
            msg = f"Key not found: '{key}'"
            self._format_and_raise(key=key, value=None, cause=e, type_override=ConfigKeyError)
```

omry (changes requested):

tests/test_omegaconf.py
Jasha10: The behavior is changed when `DictConfig.get` is called without a default value and `key` is not found. Previously a `ConfigKeyError` was raised; now `None` is returned.

I'll add a news fragment.

omegaconf/dictconfig.py
Jasha10: I was able to simplify to
```python
        try:
            del self.__dict__["_content"][key]
        except KeyError:
            msg = "Key not found: '$KEY' (path: '$FULL_KEY')"
            self._format_and_raise(key=key, value=None, cause=ConfigKeyError(msg))
```

I did not have success with setting `cause=e` and `type_override=ConfigKeyError` because `Node._format_and_raise` sets `msg=str(cause)` which in this case would be contain the correct string template (since the original `cause` is of type `KeyError`, not type `ConfigKeyError`).

tests/test_omegaconf.py
Jasha10: I also created an Issue documenting the inconsistency: #527

tests/test_compare_dictconfig_vs_dict.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/484/commits/54ae46390c4967be09e6ed1da51afb2ff71bf83b and https://github.com/omry/omegaconf/pull/484/commits/19e31de31ec6ae88e978d06914bd282b897b1cda

tests/test_compare_dictconfig_vs_dict.py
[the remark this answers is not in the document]
Jasha10: > I think this approach scales better if we can pull it off.

I think that pulling this off will be a good exercise for me to learn more about `pytest`.

omegaconf/dictconfig.py
omry: Okay. I still think you can drop the (path: ..) component. it's going to be in the exception anyway. (try to print the resulting exception).

tests/test_compare_dictconfig_vs_dict.py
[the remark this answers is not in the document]
omry: I spent 15 minutes now to see if I can get something decent.
this is what I came up with:
```python
from typing import MutableMapping, Any, List
from pytest import raises, mark

from omegaconf import OmegaConf, DictConfig


def struct(proto: Any) -> DictConfig:
    cfg = OmegaConf.create(proto)
    OmegaConf.set_struct(cfg, True)
    return cfg


def dictconfig(proto: Any) -> DictConfig:
    return OmegaConf.create(proto)


def dicts(proto: Any, factories: List[Any]) -> Any:
    return [f(proto) for f in factories]


class TestDict:
    @mark.parametrize("cfg", dicts({}, [dict, OmegaConf.create]))
    def test_pop_empty(self, cfg: MutableMapping) -> None:
        with raises(KeyError):
            cfg.pop("not_found")

    @mark.parametrize("cfg", dicts({"foo": 10}, [dict, dictconfig]))
    def test_pop(self, cfg: MutableMapping):
        assert cfg.pop("foo") == 10

    @mark.parametrize("cfg", dicts({"foo": 10}, [dict, dictconfig]))
    def test_pop_with_default(self, cfg: MutableMapping):
        assert cfg.pop("foo", 20) == 10

    @mark.parametrize("cfg", dicts({"foo": 10}, [dict, dictconfig]))
    def test_pop_with_default2(self, cfg: MutableMapping):
        assert cfg.pop("not_found", 10) == 10
        assert cfg == {"foo": 10}

    @mark.parametrize("cfg", dicts({"foo": 10}, [dict, dictconfig, struct]))
    def test_getitem(self, cfg: MutableMapping):
        assert cfg["foo"] == 10
```

omry:
> I ended up simplifying the logic and removing this catch clause in [af8a78a](https://github.com/omry/omegaconf/commit/af8a78a25f500c20a13828112275412c64618ba0) by changing the signature from
> 
> ```
> default_value = DEFAULT_VALUE_MARKER
> ```
> 
> to
> 
> ```
> default_value = None
> ```

Nice.
I like it when things suddenly snaps into palace :)

omegaconf/dictconfig.py
Jasha10: Done, in https://github.com/omry/omegaconf/pull/484/commits/6211d9b2899a7197fbc1131b9c81a2ae3ddffb28.

docs/source/usage.rst
```diff
@@ -231,6 +231,12 @@ You can provide default values directly in the accessing code:
     >>> conf.get('missing_key', 'a default value')
     'a default value'
 
+By default, `conf.get` returns `None` if the key is not found:
+.. doctest:: loaded
+
+    >>> conf.get('missing_key') is None
+    True
+
```
omry, line 239: I think since this is the behavior from dict we shouldn't document it.

omry (changes requested):
Can you pull out the changes to get() to a standalone PR?
We can merge it quickly and then you will be able to rebase this one on it.

tests/test_compare_dictconfig_vs_dict.py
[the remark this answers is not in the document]
Jasha10: Hmm...

I'm not sure if it will be so easy to convert the existing test file into this format.
In the current setup, a large number of test cases are automatically generated in `test_compare_dictconfig_vs_dict.py`.
For example, many of the tests in the `TestUntypedDictConfig` class are run once with each possible combination of the following inputs:
```
for key in ["a", "b", 1, 2, 123.45, 67.89, True, False, Enum1.FOO, Enum1.BAR, "FOO", "BAR"]:
    for data in [{"a": 10}, {1: "a"}, {123.45: "a"}, {True: "a"}, {Enum1.FOO: "foo"}]:
        for struct_mode in [True, False]:
            ...
```

Thus, most of the methods in the `TestUntypedDictConfig` class gets called 120 times:
12 keys x 5 data x 2 struct_mode. (A few of the methods get called only 10 times each because they don't depend on the `key` fixture). So in total, pytest runs 1110 tests for the `TestUntypedDictConfig` class. And then there are the other classes, giving a total of 2220 tests in the whole `test_compare_dictconfig_vs_dict.py` file. This number will increase once more of the tests have been split to handle separate cases in atomic tests.

My feeling is that this large amount of tests is made possible by making the tests depend on both the `DictConfig` object and on the standard python dict at the same time. This way, the programmer does not have to write down what is expected in every test case. Instead, the python dict is used to figure out what is expected, and the programmer just has to define how the particular (key, python_dict) combination should be used to figure out what is expected of the `DictConfig` object.

docs/source/usage.rst
Jasha10: Ok

Jasha10:
> Can you pull out the changes to get() to a standalone PR?
> We can merge it quickly and then you will be able to rebase this one on it.

I can do that, but the tests that led me to find the `DictConfig.get` vs `dict.get` inconsistency are in the `test_compare_dictconfig_vs_dict.py` file. If we introduce the change to `DictConfig.get` in a separate PR, the new behavior will not be heavily tested until we merge the `dictconfig_vs_dict` tests.

omegaconf/dictconfig.py
```diff
@@ -401,14 +400,18 @@ def __delitem__(self, key: DictKeyType) -> None:
...
-        del self.__dict__["_content"][key]
-
-    def get(self, key: DictKeyType, default_value: Any = DEFAULT_VALUE_MARKER) -> Any:
+        try:
+            del self.__dict__["_content"][key]
+        except KeyError:
+            msg = "Key not found: '$KEY'"
+            self._format_and_raise(key=key, value=None, cause=ConfigKeyError(msg))
+
+    def get(self, key: DictKeyType, default_value: Any = None) -> Any:
```
omry, line 409: can you rebase on the other diff and remove the get changes from this one?

omry (changes requested):

omry:
1. rebase on other diff and cleanup get changes
2. update the docs a bit (see my comment)

tests/test_compare_dictconfig_vs_dict.py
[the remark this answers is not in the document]
Jasha10: Fair enough :P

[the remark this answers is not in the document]
Jasha10: > Tests that are smart are easier to write but harder to maintain and understand.
> We are only writing them once, but we will be maintaining them forever.
> They also provide better documentation to the expected behavior.

Good points.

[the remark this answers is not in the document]
Jasha10: > Also, not all tests behaves the same with and without struct mode (which can be expressed in my version by opting out of testing struct, and adding an explicit test to cover it).

I found a cool way to opt out of specific cases using the "smart" tests: you write a general `struct_mode` fixture, then override it for specific tests:
```python3
from pytest import fixture, mark

data_dict_examples = [{"a": 123}]

@fixture(params=[True, False])
def struct_mode(request) -> bool:
    struct_mode: bool = request.param
    return struct_mode

@fixture(params=data_dict_examples)
def dictconfig(request, struct_mode) -> bool:
    cfg = OmegaConf.create(request.param)
    if struct_mode:
        OmegaConf.set_struct(cfg, True)
    return cfg

...  # here are lots of tests that use dictconfig with and without struct_mode

# Here we override struct_mode for specific tests:

@mark.parametrize("struct_mode", [False])
def test_for_specific_non_struct_behavior(dictconfig):
    ...  # Here dictconfig is constructed with struct_mode=False only

@mark.parametrize("struct_mode", [True])
def test_for_specific_non_struct_behavior(dictconfig):
    ...  # Here dictconfig is constructed with struct_mode=True only
```
Anyway, if we're going to switch to a simpler testing strategy then this isn't totally relevant. I just wanted to share the pytest trick :)

omegaconf/dictconfig.py
Jasha10: Done.

docs/source/usage.rst
```diff
@@ -41,17 +41,19 @@ From a dictionary
 
 .. doctest::
 
-    >>> conf = OmegaConf.create({"k" : "v", "list" : [1, {"a": "1", "b": "2", 3: "c"}]})
+    >>> conf = OmegaConf.create({"k" : "v", "list" : [1, {"a": "1", "b": "2", 3: "c", True: "bool key", 1.2: "float key"}]})
```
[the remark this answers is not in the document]
Jasha10: Filed issue #531 regarding `OmegaConf.to_yaml(enum_key_type)`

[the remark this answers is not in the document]
Jasha10: Updated the docs: https://github.com/omry/omegaconf/pull/484/commits/a914486e8dbfb757bff9c5fac238b70683f7eb5c

I'm having trouble getting the doctest to pass for some reason...

Jasha10:
I can't figure out why the doctest is failing. It's complaining about an indentation error...
```
running tests...

Document: usage
---------------

Warning, treated as error:
**********************************************************************
File "usage.rst", line 57, in default
Failed example:
    class Color(Enum):
Exception raised:
      File "/usr/local/lib/python3.9/doctest.py", line 1336, in __run
        exec(compile(example.source, filename, "single",
      File "/home/circleci/project/.nox/docs/lib/python3.9/site-packages/sphinx/ext/doctest.py", line 487, in compile
        return compile(code, name, self.type, flags, dont_inherit)
      File "<doctest default[1]>", line 1
        class Color(Enum):
                          ^
    IndentationError: expected an indented block
```

omry:
```diff
$ git show
commit 02394230d01474d324e0d64eba98abe60d8e4710 (HEAD -> 7iW-more-dict-key-types)
Author: Omry Yadan <omry@fb.com>
Date:   26 seconds ago

    fixed docs

diff --git a/docs/source/usage.rst b/docs/source/usage.rst
index bf27619..4f4c49c 100644
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -57,15 +57,15 @@ Here is an example of various supported key types:
 
     >>> from enum import Enum
     >>> class Color(Enum):
-    >>>     RED = 1
-    >>>     BLUE = 2
+    ...   RED = 1
+    ...   BLUE = 2
     >>> 
     >>> conf = OmegaConf.create(
-    >>>     {"key": "str", 123: "int", True: "bool", 3.14: "float", Color.RED: "Color"}
-    >>> )
+    ...   {"key": "str", 123: "int", True: "bool", 3.14: "float", Color.RED: "Color"}
+    ... )
     >>> 
     >>> print(conf)
-    {'key': 'str', 666: 'int', True: 'bool', 3.14: 'float', <Color.RED: 1>: 'Color'}
+    {'key': 'str', 123: 'int', True: 'bool', 3.14: 'float', <Color.RED: 1>: 'Color'}
 
 OmegaConf supports `str`, `int`, `bool`, `float` and Enums as dictionary key types.
```

omry:
Okay, looks good to me.
I want to do it after we merge #445 which has been in the works for months and is touching many files touched by this as well.
It's close to being merged.

In the mean time, @odelalleau  - can you also make a review pass on this one?

Jasha10:
Rebased onto master.

odelalleau (commented):
Well, that's a lot of tests -- I admit I haven't looked at all of them in details (in particular the large file comparing to Python dicts).
Overall looks quite good to me, just a few questions.

news/483.feature
```diff
@@ -0,0 +1 @@
+Add DictConfig support for keys of type float and bool
```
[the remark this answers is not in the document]
omry: There is no common practice :)
when we finally release 2.1, I will take a look and see if it makes sense to merge them manually. for now let's leave it as it.

omegaconf/dictconfig.py
```diff
@@ -275,28 +275,26 @@ def _validate_and_normalize_key(self, key: Any) -> DictKeyType:
...
-                )
-
-            return key
-        elif key_type == int:
-            if not isinstance(key, int):
+        elif key_type == bool and (key == 0 or key == 1):
+            # Python treats True as 1 and False as 0 when used as dict keys
+            #   assert hash(0) == hash(False)
+            #   assert hash(1) == hash(True)
+            return key  # type: ignore
```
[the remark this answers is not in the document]
Jasha10: Yes, that makes sense. Done in https://github.com/omry/omegaconf/pull/484/commits/2834064440855ab17affa5a76678deb59d2b5baf.

omegaconf/basecontainer.py
```diff
@@ -713,6 +713,8 @@ def prepand(full_key: str, parent_type: Any, cur_type: Any, key: Any) -> str:
                 key = _slice_to_str(key)
             elif isinstance(key, Enum):
                 key = key.name
+            elif isinstance(key, (int, float, bool, type(None))):
```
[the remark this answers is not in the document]
Jasha10: Yes, you're right.
Changing to `isinstance(key, (int, float, bool))`.

news/483.feature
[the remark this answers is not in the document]
odelalleau: Ok, resolving then :)

odelalleau (approved):
Good as far as I'm concerned (I haven't looked at other unresolved discussions though)

omegaconf/dictconfig.py
```diff
@@ -377,6 +375,7 @@ def __getitem__(self, key: DictKeyType) -> Any:
             self._format_and_raise(key=key, value=None, cause=e)
 
     def __delitem__(self, key: DictKeyType) -> None:
+        key = self._validate_and_normalize_key(key)
```
omry, line 378: This looks like a bugfix. It would be better to include it in a standalone PR but I will not insist here.

1. Do you have a test covering this?
2. Add a news fragment?

Jasha10: >     1. Do you have a test covering this?
There is some test coverage in the new `test_compare_dictconfig_vs_dict.py` file.

tests/test_compare_dictconfig_vs_dict.py
```diff
@@ -109,10 +109,7 @@ def test__delitem__(
         self, python_dict: Any, cfg: DictConfig, key: Any, struct_mode: bool
     ) -> None:
         """Ensure that __delitem__ has same result with python dict as with DictConfig."""
-        if struct_mode:
-            with raises(ConfigTypeError):
-                del cfg[key]
-        else:
+        if not struct_mode:
```
omry, line 112: If this test should not be parameterized by struct_mode why are you passing it?

tests/test_compare_dictconfig_vs_dict.py
```diff
@@ -123,6 +120,14 @@ def test__delitem__(
...
                 assert key not in cfg
 
+    def test__delitem__struct_mode(
+        self, python_dict: Any, cfg: DictConfig, key: Any, struct_mode: bool
+    ) -> None:
+        """Ensure that __delitem__ fails in struct_mode"""
+        if struct_mode:
+            with raises(ConfigTypeError):
+                del cfg[key]
+
```
omry, line 130: if this test only operates when struct_mode is True, why even pass it?
Why not this?
```suggestion
    def test__delitem__struct_mode(
        self, python_dict: Any, cfg: DictConfig, key: Any
    ) -> None:
        """Ensure that __delitem__ fails in struct_mode"""
        OmegaConf.set_struct(cfg, True)    
        with raises(ConfigTypeError):
            del cfg[key]

```

Jasha10: The reason is that the `cfg` fixture is parametrized by `struct_mode`; by passing in `cfg` in the test function signature, the test gets called once with `struct_mode == True` and once with `struct_mode == False`.

I guess a clearer setup could be to have `cfg_struct_mode` and `cfg_no_struct_mode` fixtures in addition to the `cfg` fixture.

omry: or cfg and cfg_struct:
```
cfg - just a default DictConfig
cfg_struct:
  cfg with struct False
  cfg with struct True
```

omegaconf/dictconfig.py
Jasha10: Specifically the test coverage is in the `test__delitem__enum_typed` and `test__delitem__enum_typed_struct_mode` methods.

Jasha10: I created an issue to track the bug: #554

Jasha10: And here is the news fragment: https://github.com/omry/omegaconf/pull/484/commits/91e13a94d1166da2501becb14c79df2b93b2b287

tests/test_compare_dictconfig_vs_dict.py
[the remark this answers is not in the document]
omry: Very nice.

By the way:
flags has 3 states:
None, True and False.

tests/test_compare_dictconfig_vs_dict.py
```diff
@@ -0,0 +1,678 @@
...
+        1,
+        2,
+        123.45,
+        67.89,
+        True,
+        False,
+        Enum1.FOO,
+        Enum1.BAR,
+        "FOO",
+        "BAR",
```
omry, line 45: I suggest one key per supported key type.
```suggestion
        "str",
        1,
        3.1415,
        True,
        Enum1.FOO,
```

tests/test_compare_dictconfig_vs_dict.py
```diff
@@ -0,0 +1,647 @@
...
+        """Ensure that __delitem__ has same result with python dict as with DictConfig."""
+        try:
+            del python_dict[key]
+            assert key not in python_dict
+        except KeyError:
+            with raises(ConfigKeyError):
+                del cfg[key]
+        else:
+            del cfg[key]
+            assert key not in cfg
```
Jasha10, line 111: Tests for non-struct behaviors are marked with `@mark.parametrize("struct_mode", [False, None])`, which will override the default values for `struct_mode in [True, False, None]`.

tests/test_compare_dictconfig_vs_dict.py
```diff
@@ -0,0 +1,647 @@
...
+
+    @mark.parametrize("struct_mode", [True])
+    def test__delitem__struct_mode(
+        self, python_dict: Any, cfg: DictConfig, key: Any
+    ) -> None:
+        """Ensure that __delitem__ fails in struct_mode"""
+        with raises(ConfigTypeError):
+            del cfg[key]
```
Jasha10, line 119: Tests that should only pass for `struct_mode==True` are marked with `@mark.parametrize("struct_mode", [True])`.

Jasha10 (commented):
In the most recent commit I've switched to a new way of parametrizing `struct_mode` in the `test_compare_dictconfig_vs_dict.py` file. @omry I'm not sure if you will like it. If you don't I'll switch it to something simpler, like using separate fixtures for `cfg` and `cfg_struct`.

Explanation below:

tests/structured_conf/data/dataclasses.py
```diff
@@ -403,10 +403,10 @@ class DictExamples:
         }
     )
     int_keys: Dict[int, str] = field(default_factory=lambda: {1: "one", 2: "two"})
-
-
-@dataclass
-class DictWithEnumKeys:
+    float_keys: Dict[float, str] = field(
```
omry, line 406: DictExamples becomes very large.
Maybe break it into `DictValueTypes` and `DictKeyTypes`?

tests/test_compare_dictconfig_vs_dict.py
```diff
@@ -0,0 +1,647 @@
...
+@fixture
+def python_dict(data: Dict[Any, Any]) -> Dict[Any, Any]:
+    """Just a standard python dictionary, to be used in comparison with DictConfig."""
+    return deepcopy(data)
+
+
+@fixture(params=[True, False, None])
+def struct_mode(request: Any) -> Optional[bool]:
+    struct_mode: Optional[bool] = request.param
+    return struct_mode
```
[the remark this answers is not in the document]
omry: I am not sure we need to test None because at least for Struct Mode, it supposed to behave like False.
Did it catch anything interesting?

In terms of order: I would go with `[None, False, True]`.

tests/test_compare_dictconfig_vs_dict.py
```diff
@@ -0,0 +1,647 @@
...
+
+
+@mark.parametrize(
+    "data",
+    [
+        param({"a": 10}, id="str"),
+        param({1: "a"}, id="int"),
+        param({123.45: "a"}, id="float"),
+        param({True: "a"}, id="bool"),
+        param({Enum1.FOO: "foo"}, id="Enum1"),
```
omry, line 65: Tests here seems to be focusing on the type of the key only.
What is the purpose of this test? why is the type of the key so critical?

tests/test_compare_dictconfig_vs_dict.py
omry: I would differentiate the case for a key that is found and a key that is not found into two tests.
This probably means that it's easier to not use the key fixture at all here.
```python
@mark.parameterize(("cfg", "key"), [
  ({"key": 1}, "key"),
  ({True: 1}, True),
  ...
])
def test__delitem__found(cfg, key):
  del cfg[key]
  assert key not in cfg

@mark.parameterize(("cfg", "key"), [
  ({"key": 1}, "key"),
  ({True: 1}, True),
  ...
])
def test__delitem__not_found(cfg):
  with raises(KeyError):
    del cfg[key]
```

I don't think we need to worry about testing that the exception type is ConfigKeyError here, we just care about similarity to plain dict.

This can be a part of the followup reffactor of this class.

tests/test_compare_dictconfig_vs_dict.py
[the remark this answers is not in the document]
Jasha10: > I am not sure we need to test None because at least for Struct Mode, it supposed to behave like False.
> Did it catch anything interesting?

No, nothing interesting.

I did notice that DictConfig.pop and DictConfig.__delitem__ have some special logic for if `struct_mode is False`: if `cfg._is_typed()` is `True`, i.e. if `object_type` is set, then `struct_mode` must be `False` for `__delitem__` or `pop` to work. If `struct_mode` is `None`, then `__delitem__` and `pop` raise `ConfigTypeError`. This use case does not come up in the tests from this file.

> In terms of order: I would go with `[None, False, True]`.

Updated to `[None, False, True]` in https://github.com/omry/omegaconf/pull/484/commits/baa918c022969dd8960e9a41f7aadee10b12a60e

PR #490, opened 2021-01-23
Fixed to not serialize flags cache when pickling

omry:
No point in serializing the flags cache, it can just be re-initialized lazily later on access.
Serializing it is also wasteful and breaking protocol compatibility.


omry:
cc @samuelstanton, @jieru-hu

omry:
Released 2.10dev17 with this change.

PR #493, opened 2021-01-24
Improve docs regarding interpolations in structured config.  Fixes #488

### LINKED ISSUE #488: Overriding configuration using an environment doesn't trigger type checks when merging into Structured config
**Describe the bug**

If I create a structured configuration with an integer field and then merge in a configuration where that fields value is taken from an environment variable, the type checks usually associated with the structured field are not triggered.

**To Reproduce**

```
def test_strange_overriding_behaviour():
    @dataclass
    class Base:
        foo: int = omegaconf.MISSING

    override = textwrap.dedent("""
        foo: ${env:FOO_OVERRIDE}
    """)

    with MonkeyPatch.context() as mp:
        mp.setenv("FOO_OVERRIDE", "This is not an integer")
        base_conf = omegaconf.OmegaConf.structured(Base)
        override_conf = omegaconf.OmegaConf.create(override)

        # This does not raise ValidationError but merges successfully, and merged_conf.foo is set to the string
        # "This is not an integer"
        with pytest.raises(omegaconf.ValidationError):
            merged_conf = omegaconf.OmegaConf.merge(base_conf, override_conf)

```

**Expected behavior**

I expected a validation error to be raised on merge, as it would have been had I set:
`override = "foo: "This is not an integer"`

**Additional context**
- [x] OmegaConf version: 2.0.5
- [x] Python version: 3.8.5
- [x] Operating system : Ubuntu 16.04.7 LTS

SimonStJG:


docs/source/structured_config.rst
```diff
@@ -332,6 +332,21 @@ To work around it, use SI and II described below.
     >>> assert conf.b == 100
     >>> assert conf.c == 100
 
+Runtime type validation is not performed on interpolated variables, see below.
```
omry, line 335: I think this is a bit better (type validation IS performed on assignment).
---
Runtime type validation is not performed on values returned by interpolation, e.g:

```python
@dataclass
class Interpolation:
    int_key: int = II("str_key")
    str_key: str = "string"

cfg = OmegaConf.structured(Interpolation)
assert cfg.int_key == "string"
```

SimonStJG: Sure, updated :)

omry (approved):
lgtm! can followup by cherry picking it to the 2.0 branch?

SimonStJG:
Yup, here it is https://github.com/omry/omegaconf/pull/500

PR #495, opened 2021-01-25
Fix flags cache inconsistency when serializing a container

### LINKED ISSUE #494: The flags cache may become inconsistent after serializing a container config
**Describe the bug**

This test currently fails:

```python
def test_pickle_flags_consistency() -> None:
    cfg = DictConfig({"a": 0})
    cfg._set_flag("test", True)
    a_node = cfg._get_node("a")
    assert a_node is not None
    assert a_node._get_flag("test")

    cfg2 = pickle.loads(pickle.dumps(cfg))
    cfg2._set_flag("test", None)
    assert cfg2._get_flag("test") is None
    a2_node = cfg2._get_node("a")
    assert a2_node._get_flag("test") is None  # <-- FAIL!
```

This is because the flags cache of `cfg` is not preserved during serialization, while the flags cache of `cfg.a` is. As a result, when `cfg2`'s cache is invalidated (when setting its "test" flag to `None`), this is not propagated to its children, and the last line re-uses the (now invalid) cache.

**Potential solutions**

I can see at least two options to fix this issue:
1. Serialize the flags cache as an empty dict if it is not None (instead of ignoring it and setting it to `None` when de-serializing)
2. Also wiping the flags cache of non-container nodes during serialization

I'm leaning towards option 2, but I'm going to benchmark both options with the current benchmark and submit a PR with the fix

odelalleau:
Fixes #494

### DIFF
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -290,3 +290,14 @@ def test_pickle_none() -> None:
         fp.seek(0)
         cfg2 = pickle.load(fp)
         assert cfg == cfg2
+
+
+def test_pickle_flags_consistency() -> None:
+    cfg = DictConfig({"a": 0})
+    cfg._set_flag("test", True)
+    assert cfg._get_node("a")._get_flag("test")  # type: ignore
+
+    cfg2 = pickle.loads(pickle.dumps(cfg))
+    cfg2._set_flag("test", None)
+    assert cfg2._get_flag("test") is None
+    assert cfg2._get_node("a")._get_flag("test") is None
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -293,7 +293,7 @@ def test_pickle_none() -> None:
 
 
 def test_pickle_flags_consistency() -> None:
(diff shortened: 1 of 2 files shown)


tests/test_serialization.py
```diff
@@ -290,3 +290,14 @@ def test_pickle_none() -> None:
         fp.seek(0)
         cfg2 = pickle.load(fp)
         assert cfg == cfg2
+
+
+def test_pickle_flags_consistency() -> None:
+    cfg = DictConfig({"a": 0})
```
omry, line 296: Prefer OmegaConf.create to direct DictConfig usage.
```suggestion
    cfg = OmegaConf.create({"a": 0})
```

omry (changes requested):

omry:
Nice catch, how did you notice?

odelalleau:
> Nice catch, how did you notice?

Hunting down a tricky bug while playing around with a simpler deepcopy mechanism (next PR). It turns out that the default deepcopy implementation relies on __getstate__ and __setstate__ (this is when I ran into it)

PR #502, opened 2021-01-30
Instantiate structured configs

### LINKED ISSUE #472: Convert Structured Config back to dataclass instance

`OmegaConf.structured()` returns a `DictConfig`, but it is recommended to duck type that as the original structured config type to get a static type checker to catch errors. But this seems like a leaky abstraction:

```python
import attr
from typing import List

@attr.s(auto_attribs=True)
class Widget:
   foo: str
   bar: int

@attr.s(auto_attribs=True)
class Quux:
   widget_list: List[Widget]

In [19]:
qq: Quux = OmegaConf.structured(Quux(widget_list=[Widget(foo='a', bar=1)]))

In [27]:
type(qq.widget_list[0])
Out [27]:
omegaconf.dictconfig.DictConfig

def do_something(conf: Quux):
   print(conf.widget_list[0].foo)
```

For example, in the above `do_something` function, we are annotating `conf: Quux`, and this is ok _as long_ as we don't do anything expecting `conf.widget_list[0]` to be an actual `Widget` instance. If I wanted to convert a `conf.widget_list` to JSON, I would do something different for a "real" `Quux` instance and for a duck-typed instance.

My naive understanding is that this is so you can add runtime validation for merging etc. 

I think it would be nice to have a way to re-instantiate the original type after merging etc: 

```python
qq = OmegaConf.instantiate(
   OmegaConf.merge(OmegaConf.structured(Quux), OmegaConf.load(...))
)
assert type(qq) is Quux
```
The benefit I'm imagining is that code that is annotated with `Quux` would not have any "leaks".  Code that is concerned with manipulating configs can continue to use DictConfigs and their runtime features, but code that is using configs can work with the original types.

What stops us from doing so?

- Another nice-to-have would be for `OmegaConf.structured` to return a `DictConfig[Quux]` -- ie. a generic so that this can be annotated clearly.

- I think that if I were implementing this library from scratch, I would start with `attrs`/structured configs only, and then add the runtime validation via `attrs`'s hooks like `on_setattr` etc. - we would never have to enter `DictConfig`-land. What are the problems with this approach (setting side the fact that dictconfigs are widely used already)?

**Additional context**

Jasha10:
Closes #472 

The idea is to add an `instantiate_structured_configs` keyword argument to the `OmegaConf.to_container` method.
If `OmegaConf.to_container(cfg, instantiate_structured_configs=True)` is called with this new flag, the returned python object will include instantiated `dataclass` or `attrs` instances.


omry:
Looking at the code, I believe this does not handle nested structured configs right now.
Also, this means to_container() can actually return an instance now, which is a bit weird:
1. The same is not great in such a case.
2. I did not see see the expected type changes demonstrating you are accounting to this at a glance.

We should probably introduce an alias to to_container(), maybe to_object() which will call to_container but will allow Any return type.

Jasha10:
> Looking at the code, I believe this does not handle nested structured configs right now.
I am having success with nested structured configs:
```
@dataclass
class Inner:
    cost: float

@dataclass
class Outer:
    i: Inner

cfg = OmegaConf.create(Outer(Inner(1.23)))
print(OmegaConf.to_container(cfg, instantiate_structured_configs=True))
# prints Outer(i=Inner(cost=1.23))
```
It is also working for e.g. structured config as a dict value type.

> Also, this means to_container() can actually return an instance now, which is a bit weird:
> 
>     1. The same is not great in such a case.
> 
>     2. I did not see see the expected type changes demonstrating you are accounting to this at a glance.
I see what you mean. I'm surprised that mypy did not catch that.

> We should probably introduce an alias to to_container(), maybe to_object() which will call to_container but will allow Any return type.
Ok, sounds good. So `to_container` will keep the old behavior (return python primitives) and `to_object` will instantiate structured configs.

tests/structured_conf/test_structured_config.py
```diff
@@ -864,6 +864,85 @@ def test_create_untyped_list(self, class_type: str) -> None:
...
+            assert user.age is MISSING
+
+            user = self.round_trip_to_container(module.User("Bond", 7))
+            assert isinstance(user, module.User)
+            assert type(user) is module.User
+            assert user.name == "Bond"
+            assert user.age == 7
+
+        def test_nested(self, module: Any) -> None:
+            data = self.round_trip_to_container({1: module.User()})
```
omry, line 894: I know you like using integers as dict key but it's hardly the most common pattern.
test with a string as the dict key.
```suggestion
            data = self.round_trip_to_container({"user": module.User()})
```

tests/structured_conf/test_structured_config.py
```diff
@@ -864,6 +864,85 @@ def test_create_untyped_list(self, class_type: str) -> None:
...
+            assert isinstance(user, module.User)
+            assert type(user) is module.User
+            assert user.name is MISSING
+            assert user.age is MISSING
+
+            user = self.round_trip_to_container(module.User("Bond", 7))
+            assert isinstance(user, module.User)
+            assert type(user) is module.User
+            assert user.name == "Bond"
+            assert user.age == 7
```
omry, line 891: One use case per test please. (you are testing User() and User(...) here.

tests/structured_conf/test_structured_config.py
omry: This test is testing an object nested in a dict.
Add a test for an object nested in an object.

omry (commented):
> I am having success with nested structured configs:

Please add actual tests.
Can you explain why this works?

I also noticed that you are using get_structured_config_data() which is not designed for this (it's used to convert from an sc to a dictconfig) and is actually a rather weird function.
Can you explain what you are using it for exactly, it's not clear to me looking at the code what value it provides you. (it's returning a dictionary where values can be DictConfigs).

> Ok, sounds good. So to_container will keep the old behavior (return python primitives) and to_object will instantiate structured configs?

Seems reasonable.
to_container() can still have the new parameter instantiate_structured_configs. (I think a better name for it is instantiate). Doc string should explain it in more detail.

Jasha10:
Rebased onto master

tests/structured_conf/test_structured_config.py
Jasha10: Resolved in https://github.com/omry/omegaconf/pull/502/commits/eca1dc0f410ea5aa31d1444e1c5274a3b1ab373c

Jasha10:
> > I am having success with nested structured configs:
> 
> Please add actual tests.

I've added tests for nested structured config (https://github.com/omry/omegaconf/pull/502/commits/f02261eaf6b395728eaaecc20fe9280713aa899c).

Jasha10:
> I also noticed that you are using get_structured_config_data() which is not designed for this (it's used to convert from an sc to a dictconfig) and is actually a rather weird function.

The `get_structured_config_data` function is only used in the case where a dataclass subclasses a Dict:
```
@dataclass
class Str2StrWithField(Dict[str, str]):
    foo: str = "bar"
```
When the `_instantiate_structured_config_impl` function instantiates a `Str2StrWithField` object, it first does
```
result = Str2StrWithField(foo=...)
```
to create the object, then it does
```
result.update(remaining keys that are not "foo")
```

The `get_structured_config_data` function is used to figure out which keys should be passed as keyword arguments in the first step, and which keys should be passed to `result.`update` in the second step.

> Can you explain what you are using it for exactly, it's not clear to me looking at the code what value it provides you. (it's returning a dictionary where values can be DictConfigs).

Hmm...

omegaconf/basecontainer.py
```diff
@@ -193,10 +193,16 @@ def _to_content(
...
+        instantiate: bool = False,
     ) -> Union[None, Any, str, Dict[DictKeyType, Any], List[Any]]:
         from .dictconfig import DictConfig
         from .listconfig import ListConfig
 
+        if exclude_structured_configs and instantiate:
+            raise ValueError(
+                "Cannot both exclude and and instantiate structured configs"
+            )
+
```
omry, line 205: exclude_structured_configs is new in 2.1 and we can still change it.
Looks like an enum can replace exclude_structured_configs and instantiate:
```python
class SCMode(Enum):
  DICT = 1            # convert to plain dict
  DICT_CONFIG = 2     # Keep as OmegaConf DictConfig
  INSTANTIATE = 3     # instantiate underlying object.
```

omegaconf/basecontainer.py
```diff
@@ -798,3 +815,27 @@ def _update_types(node: Node, ref_type: type, object_type: Optional[type]) -> No
         if new_ref_type is not Any:
             node._metadata.ref_type = new_ref_type
             node._metadata.optional = new_is_optional
+
+
+def _instantiate_structured_config_impl(
+    retdict: Dict[str, Any], object_type: Type[Any]
```
omry, line 821: retdict is not a good argument name here.
Also, just pass conf as is. no need to break the object_type out, it's inside.

omegaconf/basecontainer.py
```diff
@@ -798,3 +815,27 @@ def _update_types(node: Node, ref_type: type, object_type: Optional[type]) -> No
...
+            k: v for k, v in retdict.items() if k in object_type_field_names
+        }
+        retdict_nonfield_items = {
+            k: v for k, v in retdict.items() if k not in object_type_field_names
+        }
+        result = object_type(**retdict_field_items)
+        result.update(retdict_nonfield_items)
+    else:
+        # normal structured config
+        assert set(retdict.keys()) <= set(object_type_field_names)
```
omry, line 839: Does this handle the case where the user add an additional field to the Structurec Config that is not in the underlying type?

omegaconf/basecontainer.py
```diff
@@ -798,3 +815,27 @@ def _update_types(node: Node, ref_type: type, object_type: Optional[type]) -> No
...
             node._metadata.ref_type = new_ref_type
             node._metadata.optional = new_is_optional
+
+
+def _instantiate_structured_config_impl(
+    retdict: Dict[str, Any], object_type: Type[Any]
+) -> Any:
+    from ._utils import get_structured_config_data
+
+    object_type_field_names = get_structured_config_data(object_type).keys()
```
omry, line 825: flip the order of the two cases. in terms of importance, the regular object case is the important one.

omry, line 825: get_structured_config_data is recursive and expensive.
looks like all you need is the keys. can you create a few helper functions that will just give you the keys?

something like:
```python
get_structured_config_field_names()
get_dataclass_field_names()
get_attr_class_field_names()
```

tests/structured_conf/test_structured_config.py
```diff
@@ -864,6 +864,135 @@ def test_create_untyped_list(self, class_type: str) -> None:
         assert cfg.list == [1, 2]
         assert cfg.opt_list is None
 
+    class TestInstantiateStructuredConfigs:
```
omry, line 867: Those tests should probably go to test_to_container()

tests/structured_conf/test_structured_config.py
```diff
@@ -864,6 +864,135 @@ def test_create_untyped_list(self, class_type: str) -> None:
...
+            assert type(user) is module.User
+            assert user.name == "Bond"
+            assert user.age == 7
+
+        def test_basic_with_missing(self, module: Any) -> None:
+            user = self.round_trip_to_object(module.User())
+            assert isinstance(user, module.User)
+            assert type(user) is module.User
+            assert user.name == MISSING
+            assert user.age == MISSING
```
omry, line 890: I think instantiation should fail if there are MISSING fields that should be used to instantiate objects.
Is this the motivation for the `to_container throw_if_missing` PR?

tests/structured_conf/test_structured_config.py
```diff
@@ -864,6 +864,135 @@ def test_create_untyped_list(self, class_type: str) -> None:
...
+
+        def test_nested(self, module: Any) -> None:
```
omry, line 892: When I said that I don't thin you are handling nesting, you said that you were.
looking at this test, the nesting you were thinking about is a Structured Config nested in a dict/list.
I was thinking about a Structured Config nested in another Structured Config.

e.g: a Plugin inside a PluginHolder.

tests/test_to_container.py
```diff
@@ -63,6 +63,9 @@ def test_exclude_structured_configs(cfg: Any, ex_false: Any, ex_true: Any) -> No
     ret1 = OmegaConf.to_container(cfg, exclude_structured_configs=False)
     assert ret1 == ex_false
 
+    ret1 = OmegaConf.to_container(cfg, instantiate=True)
+    assert ret1 == ex_true
+
```
omry, line 68: avoid overloading tests.

You can refactor this to take:
cfg, exclude_structured_configs, expected_result

Also, OmegaConf is overriding __eq__ and return True when a structured config is compared to a matching underlying object.
Your test may be passing because of that. Should probably not use `==` here.
(You can write a more precise eq function for the purpose of this test that is also validating the field type).

omegaconf/basecontainer.py
Jasha10: It does not handle that case. What is the desired behavior when there are extra fields: Do we want to call `setattr` on the dataclass instance?

omegaconf/basecontainer.py
Jasha10: Looks good.

tests/structured_conf/test_structured_config.py
Jasha10: This case is handled by the `test_nested_object` test defined further down in the file:
```python3
        def test_nested_object(self, module: Any) -> None:
            nested = self.round_trip_to_object(module.NestedConfig)
            assert isinstance(nested, module.NestedConfig)
            assert type(nested) is module.NestedConfig

            assert nested.default_value == MISSING

            assert isinstance(nested.user_provided_default, module.Nested)
            assert type(nested.user_provided_default) is module.Nested
            assert nested.user_provided_default.with_default == 42
```

tests/structured_conf/test_structured_config.py
Jasha10: > Is this the motivation for the `to_container throw_if_missing` PR?

Yes, exactly.

omegaconf/basecontainer.py
omry: What is the purpose of the assert?


> What is the desired behavior when there are extra fields: Do we want to call setattr on the dataclass instance?

```python
In [1]: from dataclasses import *
   ...: from typing import *
   ...: from omegaconf import OmegaConf
   ...: @dataclass
   ...: class Foo:
   ...:     x: int
   ...: foo = Foo(x=10)

In [2]: foo
Out[2]: Foo(x=10)

In [3]: setattr(foo, "y", 20)

In [4]: foo
Out[4]: Foo(x=10)

In [5]: foo.y
Out[5]: 20
```

I actually had no idea this was possible.
Yes, I think it would be nice to use setattr here.
Other alternatives are to error out, or  to outright ignore the extra fields. But I think setattr is matching the OmegaConf behavior very well so we can try it.

tests/structured_conf/test_structured_config.py
omry: Okay, cool.

omegaconf/omegaconf.py
```diff
@@ -600,6 +604,34 @@ def to_container(
...
             enum_to_str=enum_to_str,
             exclude_structured_configs=exclude_structured_configs,
+            instantiate=instantiate,
+        )
+
+    @staticmethod
+    def to_object(
+        cfg: Any,
+        *,
+        resolve: bool = False,
```
omry, line 614: resolve should probably default to True here.

odelalleau:
@omry I noticed the review request, but since I'm just discovering this PR now, I'd rather wait until your current comments have been addressed by @Jasha10

Jasha10:
@odelalleau I'll ping you when I've finished addressing the comments.

tests/structured_conf/test_structured_config.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/841ac01b73197e22f8195ece8c76649eb305d237

omegaconf/basecontainer.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/fa81a4d4bc35342ebd2a76942ce68653d1f41078

tests/test_utils.py
```diff
@@ -192,6 +192,24 @@ def test_get_structured_config_data(test_cls_or_obj: Any, expectation: Any) -> N
...
+        (_TestAttrsClass(), does_not_raise()),
+        ("invalid", raises(ValueError)),
+    ],
+)
+def test_get_structured_config_field_names(
+    test_cls_or_obj: Any, expectation: Any
+) -> None:
+    with expectation:
+        field_names = _utils.get_structured_config_field_names(test_cls_or_obj)
+        assert field_names == ["x", "s", "b", "f", "e", "list1", "dict1"]
```
omry, line 210: separate to two tests: one positive and one for error conditions.

omegaconf/omegaconf.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/872850b96ada2890a5ce45248e23c50eb23a72d5.

omegaconf/basecontainer.py
Jasha10: Ok, sounds good!

omegaconf/basecontainer.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/072e8d861ce6a8df13fdbe0d67979be3874068cd.

tests/test_utils.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/1953a47fdfb5b65725febeb3fec213e9641f2cc1

omegaconf/basecontainer.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/a4af7f5466cb74cbc2cde39e6f6e447a1496a992

omegaconf/basecontainer.py
Jasha10: > What is the purpose of the assert?

Sorry that I missed this.

If object_type is `MyDataclass`, we want to make sure that calling `MyDataclass(k1=v1, k2=v2, ...)` will not raise a TypeError.

Thus, before calling `MyDataclass(**instance_data)`, we assert that `instance_data.keys()` is a subset of the keywords arguments that `MyDataclass.__init__` is able to accept.

omegaconf/basecontainer.py
```diff
@@ -769,3 +778,30 @@ def _update_types(node: Node, ref_type: type, object_type: Optional[type]) -> No
...
+
+def _instantiate_structured_config_impl(
+    conf: "DictConfig", instance_data: Dict[str, Any]
+) -> Any:
+    """Instantiate an instance of `conf._metadata.object_type`, populated by `instance_data`."""
+    from ._utils import get_structured_config_field_names
+
+    object_type = conf._metadata.object_type
+
+    object_type_field_names = get_structured_config_field_names(object_type)
```
odelalleau, line 791: Can you make it a set here? This will speed up the `in` checks below in case of a large dataclass.

omegaconf/_utils.py
```diff
@@ -192,6 +192,12 @@ def _resolve_forward(type_: Type[Any], module: str) -> Type[Any]:
         return type_
 
 
+def get_attr_class_field_names(obj: Any) -> List[str]:
+    is_type = isinstance(obj, type)
+    obj_type = obj if is_type else type(obj)
+    return [name for name in attr.fields_dict(obj_type).keys()]
```
odelalleau, line 198: Minor simplification:
```suggestion
    return list(attr.fields_dict(obj_type))
```

omegaconf/omegaconf.py
```diff
@@ -620,6 +623,32 @@ def to_container(
...
+        *,
+        resolve: bool = True,
+        enum_to_str: bool = False,
+    ) -> Union[Dict[DictKeyType, Any], List[Any], None, str, Any]:
+        """
+        Resursively converts an OmegaConf config to a primitive container (dict or list).
+        Any DictConfig objects backed by dataclasses or attrs classes are instantiated
+        as instances of those backing classes.
+
+        This is an alias for OmegaConf.to_container(..., resolve=True, structured_config_mode=SCMode.INSTANTIATE)
```
odelalleau, line 638: Remove `resolve=True`?

tests/structured_conf/test_structured_config.py
```diff
@@ -1010,6 +1010,26 @@ def test_str2user(self, module: Any) -> None:
...
+        assert cfg.mp.age == 11
+        assert isinstance(cfg.mp, DictConfig)
+
+        with raises(ValidationError):
+            # bad value
+            cfg.hello = "world"
+
+        with raises(KeyValidationError):
+            # bad key
+            cfg[Color.BLUE] = "nope"
```
odelalleau, line 1031: Can you assign a valid user here instead of a string? (ex: cfg.mp)

tests/test_to_container.py
```diff
@@ -39,19 +41,21 @@ def assert_container_with_primitives(item: Any) -> None:
 
 
 @mark.parametrize(
-    "src,ex_dict,ex_dict_config,key",
+    "src,ex_dict,ex_dict_config,ex_instantiate,key",
```
odelalleau, line 44: Why add `ex_instantiate`? It seems to always be equal to `ex_dict_config`, it's a bit confusing

odelalleau (changes requested):
Looks good, just a few small things

tests/test_to_container.py
Jasha10: Hmm.. Maybe I can just use `src` instead of `ex_dict_config` and `ex_instantiate` since all three are equal.

omegaconf/omegaconf.py
Jasha10: Remove from the docstring?
Or do you mean remove from the function signature?

Jasha10: The motivation for resolve=True by default:
This enables the result of `to_object(cfg)` to be immediately used in the client's application logic, including e.g. calling methods belonging to the dataclasses that were defined by the client.

tests/test_to_container.py
odelalleau: Ah wait, looking at the test again, now I understand why you did it that way. It makes sense now => resolving.

omegaconf/omegaconf.py
odelalleau: I meant remove from the docstring, since my understanding is that `...` represents the arguments of `to_object()` (otherwise I would have expected `enum_to_str=False` to also be present in the docstring)

Jasha10: Oh, that makes sense. Will do.

tests/structured_conf/test_structured_config.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/7f8addbb7ca846edd672060c6d64fae494d167f6.

omegaconf/omegaconf.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/a10fa5bdca987120e33a26a7e57fcd588aa8f2e8.

omegaconf/basecontainer.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/6b014aea0ad0185722942a1315e6af22b2a9bdfe.

Jasha10: And a related commit here: https://github.com/omry/omegaconf/pull/502/commits/4cefc6d859546949f65917b98ee3f36166f2104c.

omegaconf/basecontainer.py
Jasha10: Commit https://github.com/omry/omegaconf/pull/502/commits/62f34cf190f4ae6cb439a05cabcf7b1cbac791a3 implements `setattr(instance, k, v)` for extra fields appearing in a structured config.

tests/structured_conf/test_structured_config.py
Jasha10: Instantiation now fails of there are MISSING fields.

Jasha10:
Summary of API changes made in recent commits:
- `to_object` fails with `MissingMandatoryValue` if any fields belonging to a Structured Config are MISSING
- If a DictConfig has keys that do not appear as fields of the target dataclass, those extra key/value fields are handled as follows:
  - If the dataclass is a subclass of `Dict`: the extra keys/values are added to the dict using `dataclass_instance.update(...)`.
  - If the dataclass is not a subclass of `Dict`: the extra keys/values are set as attributes using `setattr(dataclass_instance, key, value)`.

TODO: Update documentation:
- [ ] a description and example of OmegaConf.to_object
- [ ] a description and example of the new `SCMode.INSTANTIATE` Enum member

omegaconf/basecontainer.py
odelalleau: > Commit [62f34cf](https://github.com/omry/omegaconf/commit/62f34cf190f4ae6cb439a05cabcf7b1cbac791a3) implements `setattr(instance, k, v)` for extra fields appearing in a structured config.

What's the motivation? The assert made sense to me: I thought only "dict-like" dataclasses were allowed to be extended with new fields?

odelalleau (changes requested):

omegaconf/basecontainer.py
```diff
@@ -232,7 +232,18 @@ def convert(val: Node) -> Any:
                     )
                 else:
                     retdict[key] = convert(node)
-            return retdict
+
+            if structured_config_mode == SCMode.INSTANTIATE and is_structured_config(
+                conf._metadata.object_type
+            ):
+                if any(_is_missing_literal(value) for value in retdict.values()):
+                    raise MissingMandatoryValue()
```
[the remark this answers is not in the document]
Jasha10: Oh yes, definitely. I'll add a test case to `test_errors.py`.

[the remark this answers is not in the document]
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/3a471320f627a67de71a8297c41bf0f704dfd529.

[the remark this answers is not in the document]
odelalleau: Thanks, can you add the class name as well to the error message to make it easier to identify? Something like:
```Structured config `User` has missing mandatory malue: name```
(also removing the capitazliation to be consistent with other similar error messages)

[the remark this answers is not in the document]
Comment (Jasha10) [reply omegaconf/basecontainer.py]: Done in https://github.com/omry/omegaconf/pull/502/commits/c581548edb11709db46ff8982b25c12a50bca625.

**Review by odelalleau — COMMENTED** [review]
Only thing that still bugs me is https://github.com/omry/omegaconf/pull/502#discussion_r593209704

**Comment by omry** [comment]
@odelalleau , @Jasha10 : Just want to make sure you are both on the same page:
this functionality is not scheduled for 2.1, I would rather to keep the focus on thins with the 2.1 milestone.

**Comment by odelalleau** [comment]
> @odelalleau , @Jasha10 : Just want to make sure you are both on the same page:
> this functionality is not scheduled for 2.1, I would rather to keep the focus on thins with the 2.1 milestone.

Doesn't really matter to me, but as far as I'm concerned I have no objection to have it included in 2.1.

**Comment by omry** [comment]
Okay, I am adding this to 2.1 tentatively. Depending on time constraints because it's mostly new functionality and low risk to existing functionality.

File: omegaconf/basecontainer.py (hunk 5 above)
Comment (Jasha10) [reply omegaconf/basecontainer.py]: As I see it, there are both pros and cons to this `setattr` behavior.

The major cons are:
1. type safety is reduced, and
2. accidental calls to `UserDataclass.__setattr__` can have serious consequences:
  - if `UserDataclass` has any methods, those methods could be overwritten by
    our calling `setattr(instance, 'methodname', value)`
  - if `UserDataclass` implements a custom `__setattr__` method, our calling
    `setattr(instance, 'key', value)` could trigger side effects.

The main pro is:

OmegaConf and Hydra both make it easy for clients to add extra fields to
Structured Configs, using `omegaconf.open_dict` and the Hydra `+key=value` CLI
syntax, respectively. Not implementing the `setattr` behavior would mean that
users are restricted from using `open_dict`/`+key=value` in combination with
`to_object` when structured configs are involved.

I'd like to propose the following: enforce the schema strictly by default
(raising an error if there are extra fields), and allow users to opt in to the
`setattr` behavior. For example, we could add another member to the `SCMode`
Enum, say `SCMode.INSTANTIATE_SETATTR`, so that users can opt in to the
behavior by selecting this member (with `SCMode.INSTANTIATE` (no `setattr`)
being the default).

Comment (Jasha10) [reply omegaconf/basecontainer.py]: Here is some additional motivation for allowing users to opt in to `setattr`
behavior:

For Python objects in general (not just for dataclasses), the default
implementation of `__setattr__` allows for flexible attribute assignment:
```python
class A:
    pass

A.x = 123  # A.x becomes a class variable
a = A()
a.y = 456  # A.y becomes an instance variable
assert a.x == 123
assert a.y == 456
```
This is pythonic; it works because `A` is a subclass of the `object` class and
`object.__setattr__` implements this behavior. Users can take advantage of this
to achieve something similar to the `namespace` from C++.

Here is an example of how this might look with OmegaConf:
```python
@dataclass
class MyNamespace:
    required_id: str

    @dataclass
    class SubNamespace:
        pass

    sub1: SubNamespace = field(default_factory=SubNamespace)
    sub2: SubNamespace = field(default_factory=SubNamespace)

cfg = OmegaConf.structured(MyNamespace)
cfg.required_id = "bond"
with open_dict(cfg):
    cfg.foo = "bar"

with open_dict(cfg.sub1):
    cfg.sub1.baz = "qux"

data: MyNamespace = OmegaConf.to_object(cfg, structured_config_mode=SCMode.INSTANTIATE_SETATTR)

assert data.foo == "bar"
assert data.sub1.baz == "qux"
```

odelalleau: > OmegaConf and Hydra both make it easy for clients to add extra fields to Structured Configs, using `omegaconf.open_dict` and the Hydra `+key=value` CLI syntax, respectively.

Just that is enough to convince me. However, in that case I don't think we should treat differently a dataclass that's been extended through `open_dict()` vs a dataclass that is declared "extendable" by inheriting from Dict => I would suggest to make all extra fields new attributes in `to_object()`.

Jasha10: > Edit: actually, in addition to that, I think that if the dataclass inherits from Dict then the instance returned by `to_object()` should mirror _all_ attributes in its "dictionary" content. This way one can choose to use either `obj.key` or `obj["key"]`.

Interesting idea :)

I'd like to point out a drawback to this approach: if the client mutates the Dict subclass after calling `to_object`, e.g. by calling `obj.key = new_value` or `obj["key"] = new_value` or `del obj.key` or `del obj["key"]`, then the dict keys of `obj` will no longer be "in sync" with the attributes of `obj` (i.e. we might get `obj.key != obj["key"]`).

odelalleau: > I'd like to point out a potential drawback to this approach: if the client mutates the Dict subclass after calling `to_object`, e.g. by calling `obj.key = new_value` or `obj["key"] = new_value` or `del obj.key` or `del obj["key"]`, then the dict keys of `obj` will no longer be "in sync" with the attributes of `obj` (i.e. we might get `obj.key != obj["key"]`).

Yeah, I also thought it may not play too well with modifiying the object afterwards, but I think there are similar issues with other approaches as well. Even if initially attributes aren't mirrored, someone may call `obj.key = v` or `obj["key"] = v` thinking they are overriding their current value, while it may actually create a distinct entry (in the first case if `"key"` was actually in the dict and not an attribute, and vice-versa in the second case). I doubt we can fully avoid this kind of problem without messing with the dataclass internals (which would go against the spirit of this PR, which is to obtain a "normal" dataclass instance).

Jasha10: > I doubt we can fully avoid this kind of problem without messing with the dataclass internals (which would go against the spirit of this PR, which is to obtain a "normal" dataclass instance).

Agreed.

Jasha10: My opinion is that, once the data has been "exported" from OmegaConf using `to_container`/`to_object`, the responsibility for using that data in a correct way should fall on the client.

odelalleau: Just to recap, I'm fine with either of these options for dataclasses inheriting from Dict:
- put everything as attributes (nothing in the dict)
- mirror everything in the dict

But the current hybrid approach seems a bit more confusing and dangerous (e.g., someone could start from a regular structured config and extend it with `open_dict()`, then later decide to make it inherit from Dict instead for convenience, and suddenly the behavior might change unexpectedly)

Jasha10: > But the current hybrid approach seems a bit more confusing and dangerous (e.g., someone could start from a regular structured config and extend it with `open_dict()`, then later decide to make it inherit from Dict instead for convenience, and suddenly the behavior might change unexpectedly)

Hmm, I see your point...

Jasha10: > Just to recap, I'm fine with either of these options for dataclasses inheriting from Dict:
> 
>     * put everything as attributes (nothing in the dict)
> 
>     * mirror everything in the dict

@omry, what do you think?

omegaconf/basecontainer.py
```diff
@@ -232,7 +232,16 @@ def convert(val: Node) -> Any:
                     )
                 else:
                     retdict[key] = convert(node)
-            return retdict
+
+            if structured_config_mode == SCMode.INSTANTIATE and is_structured_config(
+                conf._metadata.object_type
+            ):
+                retstruct = _instantiate_structured_config_impl(
```
omry, line 239: I'd simplify by changing retdict to ret and type is as Any.

omegaconf/basecontainer.py
```diff
@@ -769,3 +778,39 @@ def _update_types(node: Node, ref_type: type, object_type: Optional[type]) -> No
         if new_ref_type is not Any:
             node._metadata.ref_type = new_ref_type
             node._metadata.optional = new_is_optional
+
+
+def _instantiate_structured_config_impl(
+    conf: "DictConfig", instance_data: Dict[str, Any]
+) -> Any:
```
omry, line 785: should this be a method on DictConfig?

omegaconf/basecontainer.py
```diff
@@ -769,3 +778,39 @@ def _update_types(node: Node, ref_type: type, object_type: Optional[type]) -> No
...
+            nonfield_items[k] = v
+
+    result = object_type(**field_items)
+    if not issubclass(object_type, dict):
+        # normal structured config
+        for k, v in nonfield_items.items():
+            setattr(result, k, v)
+    else:
+        # Extending dict as a subclass
+        result.update(nonfield_items)
```
omry, line 815: You can also force fields into non-dict subclasses with open_dict():

```python
with open_dict(cfg.user):
  cfg.user.root = True
```

dataclasses and attr classes supports adding additional fields. Maybe we should just allow additional fields in the same way?

```python
In [1]: from dataclasses import dataclass

In [2]: @dataclass
   ...: class User:
   ...:     name: str
   ...:

In [3]: user = User(name="joe")

In [4]: user.age = 10

In [5]: user
Out[5]: User(name='joe')

In [6]: user.age
Out[6]: 10

In [7]: import attr

In [8]: @attr.s(auto_attribs=True)
   ...: class User2:
   ...:     name: str
   ...:

In [9]: user2 = User2(name="Bob")

In [10]: user2.age = 11

In [11]: user2
Out[11]: User2(name='Bob')

In [12]: user2.age
Out[12]: 11
```

omegaconf/omegaconf.py
```diff
@@ -620,6 +623,32 @@ def to_container(
             structured_config_mode=structured_config_mode,
         )
 
+    @staticmethod
+    def to_object(
+        cfg: Any,
+        *,
+        resolve: bool = True,
```
omry, line 630: I think resolve here should not be a member and should be forced to True.

omegaconf/omegaconf.py
```diff
@@ -591,6 +591,9 @@ def to_container(
         :param structured_config_mode: Specify how Structured Configs (DictConfigs backed by a dataclass) are handled.
             By default (`structured_config_mode=SCMode.DICT`) structured configs are converted to plain dicts.
             If `structured_config_mode=SCMode.DICT_CONFIG`, structured config nodes will remain as DictConfig.
+            If `structured_config_mode=SCMode.INSTANTIATE`, this function will instantiate structured configs
+               (DictConfigs backed by a dataclass), by creating an instance of the underlying dataclass.
+               See also OmegaConf.to_object.
```
omry, line 596: This should require resolve=True.
I don't think we should support instantiate without resolve as it may be used to create malformed objects.
(interpolation type is string and the underlying type may be something else, it could even an interpolation to another object).

omry (commented):
Did not read the tests yet but got some initial comments.

omegaconf/omegaconf.py
odelalleau: I see your point, but the dataclasses we use right now to define structured configs can already be "malformed", and they are still useful. So I can imagine someone wanting to use resolve=False (probably to do some manipulation on the dataclass then convert it back into a structured config).

omegaconf/basecontainer.py
odelalleau: This point is waiting for your input here ;) https://github.com/omry/omegaconf/pull/502#discussion_r594276220

omegaconf/basecontainer.py
Jasha10: I don't see why not.

omry: seems to make more sense there. move it?

Jasha10: https://github.com/omry/omegaconf/pull/502/commits/2cf460f9b2187369bce8576ea30d67ff7920ceca

omegaconf/omegaconf.py
omry: Why manipulate the dataclass instead of the config in that case?

omegaconf/basecontainer.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/d6e9749cacf127bcd3196854b6c9945014537cbd.

omegaconf/basecontainer.py
omry: Extending from Dict in Structured Configs was always is a way to mark the object as open to additions. Not a way to declare that the object should contain an actual dict fields.

I would go with the simpler solution of putting everything as attributes and ignore the fact that there the instantiated object can actually be used as a dictionary as well.

An alternative I am somewhat okay with is to put extra fields added to a Structured config extending Dict into into the inherited dictionary (__setitem__), and as attributes otherwise, but this is really getting subtle and for simplicity I prefer the first option.

I don't like replicating everything to both because it's confusing as hell and can also enable the same field to have two different values.

Bottom line, let's put it all into attributes and not treat dataclasses extending Dicts any differently.

omegaconf/basecontainer.py
```diff
@@ -265,6 +265,39 @@ def convert(val: Node) -> Any:
 
         assert False
 
+    def _instantiate_structured_config_impl(conf, instance_data: Dict[str, Any]) -> Any:
```
omry, line 264: rename conf to self in the signature?

omegaconf/omegaconf.py
odelalleau: I don't know, maybe the function doing the manipulation is actually a method of the dataclass? Or the goal is to modify / extend the dataclass in a way that is not supported by the public DictConfig API?
(anyway, I don't actually have a concrete use case I need this for myself, so it's not a huge deal, I just wanted to point out that "malformed" dataclass objects are not necessarily always bad)

omry: I think the primary reason to convert to a dataclass instance is to lose all traces of OmegaConf and have a good old dataclass you can use normally.
Having interpolation values left-over seems to go against it.

I am okay with allowing this combination in the (now) lower-level `to_container()`, but I think to_object() should be more opinionated.

odelalleau: Makes sense :)

omegaconf/basecontainer.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/f16ad229015723c6e211d6e285e6566f0c698679.

omegaconf/omegaconf.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/2bf73b024af397bf60baaf7f713cb3af0c3d4def.

docs/source/usage.rst
```diff
@@ -757,6 +757,18 @@ as DictConfig, allowing attribute style access on the resulting node.
     >>> assert type(container["structured_config"]) is DictConfig
     >>> assert container["structured_config"].port == 80
 
+Using **structured_config_mode=SCMode.INSTANTIATE** causes such nodes to
+be converted to instances of the backing dataclass.
```
omry, line 761: 1. document to_object instead and mention that it's calling to_container in with SCMode.INTANTIATE?
2. backing dataclass is inaccurate (attr classes too, maybe other things later). Underlying Structured Config type?
3. This could be too long to fit nicely in the code block, check it in a browser.

Jasha10: Ok, I made some updates in https://github.com/omry/omegaconf/pull/502/commits/32f6c68c6dde51365bc15d94fa99e54388bbfa41.

Jasha10: >     3. This could be too long to fit nicely in the code block, check it in a browser.

You were right, it's fixed now.

I also noticed that the `OmegaConf.to_container` section has a very long line that does not fit in the code block. I opened up PR https://github.com/omry/omegaconf/pull/614 to address it.

tests/structured_conf/test_structured_config.py
```diff
@@ -1010,6 +1010,26 @@ def test_str2user(self, module: Any) -> None:
             # bad key
             cfg[Color.BLUE] = "nope"
 
+    def test_str2user_with_field(self, module: Any) -> None:
```
omry, line 1013: Is this testing anything not already covered?

omegaconf/basecontainer.py
omry: @Jasha10, you didn't address this one, right?

tests/structured_conf/test_structured_config.py
Jasha10: No.
I'll remove it.

omry: I am trying to focus on specific aspects in tests, not on specific test classes (that might have a bunch of different aspects in them).
Let's remove it unless you think it's covering something not yet covered (and in that case, we should probably be more pointed in the test).

Jasha10: Removed in https://github.com/omry/omegaconf/pull/502/commits/1019df6dac36e827789a6a3040557c10f5d3cf1d.

omegaconf/basecontainer.py
Jasha10: As of commit https://github.com/omry/omegaconf/pull/502/commits/3fef7f0563bb1b304bf625e0a54a5dd42b3f9cfd, dict subclass items become instance attributes.

omegaconf/basecontainer.py
Jasha10: > @Jasha10, you didn't address this one, right?

It was not addressed when you wrote that; it is now addressed (in https://github.com/omry/omegaconf/pull/502/commits/3fef7f0563bb1b304bf625e0a54a5dd42b3f9cfd).

docs/source/usage.rst
```diff
@@ -806,6 +806,32 @@ as DictConfig, allowing attribute style access on the resulting node.
...
+
+.. doctest::
+
+    >>> container = OmegaConf.to_object(conf)
+    >>> show(container)
+    type: dict, value: {'structured_config': MyConfig(port=80, host='localhost')}
+    >>> show(container["structured_config"])
+    type: MyConfig, value: MyConfig(port=80, host='localhost')
+
+Note that here, ``container.structured_config`` is actually an instance of
```
odelalleau, line 822:
```suggestion
Note that here, ``container["structured_config"]`` is actually an instance of
```

odelalleau (changes requested):
Just a minor doc fix

docs/source/usage.rst
Jasha10: Great, thanks.

odelalleau (approved):
Good stuff! :)

omry (changes requested):
Putting back into you queue.
I have some pending questions.
Also, I see new conflicts.

docs/source/usage.rst
```diff
@@ -806,6 +806,32 @@ as DictConfig, allowing attribute style access on the resulting node.
     >>> show(container["structured_config"])
     type: DictConfig, value: {'port': 80, 'host': 'localhost'}
 
+OmegaConf.to_object
+^^^^^^^^^^^^^^^^^^^^^^
+The ``OmegaConf.to_object`` method is very similar to the
+``OmegaConf.to_container`` method.
```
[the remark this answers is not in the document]
Jasha10: Ok, I cleaned up the intro in https://github.com/omry/omegaconf/pull/502/commits/80284f4fc3209845fc380e40203973af0d3e0beb.

docs/source/usage.rst
```diff
@@ -806,6 +806,32 @@ as DictConfig, allowing attribute style access on the resulting node.
...
+``DictConfig`` object that was duck-typed to look like an instance of
+``MyConfig``.
+
+The call ``OmegaConf.to_object(conf)`` is equivalent to
+``OmegaConf.to_container(conf, resolve=True,
+structured_config_mode=SCMode.INSTANTIATE)``.  The ``resolve=True`` keyword
+argument means that string interpolations are resolved before conversion to a
+container, and ``structured_config_mode=SCMode.INSTANTIATE`` means that each
+structured config node is converted into an actual instance of the underlying
+structured config type.
```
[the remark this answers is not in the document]
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/8f17b9a45739d748ee3028b7f198e0b4317786c5.

omegaconf/basecontainer.py
```diff
@@ -257,6 +262,34 @@ def convert(val: Node) -> Any:
...
+        """Instantiate an instance of `self._metadata.object_type`, populated by `instance_data`."""
+        from ._utils import get_structured_config_field_names
+
+        object_type = self._metadata.object_type
+        object_type_field_names = set(get_structured_config_field_names(object_type))
+
+        field_items: Dict[str, Any] = {}
+        nonfield_items: Dict[str, Any] = {}
+        for k, v in instance_data.items():
+            if _is_missing_literal(v):
```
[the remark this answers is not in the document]
Jasha10: I don't think that would work. The values of `self` are subconfig objects that are not yet converted to python primitives (possibly DictConfig or ListConfig, and possibly containing unresolved interpolations). The values of `instance_data` are already resolved and converted to dict or list or to dataclass/attrclass instances.

docs/source/usage.rst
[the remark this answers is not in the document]
Jasha10: I've also added `SCMode.INSTANTIATE` documentation to the `to_container` docs, in https://github.com/omry/omegaconf/pull/502/commits/29323a4807789d35bf089b0cf06f247cf7fde541.

omegaconf/omegaconf.py
Jasha10: Reverted in https://github.com/omry/omegaconf/pull/502/commits/fe5df1d6202ff18f92d3352be3c8d6e7bcd8ff80, since the `to_object` signature no longer takes a `resolve` argument.

docs/source/usage.rst
```diff
@@ -788,6 +793,30 @@ as DictConfig, allowing attribute style access on the resulting node.
     >>> show(container["structured_config"])
     type: DictConfig, value: {'port': 80, 'host': 'localhost'}
 
+OmegaConf.to_object
+^^^^^^^^^^^^^^^^^^^^^^
+The ``OmegaConf.to_object`` method recursively converts DictConfig and ListConfig objects
+into dicts and lists, with the execption that Structured Config objects are
+converted into instances of the backing dataclass or attr class.  All OmegaConf
+interpolations are resolved before conversion to python containers.
```
odelalleau, line 801:
```suggestion
into dicts and lists, with the exception that Structured Config objects are
converted into instances of the backing dataclass or attr class.  All OmegaConf
interpolations are resolved before conversion to Python containers.
```

docs/source/usage.rst
```diff
@@ -776,6 +776,11 @@ Structured Config nodes using the ``structured_config_mode`` option.
 By default, Structured Config nodes are converted to plain dict.
 Using ``structured_config_mode=SCMode.DICT_CONFIG`` causes such nodes to remain
 as DictConfig, allowing attribute style access on the resulting node.
+Using ``structured_config_mode=SCMode.INSTANTIATE``, Structured Config nodes
+are converted to instances of the backing dataclass or attrs class. Note that
+typically ``structured_config_mode=SCMode.INSTANTIATE`` makes the most sense
+when combined with ``resolve=True``, so that interpolations are resolved before
+being using to instantiate dataclass/attr class instances.
```
odelalleau, line 783:
```suggestion
being used to instantiate dataclass/attr class instances.
```

odelalleau (changes requested):
Just some minor typos

odelalleau (approved):
No more comment from me at this point, thanks!

omegaconf/basecontainer.py
```diff
@@ -257,6 +262,34 @@ def convert(val: Node) -> Any:
...
+                )
+            if k in object_type_field_names:
+                field_items[k] = v
+            else:
+                nonfield_items[k] = v
+
+        result = object_type(**field_items)
+        for k, v in nonfield_items.items():
+            setattr(result, k, v)
+        return result
```
omry, line 291: I realized that this would not work for `ClassExtendingDictIntKeys(Dict[int, VT])`.

I wonder if we should only support extending `Dict[str, VT]` in general, eliminating this problem and simplifying a handful of other cases which are ill defined.

tests/test_to_container.py
```diff
@@ -39,19 +48,21 @@ def assert_container_with_primitives(item: Any) -> None:
...
             {"user": User(age=7, name="Bond")},
+            {"user": User(age=7, name="Bond")},
             "user",
             id="structured-inside-dict",
         ),
         param(
             [1, User(age=7, name="Bond")],
             [1, {"name": "Bond", "age": 7}],
             [1, User(age=7, name="Bond")],
+            [1, User(age=7, name="Bond")],
```
omry, line 65: This test in general is a bit odd.
We are passing everything to all functions, and each one is testing just one input.
can probably be refactored.

omegaconf/basecontainer.py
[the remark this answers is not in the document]
omry: The data passed in is derived from the content of self, which is why I am asking it.
It feels redundant.
You have custom code here that is handling things, it can resolve interpolations and deal with containers.

am I missing anything?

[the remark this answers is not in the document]
Jasha10: > The data passed in is derived from the content of self, which is why I am asking it.

This is true.

> It feels redundant.

Take a look at how `instance_data` is calculated in the calling function (`DictConfig._to_content`)... The calculation is nontrivial, and it depends on several things that are defined in the scope of the calling function.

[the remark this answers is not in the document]
Jasha10: Since the calling function has to perform this calculation anyway (for the case where `structured_config_mode == SCMode.DICT`), I think it makes sense to reuse the calculation by passing in `instance_data` for the `structured_config_mode == SCMode.INSTANTIATE` case.

tests/test_to_container.py
Jasha10: Refactored in https://github.com/omry/omegaconf/pull/502/commits/a17a11fcd1ebbfb5a5d42c47deea0e060dbd18a3.

omegaconf/basecontainer.py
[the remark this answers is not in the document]
omry: I see a few things there that does not make sense when converting a DictConfig backed by a dataclass to an object:
1. Support for not resolving interpolations
2. Support for converting enums to string (doesn't make sense if we are creating an object).

What do you think about this high level approach:
```
# converts THIS dict config to an object if it's backed by a Structured Config (error otherwise):
DictConfig._to_object()
```

The implementation will use `_to_content(scmode=Instantiate)` for any nested containers (ListConfig or DictConfig).
_to_content() can call `DictConfig._to_object()` if the container it is called on is a DictConfig backed by a Structured Config.

omegaconf/basecontainer.py
omry: Let's put down the motivation initial motivations help guide the decision:
1. DictConfig now supports non string keys.
2. Extending a Dict in a Structured config (dataclass/attr class) is used (one can say abused) as a way to flag that node as open to additions (either via setattr or via setitem).

The motivation for the second case only makes sense for string keys.
You may remember I brought up the case for extending Dict when you added support for additional key types.
I am wondering now if we should only support extending Dict[str, VT].
`Dict[Any, VT]` can issue a deprecation warning saying only `Dict[str, VT]` is supported, and it can treat it like `Dict[str, VT]`.

At that point, the problem I am bringing up here will disappear.
If that plan makes sense to you, it should be in a followup PR, not this one.

tests/test_to_container.py
```diff
@@ -39,49 +48,88 @@ def assert_container_with_primitives(item: Any) -> None:
...
-    def test_scmode_dict_config(
-        self, cfg: Any, ex_dict: Any, ex_dict_config: Any, key: Any
-    ) -> None:
-        ret = OmegaConf.to_container(cfg, structured_config_mode=SCMode.DICT_CONFIG)
-        assert ret == ex_dict_config
-        assert isinstance(ret[key], DictConfig)
+    else:
+        ret = OmegaConf.to_container(cfg, structured_config_mode=structured_config_mode)
+    assert ret == expected
+    assert isinstance(ret[key], expected_value_type)
```
omry, line 132: I like this version much more :)

omegaconf/basecontainer.py
Jasha10: > At that point, the problem I am bringing up here will disappear.
> If that plan makes sense to you, it should be in a followup PR, not this one.

Fine with me.

Jasha10: I've created issue https://github.com/omry/omegaconf/issues/657 to keep track of this.

omegaconf/basecontainer.py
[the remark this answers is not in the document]
Jasha10: Hmm... Then what would be the behavior if a client calls like this?
```python
OmegaConf.to_container(cfg, resolve=False, enum_to_str=True, structured_config_mode=SCMode.INSTANTIATE)
```
Such a call is possible with current `to_container` API.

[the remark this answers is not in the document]
omry: I it will not apply resolve and enum_to_str to the instantiated objects (but will apply it to unstructured config nodes).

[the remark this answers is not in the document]
Jasha10: Sure, makes sense to me.
So:
- the structured config gets processed with `enum_to_str=False` and `resolve=False`,
- nested configs inside the structured config get processed with `enum_to_str=False` and `resolve=False`, and
- non-structured parent configs have `enum_to_str` and `resolve` according to keyword args chosen by the client.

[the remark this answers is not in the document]
omry: > Sure, makes sense to me.
> So:
> 
> * the structured config gets processed with `enum_to_str=False` and `resolve=False`,
you mean resolve=True, right?
> * nested configs inside the structured config get processed with `enum_to_str=False` and `resolve=False`, and
> * non-structured parent configs have `enum_to_str` and `resolve` according to keyword args chosen by the client.

Since a Structured Config may contain an unstructured DictConfig, we could consider inheriting the resolve flag for them from the surrounding call, but I am not sure it's worth the added complexity.

in that case, the signature of to_object() will look something like:
```
def to_object(self, resolve_nested_configs: bool, enum_to_str_nested_configs: bool) -> Any:
  ...
```

An alternative is to just say that everything under a Structured Config is always resolved (and enums are not converted to strings).

I think I prefer the second option.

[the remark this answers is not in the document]
Jasha10: >   you mean resolve=True, right?

Yes.

> I think I prefer the second option.

Me too. I'll get started on the diff.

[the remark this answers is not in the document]
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/c672c100f9bfd07f1e753d448118603abd71eea2.

omry (commented):
I like what I am seeing. A few nits inline.
@odelalleau, can you take a second look? there have been some significant changes.

omegaconf/dictconfig.py
```diff
@@ -682,3 +683,49 @@ def _dict_conf_eq(d1: "DictConfig", d2: "DictConfig") -> bool:
...
+            assert isinstance(node, Node)
+            node = node._dereference_node(throw_on_resolution_failure=True)
+            assert node is not None
+            if isinstance(node, Container):
+                v = BaseContainer._to_content(
+                    node,
+                    resolve=True,
+                    enum_to_str=False,
+                    structured_config_mode=SCMode.INSTANTIATE,
+                    # TODO: throw_on_missing=True,
```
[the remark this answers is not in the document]
Jasha10: I think we need it for the case where a missing value is nested deep within the structured config (e.g. within another DictConfig or ListConfig).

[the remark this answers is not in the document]
omry: If we end up adding throw_on_missing, I agree.
The original motivation for it is already addressed: instantiated Structured Configs will not have MISSING assigned to their fields.

omegaconf/dictconfig.py
```diff
@@ -682,3 +683,49 @@ def _dict_conf_eq(d1: "DictConfig", d2: "DictConfig") -> bool:
                 return False
 
         return True
+
+    def _to_object(self) -> Any:
+        """Instantiate an instance of `self._metadata.object_type`.
+        This requires `self` to be a structured config.
+        Nested subconfigs are converted to_container with resolve=True."""
+        from ._utils import get_structured_config_field_names
```
odelalleau, line 691: You should be able to move this import at top of the file, no?

Also, minor formatting comment: most multiline docstrings in the code have `"""` on their own line.

omegaconf/dictconfig.py
```diff
@@ -682,3 +683,49 @@ def _dict_conf_eq(d1: "DictConfig", d2: "DictConfig") -> bool:
...
+        assert is_structured_config(object_type)
+        object_type_field_names = set(get_structured_config_field_names(object_type))
+
+        field_items: Dict[str, Any] = {}
+        nonfield_items: Dict[str, Any] = {}
+        for k in self.keys():
+            node = self._get_node(k)
+            assert isinstance(node, Node)
+            node = node._dereference_node(throw_on_resolution_failure=True)
+            assert node is not None
```
odelalleau, line 703: @omry calling `get_node()` followed by `_dereference_node()` is a somewhat common operation and it's a bit annoying right now with the mypy asserts.
Would it make sense to add a `get_dereferenced_node()` function that would return a `Node` (not optional) to streamline such code?

odelalleau (changes requested):
Just some small comments.

Overall looks good to me. I'm not entirely sure we should ignore the `resolve` flag for structured configs (if someone wants an unresolved dataclass to later convert it back to a DictConfig, why not?), but I don't care much about it so let's go :)

omegaconf/dictconfig.py
Jasha10: That would make sense to me.
I'll do it in another PR :)

docs/source/usage.rst
```diff
@@ -776,6 +776,17 @@ Structured Config nodes using the ``structured_config_mode`` option.
...
+are converted to instances of the backing dataclass or attrs class. Note that
+when ``structured_config_mode=SCMode.INSTANTIATE``, interpolations nested within
+a structured config node will be resolved, even if ``OmegaConf.to_container`` is called
+with the the keyword argument ``resolve=False``, so that interpolations are resolved before
+being used to instantiate dataclass/attr class instances. Interpolations within
+non-structured parent nodes will be resolved (or not) as usual, according to
+the ``resolve`` keyword arg.
+Similarly, when ``structured_config_mode=SCMode.INSTANTIATE``, enum values nested within a
+structured config node will not be converted to ``str``, even if ``OmegaConf.to_container``
+is called with ``enum_to_str=True``.
```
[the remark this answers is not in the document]
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/4787e8d34cc866e0a99e8744c5cedf5686c4a683.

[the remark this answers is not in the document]
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/c1d13f8a5419fe4c5aa5be714b742b996bf671a1.

omegaconf/dictconfig.py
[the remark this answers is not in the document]
Jasha10: I'll be working on the throw_on_missing PR soon :P

Anyway, I've removed the TODO comment in https://github.com/omry/omegaconf/pull/502/commits/bc2f610322b3f55833cc3a115862a8e51af0efd3.

omegaconf/dictconfig.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/502/commits/b51d33f1135197cb1e204d2588b997f9e2b4f46c and https://github.com/omry/omegaconf/pull/502/commits/f1a4270bdae1c8f6fc219e8afe5a33c098892823.

odelalleau: Thanks! The first `"""` is still on the same line but no biggie :)

omry (approved):
shipit!

omegaconf/dictconfig.py
Jasha10: Fixed in https://github.com/omry/omegaconf/pull/502/commits/d12701e9add8ee7084b9ca8598870e441e974d6c :)

omegaconf/dictconfig.py
Jasha10: First step is here: #668

PR #503, opened 2021-01-30
To container throw on missing

### LINKED ISSUE #501: Feature: throw_on_missing keyword argument for OmegaConf.to_container
# Feature:
Currently, calling `OmegaConf.to_container(cfg)` can result in a python object that contains some `"???"` MISSING strings.

I propose to add a `throw_on_missing` keyword argument to `OmegaConf.to_container`, so that the structure returned by `OmegaConf.to_container(cfg, throw_on_missing=True)` will raise `omegaconf.MissingMandatoryValue` if any missing values are encountered.

This way, if no `MissingMandatorValue` is raised, clients can be assured that the result of `to_container` does not contain any `"???"` strings.

To maintain backward compatibility, the default would be `throw_on_missing=False`.

Jasha10:
Closes #501

### DIFF
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -193,6 +193,7 @@ class BaseContainer(Container, ABC):
         resolve: bool,
         enum_to_str: bool = False,
         exclude_structured_configs: bool = False,
+        throw_on_missing: bool = False,
     ) -> Union[None, Any, str, Dict[DictKeyType, Any], List[Any]]:
         from .dictconfig import DictConfig
         from .listconfig import ListConfig
@@ -212,6 +213,12 @@ class BaseContainer(Container, ABC):
             assert isinstance(inter, str)
             return inter
         elif conf._is_missing():
+            if throw_on_missing:
+                conf._format_and_raise(
+                    key=None,
+                    value="???",
+                    cause=MissingMandatoryValue("Missing mandatory value"),
+                )
             return "???"
         elif isinstance(conf, DictConfig):
             if conf._metadata.object_type is not None and exclude_structured_configs:
@@ -219,7 +226,7 @@ class BaseContainer(Container, ABC):
 
             retdict: Dict[str, Any] = {}
             for key in conf.keys():
-                node = conf._get_node(key)
+                node = conf._get_node(key, throw_on_missing_value=throw_on_missing)
                 assert isinstance(node, Node)
                 if resolve:
                     node = node._dereference_node(
@@ -233,6 +240,7 @@ class BaseContainer(Container, ABC):
                         resolve=resolve,
                         enum_to_str=enum_to_str,
                         exclude_structured_configs=exclude_structured_configs,
+                        throw_on_missing=throw_on_missing,
                     )
                 else:
                     retdict[key] = convert(node)
@@ -240,7 +248,7 @@ class BaseContainer(Container, ABC):
         elif isinstance(conf, ListConfig):
             retlist: List[Any] = []
             for index in range(len(conf)):
-                node = conf._get_node(index)
+                node = conf._get_node(index, throw_on_missing_value=throw_on_missing)
                 assert isinstance(node, Node)
                 if resolve:
                     node = node._dereference_node(
@@ -253,6 +261,7 @@ class BaseContainer(Container, ABC):
                         resolve=resolve,
                         enum_to_str=enum_to_str,
                         exclude_structured_configs=exclude_structured_configs,
+                        throw_on_missing=throw_on_missing,
                     )
                     retlist.append(item)
                 else:
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -229,9 +229,13 @@ class BaseContainer(Container, ABC):
                 node = conf._get_node(key, throw_on_missing_value=throw_on_missing)
                 assert isinstance(node, Node)
                 if resolve:
-                    node = node._dereference_node(
-                        throw_on_missing=False, throw_on_resolution_failure=True
-                    )
+                    try:
+                        node = node._dereference_node(
+                            throw_on_missing=False, throw_on_resolution_failure=True
+                        )
+                    except MissingMandatoryValue as e:
+                        assert node is not None
+                        conf._format_and_raise(key=key, value=node._value(), cause=e)
 
                 assert node is not None
                 if isinstance(node, Container):
@@ -251,9 +255,13 @@ class BaseContainer(Container, ABC):
                 node = conf._get_node(index, throw_on_missing_value=throw_on_missing)
                 assert isinstance(node, Node)
                 if resolve:
-                    node = node._dereference_node(
-                        throw_on_missing=False, throw_on_resolution_failure=True
-                    )
+                    try:
+                        node = node._dereference_node(
+                            throw_on_missing=False, throw_on_resolution_failure=True
+                        )
+                    except MissingMandatoryValue as e:
+                        assert node is not None
+                        conf._format_and_raise(key=index, value=node._value(), cause=e)
                 assert node is not None
                 if isinstance(node, Container):
                     item = BaseContainer._to_content(
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -447,7 +447,11 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
             if throw_on_missing_key:
                 raise ConfigKeyError(f"Missing key {key}")
         elif throw_on_missing_value and value._is_missing():
-            raise MissingMandatoryValue("Missing mandatory value")
+            self._format_and_raise(
+                key=key,
+                value=None,
+                cause=MissingMandatoryValue("Missing mandatory value"),
+            )
         return value
 
     def pop(self, key: DictKeyType, default: Any = DEFAULT_VALUE_MARKER) -> Any:
--- a/omegaconf/listconfig.py
+++ b/omegaconf/listconfig.py
@@ -390,13 +390,22 @@ class ListConfig(BaseContainer, MutableSequence[Any]):
             if value is not None:
                 if isinstance(key, slice):
                     assert isinstance(value, list)
-                    for v in value:
+                    indices = range(len(self))[key]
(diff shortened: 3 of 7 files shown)


Jasha10:
Still needs test coverage for the new `throw_on_missing_flag`.

omry (commented):
It will probably be simpler to 
node = conf._get_node(key, throw_on_missing=True)

Try to line the error message with the normal error for missing values (`raise MissingMandatoryValue("Missing mandatory value: $FULL_KEY")`.

Jasha10:
Rebased onto master.

Jasha10:
> It will probably be simpler to
> node = conf._get_node(key, throw_on_missing=True)

`conf._get_node` does not have a `throw_on_missing` keyword option.
Do you mean that I should add a `throw_on_missing` keyword argument to the `_get_node` signature?

omry:
> Do you mean that I should add a throw_on_missing keyword argument to the _get_node signature?

This is what I had in mind, yes. I added it in #513.

tests/test_errors.py
```diff
@@ -1070,6 +1071,85 @@ def finalize(self, cfg: Any) -> None:
         ),
         id="list,readonly:del",
     ),
+    # to_container: throw_on_missing
```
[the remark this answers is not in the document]
Jasha10: > They should probably be moved to test_omegaconf.py, or maybe even to a dedicated test_to_container.py file.

Ok, I'll move the `to_container` tests into a `test_to_container.py` file.

> As a side note:
> test_errors should probably be reorganized and broken into smaller functions (while reusing the error validation logic).

I'll open a draft PR for refactoring `test_errors.py`

omegaconf/basecontainer.py
```diff
@@ -211,6 +212,12 @@ def convert(val: Node) -> Any:
             assert isinstance(inter, str)
             return inter
         elif conf._is_missing():
+            if throw_on_missing:
+                conf._format_and_raise(
+                    key=None,
+                    value="???",
+                    cause=MissingMandatoryValue("Missing mandatory value: $FULL_KEY"),
+                )
```
omry, line 220: can you refactor out a small local raise_missing_value_exception() function to reduce the amount of new code?

omegaconf/omegaconf.py
```diff
@@ -502,13 +502,16 @@ def to_container(
...
+        throw_on_missing: bool = False,
         enum_to_str: bool = False,
         exclude_structured_configs: bool = False,
     ) -> Union[Dict[DictKeyType, Any], List[Any], None, str]:
         """
         Resursively converts an OmegaConf config to a primitive container (dict or list).
         :param cfg: the config to convert
         :param resolve: True to resolve all values
+        :param throw_on_missing: Whether an error should be thrown when "???"
+                values are encountered in the given config object.
```
omry, line 514: Side note:
Currently we do not generate API docs.
It would actually be useful as the API surface area increases.

omry (changes requested):
Marking as request-changes (see prev comments)

Jasha10:
Rebased onto `master`

omegaconf/basecontainer.py
Jasha10: Done, in https://github.com/omry/omegaconf/pull/503/commits/7b49aea8781c97a1515a7f723f78a141054e4dfe.

omegaconf/omegaconf.py
Jasha10: Created issue #536 to keep track of this idea.

tests/test_errors.py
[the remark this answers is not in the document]
Jasha10: > They should probably be moved to test_omegaconf.py, or maybe even to a dedicated test_to_container.py file.

I've submitted PR #538 to move `to_container` tests into a new `test_to_container.py` file.

omry (commented):
Can't comment on code you didn't change but I found what I was thinking about with the _get_node() comment:

```python
            retdict: Dict[str, Any] = {}
            for key in conf.keys():	            for key in conf.keys():
                node = conf._get_node(key)	                node = conf._get_node(key)
```

In above _get_node(), what happens if you use `conf._get_node(key, throw_on_missing=throw_on_missing)`?

(You will probably need to do the same for ListConfig somwehere).

Jasha10:
> In above _get_node(), what happens if you use `conf._get_node(key, throw_on_missing=throw_on_missing)`?

That works! There are a few test failures in `test_errors.py` which look like this:
```
AssertionError: Regex pattern 'Missing mandatory value: subcfg.x' does not match 'Missing mandatory value'.
```
If it's ok with you to change the error message, I'll make the change to using `conf._get_node(key, throw_on_missing_value=throw_on_missing)` and will update the `test_errors.py` error messages.

omegaconf/dictconfig.py
```diff
@@ -447,7 +447,11 @@ def _get_node(
             if throw_on_missing_key:
                 raise ConfigKeyError(f"Missing key {key}")
         elif throw_on_missing_value and value._is_missing():
-            raise MissingMandatoryValue("Missing mandatory value")
+            self._format_and_raise(
+                key=key,
+                value=None,
+                cause=MissingMandatoryValue("Missing mandatory value"),
+            )
```
Jasha10, line 454: I had to change `DictConfig._get_node` (and a similar change for `ListConfig._get_node` so that the `MissingMandatoryValue` will have more information about what `key` and what `node` caused the error.

omegaconf/listconfig.py
```diff
@@ -390,13 +390,21 @@ def _get_node(
...
+                            )
                 else:
                     assert isinstance(value, Node)
                     if throw_on_missing_value and value._is_missing():
-                        raise MissingMandatoryValue("Missing mandatory value")
+                        self._format_and_raise(
+                            key=key,
+                            value=None,
+                            cause=MissingMandatoryValue("Missing mandatory value"),
+                        )
```
omry, line 407: Nice, but add corresponding tests of _get_node that confirms that this is working (especially the slice case).
Feel free to improve test_get_node_throw_on_missing_value to do that instead of making test_errors even bigger.

omry (changes requested):
much cleaner.

omegaconf/listconfig.py
Jasha10: By adding tests (https://github.com/omry/omegaconf/pull/503/commits/7d7026fb8cc8463e0a5c291abdca6a6549176a25) I discovered that there was an error in the slice case. I have seen the light. Test everything.

omegaconf/listconfig.py
```diff
@@ -390,13 +390,22 @@ def _get_node(
...
-                    for v in value:
+                    indices = range(len(self))[key]
+                    for idx, v in zip(indices, value):
                         if throw_on_missing_value and v._is_missing():
-                            raise MissingMandatoryValue("Missing mandatory value")
+                            self._format_and_raise(
+                                key=idx,
+                                value=None,
+                                cause=MissingMandatoryValue("Missing mandatory value"),
+                            )
```
omry, line 400: This is a lot of custom logic for pretty low return.
I can live with a simpler handling for slice errors with missing (no need to report the exact first missing key).
I think just use the slice as is without reporting the exact index. it would also simplify your tests.

omry (changes requested):

Jasha10:
Currently code coverage tests are failing because the `except` blocks from the latest commit https://github.com/omry/omegaconf/pull/503/commits/fe32082fca206722323c2387f04d2074108c1297 have no coverage. Those try/except statements were added in anticipation of #545, and I anticipate that the coverage issue will be resolved one this branch is rebased onto that one.

omegaconf/listconfig.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/503/commits/01d53955351a9db389cf7ac3ee42b3f0085c081e

omegaconf/basecontainer.py
```diff
@@ -229,9 +229,13 @@ def convert(val: Node) -> Any:
...
-                    node = node._dereference_node(
-                        throw_on_missing=False, throw_on_resolution_failure=True
-                    )
+                    try:
+                        node = node._dereference_node(
+                            throw_on_missing=False, throw_on_resolution_failure=True
+                        )
+                    except MissingMandatoryValue as e:
+                        assert node is not None
+                        conf._format_and_raise(key=key, value=node._value(), cause=e)
```
omry, line 238: 1. Why is _dereference_node throwing a MandatoryMissingValue is throw_on_missing is False?
2. Why are you passing False to throw_on_missing and not the input flag?

Jasha10: > 1. Why is _dereference_node throwing a MandatoryMissingValue is throw_on_missing is False?

This is because we are calling `_dereference_node` with `throw_on_resolution_failure=True`.
If #545 is merged in it's current state, `_dereference_node` will raise `MandatoryMissingValue` if `throw_on_resolution_failure==True` and the interpolation cannot be resolved.

This being said, I actually think it makes more sense to change this from
```python
_dereference_node(..., throw_on_resolution_failure=True)
```
to
```python
_dereference_node(..., throw_on_resolution_failure=throw_on_missing)
```
so that resolution failure will not cause error if the user does not specify the `throw_on_missing` option. I'll add tests to this effect...

Jasha10: > 2. Why are you passing False to throw_on_missing and not the input flag?

The current workflow is:
- first, get the node and (optionally) throw if missing:
```python
node = conf._get_node(key, throw_on_missing_value=throw_on_missing)
```
- second: if `resolve==True`, dereference the node:
```python
node = node._dereference_node(throw_on_missing=False, throw_on_resolution_failure=True)
```
Since the `_get_node` call already has `throw_on_missing_value=throw_on_missing`, it would be redundant to use the `throw_on_missing` flag in the `_dereference_node` call as well.

Jasha10:
#545 will change the behavior of `_dereference_node` significantly enough that it makes sense to put this PR on hold until #545 gets merged.

Jasha10:
Marking this PR as needing possible API redesign after #502 gets merged.

Jasha10:
This PR is old; closing in favor of #730.

PR #505, opened 2021-02-01
Changing default ref_type to Any

omry:
ref_type was meant to be equivalent to type annotation:
```
x : Foo = Bar()
```

Above, x should have the ref_type Foo.
ref_type is consulted when assigning values to x, to ensure that they are of the right type.
(such assignment results in a runtime exception in OmegaConf).

Over time, ref_type was used for additional things like encoding the key type and element type of a DictConfig.
In addition, for implementation reasons, ref_type of Structured Configs automatically matched the object type:

```
cfg = OmegaConf.create({"user":  User})
```
above, cfg.user has the ref_type User, which means it cannot be assigned anything else.
This is not intentional.
Such behavior is justified using Structured Config:
```python
@dataclass
class Config
  user: User = User()
```

but not when using the previous form.

This PR is:
1. Simplifying ref_type back to what i was always meant to be.
2. Simplifies error messages to only contain reference_type if it's not Any.
3. Unrelated: Fixes exceptions raised to contain the correct stack trace (debugging OmegaConf is now far less annoying).

### DIFF
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -527,57 +527,23 @@ def _get_value(value: Any) -> Any:
 
 
 def get_ref_type(obj: Any, key: Any = None) -> Optional[Type[Any]]:
-    from omegaconf import DictConfig, ListConfig
-    from omegaconf.base import Container, Node
-    from omegaconf.nodes import ValueNode
+    from omegaconf import Container, Node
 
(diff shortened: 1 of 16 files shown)


omry:
cc @odelalleau , @pereman2 ,  @Jasha10:
This might cause merge or other issues with outstanding pull requests.
Please rebase once I merge and let me know if you need help.

PR #507, opened 2021-02-02
Support YAML merge tags

### LINKED ISSUE #470: using YAML merge tag (<<) results in exception
**Describe the bug**

This YAML code causes an exception with ``OmegaConf.create()```:
```python
OmegaConf(create("""
        a: &A
            x: 1
            y: 2

        b:
            <<: *A
            x: 3
            z: 1
"""))
# yaml.constructor.ConstructorError: could not determine a constructor for the tag 'tag:yaml.org,2002:merge'
```
It loads fine with the standard loader

**To Reproduce**
See above.

**Expected behavior**
Excpected to load without error.

**Additional context**
- [x] OmegaConf version: master
- [x] Python version: 3.6.9
- [x] Operating system : Ubuntu 18.04

jgehring:
This adds support for YAML merge tags (<< *ref) while retaining the
sanity check for duplicate keys. Note that the spec for merge keys
(https://yaml.org/type/merge.html) explicitly states that keys in the
current mapping override the ones in the merged mapping. Hence, the
check for duplicates is applied to scalar keys of the current mapping
only.

Fixes #470.

### DIFF
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -2,6 +2,7 @@
 import re
 import sys
 from typing import Any, Dict, List, Optional
+from textwrap import dedent
 
 import pytest
 import yaml
@@ -293,3 +294,47 @@ def test_create_untyped_dict() -> None:
 
     cfg = DictConfig(ref_type=Dict, content={})
     assert get_ref_type(cfg) == Optional[Dict]
+
+
+def test_yaml_duplicate_keys() -> None:
+    from yaml.constructor import ConstructorError
+
+    with pytest.raises(ConstructorError):
+        OmegaConf.create(
+            dedent(
+                """\
+                a:
+                  b: 1
+                  c: 2
+                  b: 3
+                """
+            )
+        )
+    with pytest.raises(ConstructorError):
+        OmegaConf.create(
+            dedent(
+                """\
+                a:
+                  b: 1
+                a:
+                  b: 2
+                """
+            )
+        )
+
+    cfg = OmegaConf.create(
+        dedent(
+            """\
+                a: &A
+                    x: 1
+                b: &B
+                    y: 2
+                c:
+                    <<: *A
+                    <<: *B
+                    x: 3
+                    z: 1
+                """
+        )
+    )
+    assert cfg == {"a": {"x": 1}, "b": {"y": 2}, "c": {"x": 3, "y": 2, "z": 1}}
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -1,8 +1,8 @@
 """Testing for OmegaConf"""
 import re
 import sys
-from typing import Any, Dict, List, Optional
 from textwrap import dedent
+from typing import Any, Dict, List, Optional
 
 import pytest
 import yaml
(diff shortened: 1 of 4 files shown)


tests/test_serialization.py
```diff
@@ -183,6 +183,29 @@ def test_load_duplicate_keys_sub() -> None:
...
+                    y: 2
+                c:
+                    <<: *A
+                    <<: *B
+                    x: 3
+                    z: 1
+                """
+            )
+            fp.write(content.encode("utf-8"))
+        cfg = OmegaConf.load(fp.name)
```
omry, line 203: no need to go through a file. you can initialize OmegaConf directly from a yaml string:

```
cfg = OmegaConf.create(content)
```

At this point, also clean up test_load_duplicate_keys_top().
I think both tests belongs in test_create().

omry (changes requested):
Very clean, thanks :)
See small request.

omry:
Please add a news fragment file in the news directory.

tests/test_create.py
```diff
@@ -293,3 +294,47 @@ def test_create_untyped_dict() -> None:
 
     cfg = DictConfig(ref_type=Dict, content={})
     assert get_ref_type(cfg) == Optional[Dict]
+
+
+def test_yaml_duplicate_keys() -> None:
```
omry, line 299: 1. do not merge the tests. two different behaviors gets two tests. (one for duplicate keys, one for merge support).
2. looks like you have two tests for duplicate keys. Can you parameterize with pytest.mark.parametereize?

omry (changes requested):
see comments.
also, please add  a news fragment for the bug fix.

omry:
Looking good.
you have some lint issues (isort).
you can run nox locally (e.g. nox -s lint-3.8).

omry (changes requested):
1. Please fix lint issues (hints in the [contributing](https://github.com/omry/omegaconf/blob/master/CONTRIBUTING.md) doc)
2. Please add a new s fragment.  This is similar to Hydra, see [this](https://hydra.cc/docs/development/documentation#news-entries).

omry (changes requested):
Almost.
We generate the NEWS.md from the files in the news directory using towncrier.

It's really simple, but you have to read the link I provided :)

basically, create a file:
```
news/ISSUE_ID.bugfix
```

With the content of the news entry.
```
Support merge tags in YAML files
```

The file name should be after the issue, not the pull request.
so in this case `news/470.bugfix`.

jgehring:
Haha, sorry for messing it up :)

omry (approved):
Awesome, thanks for fixing it!

gchhablani:
Hi @omry,

Will the next release include this feature? When is it scheduled?

omry:
This bug fix (not a feature) will be included in 2.1.
It will be released when 2.1 is ready. In the mean time you can use the 2.1 dev release. (`pip install omegaconf --pre --upgrade`) but note that it's not finalized.

gchhablani:
Hi @omry,

I installed the latest dev version. Now, interpolation doesn't work even if there is no YAML merge.

omry:
There are about 500 unit tests covering interpolation. If you think you have a big please file a new issue a minimal repro.

gchhablani:
Hi @omry,

I apologize. Interpolation works fine. I was checking it the wrong way.

Infact, interpolation with YAML merge also works:
Like this:
```
default: &default
  a: 1
  b: 2

modify1: 
  <<: *default
  c: ${default.b}
```

Output on printing:
```
{'default': {'a': 1, 'b': 2}, 'modify1': {'a': 1, 'b': 2, 'c': '${default.b}'}}
{'a': 1, 'b': 2}
1
2
{'a': 1, 'b': 2, 'c': '${default.b}'}
1
2
2
```

PR #513, opened 2021-02-03
added throw_on_missing flag to _get_node()

omry:
I also noticed that _get_node() can operate on slices as well, in which case it does not return a Node but a list.
I changed it's return type to Any to address this.

### DIFF
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -220,6 +220,7 @@ class BaseContainer(Container, ABC):
             for key in conf.keys():
                 node = conf._get_node(key)
                 assert node is not None
+                assert isinstance(node, Node)
                 if resolve:
                     node = node._dereference_node(
                         throw_on_missing=False, throw_on_resolution_failure=True
@@ -241,6 +242,7 @@ class BaseContainer(Container, ABC):
             for index in range(len(conf)):
                 node = conf._get_node(index)
                 assert node is not None
+                assert isinstance(node, Node)
                 if resolve:
                     node = node._dereference_node(
                         throw_on_missing=False, throw_on_resolution_failure=True
@@ -331,6 +333,8 @@ class BaseContainer(Container, ABC):
         for key, src_value in src.items_ex(resolve=False):
             src_node = src._get_node(key, validate_access=False)
             dest_node = dest._get_node(key, validate_access=False)
+            assert src_node is None or isinstance(src_node, Node)
+            assert dest_node is None or isinstance(dest_node, Node)
 
             if isinstance(dest_node, DictConfig):
                 dest_node._validate_merge(value=src_node)
@@ -558,6 +562,7 @@ class BaseContainer(Container, ABC):
                     if is_structured_config(val):
                         ref_type = self._metadata.element_type
                 else:
+                    assert isinstance(target, Node)
                     is_optional = target._is_optional()
                     ref_type = target._metadata.ref_type
             return _maybe_wrap(
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -608,6 +608,9 @@ class BaseContainer(Container, ABC):
         v2 = c2._get_node(k2)
         assert v1 is not None and v2 is not None
 
+        assert isinstance(v1, Node)
+        assert isinstance(v2, Node)
+
         if v1._is_none() and v2._is_none():
             return True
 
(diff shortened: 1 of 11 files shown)


omry:
@Jasha10, can you review?

Jasha10:
Will take a look now.

omry:
> Will take a look now.

Thanks.
I just updated the typing.
By the way, can you join the Hydra chat?

Jasha10:
Yes, I've joined the chat.

omegaconf/basecontainer.py
```diff
@@ -220,6 +220,7 @@ def convert(val: Node) -> Any:
             for key in conf.keys():
                 node = conf._get_node(key)
                 assert node is not None
+                assert isinstance(node, Node)
```
Jasha10, line 223:
```suggestion
                assert isinstance(node, Node)
```

No need for the first assertion, since the second one implies `node is not None`.

omegaconf/basecontainer.py
```diff
@@ -241,6 +242,7 @@ def convert(val: Node) -> Any:
             for index in range(len(conf)):
                 node = conf._get_node(index)
                 assert node is not None
+                assert isinstance(node, Node)
```
Jasha10, line 245:
```suggestion
                assert isinstance(node, Node)
```

omegaconf/listconfig.py
```diff
@@ -379,8 +384,24 @@ def _get_node(
             assert isinstance(self.__dict__["_content"], list)
             if validate_access:
                 self._validate_get(key)
-            return self.__dict__["_content"][key]  # type: ignore
+            assert self.__dict__["_content"] is not None
+            assert not isinstance(self.__dict__["_content"], str)
```
Jasha10, line 388: I am confused by these three assertions:
  - `isinstance(_content, list)`
  - `_content is not None`
  - `not isinstance(_content, str)`

Could these maybe be replaced with a single assertion?

Jasha10: On my computer, the tests + mypy still pass if the later two assertions are deleted.

Jasha10, line 388:
```suggestion
```

omegaconf/omegaconf.py
```diff
@@ -901,8 +902,10 @@ def _select_one(
     assert isinstance(c, (DictConfig, ListConfig)), f"Unexpected type : {c}"
     if isinstance(c, DictConfig):
         assert isinstance(ret_key, str)
-        val: Optional[Node] = c._get_node(ret_key, validate_access=False)
+        val = c._get_node(ret_key, validate_access=False)
+        assert val is None or isinstance(val, Node)
         if val is not None:
+            assert isinstance(val, Node)
```
Jasha10, line 908:
```suggestion
        assert val is None or isinstance(val, Node)
        if val is not None:
```

Technically the second assertion here is not necessary: mypy can infer that `val` has type `Node` based on the previous two lines.

Jasha10: (If you type `reveal_type(val)` inside of the `if` block, mypy prints `Revealed type is 'omegaconf.base.Node'`)

Jasha10 (commented):
Looks good!

omegaconf/basecontainer.py
omry: roger.

omegaconf/basecontainer.py
omry: roger roger

omegaconf/listconfig.py
omry: Good spot, the first one is sufficient.

omry (commented):
Thanks, fixed.

Many of those are because the first iteration was returning Any and later I decided that returning a Union is a better idea.

tests/test_base_config.py
```diff
@@ -739,3 +739,18 @@ def test_assign(parent: Any, key: Union[str, int], value: Any, expected: Any) ->
...
+    [
+        # dict
+        (OmegaConf.create({"foo": "???"}), "foo"),
+        # list
+        (OmegaConf.create([10, "???", 30]), 1),
+        (OmegaConf.create([10, "???", 30]), slice(1, 2)),
+    ],
+)
+def test_get_node_throw_on_missing(cfg: Any, key: Any) -> None:
+    with pytest.raises(MissingMandatoryValue):
```
Jasha10, line 755:
```suggestion
    with pytest.raises(
        MissingMandatoryValue,
        match="Missing (mandatory value|node) " + re.escape(f"'{key}'"),
    ):
```

omry: Meh, I just normalized the exception message :)
Why use a toopick when you can use a BFG7000?

PR #514, opened 2021-02-04
Structured Configs IR : WIP

omry:


omry:
This PR is obsolete and would need to be redone from scratch, but the underlying idea is still important.

OmegaConf needs a backend-independent internal representation for structured config fields and metadata. Without that, every structured-config backend becomes its own pipeline, and testing behavior across dataclasses, attrs, future Pydantic support, PEP 593 metadata, validators, aliases, defaults, and constraints becomes unmanageable.

The original WIP here was an early attempt in that direction. I am closing this PR because it is no longer mergeable as-is, but I would like to keep the idea alive in a fresh design issue focused on a normalized structured config IR.

PR #516, opened 2021-02-05
Dict getitem consistency

### LINKED ISSUE #515: OmegaConf 2.1: changes in DictConfig __getitem__ and __getattr__
An historical decision was to make `__getitem__` and `__getattr__` return None if a the input key is not found.
This is inconsistent with the behavior of plain dict and regular Python objects.

Fixing this is a breaking change, users depending on it will need small code changes.

```python
if cfg["some_attribute"] is not None:
   ...
```
And

```python
if cfg.some_attribute is not None:
   ...
```

Will have to be replaced with
```python
if "some_attribute" in cfg:
  ...
```

Note that is cfg is ducktyped as a type, you can use hasattr just like you would on an real instance of Config:
```python
def foo(cfg: Config):
  # before
  if cfg.some_attribute is not None: 
    ..
  # after
  if hasattr(cfg, "some_attribute"):
    ...
```

Note:
DictConfig put into struct mode is already behaving in the new way (raising exceptions).
This change should not affect most Hydra users because Hydra puts the cfg object into struct mode by default.

omry:
Fixes #515.

This is an historical mistake. See #515 for some explanation and migration instructions.

### DIFF
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -347,15 +347,15 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
         :param key:
         :return:
         """
-        # PyCharm is sometimes inspecting __members__, be sure to tell it we don't have that.
-        if key == "__members__":
-            raise AttributeError()
-
         if key == "__name__":
             raise AttributeError()
 
         try:
             return self._get_impl(key=key, default_value=DEFAULT_VALUE_MARKER)
+        except ConfigKeyError as e:
+            self._format_and_raise(
+                key=key, value=None, cause=e, type_override=ConfigAttributeError
+            )
         except Exception as e:
             self._format_and_raise(key=key, value=None, cause=e)
 
@@ -413,9 +413,9 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
 
     def _get_impl(self, key: DictKeyType, default_value: Any) -> Any:
         try:
-            node = self._get_node(key=key)
-        except ConfigAttributeError:
-            if default_value != DEFAULT_VALUE_MARKER:
+            node = self._get_node(key=key, throw_on_missing_key=True)
+        except (ConfigAttributeError, ConfigKeyError):
+            if default_value is not DEFAULT_VALUE_MARKER:
                 node = default_value
             else:
                 raise
@@ -427,7 +427,8 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
         self,
         key: DictKeyType,
         validate_access: bool = True,
-        throw_on_missing: bool = False,
+        throw_on_missing_value: bool = False,
+        throw_on_missing_key: bool = False,
     ) -> Union[Optional[Node], List[Optional[Node]]]:
         try:
             key = self._validate_and_normalize_key(key)
@@ -440,10 +441,12 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
         if validate_access:
             self._validate_get(key)
 
-        value: Node = self.__dict__["_content"].get(key)
-        if throw_on_missing and value._is_missing():
+        value: Optional[Node] = self.__dict__["_content"].get(key)
+        if value is None:
+            if throw_on_missing_key:
+                raise ConfigKeyError(f"Missing key {key}")
+        elif throw_on_missing_value and value._is_missing():
             raise MissingMandatoryValue("Missing mandatory value")
-
         return value
 
     def pop(self, key: DictKeyType, default: Any = DEFAULT_VALUE_MARKER) -> Any:
--- a/omegaconf/listconfig.py
+++ b/omegaconf/listconfig.py
@@ -372,7 +372,8 @@ class ListConfig(BaseContainer, MutableSequence[Any]):
         self,
         key: Union[int, slice],
         validate_access: bool = True,
-        throw_on_missing: bool = False,
+        throw_on_missing_value: bool = False,
+        throw_on_missing_key: bool = False,
     ) -> Union[Optional[Node], List[Optional[Node]]]:
         try:
             if self._is_none():
@@ -390,15 +391,15 @@ class ListConfig(BaseContainer, MutableSequence[Any]):
                 if isinstance(key, slice):
                     assert isinstance(value, list)
                     for v in value:
-                        if throw_on_missing and v._is_missing():
+                        if throw_on_missing_value and v._is_missing():
                             raise MissingMandatoryValue("Missing mandatory value")
                 else:
                     assert isinstance(value, Node)
-                    if throw_on_missing and value._is_missing():
+                    if throw_on_missing_value and value._is_missing():
                         raise MissingMandatoryValue("Missing mandatory value")
             return value
         except (IndexError, TypeError, MissingMandatoryValue, KeyValidationError) as e:
-            if isinstance(e, MissingMandatoryValue) and throw_on_missing:
+            if isinstance(e, MissingMandatoryValue) and throw_on_missing_value:
                 raise
             if validate_access:
                 self._format_and_raise(key=key, value=None, cause=e)
--- a/tests/test_basic_ops_dict.py
+++ b/tests/test_basic_ops_dict.py
@@ -49,10 +49,10 @@ def test_setattr_deep_from_empty() -> None:
     assert {"a": {"b": 9}} == c
 
 
-def test_setattr_deep_map() -> None:
-    c = OmegaConf.create(dict(a=dict(b=dict(c=1))))
+def test_setattr_dict_nested() -> None:
+    c = OmegaConf.create({"a": {"b": {"c": 1}}})
     c.a.b = {"z": 10}
-    assert {"a": {"b": {"z": 10}}} == c
+    assert c == {"a": {"b": {"z": 10}}}
 
 
 def test_getattr() -> None:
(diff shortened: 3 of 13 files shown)


tests/test_basic_ops_dict.py
```diff
@@ -613,7 +620,7 @@ def test_creation_with_invalid_key() -> None:
 def test_set_with_invalid_key() -> None:
     cfg = OmegaConf.create()
     with pytest.raises(KeyValidationError):
-        cfg[object()] = "a"  # type: ignore
+        cfg[object()]  # type: ignore
```
Jasha10, line 623: The `test_set_with_invalid_key` test is now identical to the `test_get_with_invalid_key` that follows it.
I would suggest to revert this change. (The tests still pass after reverting)

```suggestion
        cfg[object()] = "a"  # type: ignore
```

omegaconf/listconfig.py
```diff
@@ -372,7 +372,8 @@ def _get_node(
         self,
         key: Union[int, slice],
         validate_access: bool = True,
-        throw_on_missing: bool = False,
+        throw_on_missing_value: bool = False,
+        throw_on_missing_key: bool = False,
```
Jasha10, line 376: Currently `ListConfig._get_node` gives the same behavior whether `throw_on_missing_key` is `True` or `False` (because the method does not make any use of this keyword argument).

omry: Yes, this is coming from the interface. there was no need yet to support it.
This is unrelated to this PR though, feel free to post a followup issue fixing it (there aren't a lot of uses of throw_on_missing yet, we may eventually just revert it is there isn't any need).

tests/test_basic_ops_dict.py
omry: good catch.
I fixed the tests to call `__getitem__` and `__setitem__` explicitly.
I didn't initially realize that those two tests actually were different at a glance.

omegaconf/dictconfig.py
```diff
@@ -413,8 +413,8 @@ def get(self, key: DictKeyType, default_value: Any = DEFAULT_VALUE_MARKER) -> An
 
     def _get_impl(self, key: DictKeyType, default_value: Any) -> Any:
         try:
-            node = self._get_node(key=key)
-        except ConfigAttributeError:
+            node = self._get_node(key=key, throw_on_missing_key=True)
+        except (ConfigAttributeError, ConfigKeyError):
             if default_value != DEFAULT_VALUE_MARKER:
```
odelalleau, line 418: Unrelated to this PR, but `is not` is a bit better than `!=` (avoids calling `__eq__()`)

tests/test_errors.py
```diff
@@ -196,7 +196,7 @@ def finalize(self, cfg: Any) -> None:
         Expected(
             create=lambda: OmegaConf.create({"foo": "${missing}"}),
             op=lambda cfg: getattr(cfg, "foo"),
-            exception_type=ConfigKeyError,
+            exception_type=ConfigAttributeError,
```
odelalleau, line 199: I wonder if missing interpolations should have their own exception? (that would neither be a KeyError nor an AttributeError)

Reason is that people may use exception-based programming with the new API, to detect missing keys:

```python
try:
    foo = cfg.foo
except AttributeError:
    foo = cfg.foo = 42
print(foo)
```
and if the `AttributeError` is raised because `foo` is referring to a missing interpolation, then bad things may happen.

odelalleau (commented):
A few comments

tests/test_basic_ops_dict.py
```diff
@@ -610,16 +617,16 @@ def test_creation_with_invalid_key() -> None:
         OmegaConf.create({object(): "a"})
 
 
-def test_set_with_invalid_key() -> None:
+def test_setitem_with_invalid_key() -> None:
     cfg = OmegaConf.create()
     with pytest.raises(KeyValidationError):
-        cfg[object()] = "a"  # type: ignore
+        cfg.__setitem__(object(), "a")  # type: ignore
```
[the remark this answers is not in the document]
odelalleau: Ah never mind, saw your other comments => resolving

omegaconf/listconfig.py
omry: _get_node is an internal API. As long as there is no external API that can expose an issue because of it I don't think it's particularly important.

omegaconf/dictconfig.py
```diff
@@ -427,7 +427,8 @@ def _get_node(
         self,
         key: DictKeyType,
         validate_access: bool = True,
-        throw_on_missing: bool = False,
+        throw_on_missing_value: bool = False,
+        throw_on_missing_key: bool = False,
```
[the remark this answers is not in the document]
omry: The way this code has evolved means things are a bit hacky now.
initially I tried to use validate_access for this but it did fail some other cases.
Overall we can clean things up but given that it's an internal API we can always do so later.

omegaconf/dictconfig.py
omry: fixed

tests/test_errors.py
omry: Why is that bad?
you accessed a field, it was pointing to something that is not there. you assign an actual value to it. It's no longer a pointer but a real value. underlying pointed to field is unchanged.

odelalleau: Someone writing such code most likely only intended to override `cfg.foo` if there was no such key in the config already. This could lead to subtle bugs if suddenly an interpolation gets broken (e.g., because a variable was renamed in the config), because instead of seeing an error the code will run just fine (ignoring the broken interpolation).

omry: Create a concrete example demonstrating a problem.
writing values on an interpolation is not writing through the interpolation.
Also, your example will not even work now because the exception raised is not an ConfigAttributeError but a ConfigKeyError right now:

```
cfg = OmegaConf.create({"a" : "${b}"})
cfg.a # ConfigKeyError
cfg["a"] # ConfigKeyError
```
and in any case writing on an interpolation is not write through:
```python
﻿In [11]: cfg
Out[11]: {'a': '${b}'}
In [12]: cfg.a = 10
In [13]: cfg
Out[13]: {'a': 10}
```
I am open to changes here but it's a different topic (and does not much much to do with standardizing with dict and object (given that they do not support interpolations).

odelalleau: Are you on master? I'm at d018678 and in your example `cfg.a` raises a `ConfigAttributeError` for me.

Here's an example illustrating what I have in mind: note how the typo in the interpolation makes it being silently ignored.

```python
cfg = OmegaConf.create("""
default_ram: 32
super_computer_ram: 128

machines: [m1, m2, m3, m4, m5]

ram_per_machine:
    m3: 64
    m5: ${super_computerr_ram}
""")

# Ensure `cfg.ram_per_machine` contains RAM data for all machines.
for machine in cfg.machines:
    ram_per_machine = cfg.ram_per_machine
    try:
        ram = getattr(ram_per_machine, machine)
    except AttributeError:
        ram = cfg.default_ram
        setattr(ram_per_machine, machine, cfg.default_ram)
    print(f"RAM for {machine}: {ram}")

print("\nConfig:\n")
print(OmegaConf.to_yaml(cfg, resolve=True))
```

It's not a huge deal, I just thought I'd point it out (it wasn't a concern before because there were no exceptions to catch)

omry: > Are you on master? 

nope, I was not. Exceptions are correct.

Looking at your example, I am not sure what you are trying to say.
1. What is the expected output?
2. Is the typo in the config intentional?

If the typo (`computerr`) is intentional, I don't think I care.
Your config is wrong, you are applying custom logic and it does not do what you wanted because your config is wrong.

Jasha10: >     1. What is the expected output?

If you fix the typo in the config, the output is:
```
RAM for m1: 32
RAM for m2: 32
RAM for m3: 64
RAM for m4: 32
RAM for m5: 128
```
>     2. Is the typo in the config intentional?

I think the idea is that the typo was not intentional. The `AttributeError` is being interpreted by this custom logic as meaning that `ram_per_computer.m5` was not set, but the reality is that `ram_per_computer.m5` *was* set, but there was an accidental typo in the `ram_per_computer.m5` setting (preventing the interpolation from being resolved).

Jasha10: > I wonder if missing interpolations should have their own exception? (that would neither be a KeyError nor an AttributeError)

So the idea is that each of the following would raise a different type of exception:
```
OmegaConf.create("").a  # raises ConfigAttributeError
OmegaConf.create("a: ${xyz}").a  # also raises ConfigAttributeError
```

odelalleau: @Jasha10 said it better than I could have (thank you!)

Again, I don't feel strongly about it, since it should be pretty rare (combination of exception-based programming + interpolation + user error) => just wanted to mention the idea of maybe using a different exeption, but up to you @omry

Jasha10: We already have
```class InterpolationResolutionError(OmegaConfBaseException, ValueError): ...```
defined in `omegaconf.errors`. Why no throw that one?

omry: Ha, looks like I added it when I added [relative interpolations](https://github.com/omry/omegaconf/commit/3c208f948da2569c475baba1aff70338ece8210b) but somehow never used it.

I am fine with using it to indicate an that an interpolation could not be resolved.

Jasha10: I've opened a PR (https://github.com/omry/omegaconf/pull/523) to use `InterpolationResolutionError` for this purpose.

PR #518, opened 2021-02-05
`to_pathdict()` serialization functionality

mivanit:
`OmegaConf.to_pathdict()` takes the same arguments as `OmegaConf.to_container()` (plus separator) 
and outputs a "flattened" dict. Useful for serializing when you can't have a nested hash map.
(Essentially the inverse operation of creating a config from a dotlist)

I'm not sure if the tests are in the right place, or if I should write more than just one.
Let me know what I need to change to get this accepted.

### DIFF
--- a/omegaconf/omegaconf.py
+++ b/omegaconf/omegaconf.py
@@ -528,13 +528,13 @@ class OmegaConf:
 
     @staticmethod
     def to_pathdict(
-            cfg: Any,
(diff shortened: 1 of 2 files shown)


omry:
Please file an issue so we can discuss the need before reviewing the code.

omry:
Thanks.
Closing because #520 is a cleaner implementation.

PR #523, opened 2021-02-07
Use interpolation resolution error

Jasha10:
Based on comment https://github.com/omry/omegaconf/pull/516#discussion_r571080800
we use a separate exception type `InterpolationResolutionError` to indicate failure of interpolation, instead of using `ConfigKeyError`.


omry (approved):
Nice, thanks!

omry:
Ohh, I forgot:
can you followup with adding a news fragment? this is an API change.
Normally we use issue numbers for the file name but in this case we can go with the PR number. Something like:

news/523.api_change 
```
Interpolation resolution errors now raises InterpolationResolutionError instead of ConfigKeyError
```

PR #525, opened 2021-02-07
Fix optional flag for interpolated nodes

### LINKED ISSUE #524: Can't assign None to optional field currently interpolated to a non-optional one
**Describe the bug**

The following code should work, but currently crashes with `ValidationError: child 'b' is not Optional`:

```python
@dataclass
class Config:
    a: int = 3
    b: Optional[int] = "${a}"
        
cfg.b = None
```

**To Reproduce**
See above

**Expected behavior**
No error

**Additional context**
This is on current master (8103a84)

odelalleau:
Fixes #524

### DIFF
--- a/news/524.bugfix
+++ b/news/524.bugfix
@@ -1,0 +1,1 @@
+Fix ValidationError when setting to None an optional field currently interpolated to a non-optional one
--- a/omegaconf/nodes.py
+++ b/omegaconf/nodes.py
@@ -93,9 +93,7 @@ class ValueNode(Node):
         return node._value() is None
 
     def _is_optional(self) -> bool:
-        node = self._dereference_node()
-        assert node is not None
-        return node._metadata.optional
+        return self._metadata.optional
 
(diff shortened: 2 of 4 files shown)


omry:
I think we can go with this, but a followup should be to validate_and_convert interpolation results to ensure that the type returned by a field is always compatible even if the field is an interpolation.

odelalleau:
> I think we can go with this, but a followup should be to validate_and_convert interpolation results to ensure that the type returned by a field is always compatible even if the field is an interpolation.

Yes, that's actually what I was working on when I came across this issue. I thought this one could be fixed right away, but I don't mind including this fix in a follow-up PR -- as you prefer.

omry:
> Yes, that's actually what I was working on when I came across this issue. I thought this one could be fixed right away, but I don't mind including this fix in a follow-up PR -- as you prefer.

merging this one.
Let's the do the followup for this separately from 455.

PR #530, opened 2021-02-09
Optimizing ListConfig.__contains__

### LINKED ISSUE #529: ListConfig.__contains__ significantly slower than list.__contains__

omry:
Speedup from 1200x slower than standard list contains to only 68 times slower (~15x).

Closes #529

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict
+from typing import Any, Dict, List
 
 from pytest import lazy_fixture  # type: ignore
 from pytest import fixture, mark, param
@@ -6,7 +6,7 @@ from pytest import fixture, mark, param
 from omegaconf import OmegaConf
 
 
-def build(
+def build_dict(
     d: Dict[str, Any], depth: int, width: int, leaf_value: Any = 1
 ) -> Dict[str, Any]:
     if depth == 0:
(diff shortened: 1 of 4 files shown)


omegaconf/listconfig.py
```diff
@@ -539,7 +539,12 @@ def __iadd__(self, other: Iterable[Any]) -> "ListConfig":
         return self
 
     def __contains__(self, item: Any) -> bool:
-        for x in iter(self):
+        if self._is_none() or self._is_missing():
+            return False
```
odelalleau, line 543: Looking at the previous code, I think it was crashing in this scenario. I believe this would be safer, no?

benchmark/benchmark.py
```diff
@@ -79,3 +93,26 @@ def test_omegaconf_create(data: Any, benchmark: Any) -> None:
...
+
+@mark.parametrize(
+    "lst",
+    [
+        lazy_fixture("small_list"),
+        lazy_fixture("small_listconfig"),
+    ],
+)
+def test_list_iter(lst: List[Any], benchmark: Any) -> None:
+    # Performance is really bad for list config iteration, not sure why.
```
odelalleau, line 117: Why is that surprising? If you compare to a regular list, ListConfig is doing a lot more stuff to check if any interpolation needs to be resolved, while a regular list can instantly return the reference to the object.

odelalleau (commented):
Just not sure about returning False for missing/None

omegaconf/listconfig.py
omry: Yes, you are right.
I actually didn't realize that. in any case, no one complained about the previous behavior (probably because you are not supposed to be doing it anyway).

benchmark/benchmark.py
omry: Because it's 500 times slower :).
I don't think it's justified. I tried to profile it a bit and do some blind optimizations but was not able to pinpoint the cause.

omegaconf/listconfig.py
omry: Oh, by "this" do you mean return False or raising an exception?

odelalleau: I meant that crashing seems safer to me (and noone should complain about a crash when trying to iterate on None / missing)

benchmark/benchmark.py
odelalleau: Ok -- actually given the speed-up that you got here, I do agree that it should be possible to iterate faster (`x in lst` shouldn't be much faster than `for y in lst` -- at least when iterating through the whole list in both cases)

omegaconf/listconfig.py
omry: yeah, I agree.
fixed it back.

odelalleau (approved):
Looks good!

PR #542, opened 2021-02-12
Add support for dots in resolver names (for namespaces)

### LINKED ISSUE #539: Support dots in resolver names for namespaces

Introducing new resolvers in libraries may clash with user-defined custom resolvers.

Supporting dots in resolver names to denote namespaces would avoid such clashes, e.g., `env` could become `oc.env` (although this specific change to the env resolver is not part of this issue).

Nested namespaces will also be supported (ex: `${ns1.ns2.func:123}`).

Also considered `_`.

**Additional context**

Related to discussion in https://github.com/omry/omegaconf/issues/383#issuecomment-774608613

odelalleau:
Fixes #539


omegaconf/grammar/OmegaConfGrammarParser.g4
```diff
@@ -48,8 +48,9 @@ sequence: element (COMMA element)*;
 
 interpolation: interpolationNode | interpolationResolver;
 interpolationNode: INTER_OPEN DOT* configKey (DOT configKey)* INTER_CLOSE;
-interpolationResolver: INTER_OPEN (interpolation | ID) COLON sequence? BRACE_CLOSE;
+interpolationResolver: INTER_OPEN resolverName COLON sequence? BRACE_CLOSE;
 configKey: interpolation | ID | INTER_KEY;
+resolverName: (interpolation | ID) (DOT (interpolation | ID))* ;  // oc.env, myfunc, ns.${x}, ns1.ns2.f
```
omry, line 53: I had no idea we supported interpolations as function names here.
```
oc.env, myfunc, ns.${x}, ns1.ns2.f
```

```python
OmegaConf.register_resolver("aaa", lambda _ : "aaa" + _)
OmegaConf.register_resolver("bbb", lambda _ : "bbb" + _)
cfg = OmegaConf.create({"a": "${${b}:10}", "b": "aaa"})

cfg.a        # -> 'aaa10'

cfg.b = "bbb"

cfg.a        # -> 'bbb10'
```

Is this documented?

news/539.feature
```diff
@@ -0,0 +1 @@
+Dots may now be used in resolver names to denote namespaces (ex: `${ns1.ns2.myfunc:123}`)
```
omry, line 1:
```suggestion
Dots may now be used in resolver names to denote namespaces (e.g: `${namespace.my_func:123}`)
```

omegaconf/grammar_visitor.py
```diff
@@ -168,37 +168,22 @@ def visitInterpolationNode(
...
-                    args_str.append(txt)
-            else:
-                assert isinstance(child, TerminalNode)
+        if isinstance(maybe_seq, TerminalNode):  # means there are no args
+            assert maybe_seq.symbol.type == OmegaConfGrammarLexer.BRACE_CLOSE
+        else:
+            assert isinstance(maybe_seq, OmegaConfGrammarParser.SequenceContext)
+            for val, txt in self.visitSequence(maybe_seq):
+                args.append(val)
+                args_str.append(txt)
```
omry, line 185: can you explain the change here?

tests/test_grammar.py
```diff
@@ -221,6 +221,8 @@
     ("ws_inter_res_around_colon", "${test\t  : \tfoo}", "foo"),
     ("ws_inter_res_inside_id", "${te st:foo}", GrammarParseError),
     ("ws_inter_res_inside_args", "${test:f o o}", "f o o"),
+    ("ws_inter_res_namespace", "${ns1 .\t ns2 . test:0}", 0),
+    ("ws_inter_res_no_args", "${test: \t}", []),
```
omry, line 225: Why are whitespaces inside the function name legal?

omegaconf/grammar/OmegaConfGrammarParser.g4
odelalleau: > Is this documented?

No. It's kind of natural, but I guess it should be :)

news/539.feature
odelalleau: Taking your suggestion

tests/test_grammar.py
odelalleau: I thought it made sense because we allow whitespaces in node interpolations (`${foo . bar}`). But I don't mind either way.

omegaconf/grammar_visitor.py
odelalleau: In the previous version, `visitInterpolationResolver()` was computing everything: the resolver name (which could be either an `ID` or an interpolation), then the sequence of arguments.

In the new version, to make the grammar more readable, I moved the parsing of the resolver name to a new parser rule, and the first two `if`/`elif` statements have thus been moved to a new visitor method `visitResolverName()`. The only part that's left in `visitInterpolationResolver()` is the computation of the list of arguments.

omegaconf/grammar/OmegaConfGrammarParser.g4
odelalleau: Also, I wasn't sure if it was worth documenting the namespace feature (since it will become obvious when `oc.env` is introduced), but I think I'll create a new doc entry that shows both namespace + interpolation in the resolver name.

odelalleau: > I'll create a new doc entry that shows both namespace + interpolation in the resolver name.

Done in 1d82ae8

docs/source/usage.rst
```diff
@@ -457,6 +458,19 @@ You can take advantage of nested interpolations to perform custom operations ove
...
 
+
+More advanced resolver naming features include the ability to prefix a resolver name with a
+namespace, and to use interpolations in the name itself. The following example demonstrates both:
+
+.. doctest::
+
+    >>> OmegaConf.register_new_resolver("mylib.plus1", lambda x: x + 1)
+    >>> c = OmegaConf.create({"func": "plus1",
+    ...                       "x": "${mylib.${func}:3}"})
```
omry, line 469: Use Black to determine formatting of code snippets in the docs:

```python
c = OmegaConf.create({"func": "plus1", "x": "${mylib.${func}:3}"})
```
or
```python
c = OmegaConf.create(
    {
        "func": "plus1",
        "x": "${mylib.${func}:3}",
    }
)

```

I prefer the second. your call.

news/445.feature.1
```diff
@@ -1 +1 @@
-Add ability to nest interpolations, e.g. ${env:{$var}} or ${foo.${bar}.${baz}}
+Add ability to nest interpolations, e.g. ${foo.${bar}}}, ${env:{$var1},${var2}}, or ${${func}:}
```
omry, line 1:
```suggestion
Add ability to nest interpolations, e.g. ${foo.${bar}}}, ${env:{$var1},${var2}}, or ${${func}:x1,x2}
```

omegaconf/grammar/OmegaConfGrammarParser.g4
omry: I am considering this on the verge of too powerful, but willing to go with it because it's in already.
Do you have an actual use case?

tests/test_grammar.py
omry: I think both should be illegal.
Whitespace in interpolation keys is not supported in OmegaConf 2.0:

```python
In [9]:  OmegaConf.create({"x" : {"a": 10, "b": "${x. a}"}}).x.b
...
ConfigKeyError: str interpolation key 'x. a' not found
        full_key: x.b
        reference_type=Any
        object_type=dict
```

This can be addressed in a followup diff to keep things manageable :)

docs/source/usage.rst
odelalleau: Done in 18ec4cb (went for second one because first one went slightly over the doc width limit, forcing scrolling)

news/445.feature.1
odelalleau: Taking your suggestion.

omegaconf/grammar/OmegaConfGrammarParser.g4
odelalleau: > Do you have an actual use case?

It's not critical since you can usually replace `${${func}:x}` with `${master:${func},x}` where `master` is a high-level resolver that dispatches calls to other functions based on its first argument. The advantage of a config-based version would be to not have to maintain this high-level resolver (also, it is easier if the resolvers are spread around the codebase and it's hard to access them from a single point, but that's probably a rare case).

Generally speaking, I can see use cases for dynamically choosing a different resolver, when there are multiple ways to compute a quantity and you want to be able to easily switch between them.

tests/test_grammar.py
odelalleau: I think this dates back from the old Hydra grammar where whitespaces were systematically ignored in the lexer.
But I don't mind removing that -- preferably in a different PR indeed since this also touches non resolvers.

omry: Yup, let's clean this up.
Hydra should also not accept it as valid.

PR #544, opened 2021-02-12
Update notebook to synch with the docs

odelalleau:
The main purpose is to get (almost) the same content in the docs and in the notebook.

There is also a minor change to the doc on node interpolation, to make
it clearer that we can use them as config objects (they are not plain
Python dictionaries).

### DIFF
--- a/docs/notebook/Tutorial.ipynb
+++ b/docs/notebook/Tutorial.ipynb
@@ -503,13 +503,23 @@
      "name": "#%%\n"
     }
    },
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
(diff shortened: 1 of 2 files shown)


omry:
it's hard to review in the current form, but I didn't see a change removing the pytest usage.
pytest is not a runtime dependency of OmegaConf so it's not available in the binder by default. we should avoid using it in the example.

odelalleau:
> it's hard to review in the current form, but I didn't see a change removing the pytest usage.

Right, this first PR was just to synch the docs. But I can include the pytest fix in it as well, will be done in a minute.

omry:
no rush.

odelalleau:
Fixed the pytest / binder issue in 37e96d3

PR #546, opened 2021-02-16
documenting relative interpolations

### LINKED ISSUE #421: Document relative interpolation
#48

omry:
closes #421

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -296,6 +296,10 @@ Variable interpolation
 ----------------------
 
 OmegaConf supports variable interpolation. Interpolations are evaluated lazily on access.
+Interpolations are absolute by default.
+Relative interpolation are prefixed by one or more dots:
+The first dot denotes the level of the node itself and additional dots are going up the parent hierarchy.
+e.g. ..foo goes to the foo node of the parent of the containing node.
 
 Config node interpolation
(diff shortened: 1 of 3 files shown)


docs/source/usage.rst
```diff
@@ -296,6 +296,10 @@ Variable interpolation
 ----------------------
 
 OmegaConf supports variable interpolation. Interpolations are evaluated lazily on access.
+Interpolations are absolute by default.
+Relative interpolation are prefixed by one or more dots:
+The first dot denotes the level of the node itself and additional dots are going up the parent hierarchy.
+e.g. ..foo goes to the foo node of the parent of the containing node.
```
odelalleau, line 302: * Should this be under `Config node interpolation` below instead? (since this prefix doesn't apply to custom interpolations)
* The formatting / case could be improved to make it look better when compiled to HTML.
* "the foo node of the parent of the containing node" => a bit confusing (at least to me), would that make sense: "the `foo` sibling of the parent of the current node"?

odelalleau (changes requested):

docs/source/usage.rst
omry: formatted. looks like this now:

image

odelalleau (approved):
All good, thanks! :)

PR #549, opened 2021-02-17
enum_to_str: convert enum keys, not just enum values

### LINKED ISSUE #531: OmegaConf.to_yaml does not convert Enum DictConfig keys to string
```python3
from enum import Enum

class Color(Enum):
    RED = 1
    BLUE = 2

conf = OmegaConf.create({Color.RED: Color.BLUE})

print(OmegaConf.to_yaml(conf))
```
Outputs:
```
? !!python/object/apply:__main__.Color
- 1
: BLUE
```
Expected output:
```
RED: BLUE
```

Note that the DictConfig value `Color.BLUE` *is* converted to a string, but the key `Color.RED` is not.

### LINKED ISSUE #550: Behavior change: to_container(..., enum_to_str=True) to convert both keys and values
```python
$ cat tmp.py
from enum import Enum


class Color(Enum):
    RED = 1
    BLUE = 2


conf = OmegaConf.create({Color.RED: Color.BLUE})
print(OmegaConf.to_container(conf, enum_to_str=True))

$ python tmp.py
{<Color.RED: 1>: 'BLUE'}
```

Desired output is for both key and value to be converted when `enum_to_str` is set to True:
```
{'RED': 'BLUE'}
```

Jasha10:
Closes #531
Closes #550

### DIFF
--- a/tests/test_to_container.py
+++ b/tests/test_to_container.py
@@ -200,7 +200,7 @@ class TestEnumToStr:
     ) -> None:
         cfg = OmegaConf.create(src)
         container: Dict[Any, Any] = OmegaConf.to_container(cfg, enum_to_str=enum_to_str)  # type: ignore
-        value = list(container.values())[0]
+        value = container["enum val"]
         assert value == expected
         assert type(value) == type(expected)
 
--- a/tests/test_to_container.py
+++ b/tests/test_to_container.py
@@ -1,25 +1,25 @@
 import re
 from enum import Enum
-from typing import Any, Dict
+from typing import Any, Dict, List
 
-import pytest
+from pytest import mark, param, raises
 
 from omegaconf import DictConfig, ListConfig, OmegaConf
 from tests import Color, User
 
 
-@pytest.mark.parametrize(
+@mark.parametrize(
     "input_",
     [
-        pytest.param([1, 2, 3], id="list"),
-        pytest.param([1, 2, {"a": 3}], id="dict_in_list"),
-        pytest.param([1, 2, [10, 20]], id="list_in_list"),
-        pytest.param({"b": {"b": 10}}, id="dict_in_dict"),
-        pytest.param({"b": [False, 1, "2", 3.0, Color.RED]}, id="list_in_dict"),
-        pytest.param({"b": DictConfig(content=None)}, id="none_dictconfig"),
-        pytest.param({"b": ListConfig(content=None)}, id="none_listconfig"),
-        pytest.param({"b": DictConfig(content="???")}, id="missing_dictconfig"),
-        pytest.param({"b": ListConfig(content="???")}, id="missing_listconfig"),
+        param([1, 2, 3], id="list"),
+        param([1, 2, {"a": 3}], id="dict_in_list"),
+        param([1, 2, [10, 20]], id="list_in_list"),
+        param({"b": {"b": 10}}, id="dict_in_dict"),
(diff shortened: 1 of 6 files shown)


Jasha10:
I created some tests for the `enum_to_str` argument to `to_container`.
Previously there were no tests for this.

tests/test_to_container.py
```diff
@@ -167,3 +167,121 @@ def test_string_interpolation_with_readonly_parent() -> None:
...
+                id="int_key:F-unaffected",
+            ),
+        ],
+    )
+    def test_enum_to_str_for_keys(
+        self, src: DictConfig, enum_to_str: bool, expected: Any
+    ) -> None:
+        container = OmegaConf.to_container(src, enum_to_str=enum_to_str)
+        assert isinstance(container, dict)
+        key = list(container.keys())[0]
```
omry, line 209: If you use the same key name in all inputs you would be able to use it directly here.

tests/test_to_container.py
```diff
@@ -167,3 +167,121 @@ def test_string_interpolation_with_readonly_parent() -> None:
 def test_to_container_missing_inter_no_resolve(src: Any, expected: Any) -> None:
     res = OmegaConf.to_container(src, resolve=False)
     assert res == expected
+
```
omry, line 170: That's a bit too many tests.
Let's try to keep tests high value, a lot of test code becomes a maintenance burden.

I think we can go with a single test function testing:
1. with and without enum_to_str
2. A single config containing dict and list, enums as keys and enums as values.

Maybe:
```python
cfg = {
  "d": {
    Color.RED: "red",
    "red" : Color.RED,   
  },
  "l" : ["red", Color.RED]
}
```

Testing conversion of a config derived from this, with the flag true and false should give solid coverage and will be much more compact.

Jasha10: Done in https://github.com/omry/omegaconf/pull/549/commits/812ce8018e466308ef80d6ad3ab7812971c86da6

tests/test_to_container.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/549/commits/910e4d9df9acfee9738ab3226199b855936a13d0.

tests/test_to_container.py
```diff
@@ -167,3 +167,55 @@ def test_string_interpolation_with_readonly_parent() -> None:
...
+        ],
+    )
+    def test_enum_to_str_for_list(
+        self, src: Any, enum_to_str: bool, expected: Any
+    ) -> None:
+        cfg = OmegaConf.create(src)
+        container: List[Any] = OmegaConf.to_container(cfg, enum_to_str=enum_to_str)  # type: ignore
+        value = container[0]
+        assert value == expected
+        assert type(value) == type(expected)
```
omry, line 221: ditto.

omry (commented):
looking good. a few nits.

omry:
Oh, I thought about #550 a bit:
Every bugfix is a behavior change, I don't think this one justifies an API change entry. (I doubt anyone have actually ran into it).

Jasha10:
> I don't think this one justifies an API change entry.

Ok, I'll scrap the #550 news fragment.

tests/test_to_container.py
Jasha10, line 221:
```suggestion
        assert container == [expected]
```

omry (approved):
Great, thanks.

Issue #543, opened 2021-02-12
Better and more consistent handling of the "missing" status


odelalleau:
**Describe the bug**

Currently the handling of a missing node in interpolations can lead to surprising behavior like:
```python
cfg = OmegaConf.create({"x": "${missing}", "y": "abc ${missing}", "missing": "???"})
print(f'{cfg._get_node("x")._is_missing() = }')
print(f'{cfg._get_node("x")._dereference_node() = }')
print(f'{cfg._get_node("y")._is_missing() = }')
print(f'{cfg._get_node("y")._dereference_node() = }')
```
which prints:
```
cfg._get_node("x")._is_missing() = True
cfg._get_node("x")._dereference_node() = '???'
cfg._get_node("y")._is_missing() = False
cfg._get_node("y")._dereference_node() = 'abc ???'
```

The two potentially surprising things are:
1. The fact that `x` is considered as missing while `y` isn't (while both refer to a missing node)
2. The fact that `y` is happily dereferenced into a `StringNode` containing the string `"abc ???"`

**Expected behavior**

I propose the following behavior:
1. A node is *never* missing if it is an interpolation (even if it interpolates into a missing node). This is consistent with the spirit of #462, where we want "missing" to mean only being set to "???"
2. Dereferencing a node containing an interpolation to a missing node should return `None` (unless ~~`throw_on_missing`~~ `throw_on_resolution_failure` is True, in which case it should raise the `InterpolationToMissingValueError` exception)

(Edit: I replaced `throw_on_missing` with `throw_on_resolution_failure` in point 2 because `throw_on_missing` when dereferencing a node should only refer to the status of the node being referenced, not to what happens when trying to resolve interpolations, while `throw_on_resolution_failure` relates to errors encountered when resolving interpolations)

omry:
The only surprise to me is `cfg._get_node("y")._dereference_node() = 'abc ???'`.
This should raise an exception while resolving the string interpolation.

This behavior is driving the __contains__ API behavior for missing fields:

```
"missing" in cfg`
````
which returns False.
In a previous version string interpolation using a missing field was also considered missing, but I changed that at some point.

I am open to changing this behavior but we should consider it carefully and check in downstream apps (at least in Hydra).

odelalleau:
> In a previous version string interpolation using a missing field was also considered missing, but I changed that at some point.

[...]

PR #553, opened 2021-02-21
Workaround for Dataframe object in the config: check whether value is string before mandatory missing check

### LINKED ISSUE #566: Illegal Comparisons when handling pandas DataFrame objects
**Describe the bug**
In several places, the code attempts to check if a node is missing with a comparison + a conditional (eg `x = "a" if node == "???" else "b"`). This conditional breaks if the node value is a pandas DataFrame, as DataFrames implement additional `__bool__` logic that prevents such a conditional (see: https://github.com/pandas-dev/pandas/blob/v1.2.2/pandas/core/generic.py#L1440-L1446).

**To Reproduce**
creating a DictConfig:
```
from pandas import DataFrame

kwargs = OmegaConf.create({"a": DataFrame()}, flags={"allow_objects": True})
```

when attempting to set an attribute of an existing config:
```
from pandas import DataFrame

c = OmegaConf.create()
c._set_flag("allow_objects", True)
c.a = DataFrame()
```

**Expected behavior**
when running creating DictConfig or setting a new attr, there should simply be an update to the config where the DF object is now included (assuming allow_object flag is set to True).

**Additional context**
- [X] OmegaConf version: master branch
- [X] Python version: 3.7.6
- [X] Operating system : ubuntu

jeffknaide:
Closes #566 

sidestep source of https://github.com/facebookresearch/hydra/issues/1285 at the omegaconf level

### DIFF
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -328,10 +328,6 @@ def get_structured_config_data(
         raise ValueError(f"Unsupported type: {type(obj).__name__}")
 
 
-def is_mandatory_missing(value: Any) -> bool:
-    return isinstance(value, str) and value == "???"
-
-
 class ValueKind(Enum):
     VALUE = 0
     MANDATORY_MISSING = 1
@@ -357,7 +353,7 @@ def get_value_kind(
 
     value = _get_value(value)
 
-    if is_mandatory_missing(value):
+    if isinstance(value, str) and value == "???":
         return ValueKind.MANDATORY_MISSING
 
     # We identify potential interpolations by the presence of "${" in the string.
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -353,6 +353,9 @@ def get_value_kind(
 
     value = _get_value(value)
 
+    if not isinstance(value, str):
+        return ValueKind.VALUE
+
     if value == "???":
         return ValueKind.MANDATORY_MISSING
 
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -328,6 +328,10 @@ def get_structured_config_data(
         raise ValueError(f"Unsupported type: {type(obj).__name__}")
 
 
+def is_mandatory_missing(value: Any) -> bool:
+    return isinstance(value, str) and value == "???"
+
+
 class ValueKind(Enum):
     VALUE = 0
     MANDATORY_MISSING = 1
@@ -353,7 +357,7 @@ def get_value_kind(
 
     value = _get_value(value)
 
-    if isinstance(value, str) and value == "???":
+    if is_mandatory_missing(value):
         return ValueKind.MANDATORY_MISSING
 
     # We identify potential interpolations by the presence of "${" in the string.
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -353,10 +353,7 @@ def get_value_kind(
 
     value = _get_value(value)
 
-    if not isinstance(value, str):
-        return ValueKind.VALUE
-
-    if value == "???":
+    if isinstance(value, str) and value == "???":
         return ValueKind.MANDATORY_MISSING
 
     # We identify potential interpolations by the presence of "${" in the string.
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -9,17 +9,18 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 import yaml
 
 from ._utils import (
+    ValueKind,
     _ensure_container,
     _get_value,
     _is_interpolation,
     _resolve_optional,
     get_ref_type,
     get_structured_config_data,
+    get_value_kind,
     get_yaml_loader,
     is_container_annotation,
     is_dict_annotation,
     is_list_annotation,
-    is_mandatory_missing,
     is_primitive_dict,
     is_primitive_type,
     is_structured_config,
@@ -48,6 +49,9 @@ class BaseContainer(Container, ABC):
         default_value: Any = DEFAULT_VALUE_MARKER,
     ) -> Any:
         """returns the value with the specified key, like obj.key and obj['key']"""
+
+        def is_mandatory_missing(val: Any) -> bool:
+            return bool(get_value_kind(val) == ValueKind.MANDATORY_MISSING)
 
         val = _get_value(value)
         has_default = default_value is not DEFAULT_VALUE_MARKER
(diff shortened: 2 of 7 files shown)


omegaconf/basecontainer.py
```diff
@@ -784,7 +784,7 @@ def _create_structured_with_missing_fields(
 def _is_missing_value(value: Any) -> bool:
     if isinstance(value, Container):
         value = value._value()
-    ret = value == "???"
+    ret = value == "???" if isinstance(value, str) else False
```
Jasha10, line 787: So the idea is to sidestep the custom `DataFrame.__eq__` method by first deciding if `value` is a `str`, then only calling `value.__eq__("???")` if we are sure that `value` is a `str`.

jeffknaide: yep, exactly.

omry:
Thanks @jeffknaide.
We have some outstanding discussions around the internal behavior of MISSING, we will address this once we comes to a conclusion there.

omry:
Blocked by #543.

omegaconf/_utils.py
```diff
@@ -353,6 +353,9 @@ def get_value_kind(
 
     value = _get_value(value)
 
+    if not isinstance(value, str):
+        return ValueKind.VALUE
+
```
omry, line 358: Change the following check to:
```
    if isinstance(value,str) and value == "???":
        return ValueKind.MANDATORY_MISSING
```
instead.

omegaconf/basecontainer.py
omry, line 787:
```suggestion
    ret = isinstance(value, str) and value == "???"
```

tests/__init__.py
```diff
@@ -34,6 +34,17 @@ def __deepcopy__(self, memo: Any) -> Any:
         raise NotImplementedError()
 
 
+class Dataframe:
+    def __init__(self) -> None:
+        pass
+
+    def __eq__(self, other: Any) -> Any:
+        return self
```
omry, line 42: I find `__eq__` returning self very odd. what's going on here?

tests/__init__.py
```diff
@@ -34,6 +34,17 @@ def __deepcopy__(self, memo: Any) -> Any:
         raise NotImplementedError()
 
 
+class Dataframe:
```
omry, line 37: Add a comment here.
something like:

```
Emulates Pandas Dataframe equality behavior.
```

tests/__init__.py
```diff
@@ -34,6 +34,17 @@ def __deepcopy__(self, memo: Any) -> Any:
...
+    def __init__(self) -> None:
+        pass
+
+    def __eq__(self, other: Any) -> Any:
+        return self
+
+    def __bool__(self) -> None:
+        raise ValueError
```
omry, line 45: add a comment explaining what this achieves.

tests/test_merge.py
```diff
@@ -599,6 +600,14 @@ def test_merge_allow_objects(merge: Any) -> None:
     assert ret == {"a": 10, "foo": iv}
 
 
+def test_merge_with_allow_Dataframe() -> None:
+    kwargs = OmegaConf.create({"a": Dataframe()}, flags={"allow_objects": True})
+    cfg = OmegaConf.create()
+
+    cfg.merge_with(kwargs)
+    assert isinstance(cfg.a, Dataframe)
```
omry, line 608: If this still reproduces the problem, please switch to it:
```suggestion
def test_merge_with_allow_Dataframe() -> None:
    cfg = OmegaConf.create({"a": Dataframe()}, flags={"allow_objects": True})
    ret = OmegaConf.merge({}, cfg)
    assert isinstance(ret.a, Dataframe)
```

tests/test_nodes.py
```diff
@@ -559,3 +559,10 @@ def test_allow_objects() -> None:
     iv = IllegalType()
     c.foo = iv
     assert c.foo == iv
+
+
+def test_allow_Dataframe_objects() -> None:
+    c = OmegaConf.create()
+    c._set_flag("allow_objects", True)
+    c.a = Dataframe()
+    assert isinstance(c.a, Dataframe)
```
omry, line 568: allow_objects is already tested, remove this test.

omry (changes requested):
Thanks @jeffknaide.
See inline comments.

tests/__init__.py
jeffknaide: when a pandas DataFrame runs `df.__eq__(val)`, it returns a DataFrame of booleans with the same shape as the original DataFrame. Here I was trying to mimic the behavior of a df comparison resulting in the same class being returned, as this is partially what caused issues originally. perhaps this should be made clearer with a comment?

omry: Gotcha.
Yes, add a comment what the behavior we are mimicking is.

jeffknaide: i don't think so -- if we don't override eq, the comparison to string would return a bool and then not raise the expected error on the conditional. eg `if Dataframe() == "???"` would equate to `if False` (which wouldn't raise the expected error) without the eq override.

tests/__init__.py
jeffknaide: let me know if the comments I added here work

omry:
lgtm.
@odelalleau, this could interact with your outstanding PR to change the semantics of missing.

odelalleau:
> lgtm.
> @odelalleau, this could interact with your outstanding PR to change the semantics of missing.

At first glance I don't foresee any issue (*crossing fingers*). Just mentioning that maybe there could be a new utility function to avoid repeating `if isinstance(x, str) and x == "???"` everywhere?

omry:
Yeah, was thinking that too.

jeffknaide:
@odelalleau @omry oh, yes, that makes sense. I just added `is_mandatory_missing` to `_utils` and referenced it where i had been using `if isinstance(x, str) and x == "???"`.

There was also an inner function defined in `BaseContainer._resolve_with_default` called `is_mandatory_missing` which was returned of `get_value_kind`. I just removed that inner function in favor of `is_mandatory_missing` defined in _utils. This seems to break the tests, though, so I must be missing something that's happening within `get_value_kind`

odelalleau:
> I must be missing something that's happening within `get_value_kind`

It’s almost certainly because that function can be called on a node: it starts with `_get_value(..)` to obtain the node’s value if that’s the case. So in BaseContainer the problem is likely related to ˋ if is_mandatory_missing(resolved):`, and replacing ˋresolved` with its value here should fix it.

omry:
I think we will want to take a closer look at this in a followup.
can you undo the refactoring changes?

odelalleau:
> can you undo the refactoring changes?

Most could be kept, only the change in `_resolve_with_default()` needs reverting

omry:
Maybe. it just feel like there are a bit too many layers here. The logic can be slightly reorganized to be simpler.

jeffknaide:
@omry - i just reverted previous commit. let me know if you want me to put the utils function back while leaving `_resolve_with_default ` alone.

omry (approved):
Thanks @jeffknaide.

PR #557, opened 2021-02-22
Error message improvement for OmegaConf.structured creation

### LINKED ISSUE #435: Unclear error: incorrect type for field of dataclass instance
**Describe the bug**
When using the incorrect value type for a field in a dataclass instance and then using that to construct a config object the error message is quite unclear.

**To Reproduce**
```python3
import omegaconf
import dataclasses
@dataclasses.dataclass
class Config:
    some_field: int = 3

omegaconf.OmegaConf.create(Config(some_field='test'))
```
gives the error
```text
omegaconf.errors.ValidationError: Value 'None' could not be converted to Integer
	full_key: 
	reference_type=None
	object_type=None
```

**Expected behavior**
I would love to have the same behavior as when running
```python3
omegaconf.OmegaConf.merge(Config, {'some_field':'test'})
```
which gives
```text
omegaconf.errors.ValidationError: Value 'test' could not be converted to Integer
	full_key: some_field
	reference_type=Optional[Config]
	object_type=Config
```

**Additional context**
- [ ] OmegaConf version: 2.0.2
- [ ] Python version: 3.8.3
- [ ] Operating system : Ubuntu 18.04

This issue is not important, it was just something I discovered while testing out error messages for a framework.

Jasha10:


omegaconf/omegaconf.py
```diff
@@ -880,67 +880,81 @@ def _node_wrap(
...
         else:
-            raise ValidationError(f"Unexpected object type : {type_str(type_)}")
+            if parent is not None and parent._get_flag("allow_objects") is True:
+                node = AnyNode(
+                    value=value, key=key, parent=parent, is_optional=is_optional
+                )
+            else:
+                raise ValidationError(f"Unexpected object type : {type_str(type_)}")
+    except ValidationError as e:
+        format_and_raise(node=parent, key=key, value=value, msg=str(e), cause=e)
```
Jasha10, line 957: The main change of this PR is to wrap the body of `OmegaConf._node_wrap` in a try/catch block, so that more information can be given to the `format_and_raise` function. The goal is to close #435.

I wasn't sure if passing `node=parent` to `format_and_raise` above is the right decision. I tried passing `node=None`, which resulted in a lot more errors.

tests/test_errors.py
```diff
@@ -292,6 +292,19 @@ def finalize(self, cfg: Any) -> None:
...
+            create=lambda: None,
+            op=lambda cfg: OmegaConf.structured(User(age="foo")),  # type: ignore
+            exception_type=ValidationError,
+            msg="Value 'foo' could not be converted to Integer",
+            object_type=dict,
+            key="age",
+            low_level=True,
+        ),
+        id="structured:creation-with-incorrect-key-type",
+    ),
```
Jasha10, line 307: This is the test case motivated by #435.
The `object_type=dict` comes from passing `node=parent` to the `format_and_raise` function, I think.

tests/test_errors.py
```diff
@@ -477,8 +490,9 @@ def finalize(self, cfg: Any) -> None:
             op=lambda cfg: OmegaConf.structured(NotOptionalInt),
             exception_type=ValidationError,
             msg="Non optional field cannot be assigned None",
-            object_type_str=None,
-            ref_type_str=None,
+            object_type=dict,
+            key="foo",
+            low_level=True,
```
Jasha10, line 495: Made this change to get the tests to pass. I will review to see if there is a better way to achieve the improved error message, without having to change the above test case.

tests/test_errors.py
```diff
@@ -487,10 +501,10 @@ def finalize(self, cfg: Any) -> None:
             create=lambda: None,
             op=lambda cfg: OmegaConf.structured(NotOptionalInt),
             exception_type=ValidationError,
-            object_type=None,
             msg="Non optional field cannot be assigned None",
-            object_type_str="NotOptionalInt",
-            ref_type_str=None,
+            object_type=dict,
+            key="foo",
+            low_level=True,
```
Jasha10, line 507: ditto.

omegaconf/omegaconf.py
Jasha10: After further review recalled that, in the case of structured config creation, the `parent` is a `dummy_parent` created by the `get_dataclass_data` function.

PR #558, opened 2021-02-23
Refactor structured config tests

Jasha10:
This PR does not make any major changes -- it is just a refactoring of the `test_structured_basic.py` and `test_structured_config.py` files.

- Replace `import pytest` with `from pytest import fixture, mark, param, raises`
- Factor out a fixture for importing from the correct module:
```python
@fixture(
    params=[
        "tests.structured_conf.data.dataclasses",
        "tests.structured_conf.data.attr_classes",
    ],
    ]
)
def module(request: Any) -> Any:
    return import_module(request.param)
```

### DIFF
--- a/tests/structured_conf/test_structured_basic.py
+++ b/tests/structured_conf/test_structured_basic.py
@@ -2,7 +2,7 @@ import re
 from importlib import import_module
 from typing import Any, Optional
 
-from pytest import mark, raises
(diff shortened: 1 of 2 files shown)


Jasha10:
marking as a draft because there are some merge conflicts with `master`

omry (approved):
Neat, thanks!

PR #559, opened 2021-02-23
Improve error message when initializing from a Structured Config with incorrect type

### LINKED ISSUE #435: Unclear error: incorrect type for field of dataclass instance

Jasha10:
Closes #435.
Based on http://paste.ubuntu.com/p/d7zKBpG9rg/


Jasha10:
Marking as a draft because there are some merge conflicts with 'master'

tests/test_errors.py
```diff
@@ -532,30 +532,31 @@ def finalize(self, cfg: Any) -> None:
...
             create=lambda: None,
-            op=lambda cfg: OmegaConf.structured(NotOptionalInt),
+            op=lambda _: OmegaConf.structured(NotOptionalInt),
             exception_type=ValidationError,
             msg="Non optional field cannot be assigned None",
-            object_type_str=None,
-            ref_type_str=None,
+            key="foo",
+            object_type=NotOptionalInt,
+            parent_node=lambda _: {},  # dummy parent
```
omry, line 540: why do you need to add those parent_node fields?

Jasha10: This is because `OmegaConf.structured(...)` creates a `dummy_parent` node (in `_utils.py:get_dataclass_data` or `get_attr_data`).

Lower down in this test file we have
```python
assert ex.parent_node == expected.parent_node(cfg)
```
which results in
```
AssertionError: assert {} == None
```
unless if the parent_node field is set.

Jasha10: It turns out that only two of parent_node fields were strictly necessary; I reverted/removed two of them in https://github.com/omry/omegaconf/pull/559/commits/3f40dbb0db50416ce08d659d41c1572ba31fd718.

omry:
Can you add a test when the error is in a deeper level to see the resulting error message is reasonable?

maybe something like:
```python
            create=lambda: OmegaConf.structured(
                ConcretePlugin(params=ConcretePlugin.FoobarParams(foo="x")) # type: ignore
            ),
```

tests/test_errors.py
```diff
@@ -580,7 +581,21 @@ def finalize(self, cfg: Any) -> None:
...
+            op=lambda _: OmegaConf.structured(
+                ConcretePlugin(params=ConcretePlugin.FoobarParams(foo="x"))  # type: ignore
+            ),
+            exception_type=ValidationError,
+            msg="Value 'x' could not be converted to Integer",
+            object_type=ConcretePlugin.FoobarParams,
+            key="foo",
+            parent_node=lambda _: {},  # dummy parent
+        ),
+        id="structured:create_with_invalid_value",
```
Jasha10, line 593: I've added the test case that you mentioned.
If I write `full_key="params.foo",` this results in `AssertionError: assert 'foo' == 'params.foo'`.
I think this is related to the use of a "dummy parent" in the `get_dataclass_data`/`get_attr_data` methods.

omegaconf/_utils.py
```diff
@@ -217,13 +220,16 @@ def get_attr_data(obj: Any, allow_objects: Optional[bool] = None) -> Dict[str, A
...
+        try:
+            d[name] = _maybe_wrap(
+                ref_type=type_,
+                is_optional=is_optional,
+                key=name,
+                value=value,
+                parent=dummy_parent,
+            )
+        except ValidationError as ex:
+            dummy_parent._format_and_raise(key=name, value=value, cause=ex)
```
omry, line 232: can you try to use `format_and_raise()` directly instead of through the dummy parent?
it can probably make things a bit cleaner when testing this.

tests/test_errors.py
```diff
@@ -532,30 +532,31 @@ def finalize(self, cfg: Any) -> None:
...
+            key="foo",
+            parent_node=lambda _: {},  # dummy parent
         ),
         id="dict:create:not_optional_int_field_with_none",
     ),
     pytest.param(
         Expected(
             create=lambda: None,
-            op=lambda cfg: OmegaConf.structured(NotOptionalA),
+            op=lambda _: OmegaConf.structured(NotOptionalA),
```
omry, line 559: In general, try to avoid no-op changes that are not directly related to code you are changing. (I know that in this case you did initially change this code so this is not a change request).

omry (changes requested):

tests/test_errors.py
Jasha10: > In general, try to avoid no-op changes that are not directly related to code you are changing.

Ok, got it.

omegaconf/_utils.py
Jasha10: In https://github.com/omry/omegaconf/pull/559/commits/f2a1bfd2af08212cb2b7e294e24f63bdf8fb28a4 I've switched from `dummy_parent._format_and_raise(...)` to `format_and_raise(node=None, ...)`.

The disadvantage of this change is that we no longer have `object_type` information in the error message (as `dummy_parent` was providing the `object_type`).

Jasha10: Besides the `object_type` issue, the error messages are also not reporting `full_key` accurately. I think that reporting on `full_key` would be easiest if we pass a `parent_node` into the `get_structured_config_data` function (so that `full_key` could be computed using the parent config). This change would require some plumbing of the methods that call `get_structured_config_data`. I can give it a try if you're interested...

omegaconf/_utils.py
```diff
@@ -195,13 +196,15 @@ def _resolve_forward(type_: Type[Any], module: str) -> Type[Any]:
 def get_attr_data(obj: Any, allow_objects: Optional[bool] = None) -> Dict[str, Any]:
     from omegaconf.omegaconf import OmegaConf, _maybe_wrap
 
+    obj_type = get_type_of(obj)
+
     flags = {"allow_objects": allow_objects} if allow_objects is not None else {}
     dummy_parent = OmegaConf.create(flags=flags)
+    dummy_parent._metadata.object_type = obj_type
```
omry, line 203: Is this change still required?

omry (changes requested):
looking better.
make a pass to cleanup left overs from previous attempt.

omegaconf/_utils.py
Jasha10: Nope. Reverted in https://github.com/omry/omegaconf/pull/559/commits/703b6df88795df9b93e1dfc4c76bc135a594e6a0.

omry (approved):
awesome :)

omegaconf/_utils.py
omry: I think this will trivially easier once we have the structured_ir diff baked and ready.
for now this is good enough.

omry:
Can you add a news fragment file? feel free to merge in once you do.

PR #564, opened 2021-02-24
Fix handling of `throw_on_missing` flag for `select()` on ListConfig

### LINKED ISSUE #563: `select()` ignores `throw_on_missing` for ListConfig
**Describe the bug**
The following code should raise an exception, but does not:

```python
cfg = OmegaConf.create(["???"])
OmegaConf.select(cfg, key, default=0, throw_on_missing=True)
```

odelalleau:
Fixes #563

### DIFF
--- a/news/563.bugfix
+++ b/news/563.bugfix
@@ -1,1 +1,1 @@
-Selecting a missing node from a ListConfig with `throw_on_missing` set to True now raises the intended exception.
+Calling `OmegaConf.select()` on a missing node from a ListConfig with `throw_on_missing` set to True now raises the intended exception.
--- a/news/563.bugfix
+++ b/news/563.bugfix
@@ -1,0 +1,1 @@
+Selecting a missing node from a ListConfig with `throw_on_missing` set to True now raises the intended exception.
--- a/tests/test_select.py
+++ b/tests/test_select.py
@@ -32,6 +32,7 @@ def test_select_key_from_empty(struct: Optional[bool]) -> None:
         pytest.param([], "0", None, id="list:oob"),
         pytest.param([1, "2"], "0", 1, id="list:int"),
         pytest.param([1, "2"], "1", "2", id="list:str"),
+        pytest.param(["???"], "0", None, id="list:missing"),
         pytest.param([1, {"a": 10, "c": ["foo", "bar"]}], "0", 1),
         pytest.param([1, {"a": 10, "c": ["foo", "bar"]}], "1.a", 10),
         pytest.param([1, {"a": 10, "c": ["foo", "bar"]}], "1.b", None),
@@ -85,7 +86,7 @@ def test_select_default(
         pytest.param({"missing": "???"}, "missing", id="missing"),
     ],
 )
-def test_select_default_throw_on_missing(
+def test_select_default_throw_on_missing_dict(
     cfg: Any,
     struct: bool,
     key: Any,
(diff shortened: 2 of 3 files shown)


news/563.bugfix
```diff
@@ -0,0 +1 @@
+Selecting a missing node from a ListConfig with `throw_on_missing` set to True now raises the intended exception.
```
omry, line 1: Function name here to be more specific about what you mean by "selecting"
```suggestion
OmegaConf.select() a missing node from a ListConfig with `throw_on_missing` set to True now raises the intended exception.
```

tests/test_select.py
```diff
@@ -85,7 +86,7 @@ def test_select_default(
         pytest.param({"missing": "???"}, "missing", id="missing"),
     ],
 )
-def test_select_default_throw_on_missing(
+def test_select_default_throw_on_missing_dict(
```
omry, line 89: old function name seems better. why rename it?

tests/test_select.py
```diff
@@ -99,6 +100,23 @@ def test_select_default_throw_on_missing(
...
+    cfg: Any,
+    key: Any,
+) -> None:
+    cfg = OmegaConf.create(cfg)
+
+    # throw on missing still throws if default is provided
+    with pytest.raises(MissingMandatoryValue):
+        OmegaConf.select(cfg, key, default=0, throw_on_missing=True)
+
+
```
omry, line 119: looks like this is covered by the first test you changed?

omry (changes requested):

Jasha10 (approved):
Looks good.

tests/test_select.py
Jasha10: Before there was just the `test_select_default_throw_on_missing` function;
now there is a `test_select_default_throw_on_missing_dict` function and a `test_select_default_throw_on_missing_list` function.

tests/test_select.py
Jasha10: I think that this test covers the `throw_on_missing=True` case, the other one does not.

odelalleau: No, the `list:missing` test I added is a success case when `throw_on_missing` is False (it was working before as well, but it wasn't tested, I just added it for better test coverage)

news/563.bugfix
odelalleau: Done in b9a5db6

tests/test_select.py
omry: this function is testing both dicts and lists.

tests/test_select.py
omry: How is it a success case if the test body has this?
```
with pytest.raises(MissingMandatoryValue):
   ...
```

omry (changes requested):

tests/test_select.py
odelalleau: It was only testing a dict here, but actually you're right that I can re-use the same test for lists as well (somehow I thought that `set_struct` was only for DictConfig, but looks like I was wrong).

I changed it in cef0a23

tests/test_select.py
odelalleau: Trying to clear up the confusion:
* I thought by "first test you changed" you meant `test_select()`: I indeed added a new test here (id `list:missing`) to test the case where we select a missing value with `throw_on_missing` set to False (which worked before, and still works here)
* The test with `with pytest.raises(MissingMandatoryValue):` is `test_select_default_throw_on_missing`: this tests the case where `throw_on_missing` is True. The new test I added here would have failed before this PR.

omry (approved):
Looking good.
I think GitHub diff view played some tricks on all of us here.

PR #568, opened 2021-02-25
OmegaConf.to_yaml(bad input type): raise ValueError

### LINKED ISSUE #560: OmegaConf.to_yaml fails when passed a plain `str`
You can pass `OmegaConf.to_yaml` a `dict` or `list` with no problem:
```python
>>> OmegaConf.to_yaml(["abc"])
'- abc\n'
```
But if you pass a `str`, you get an `AssertionError`:
```python
>>> OmegaConf.to_yaml("abc")
  File "/home/Jasha10/omegaconf/omegaconf/omegaconf.py", line 815, in to_yaml
    cfg = _ensure_container(cfg)
  File "/home/Jasha10/omegaconf/omegaconf/_utils.py", line 694, in _ensure_container
    assert OmegaConf.is_config(target)
AssertionError
```

# Desired Behavior:
- a more informative error message, such as `"to_yaml got a bad input of type <class 'str'>"`.

Does it make sense to do anything besides throwing an error? The output of the `to_yaml` function is a string, so I'm not sure what would be the correct way to handle inputs that are strings.

Jasha10:
Closes #560 

Raise a `ValueError` if `OmegaConf.to_yaml` is called with a `str`-typed input.

### DIFF
--- a/news/560.api_change
+++ b/news/560.api_change
@@ -1,0 +1,1 @@
+OmegaConf.to_yaml now raises a ValueError when called on a str input. Previously an AssertionError was raised.
--- a/news/560.api_change
+++ b/news/560.api_change
@@ -1,1 +1,1 @@
-OmegaConf.to_yaml now raises a ValueError when called on a str input. Previously an AssertionError was raised.
+OmegaConf.{merge, unsafe_merge, to_yaml} now raises a ValueError when called on a str input. Previously an AssertionError was raised.
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -691,7 +691,12 @@ def _ensure_container(target: Any, flags: Optional[Dict[str, bool]] = None) -> A
         target = OmegaConf.create(target, flags=flags)
     elif is_structured_config(target):
         target = OmegaConf.structured(target, flags=flags)
-    assert OmegaConf.is_config(target)
+    if not OmegaConf.is_config(target):
+        raise ValueError(
+            "Invalid input. Supports one of "
+            + "[dict,list,DictConfig,ListConfig,dataclass,dataclass instance,attr class,attr class instance]"
+        )
+
     return target
 
 
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -691,7 +691,7 @@ def _ensure_container(target: Any, flags: Optional[Dict[str, bool]] = None) -> A
         target = OmegaConf.create(target, flags=flags)
     elif is_structured_config(target):
         target = OmegaConf.structured(target, flags=flags)
-    if not OmegaConf.is_config(target):
+    elif not OmegaConf.is_config(target):
         raise ValueError(
             "Invalid input. Supports one of "
             + "[dict,list,DictConfig,ListConfig,dataclass,dataclass instance,attr class,attr class instance]"
--- a/omegaconf/omegaconf.py
+++ b/omegaconf/omegaconf.py
@@ -829,6 +829,14 @@ class OmegaConf:
         :param sort_keys: If True, will print dict keys in sorted order. default False.
         :return: A string containing the yaml representation.
         """
+        if not (
+            is_primitive_container(cfg)
+            or is_structured_config(cfg)
+            or OmegaConf.is_config(cfg)
+        ):
+            raise ValueError(
+                "to_yaml expects input of type DictConfig / ListConfig / dict / list / StructuredConfig."
+            )
         cfg = _ensure_container(cfg)
         container = OmegaConf.to_container(cfg, resolve=resolve, enum_to_str=True)
         return yaml.dump(  # type: ignore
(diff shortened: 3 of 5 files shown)


omegaconf/omegaconf.py
```diff
@@ -829,6 +829,14 @@ def to_yaml(cfg: Any, *, resolve: bool = False, sort_keys: bool = False) -> str:
...
+            is_primitive_container(cfg)
+            or is_structured_config(cfg)
+            or OmegaConf.is_config(cfg)
+        ):
+            raise ValueError(
+                "to_yaml expects input of type DictConfig / ListConfig / dict / list / StructuredConfig."
+            )
         cfg = _ensure_container(cfg)
         container = OmegaConf.to_container(cfg, resolve=resolve, enum_to_str=True)
         return yaml.dump(  # type: ignore
```
Jasha10, line 842: The logic for the above if-statement is copied from the `_utils.py:_ensure_container` method, which looks like this:
```python
def _ensure_container(target: Any, flags: Optional[Dict[str, bool]] = None) -> Any:

    if is_primitive_container(target):
        assert isinstance(target, (list, dict))
        target = OmegaConf.create(target, flags=flags)
    elif is_structured_config(target):
        target = OmegaConf.structured(target, flags=flags)
    assert OmegaConf.is_config(target)
    return target
```

omegaconf/omegaconf.py
```diff
@@ -829,6 +829,14 @@ def to_yaml(cfg: Any, *, resolve: bool = False, sort_keys: bool = False) -> str:
...
         :return: A string containing the yaml representation.
         """
+        if not (
+            is_primitive_container(cfg)
+            or is_structured_config(cfg)
+            or OmegaConf.is_config(cfg)
+        ):
+            raise ValueError(
+                "to_yaml expects input of type DictConfig / ListConfig / dict / list / StructuredConfig."
+            )
```
omry, line 839: 1. Move the check to ensure_container.
2. Error should be something like:
```
            raise ValueError(
                "Invalid input. Supported one of [dict,list,DictConfig,ListConfig,dataclass,dataclass instance, attr class, attr class instance]"
            )
```

omry (changes requested):

omegaconf/omegaconf.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/568/commits/9336a17f0d4be1acd2bf178c3f4f1f14707730b9.

tests/test_to_yaml.py
```diff
@@ -168,3 +168,14 @@ def test_structured_configs(user: User) -> None:
...
+
+def test_raises_ValueError_on_string_input() -> None:
+    with pytest.raises(
+        ValueError,
+        match=re.escape(
+            "Invalid input. Supports one of "
+            + "[dict,list,DictConfig,ListConfig,dataclass,dataclass instance,attr class,attr class instance]"
+        ),
+    ):
+        OmegaConf.to_yaml("abc")
```
omry, line 181: The current fix should also catch similar issues with other placer we can to_content (e.g. OmegaConf.merge).
I suggest we test it directly instead of through another methods.

Jasha10: Sounds good.

Jasha10: Done in https://github.com/omry/omegaconf/pull/568/commits/61a01fc2b2b95cc21c12d7f0f9b2d162d545f300.

omry:
Add a news fragment (this is an API change).

omry (commented):
had a pending suggestion.

omegaconf/_utils.py
```diff
@@ -691,7 +691,12 @@ def _ensure_container(target: Any, flags: Optional[Dict[str, bool]] = None) -> A
         target = OmegaConf.create(target, flags=flags)
     elif is_structured_config(target):
         target = OmegaConf.structured(target, flags=flags)
-    assert OmegaConf.is_config(target)
+    if not OmegaConf.is_config(target):
```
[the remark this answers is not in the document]
Jasha10: It doesn't work. We need to catch the case where `OmegaConf.is_config(target)` is true because we have passed a DictConfig/ListConfig instance to `_ensure_container`.
I've changed the `if` into an `elif` in https://github.com/omry/omegaconf/pull/568/commits/0aac680601754807a7191a298925a80d6fedf11a.

news/560.api_change
```diff
@@ -0,0 +1 @@
+OmegaConf.to_yaml now raises a ValueError when called on a str input. Previously an AssertionError was raised.
```
omry, line 1: There are a handful of user facing function effected by this change (OmegaConf.merge is another). Can you provide a more complete list?

```
OmegaConf.{to_yaml, merge, ...} now raises ...
```

omry (approved):
One minor request. Feel free to merge once it's addressed.

news/560.api_change
Jasha10: I think that it is only `to_yaml / merge / unsafe_merge`.

The OmegaConf.update function also uses `_ensure_contianer` indirectly, but there is an if-statement to make sure that a string is not passed in.

PR #570, opened 2021-02-25
refactor: clean up MISSING checks

omry:
This clears all literal "???" under the omegaconf module except the definition of MISSING.

### DIFF
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -334,6 +334,23 @@ class ValueKind(Enum):
     INTERPOLATION = 2
 
 
+def _is_missing_value(value: Any) -> bool:
+    from omegaconf import Node
+
+    if isinstance(value, Node):
+        value = value._value()
+    return _is_missing_literal(value)
+
+
+def _is_missing_literal(value: Any) -> bool:
+    from omegaconf import MISSING, Node
+
+    assert not isinstance(value, Node)
+    ret = isinstance(value, str) and value == MISSING
+    assert isinstance(ret, bool)
+    return ret
+
+
 def get_value_kind(
     value: Any, strict_interpolation_validation: bool = False
 ) -> ValueKind:
@@ -351,10 +368,10 @@ def get_value_kind(
         this parsing step is skipped: this is more efficient, but will not detect errors.
     """
 
-    value = _get_value(value)
-
-    if isinstance(value, str) and value == "???":
+    if _is_missing_value(value):
         return ValueKind.MANDATORY_MISSING
+
+    value = _get_value(value)
 
     # We identify potential interpolations by the presence of "${" in the string.
     # Note that escaped interpolations (ex: "esc: \${bar}") are identified as
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -9,14 +9,14 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 import yaml
 
 from ._utils import (
-    ValueKind,
     _ensure_container,
     _get_value,
     _is_interpolation,
+    _is_missing_literal,
+    _is_missing_value,
     _resolve_optional,
     get_ref_type,
     get_structured_config_data,
-    get_value_kind,
     get_yaml_loader,
     is_container_annotation,
     is_dict_annotation,
@@ -50,12 +50,9 @@ class BaseContainer(Container, ABC):
     ) -> Any:
         """returns the value with the specified key, like obj.key and obj['key']"""
 
-        def is_mandatory_missing(val: Any) -> bool:
-            return bool(get_value_kind(val) == ValueKind.MANDATORY_MISSING)
-
         val = _get_value(value)
         has_default = default_value is not DEFAULT_VALUE_MARKER
-        if has_default and (val is None or is_mandatory_missing(val)):
+        if has_default and (val is None or _is_missing_value(val)):
             return default_value
 
         resolved = self._maybe_resolve_interpolation(
@@ -68,7 +65,7 @@ class BaseContainer(Container, ABC):
         if resolved is None and has_default:
             return default_value
 
-        if is_mandatory_missing(resolved):
+        if _is_missing_value(resolved):
             if has_default:
                 return default_value
             else:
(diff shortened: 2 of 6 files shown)


Jasha10 (approved):

omegaconf/_utils.py
```diff
@@ -351,11 +368,11 @@ def get_value_kind(
         this parsing step is skipped: this is more efficient, but will not detect errors.
     """
 
-    value = _get_value(value)
-
-    if isinstance(value, str) and value == "???":
+    if _is_missing_value(value):
         return ValueKind.MANDATORY_MISSING
 
+    value = _get_value(value)
```
odelalleau, line 374: What about keeping `value = _get_value(value)` first and calling `_is_missing_literal()` on it? (seems a bit more efficient, though it may not matter in practice)

omegaconf/basecontainer.py
```diff
@@ -50,12 +50,9 @@ def _resolve_with_default(
     ) -> Any:
         """returns the value with the specified key, like obj.key and obj['key']"""
 
-        def is_mandatory_missing(val: Any) -> bool:
-            return bool(get_value_kind(val) == ValueKind.MANDATORY_MISSING)
-
         val = _get_value(value)
         has_default = default_value is not DEFAULT_VALUE_MARKER
-        if has_default and (val is None or is_mandatory_missing(val)):
+        if has_default and (val is None or _is_missing_value(val)):
```
odelalleau, line 55: Can this one be `_is_missing_literal()`?

omegaconf/basecontainer.py
```diff
@@ -534,7 +530,7 @@ def _set_item_impl(self, key: Any, value: Any) -> None:
 
         input_config = isinstance(value, Container)
         target_node_ref = self._get_node(key)
-        special_value = value is None or value == "???"
+        special_value = value is None or _is_missing_value(value)
```
odelalleau, line 533: Is this fixing a bug? (for instance `DictConfig("???") == "???"` used to be False before)
If yes can a test be added for it?

omegaconf/dictconfig.py
```diff
@@ -170,7 +171,7 @@ def _validate_set(self, key: Any, value: Any) -> None:
         if vk == ValueKind.INTERPOLATION:
             return
         self._validate_non_optional(key, value)
-        if (isinstance(value, str) and value == "???") or value is None:
+        if vk is ValueKind.MANDATORY_MISSING or value is None:
```
odelalleau, line 174: Minor: enums are usually compared with `==` (since it's possible to be equal with a different object)

omegaconf/dictconfig.py
```diff
@@ -619,8 +620,8 @@ def _set_value_impl(
         elif _is_interpolation(value, strict_interpolation_validation=True):
             self.__dict__["_content"] = value
             self._metadata.object_type = None
-        elif value == "???":
-            self.__dict__["_content"] = "???"
+        elif _is_missing_value(value):
```
odelalleau, line 623: Same question as above: is it fixing a bug?

omegaconf/nodes.py
```diff
@@ -115,7 +116,7 @@ def _is_missing(self) -> bool:
                 # Will happen if interpolation is accessing keys that does not exist.
                 return False
         else:
-            ret = self._value() == "???"
+            ret = _is_missing_literal(self._value())
```
odelalleau, line 119: What about just: `return _is_missing_value(self)`

odelalleau (changes requested):
A few questions

omegaconf/_utils.py
omry: _get_value is actually only unwrapping value nodes so it's not quiet the same as what we are after here.

omegaconf/basecontainer.py
omry: Tried. no.
since _get_value does not unwrap container nodes, _is_missing_literal rejects the input in some scenarios.

omegaconf/dictconfig.py
omry: sure.

omegaconf/dictconfig.py
omry: no. this is systematically getting rid of all "???" from the codebase.

omegaconf/nodes.py
omry: good idea.

omegaconf/_utils.py
odelalleau: True, there's a subtle difference, but from what I can see it doesn't matter here (in the sense that both approaches should behave the same), and my suggestion actually better matches the previous implementation, right?

omegaconf/basecontainer.py
omry: Still seems to be False.
This is supposed to be equivalent.

omegaconf/_utils.py
omry: you are welcome to try to change it and see the test failure.

omegaconf/basecontainer.py
odelalleau: I was confused so I tried it, and now I understand (it's also related to my previous comment).

The problem is because of the `assert not isinstance(value, Node)` in `_is_missing_literal`. If you remove it, then what I suggested makes sense, but it doesn't otherwise. I'm fine with either option, then.

omegaconf/basecontainer.py
odelalleau: But `_is_missing_value(DictConfig("???"))` is True, so it's not equivalent

omegaconf/basecontainer.py
omry: gotcha.
I was trying to draw a very clear line between checking if a node is missing versus checking if an actual value is missing.
let's keep the assert.

omegaconf/nodes.py
odelalleau: Not a big deal but I see you didn't exactly take my suggestion, any reason?

omegaconf/basecontainer.py
odelalleau: Ok sounds good, resolving

omegaconf/basecontainer.py
omry: I think In practice it's equivalent due to the many other layers of logic before and after this.
If you can demonstrate a user facing change from this I am happy to respond to it.

omegaconf/_utils.py
odelalleau: Also related to the assert => resolving

omegaconf/nodes.py
omry: because I read it too fast :).

omegaconf/basecontainer.py
odelalleau: Not going to dig deeper myself, just wanted to point it out => ok :)

odelalleau (approved):
All good!

PR #571, opened 2021-02-26
Support and deprecate empty args

### LINKED ISSUE #572: Deprecation of empty resolver arguments in favor of explicit strings
In OmegaConf 2.0, it is possible to have empty elements in a list of resolver arguments, for instance:
* `${foo:a,}` calls `foo("a", "")`
* `${foo:a,,b}` calls `foo("a", "", "b")`
* `${foo:,}` calls `foo("", "")`

This has been deprecated in OmegaConf 2.1. Users should instead use explicit empty strings, for instance:
* `${foo:a,""}`
* `${foo:a, '', b}`
* `${foo:"", ""}`

odelalleau:
Fixes #572

### DIFF
--- a/omegaconf/grammar_visitor.py
+++ b/omegaconf/grammar_visitor.py
@@ -261,7 +261,7 @@ class GrammarVisitor(OmegaConfGrammarParserVisitor):
             )
 
         is_previous_comma = True  # whether previous child was a comma (init to True)
-        for i, child in enumerate(ctx.getChildren()):
+        for child in ctx.getChildren():
             if isinstance(child, OmegaConfGrammarParser.ElementContext):
                 # Also preserve the original text representation of `child` so
                 # as to allow backward compatibility with old resolvers (registered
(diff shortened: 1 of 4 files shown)


omegaconf/grammar_visitor.py
```diff
@@ -246,20 +247,41 @@ def visitSequence(
     ) -> Generator[Any, None, None]:
         from ._utils import _get_value
 
-        assert ctx.getChildCount() >= 1  # element (COMMA element)*
+        # (element (COMMA element?)*) | (COMMA element?)+
+        assert ctx.getChildCount() >= 1
+
+        def empty_str_warning() -> None:
+            txt = ctx.getText()
+            warnings.warn(
```
omry, line 255: Add a comment:
```
# DEPRECATED: remove in 2.2
```

This is not a practice I follow yet in OmegaConf but I think we should.
it makes it easier to spot those later.

omry (changes requested):
Where are your tests? :)

omegaconf/grammar_visitor.py
odelalleau: Done in 9ecac27

Also adds the missing tests and news I mentioned in the chat :p

omry (approved):
lgtm, thanks!

Issue #535, opened 2021-02-10
Follow-up plan regarding changes to resolvers


odelalleau:
#445 introduces several changes to resolvers that triggered multiple discussions on related follow-up changes. Here is an overview of what is planned once #445 is merged:

- [x] Address #488 by validating & converting the result of all interpolations (both node interpolations and resolver outputs) when the target node is typed
- [x] Do a pass on the notebook to make sure it is in synch with the doc, and works with Binder (#544)
- [x] Change the `env` resolver so that it does not parse environment variables anymore, instead keeping them as strings (that may be cast into a different type thanks to the previous bullet point) -- will also resolve #383
- [x] Add support for resolver namespaces (ex: `oc.env`) and deprecate the plain `env` resolver (#573)
- [x] Add new resolver `oc.decode` to parse strings (ex: those output by the `oc.env` resolver) (#574)
- [x] Improve API of `register_new_resolver()` according to #370
- [x] Deprecate `get_resolver()` (#608)
- [X] Forbid whitespace characters around dots in node interpolations and resolver names (#594)
- [x] Deprecate `OmegaConf.is_none()` (#547)
- [x] Implement new resolver cache behavior as discussed in #526 
- [x] Add support for $ in key names (see https://github.com/facebookresearch/hydra/issues/1437#issuecomment-789977874) => #600 (OmegaConf) / https://github.com/facebookresearch/hydra/pull/1605 (Hydra)
- [x] Fix issue with handling of quoted strings (#617)
- [x] Add resolvers `oc.dict.values` and `oc.dict.keys` (#643)
- [x] Add resolver `oc.create` (#645)
- [x] Use getitem syntax as per #179
- [X] SImplify the escaping of quotes (#695)
- [x] Simplify escaping further (#708)
- [x] Fix crash when escaped interpolations do not follow the expected interpolation syntax (#666)
- [x] Port corresponding changes to Hydra (https://github.com/facebookresearch/hydra/pull/1598 / https://github.com/facebookresearch/hydra/pull/1599 / https://github.com/facebookresearch/hydra/pull/1601 / https://github.com/facebookresearch/hydra/pull/1605)
- [x] Address thread-safety concern related to parsing of interpolations (#715)

Documentation (to be done in Q3'21):
- [x] Document OmegaConf's grammar (when doing so, also take care of https://github.com/omry/omegaconf/pull/677#discussion_r614452526)
- ~[ ] Update OmegaConf's slide deck~ (will transfer task to someone else)
[...]

PR #506, opened 2021-02-02
POC: Validating custom resolver results against type annotation

omry:

Issue #459, opened 2020-12-15
Deprecate assignment of strings to non-string nodes like IntegerNode / FloatNode


odelalleau:
**Is your feature request related to a problem? Please describe.**

Currently OmegaConf accepts the following string assignment:

```python
@dataclass
class Foo:
  x: int = 10

cfg: Foo = OmegaConf.structured(Foo)
cfg.x = "20"  # OmegaConf happily converts from string to int, though mypy would complain
```

**Describe the solution you'd like**

Deprecate this functionality to increase type safety.

**Additional context**

This issue was written following the discussion here: https://github.com/omry/omegaconf/pull/445/files#r541981251

omry:
Note that this behavior is well documented in 2.0 and the docs will have to reflect this change.

omry:
Decision for OmegaConf 2.4:

- Deprecate implicit type conversion during direct assignment and typed-container mutation. Warn in 2.4 and reject it in a future release.
- Operations such as `append()`, `insert()`, and `extend()` will also warn. There is no convenient coercive alternative for them; callers must supply values of the correct type.
- Keep conversion for `OmegaConf.update()`, merge operations, config loading, dotlists/CLI input, and interpolation resolution.
- Use `OmegaConf.update()` when explicit conversion is desired. No new coercion API is needed.

omry:
Scope clarification for OmegaConf 2.4: deprecate implicit **value coercion** during direct assignment and typed-container mutation. Warn in 2.4; `OmegaConf.update()` remains the explicit conversion path, as already decided above.

Do not include dataclass/attrs class or instance interpretation as structured configs in this issue. Existing structured-config assignment behavior stays unchanged in 2.4. The proposed `@omegaconf.config` schema marker and its migration are deferred to 3.0 in #1328.

PR #578, opened 2021-03-03
Type validation of interpolations

### LINKED ISSUE #488: Overriding configuration using an environment doesn't trigger type checks when merging into Structured config

### LINKED ISSUE #540: Allow resolvers to output a dict / list to generate a DictConfig / ListConfig
As discussed in https://github.com/facebookresearch/hydra/issues/1389, converting a resolver's output from dict / list to DictConfig / ListConfig would make it easier to generate parts of a config.

For instance:

```python
    OmegaConf.register_new_resolver("make_x", lambda: {"a": 0, "b": 1})
    cfg = OmegaConf.create({"x": "${make_x:}"})
    assert isinstance(cfg.x, DictConfig)
    assert cfg.x.a == 0 and cfg.x.b == 1
```

odelalleau:
Validate and convert interpolation results to their intended type

This PR fixes several issues:

* For nested resolvers (ex: `${f:${g:x}}`), intermediate resolver
  outputs (of `g` in this example) were wrapped in a ValueNode just to
  be unwrapped immediately, which was wasteful. This commit pushes the
  node wrapping to the very last step of the interpolation resolution.

* There was no type checking to make sure that the result of an
  interpolation had a type consistent with the node's type (when
  specified). Now a check is made and the interpolation result may be
  converted into the desired type (see #488).

* If a resolver interpolation returns a dict / list, it is now wrapped
  into a DictConfig / ListConfig, instead of a ValueNode. This makes it
  possible to generate configs from resolvers (see #540). These configs
  are read-only since they are being re-generated on each access.

Fixes #488
Fixes #540

### DIFF
--- a/docs/source/structured_config.rst
+++ b/docs/source/structured_config.rst
@@ -347,7 +347,7 @@ Interpolated values are validated, and converted when possible, to the annotated
     >>> cfg.int_key  # fails due to type mismatch
     Traceback (most recent call last):
       ...
-    omegaconf.errors.ValidationError: Value 'string' could not be converted to Integer
+    omegaconf.errors.InterpolationValidationError: Value 'string' could not be converted to Integer
         full_key: int_key
         object_type=Interpolation
     >>> cfg.str_key = "1234"  # string value
--- a/docs/source/structured_config.rst
+++ b/docs/source/structured_config.rst
@@ -309,7 +309,7 @@ Optional fields
 Interpolations
 ^^^^^^^^^^^^^^
 
-:ref:`interpolation` works normally with Structured configs but static type checkers may object to you assigning a string to an other types.
+:ref:`interpolation` works normally with Structured configs but static type checkers may object to you assigning a string to another type.
 To work around it, use SI and II described below.
 
 .. doctest::
@@ -333,18 +333,38 @@ To work around it, use SI and II described below.
     >>> assert conf.c == 100
 
 
-Type validation is performed on assignment, but not on values returned by interpolation, e.g:
+Interpolated values are validated, and converted when possible, to the annotated type when the interpolation is accessed, e.g:
 
 .. doctest::
 
-    >>> from omegaconf import SI
+    >>> from omegaconf import II
     >>> @dataclass
     ... class Interpolation:
-    ...     int_key: int = II("str_key")
     ...     str_key: str = "string"
+    ...     int_key: int = II("str_key")
 
     >>> cfg = OmegaConf.structured(Interpolation)
-    >>> assert cfg.int_key == "string"
+    >>> cfg.int_key  # fails due to type mismatch
+    Traceback (most recent call last):
+      ...
+    omegaconf.errors.ValidationError: Value 'string' could not be converted to Integer
+        full_key: int_key
+        object_type=Interpolation
+    >>> cfg.str_key = "1234"  # string value
+    >>> assert cfg.int_key == 1234  # automatically convert str to int
+
+Note however that this validation step is currently skipped for container node interpolations:
+
+.. doctest::
+
+    >>> @dataclass
+    ... class NotValidated:
+    ...     some_int: int = 0
+    ...     some_dict: Dict[str, str] = II("some_int")
+
+    >>> cfg = OmegaConf.structured(NotValidated)
+    >>> assert cfg.some_dict == 0  # type mismatch, but no error
+
 
 Frozen
 ^^^^^^
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -472,6 +472,19 @@ simply use quotes to bypass character limitations in strings.
     'Hello, World'
 
 
+Custom resolvers can return lists or dictionaries, that are automatically converted into DictConfig and ListConfig:
+
+.. doctest::
+
+    >>> OmegaConf.register_new_resolver(
+    ...     "min_max", lambda *a: {"min": min(a), "max": max(a)}
+    ... )
+    >>> c = OmegaConf.create({'stats': '${min_max: -1, 3, 2, 5, -10}'})
+    >>> assert isinstance(c.stats, DictConfig)
+    >>> c.stats.min, c.stats.max
+    (-10, 5)
+
+
 You can take advantage of nested interpolations to perform custom operations over variables:
 
 .. doctest::
--- a/news/488.api_change
+++ b/news/488.api_change
@@ -1,1 +1,1 @@
-If the value of a typed node is obtained from an interpolation, it is now validated (and possibly converted) based on the node's type.
+When resolving an interpolation of a typed config value, the interpolated value is validated and possibly converted based on the node's type.
--- a/news/488.api_change
+++ b/news/488.api_change
@@ -1,0 +1,1 @@
+If the value of a typed node is obtained from an interpolation, it is now validated (and possibly converted) based on the node's type.
--- a/news/540.api_change
+++ b/news/540.api_change
@@ -1,0 +1,1 @@
+A custom resolver interpolation whose output is a list or dictionary is now automatically converted into a ListConfig or DictConfig.
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -20,6 +20,8 @@ from .errors import (
     InterpolationKeyError,
     InterpolationResolutionError,
     InterpolationToMissingValueError,
+    InterpolationValidationError,
+    KeyValidationError,
     MissingMandatoryValue,
     OmegaConfBaseException,
     UnsupportedInterpolationType,
@@ -441,7 +443,12 @@ class Container(Node):
                 conv_value = value.validate_and_convert(res_value)
             except ValidationError as e:
                 if throw_on_resolution_failure:
-                    self._format_and_raise(key=key, value=res_value, cause=e)
+                    self._format_and_raise(
+                        key=key,
+                        value=res_value,
+                        cause=e,
+                        type_override=InterpolationValidationError,
+                    )
                 return None
 
             # If the converted value is of the same type, it means that no conversion
@@ -453,14 +460,24 @@ class Container(Node):
 
         if must_wrap:
             assert parent is None or isinstance(parent, BaseContainer)
-            return _node_wrap(
-                type_=value._metadata.ref_type,
-                parent=parent,
-                is_optional=value._metadata.optional,
-                value=resolved,
-                key=key,
-                ref_type=value._metadata.ref_type,
-            )
+            try:
+                return _node_wrap(
+                    type_=value._metadata.ref_type,
+                    parent=parent,
+                    is_optional=value._metadata.optional,
+                    value=resolved,
+                    key=key,
+                    ref_type=value._metadata.ref_type,
+                )
+            except (KeyValidationError, ValidationError) as e:
+                if throw_on_resolution_failure:
+                    self._format_and_raise(
+                        key=key,
+                        value=resolved,
+                        cause=e,
+                        type_override=InterpolationValidationError,
+                    )
+                return None
         else:
             assert isinstance(resolved, Node)
             return resolved
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -461,7 +461,7 @@ class Container(Node):
         if must_wrap:
             assert parent is None or isinstance(parent, BaseContainer)
             try:
-                return _node_wrap(
+                wrapped = _node_wrap(
                     type_=value._metadata.ref_type,
                     parent=parent,
                     is_optional=value._metadata.optional,
@@ -478,6 +478,11 @@ class Container(Node):
                         type_override=InterpolationValidationError,
                     )
                 return None
+            # Since we created a new node on the fly, future changes to this node are
+            # likely to be lost. We thus set the "readonly" flag to `True` to reduce
+            # the risk of accidental modifications.
+            wrapped._set_flag("readonly", True)
+            return wrapped
         else:
             assert isinstance(resolved, Node)
             return resolved
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -444,10 +444,10 @@ class Container(Node):
                     self._format_and_raise(key=key, value=res_value, cause=e)
                 return None
 
-            # If the same object is returned, it means the value is already valid
-            # "as is", and we can thus use it directly. Otherwise, the converted
-            # value has to be wrapped into a node.
-            if conv_value is not res_value:
+            # If the converted value is of the same type, it means that no conversion
+            # was actually needed. As a result, we can keep the original `resolved`
+            # (and otherwise, the converted value must be wrapped into a new node).
+            if type(conv_value) != type(res_value):
                 must_wrap = True
                 resolved = conv_value
 
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -417,6 +417,34 @@ class Container(Node):
         parse_tree: OmegaConfGrammarParser.ConfigValueContext,
         throw_on_resolution_failure: bool,
     ) -> Optional["Node"]:
+        """
+        Resolve an interpolation.
+
+        This happens in two steps:
+            1. The parse tree is visited, which outputs either a `Node` (e.g.,
+               for node interpolations "${foo}"), a string (e.g., for string
+               interpolations "hello ${name}", or any other arbitrary value
+               (e.g., or custom interpolations "${foo:bar}").
+            2. This output is potentially validated and converted when the node
+               being resolved (`value`) is typed.
+
+        If an error occurs in one of the above steps, an `InterpolationResolutionError`
+        (or a subclass of it) is raised, *unless* `throw_on_resolution_failure` is set
+        to `False` (in which case the return value is `None`).
+
+        :param parent: Parent of the node being resolved.
+        :param value: Node being resolved.
+        :param key: The associated key in the parent.
+        :param parse_tree: The parse tree as obtained from `grammar_parser.parse()`.
+        :param throw_on_resolution_failure: If `False`, then exceptions raised during
+            the resolution of the interpolation are silenced, and instead `None` is
+            returned.
+
+        :return: A `Node` that contains the interpolation result. This may be an existing
+            node in the config (in the case of a node interpolation "${foo}"), or a new
+            node that is created to wrap the interpolated value. It is `None` if and only if
+            `throw_on_resolution_failure` is `False` and an error occurs during resolution.
+        """
         from .basecontainer import BaseContainer
         from .nodes import AnyNode, ValueNode
         from .omegaconf import _node_wrap
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -23,6 +23,7 @@ from .errors import (
     MissingMandatoryValue,
     OmegaConfBaseException,
     UnsupportedInterpolationType,
+    ValidationError,
 )
 from .grammar.gen.OmegaConfGrammarParser import OmegaConfGrammarParser
 from .grammar_parser import parse
@@ -353,8 +354,6 @@ class Container(Node):
     ) -> Tuple[Optional["Container"], Optional[str], Optional[Node]]:
         """
         Select a value using dot separated key sequence
-        :param key:
-        :return:
         """
         from .omegaconf import _select_one
 
@@ -416,7 +415,9 @@ class Container(Node):
         parse_tree: OmegaConfGrammarParser.ConfigValueContext,
         throw_on_resolution_failure: bool,
     ) -> Optional["Node"]:
-        from .nodes import StringNode
+        from .basecontainer import BaseContainer
+        from .nodes import AnyNode, ValueNode
+        from .omegaconf import _node_wrap
 
         try:
             resolved = self.resolve_parse_tree(
@@ -429,14 +430,36 @@ class Container(Node):
                 raise
             return None
 
-        assert resolved is not None
-        if isinstance(resolved, str):
-            # Result is a string: create a new StringNode for it.
-            return StringNode(
-                value=resolved,
-                key=key,
+        # If the output is not a Node already (e.g., because it is the output of a
+        # custom resolver), then we will need to wrap it within a Node.
+        must_wrap = not isinstance(resolved, Node)
+
+        # If the node is typed, validate (and possibly convert) the result.
+        if isinstance(value, ValueNode) and not isinstance(value, AnyNode):
+            res_value = _get_value(resolved)
+            try:
+                conv_value = value.validate_and_convert(res_value)
+            except ValidationError as e:
+                if throw_on_resolution_failure:
+                    self._format_and_raise(key=key, value=res_value, cause=e)
+                return None
+
+            # If the same object is returned, it means the value is already valid
+            # "as is", and we can thus use it directly. Otherwise, the converted
+            # value has to be wrapped into a node.
+            if conv_value is not res_value:
+                must_wrap = True
+                resolved = conv_value
+
+        if must_wrap:
+            assert parent is None or isinstance(parent, BaseContainer)
+            return _node_wrap(
+                type_=value._metadata.ref_type,
                 parent=parent,
                 is_optional=value._metadata.optional,
+                value=resolved,
+                key=key,
+                ref_type=value._metadata.ref_type,
             )
         else:
             assert isinstance(resolved, Node)
@@ -483,19 +506,10 @@ class Container(Node):
     ) -> Any:
         from omegaconf import OmegaConf
 
-        from .nodes import ValueNode
-
         resolver = OmegaConf.get_resolver(inter_type)
         if resolver is not None:
             root_node = self._get_root()
-            value = resolver(root_node, inter_args, inter_args_str)
-            return ValueNode(
-                value=value,
-                parent=self,
-                metadata=Metadata(
-                    ref_type=Any, object_type=Any, key=key, optional=True
-                ),
-            )
+            return resolver(root_node, inter_args, inter_args_str)
         else:
             raise UnsupportedInterpolationType(
                 f"Unsupported interpolation type {inter_type}"
@@ -556,7 +570,7 @@ class Container(Node):
                     value=quoted_str,
                     key=key,
                     parent=parent,
-                    is_optional=False,
+                    is_optional=True,
                 ),
                 throw_on_resolution_failure=True,
             )
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -445,10 +445,6 @@ class Container(Node):
             node that is created to wrap the interpolated value. It is `None` if and only if
             `throw_on_resolution_failure` is `False` and an error occurs during resolution.
         """
-        from .basecontainer import BaseContainer
-        from .nodes import AnyNode, ValueNode
-        from .omegaconf import _node_wrap
-
         try:
             resolved = self.resolve_parse_tree(
                 parse_tree=parse_tree,
@@ -459,6 +455,24 @@ class Container(Node):
             if throw_on_resolution_failure:
                 raise
             return None
+
+        return self._validate_and_convert_interpolation_result(
+            parent=parent,
+            value=value,
+            key=key,
+            resolved=resolved,
+            throw_on_resolution_failure=throw_on_resolution_failure,
+        )
+
+    def _validate_and_convert_interpolation_result(
+        self,
+        parent: Optional["Container"],
+        value: "Node",
+        key: Any,
+        resolved: Any,
+        throw_on_resolution_failure: bool,
+    ) -> Optional["Node"]:
+        from .nodes import AnyNode, ValueNode
 
         # If the output is not a Node already (e.g., because it is the output of a
         # custom resolver), then we will need to wrap it within a Node.
@@ -487,33 +501,52 @@ class Container(Node):
                 resolved = conv_value
 
         if must_wrap:
-            assert parent is None or isinstance(parent, BaseContainer)
-            try:
-                wrapped = _node_wrap(
-                    type_=value._metadata.ref_type,
-                    parent=parent,
-                    is_optional=value._metadata.optional,
-                    value=resolved,
-                    key=key,
-                    ref_type=value._metadata.ref_type,
-                )
-            except (KeyValidationError, ValidationError) as e:
-                if throw_on_resolution_failure:
-                    self._format_and_raise(
-                        key=key,
-                        value=resolved,
-                        cause=e,
-                        type_override=InterpolationValidationError,
-                    )
-                return None
-            # Since we created a new node on the fly, future changes to this node are
-            # likely to be lost. We thus set the "readonly" flag to `True` to reduce
-            # the risk of accidental modifications.
-            wrapped._set_flag("readonly", True)
-            return wrapped
+            return self._wrap_interpolation_result(
+                parent=parent,
+                value=value,
+                key=key,
+                resolved=resolved,
+                throw_on_resolution_failure=throw_on_resolution_failure,
+            )
         else:
             assert isinstance(resolved, Node)
             return resolved
+
+    def _wrap_interpolation_result(
+        self,
+        parent: Optional["Container"],
+        value: "Node",
+        key: Any,
+        resolved: Any,
+        throw_on_resolution_failure: bool,
+    ) -> Optional["Node"]:
+        from .basecontainer import BaseContainer
+        from .omegaconf import _node_wrap
+
+        assert parent is None or isinstance(parent, BaseContainer)
+        try:
+            wrapped = _node_wrap(
+                type_=value._metadata.ref_type,
+                parent=parent,
+                is_optional=value._metadata.optional,
+                value=resolved,
+                key=key,
+                ref_type=value._metadata.ref_type,
+            )
+        except (KeyValidationError, ValidationError) as e:
+            if throw_on_resolution_failure:
+                self._format_and_raise(
+                    key=key,
+                    value=resolved,
+                    cause=e,
+                    type_override=InterpolationValidationError,
+                )
+            return None
+        # Since we created a new node on the fly, future changes to this node are
+        # likely to be lost. We thus set the "readonly" flag to `True` to reduce
+        # the risk of accidental modifications.
+        wrapped._set_flag("readonly", True)
+        return wrapped
 
     def _resolve_node_interpolation(
         self,
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -292,9 +292,7 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
             return key  # type: ignore
         elif issubclass(key_type, Enum):
             try:
-                ret = EnumNode.validate_and_convert_to_enum(key_type, key)
-                assert ret is not None
-                return ret
+                return EnumNode.validate_and_convert_to_enum(key_type, key)
             except ValidationError:
                 valid = ", ".join([x for x in key_type.__members__.keys()])
                 raise KeyValidationError(
--- a/omegaconf/errors.py
+++ b/omegaconf/errors.py
@@ -81,7 +81,7 @@ class InterpolationToMissingValueError(InterpolationResolutionError):
     """
 
 
-class InterpolationValidationError(InterpolationResolutionError):
+class InterpolationValidationError(InterpolationResolutionError, ValidationError):
     """
     Thrown when the result of an interpolation fails the validation step.
     """
--- a/omegaconf/errors.py
+++ b/omegaconf/errors.py
@@ -81,6 +81,12 @@ class InterpolationToMissingValueError(InterpolationResolutionError):
     """
 
 
+class InterpolationValidationError(InterpolationResolutionError):
+    """
+    Thrown when the result of an interpolation fails the validation step.
+    """
+
+
 class ConfigKeyError(OmegaConfBaseException, KeyError):
     """
     Thrown from DictConfig when a regular dict access would have caused a KeyError.
--- a/omegaconf/nodes.py
+++ b/omegaconf/nodes.py
@@ -1,6 +1,7 @@
 import copy
 import math
 import sys
+from abc import abstractmethod
 from enum import Enum
 from typing import Any, Dict, Optional, Type, Union
 
@@ -55,8 +56,9 @@ class ValueNode(Node):
         # Subclasses can assume that `value` is not None in `_validate_and_convert_impl()`.
         return self._validate_and_convert_impl(value)
 
+    @abstractmethod
     def _validate_and_convert_impl(self, value: Any) -> Any:
-        return value
+        ...
 
     def __str__(self) -> str:
         return str(self._val)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -20,10 +20,16 @@ def restore_resolvers() -> Any:
 
 
 @pytest.fixture(scope="function")
-def register_identity(restore_resolvers: Any) -> Any:
+def common_resolvers(restore_resolvers: Any) -> Any:
     """
     A fixture to register the common `identity` resolver.
     It depends on `restore_resolvers` to make it easier and safer to use.
     """
+
+    def cast(t: Any, v: Any) -> Any:
+        return {"str": str, "int": int}[t](v)  # cast `v` to type `t`
+
+    OmegaConf.register_new_resolver("cast", cast)
     OmegaConf.register_new_resolver("identity", lambda x: x)
+
     yield
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,6 +3,7 @@ from typing import Any
 
 import pytest
 
+from omegaconf import OmegaConf
 from omegaconf.basecontainer import BaseContainer
 
 
@@ -16,3 +17,13 @@ def restore_resolvers() -> Any:
     state = copy.deepcopy(BaseContainer._resolvers)
     yield
     BaseContainer._resolvers = state
+
+
+@pytest.fixture(scope="function")
+def register_identity(restore_resolvers: Any) -> Any:
+    """
+    A fixture to register the common `identity` resolver.
+    It depends on `restore_resolvers` to make it easier and safer to use.
+    """
+    OmegaConf.register_new_resolver("identity", lambda x: x)
+    yield
--- a/tests/test_interpolation.py
+++ b/tests/test_interpolation.py
@@ -778,6 +778,8 @@ def test_none_value_in_quoted_string(restore_resolvers: Any) -> None:
             id="intermediate_type_mismatch_ok",
         ),
         pytest.param(
+            # This example relies on the automatic casting of a string to int when
+            # assigned to an IntegerNode.
             User(name="Bond", age=SI("${cast:str,'7'}")),
             "age",
             7,
--- a/tests/test_interpolation.py
+++ b/tests/test_interpolation.py
@@ -41,6 +41,14 @@ from . import MissingDict, MissingList, StructuredWithMissing, SubscriptedList, 
 INVALID_CHARS_IN_KEY_NAMES = "\\${}()[].: '\""
 
 
+def dereference(cfg: Container, key: Any) -> Node:
+    node = cfg._get_node(key)
+    assert isinstance(node, Node)
+    node = node._dereference_node()
+    assert isinstance(node, Node)
+    return node
+
+
 @pytest.mark.parametrize(
     "cfg,key,expected",
     [
@@ -941,7 +949,7 @@ def test_interpolation_readonly_resolver_output(
 
 def test_interpolation_readonly_node() -> None:
     cfg = OmegaConf.structured(User(name="7", age=II("name")))
-    resolved = cfg._get_node("age")._dereference_node()
+    resolved = dereference(cfg, "age")
     assert resolved == 7
     # The `resolved` node must be read-only because `age` is an integer, so the
     # interpolation cannot return directly the `name` node.
@@ -971,6 +979,7 @@ def test_resolver_output_dict_to_dictconfig(
     c = OmegaConf.create({"x": "${dict:}", "y": -1})
     assert isinstance(c.x, DictConfig)
     assert c.x == expected
+    assert dereference(c, "x")._get_flag("readonly")
 
 
 @pytest.mark.parametrize(
@@ -990,3 +999,4 @@ def test_resolver_output_list_to_listconfig(
     c = OmegaConf.create({"x": "${list:}", "y": -1})
     assert isinstance(c.x, ListConfig)
     assert c.x == expected
+    assert dereference(c, "x")._get_flag("readonly")
--- a/tests/test_interpolation.py
+++ b/tests/test_interpolation.py
@@ -1,16 +1,19 @@
 import copy
 import random
 import re
-from typing import Any, Optional, Tuple
+from textwrap import dedent
+from typing import Any, Dict, List, Optional, Tuple
 
 import pytest
 from _pytest.python_api import RaisesContext
 
 from omegaconf import (
     II,
+    SI,
     Container,
     DictConfig,
     IntegerNode,
+    ListConfig,
     Node,
     OmegaConf,
     Resolver,
@@ -22,11 +25,12 @@ from omegaconf.errors import (
     GrammarParseError,
     InterpolationKeyError,
     InterpolationResolutionError,
+    KeyValidationError,
     OmegaConfBaseException,
     UnsupportedInterpolationType,
 )
 
-from . import StructuredWithMissing
+from . import MissingDict, MissingList, StructuredWithMissing, SubscriptedList, User
 
 # file deepcode ignore CopyPasteError:
 # The above comment is a statement to stop DeepCode from raising a warning on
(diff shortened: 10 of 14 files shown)


tests/test_interpolation.py
```diff
@@ -747,3 +749,136 @@ def fail_if_called(x: Any) -> None:
...
+@pytest.mark.parametrize(
+    ("cfg", "key", "expected_error"),
+    [
+        pytest.param(
+            User(name="Bond", age=SI("${cast:str,seven}")),
+            "age",
+            pytest.raises(
+                ValidationError,
+                match=re.escape(
+                    "Value 'seven' could not be converted to Integer\n    full_key: age"
```
omry, line 817: I prefer using dedent. it takes more space but makes it easier to read.
```python
  dedent("""\
  Value 'seven' could not be converted to Integer
    full_key: age
  """)
```

tests/test_interpolation.py
```diff
@@ -747,3 +749,136 @@ def fail_if_called(x: Any) -> None:
...
+    cfg = OmegaConf.structured(User(name="Bond", age=SI("${name}")))
+    bad_node = cfg._get_node("age")
+    assert bad_node._dereference_node(throw_on_resolution_failure=False) is None
+
+
+def test_resolver_output_dictconfig(restore_resolvers: Any) -> None:
+    OmegaConf.register_new_resolver("dict", lambda: {"a": 0, "b": 1})
+    cfg = OmegaConf.create({"x": "${dict:}"})
+    assert isinstance(cfg.x, DictConfig)
+    assert cfg.x.a == 0 and cfg.x.b == 1
```
omry, line 871: Test that interpolations in the the generated config are handled properly.
including relative interpolations (I think they should be relative to the node they are being accessed through).

```python
OmegaConf.register_new_resolver("dict", lambda: {"a": 0, "b": 1, "rel": "${.a}", "abs": "${y}"})
cfg = OmegaConf.create({"x": "${dict:}", "y": 10})
assert cfg.x == {"a": 0, "b": 1, "rel" : 0, "abc" : 10}
```

(List too).

omry (commented):
I know those two features are related, but this would have been easier to review as two pull requests:
1. validating and converting on interpolation access
2. converting dict and list to DictConfig and ListConfig when returned from custom resolvers.

omegaconf/dictconfig.py
```diff
@@ -292,9 +292,7 @@ def _s_validate_and_normalize_key(self, key_type: Any, key: Any) -> DictKeyType:
             return key  # type: ignore
         elif issubclass(key_type, Enum):
             try:
-                ret = EnumNode.validate_and_convert_to_enum(
-                    key_type, key, allow_none=False
-                )
+                ret = EnumNode.validate_and_convert_to_enum(key_type, key)
```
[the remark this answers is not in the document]
Jasha10: It looks like this PR has modified the `EnumNode.validate_and_convert_to_enum` function, so that it no longer accepts an `allow_none` argument. The new version of the function behaves as if `allow_none=False` had been passed in.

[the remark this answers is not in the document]
Jasha10: There is a case in `test_errors.py` that covers `__setitem__(cfg, None, ...)` for Enum key_type:
```python
    pytest.param(
        Expected(
            create=lambda: DictConfig(key_type=Color, element_type=str, content={}),
            op=lambda cfg: cfg.__setitem__(None, "bar"),
            exception_type=KeyValidationError,
            msg="Key 'None' is incompatible with the enum type 'Color', valid: [RED, GREEN, BLUE]",
            key=None,
        ),
        id="DictConfig[Color,str]:setitem_bad_key",
    ),
```

There are also a few tests that cover this code path indirectly, e.g. in `test_basic_ops_dict.py` there is the `test_in_dict` function which covers the case of `assert (None in cfg_with_Enum_keys) == False`.

Do you think we should be testing the behavior of `_s_validate_and_normalize_key` directly?

omegaconf/nodes.py
```diff
@@ -335,16 +337,13 @@ def __init__(
...
         return self.validate_and_convert_to_enum(enum_type=self.enum_type, value=value)
 
     @staticmethod
     def validate_and_convert_to_enum(
-        enum_type: Type[Enum], value: Any, allow_none: bool = True
+        enum_type: Type[Enum], value: Any
     ) -> Optional[Enum]:
-        if allow_none and value is None:
-            return None
-
```
Jasha10, line 347: Can the return type annotation for this function be changed from `Optional[Enum]` to `Enum`?

odelalleau:
> I know those two features are related, but this would have been easier to review as two pull requests:
> 
> 1. validating and converting on interpolation access
> 2. converting dict and list to DictConfig and ListConfig when returned from custom resolvers.

I hear you, although to be clear, in that case what happened is that the code I was writing for (1) turned out to also address (2). It's not like one was written on top of the other and combined in a single commit.

docs/source/structured_config.rst
```diff
@@ -333,18 +333,27 @@ To work around it, use SI and II described below.
...
 
     >>> cfg = OmegaConf.structured(Interpolation)
-    >>> assert cfg.int_key == "string"
+    >>> cfg.int_key  # fails due to type mismatch
+    Traceback (most recent call last):
+      ...
+    omegaconf.errors.ValidationError: Value 'string' could not be converted to Integer
+        full_key: int_key
+        object_type=Interpolation
+    >>> cfg.str_key = 1234  # convert int to str (assignment)
```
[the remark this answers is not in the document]
odelalleau: Yeah so that was related to how the original text also mentioned assignment... all fixed in a80ae95

docs/source/structured_config.rst
```diff
@@ -333,18 +333,27 @@ To work around it, use SI and II described below.
     >>> assert conf.c == 100
 
 
-Type validation is performed on assignment, but not on values returned by interpolation, e.g:
+Type validation (and implicit conversion when possible) is performed both on assignment and on values returned by interpolations, e.g:
```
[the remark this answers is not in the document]
odelalleau: Also fixed in a80ae95

docs/source/usage.rst
```diff
@@ -472,6 +472,19 @@ simply use quotes to bypass character limitations in strings.
     'Hello, World'
 
 
+Custom resolvers can return lists or dictionaries, that are automatically converted into config objects:
```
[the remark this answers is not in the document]
odelalleau: Taking your suggestion => resolving

news/488.api_change
```diff
@@ -0,0 +1 @@
+If the value of a typed node is obtained from an interpolation, it is now validated (and possibly converted) based on the node's type.
```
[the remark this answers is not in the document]
odelalleau: Yes, this is tested in `test_interpolation_type_validated_*`

omegaconf/base.py
```diff
@@ -413,14 +414,36 @@ def _resolve_interpolation_from_parse_tree(
...
 
-        assert resolved is not None
-        if isinstance(resolved, str):
-            # Result is a string: create a new StringNode for it.
-            return StringNode(
-                value=resolved,
-                key=key,
+        # If the output is not a Node already (e.g., because it is the output of a
+        # custom resolver), then we will need to wrap it within a Node.
+        must_wrap = not isinstance(resolved, Node)
```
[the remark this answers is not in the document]
odelalleau: The output of custom resolvers used to be wrapped within a Value (not String) node, but this PR gets rid of it (see first bullet point in the commit message)

omegaconf/base.py
```diff
@@ -413,14 +414,36 @@ def _resolve_interpolation_from_parse_tree(
...
+
+        # If the node is typed, validate (and possibly convert) the result.
+        if isinstance(value, ValueNode) and not isinstance(value, AnyNode):
```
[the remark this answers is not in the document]
odelalleau: I had actually tried to do some kind of validation based on `_validate_set()` but ran into some issues (can't remember the details), so I figured it could be left for future work.

That being said, while writing doc to explain that no such validation occurred for config nodes, I realized that it actually did. This is because I'm passing `value._metadata.ref_type` to `_node_wrap()`, which will do the validation / conversion. So instead of updating the doc I added tests for it in 4af9c8b.

omegaconf/dictconfig.py
```diff
@@ -292,9 +292,7 @@ def _s_validate_and_normalize_key(self, key_type: Any, key: Any) -> DictKeyType:
...
                 assert ret is not None
                 return ret
```
Jasha10, line 297: It should be possible to remove this `assert ret is not None` line if the return type of `validate_and_convert_to_enum
` is updated.

omegaconf/base.py
```diff
@@ -413,14 +414,36 @@ def _resolve_interpolation_from_parse_tree(
...
+            res_value = _get_value(resolved)
+            try:
+                conv_value = value.validate_and_convert(res_value)
+            except ValidationError as e:
+                if throw_on_resolution_failure:
+                    self._format_and_raise(key=key, value=res_value, cause=e)
+                return None
```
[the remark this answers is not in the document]
odelalleau: I think both your comments are related, and we should choose between:
1. Returning `None` on validation error if throw_on_resolution_failure == False, and wrapping the exception into a (subclass of) `InterpolationResolutionError` otherwise
2. Always raising the exception, in which case it should *not* be wrapped inside an `InterpolationResolutionError`. 

This is because it would be weird to say `throw_on_resolution_failure=False` then raise an `InterpolationResolutionError` (although it may depend on what we mean exactly by `throw_on_resolution_failure`: so far I've been interpreting it as catching any exception that happens during interpolation resolution)

Personally I have a slight preference for (1) since it seems more consistent with the rest of the code, but I'll wait to hear your thoughts on this before making any change.

omegaconf/base.py
```diff
@@ -413,14 +414,36 @@ def _resolve_interpolation_from_parse_tree(
...
+
+            # If the same object is returned, it means the value is already valid
+            # "as is", and we can thus use it directly. Otherwise, the converted
+            # value has to be wrapped into a node.
+            if conv_value is not res_value:
+                must_wrap = True
+                resolved = conv_value
```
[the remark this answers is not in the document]
odelalleau: Yes I think checking the type makes sense too. Done in 852ea83

omegaconf/base.py
```diff
@@ -540,7 +554,7 @@ def quoted_string_callback(quoted_str: str) -> str:
                     value=quoted_str,
                     key=key,
                     parent=parent,
-                    is_optional=False,
+                    is_optional=True,
```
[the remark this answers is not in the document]
odelalleau: This is for this case:
```python
cfg = OmegaConf.create({"x": "${test:'${none}'}", "none": None})
```

Here, the quoted string we want to evaluate is `'${none}'`. If the node we create to hold the interpolation result is not optional, then an exception will be raised by the new validation code (`Non optional field cannot be assigned None`)

omegaconf/nodes.py
```diff
@@ -40,17 +41,24 @@ def _set_value(self, value: Any, flags: Optional[Dict[str, bool]] = None) -> Non
...
         Validates input and converts to canonical form
         :param value: input value
-        :return:  converted value ("100" may be converted to 100 for example)
+        :return: converted value ("100" may be converted to 100 for example)
         """
-        return value
+        if value is None:
+            if self._is_optional():
+                return None
+            raise ValidationError("Non optional field cannot be assigned None")
```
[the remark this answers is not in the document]
odelalleau: Looking into it, they are actually not calling `validate_and_convert()`. And `None` is not properly validated: I created an issue for that (#579)

omegaconf/dictconfig.py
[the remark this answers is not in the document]
odelalleau: > It looks like this PR has modified the `EnumNode.validate_and_convert_to_enum` function, so that it no longer accepts an `allow_none` argument. The new version of the function behaves as if `allow_none=False` had been passed in.

Yes that is correct, there is no need to pass `allow_none=False` because it is now the behavior by default (this was intended).

tests/test_interpolation.py
odelalleau: Got it -- done in 93f76ce

tests/test_interpolation.py
odelalleau: Done in fc1d2c3

omegaconf/nodes.py
odelalleau: Yes that's a good point, done in cd8655b

omegaconf/dictconfig.py
odelalleau: Right, also done in b556a82

odelalleau:
@omry I think I have addressed all comments except https://github.com/omry/omegaconf/pull/578#discussion_r587975585 on which I'd like to have your input first.

Also, something I just realized, is that if we have an `AnyNode` with `optional == False`, then we will accept `None` as an interpolation result. The obvious "fix" is to remove the `and not isinstance(value, AnyNode)` condition gating the call to `validate_and_convert()`, however the problem is that currently `AnyNode.validate_and_convert()` actually rejects lists and dcitionaries.  For instance on master this currently crashes:

```python
cfg = OmegaConf.create({"x": "???"})
cfg.x = {1, 2, 3}
```

I see 4 options:
1. Keep the current code from this PR and pretend it's ok not to validate interpolation results for an AnyNode (even if it is None and the node is not supposed to be optional)
2. Only check for None vs. the is_optional flag (without calling `AnyNode.validate_and_convert()`
3. Validate interpolation results with `AnyNode.validate_and_convert()` and change it to accept lists and dicts
4. Validate interpolation results with `AnyNode.validate_and_convert()` and keep it as it is

My preference would be for (2) or (3) (option (4) seems really bad as it would break a bunch of stuff, just mentioning it for the sake of completeness)

omry:
I am not addressing any of your other comments yet, but I wanted at a high level (and in response to one of your questions):

This is what I think:
validating a container is very complicated and adds a lot of challenges. For example, how does all this interact with caching? things gets pretty dicey if you cache a DictConfig node that is actually pointing to a dynamic datastructure (through interpolation for example).

Doing it properly would require a lot of thinking and I just thinking about it (before reading your comments) made me realize we might want to drop the feature of generating DictConfig/ListConfig from 2.1 and reconsider for 2.2.

We need to wrap things up for 2.1 and this one seems like a pretty deep rabbit hole.

I think we have two options:
1. try to support this generating dictconfig and listconfig without validation at all (and maybe even produce an error when a user tries that, telling them the node type must be Any for this to be allowed.
2. Remove the generation feature completely from 2.1.

Either way we can revisit for 2.2.
What do you think?

odelalleau:
> 1. try to support this generating dictconfig and listconfig without validation at all (and maybe even produce an error when a user tries that, telling them the node type must be Any for this to be allowed.
> 2. Remove the generation feature completely from 2.1.

To be clear (there's quite a lot in in the PR commentss to it's a bit hard to follow, sorry), the new feature of "generating dictconfig and listconfig" works fine and actually has validation (because `_node_wrap()` takes care of it).

What doesn't get validated is direct node interpolation, e.g., `foo: ${bar}` where `foo` is an annotated DictConfig and `bar` is another node. This is nothing new (it's already like this), but it makes it a bit tricky to explain in the documentaton (if we consider that it is the intended behavior).

So my suggestion would be to keep it like this but open an issue regarding the validation of container nodes for direct node interpolation (which could be fixed after the 2.1 release).

I didn't understand your comment regarding caching though: which caching mechanism are you talking about?

omry:
When I say "generating", I am referring to custom resolvers returning dict or list.
This is also what I mean by caching.
- We are creating a new node on the fly, is that new node cached or will it be regenerated on every access to the custom resolver?
- What happens if the user mutates the returned DictConfig with and without caching?
- What happens when the user is accessing other nodes via interpolation in the generated DictConfig? does that change in the presence of caching?

As for validation when accessing another node via interpolation: Since the other node is already a valid DictConfig/ListConfig, I think  we can get away with shallow validation (let me know if you disagree with that assertion).
Is the top level type compatible with the type we are accessing it through?

This breaks down roughly to:
```
foo: ${bar}
bar: ...
```
Is the type of foo compatible with the type of bar?
- if foo is AnyNode, that should be yes no matter what's in bar (maybe with the exception of optional none-checks).
- if foo is Dict[str, int], bar must be a DictConfig[str,int] (exactly).
- if foo is User, bar must be a User.
- if foo is Optional[User], bar must be a none of User.

If we decide to not validate we should explain it in the docs with a sentence and consider closing the gap in a later version.
To be honest this is pretty far down the rabbit hole and the problem will effect very few people, I am not sure it justifies a lot of work (not to mention that once we start to support `Dict[str, Dict[int, int]`] and such it gets even more complicated.

odelalleau:
> When I say "generating", I am referring to custom resolvers returning dict or list.

So am I :)

> This is also what I mean by caching.
> 
> * We are creating a new node on the fly, is that new node cached or will it be regenerated on every access to the custom resolver?

It will be regenerated on every access. Even if the resolver is cached, only the dict/list output is cached, the node creation (and associated validation) will still happen dynamically.

> * What happens if the user mutates the returned DictConfig with and without caching?

Regardless of caching, the modifications will be lost next time the node is accessed, because `node_wrap()` re-creates a new object every time (at least based on my quick tests to check it out).

What should be the expected behavior is definitely a good question. I don't think it can make sense to allow modifications of such dynamically-generated configs, because the key we are overriding might not even exist anymore next time we access it. What if we made them read-only?

> * What happens when the user is accessing other nodes via interpolation in the generated DictConfig? does that change in the presence of caching?

No, because the cache is only caching the generated dict/list. So the interpolations will still be resolved dynamically regardless of whether it's cached or not.

> As for validation when accessing another node via interpolation: Since the other node is already a valid DictConfig/ListConfig, I think we can get away with shallow validation (let me know if you disagree with that assertion).

Not sure what you mean by "shallow validation".

> Is the top level type compatible with the type we are accessing it through?
> 
> This breaks down roughly to:
> 
> ```
> foo: ${bar}
> bar: ...
> ```
> 
> Is the type of foo compatible with the type of bar?
>  (...)
> If we decide to not validate we should explain it in the docs with a sentence and consider closing the gap in a later version.

Ok, and I would suggest to decide not to validate in 2.1 (because as you said, "this is pretty far down the rabbit hole").
Just to be be sure it's clear, doing this (by "this" I mean `foo: ${bar}` with `bar` being e.g. a DictConfig) is already supported and has no validation. So there would be no change here (just a short note in the doc)

omry:
> It will be regenerated on every access. Even if the resolver is cached, only the dict/list output is cached, the node creation (and associated validation) will still happen dynamically.

Okay. This could be a perf hit but I think it's okay. hopefully people will not use it to return massive dicts.

> What should be the expected behavior is definitely a good question. I don't think it can make sense to allow modifications of such dynamically-generated configs, because the key we are overriding might not even exist anymore next time we access it. What if we made them read-only?

Making then read-only seems like a good idea.
If someone insists on mutating them they can unset the flag but at least they will know that they are doing something weird.

> Not sure what you mean by "shallow validation".

I mean we should not re-create the whole node, must validate the the type match at the top level.

```python
@dataclass
class Config:
  user : User
  admin: User = II("${user}")

cfg : Config = OmegaConf.structured(Config)
cfg.admin.name = "root" # this is perfectly legal and should mutate user.
```
In the above case, we should not re-generate the node on access. we should return the user node.
in terms of validation, we already KNOW that the user node is typed as User, same as the admin node. there is no point in re-creating the node to do validation. A shallow validation is sufficient (the type matches perfectly).
We can assume that the user node is a valid User because it was already validated when we constructed it in the first place. (If it's not valid it's a bug somewhere else, or maybe even intentional if the user forced some extra field into it later).

omegaconf/base.py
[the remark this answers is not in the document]
odelalleau: I went with (1) in 92757d6 to have something working in a simple and consistent way. I'm open to change it though if you prefer a different approach (but if we decide to change the meaning of `throw_on_resolution_failure`, I'd suggest to do it in a different PR)

odelalleau:
> 1. Keep the current code from this PR and pretend it's ok not to validate interpolation results for an AnyNode (even if it is None and the node is not supposed to be optional)
> 2. Only check for None vs. the is_optional flag (without calling `AnyNode.validate_and_convert()`
> 3. Validate interpolation results with `AnyNode.validate_and_convert()` and change it to accept lists and dicts
> 4. Validate interpolation results with `AnyNode.validate_and_convert()` and keep it as it is
> 
> My preference would be for (2) or (3) (option (4) seems really bad as it would break a bunch of stuff, just mentioning it for the sake of completeness)

Just a note that I went with option (2) in 21685e3 (because it's the least disruptive). Open to considering other options though.

odelalleau:
> I mean we should not re-create the whole node, must validate the the type match at the top level.

Ok thanks for the clarification, yes I agree with you.
I tried your example to be sure, it works fine (up to minor fixes):

```python
@dataclass
class Config:
  user : User = User()
  admin: User = II("user")
    
cfg : Config = OmegaConf.structured(Config)
cfg.admin.name = "root" # this is perfectly legal and should mutate user.
assert cfg.user.name == "root"
```

odelalleau:
> Making then read-only seems like a good idea.

Done in e25937d

omry:
I owe you a full pass to see all your changes, I didn't respond to most of your comments/questions yet.

omegaconf/base.py
```diff
@@ -413,15 +416,71 @@ def _resolve_interpolation_from_parse_tree(
...
-                parent=parent,
-                is_optional=value._metadata.optional,
-            )
+        # If the output is not a Node already (e.g., because it is the output of a
+        # custom resolver), then we will need to wrap it within a Node.
+        must_wrap = not isinstance(resolved, Node)
+
+        if isinstance(value, AnyNode):
+            # If the node is untyped, we still need to validate against `None`
+            # in case the `optional` flag is set to `False`.
```
omry, line 425: AnyNode with optional False does not make a whole lot of sense to me.

In fact, it doesn't make sense to mypy as well:

1.py:
```python
from typing import *

x : Any = None
y : Optional[Any] = None
```
```shell
$ mypy --strict 1.py 
Success: no issues found in 1 source file
```

Do we have a use case for non optional Any node?
I just tested and it straightforward to remove the is_optional flag from AnyNode and always make it True.
http://paste.ubuntu.com/p/PRTffVb5vC/

omegaconf/base.py
```diff
@@ -413,15 +416,71 @@ def _resolve_interpolation_from_parse_tree(
...
+                        value=res_value,
+                        cause=e,
+                        type_override=InterpolationValidationError,
+                    )
+                return None
+
+            # If the converted value is of the same type, it means that no conversion
+            # was actually needed. As a result, we can keep the original `resolved`
+            # (and otherwise, the converted value must be wrapped into a new node).
+            if type(conv_value) != type(res_value):
```
omry, line 455: Deducing it from type of the result type seems fragile.
Maybe validate_and_convert should return a second boolean flag indicating if it converted or not?
Alternatively, we can break it into two function:

```python
def is_valid_value(self, value) -> bool:
  ...

def convert_value(self, value) -> Any:
  ...

def validate_and_convert(self, value) -> Any:
  if self.is_valid_value(value):
    return value
  else:
    return self.convert_value(value)
```

This can be a sizeable change. If we do it I suggest we do it as a standalone diff independently of this diff. Once it's in - this diff can utilize the more fine-grained API instead of calling validate_and_convert().

omegaconf/errors.py
```diff
@@ -81,6 +81,12 @@ class InterpolationToMissingValueError(InterpolationResolutionError):
     """
 
 
+class InterpolationValidationError(InterpolationResolutionError):
```
omry, line 84: make it also extend ValidationError.

omegaconf/base.py
```diff
@@ -400,7 +401,9 @@ def _resolve_interpolation_from_parse_tree(
         parse_tree: OmegaConfGrammarParser.ConfigValueContext,
         throw_on_resolution_failure: bool,
     ) -> Optional["Node"]:
```
omry, line 403: Please add docstring to this function.
In particular when it's returning None.

omegaconf/base.py
```diff
@@ -413,15 +416,71 @@ def _resolve_interpolation_from_parse_tree(
...
+                        key=key,
+                        value=resolved,
+                        cause=e,
+                        type_override=InterpolationValidationError,
+                    )
+                return None
+            # Since we created a new node on the fly, future changes to this node are
+            # likely to be lost. We thus set the "readonly" flag to `True` to reduce
+            # the risk of accidental modifications.
+            wrapped._set_flag("readonly", True)
```
omry, line 482: I can see scenarios when this can get in the way.
As a future enhancement, we can consider supporting some flexibility around not setting this flag always.

omegaconf/base.py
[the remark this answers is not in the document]
omry: I don't understand the example.
This quoted string is a string, right?
Why are you parsing it? I was expecting it to be a simple string parameter passed to test().

omegaconf/nodes.py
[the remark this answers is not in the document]
omry: Can we sort out the changes to None validation in a separate PR before this one?

omry (commented):
Reviewed the code, didn't get to the tests.
Please read all comments before fixing things (one of them is asking to split the PR).

news/540.api_change
```diff
@@ -0,0 +1 @@
+A custom resolver interpolation whose output is a list or dictionary is now automatically converted into a ListConfig or DictConfig.
```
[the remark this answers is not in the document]
odelalleau: Sure, would be easy to add. Is this specific functionality related to https://github.com/facebookresearch/hydra/issues/1389?

I wouldn't call it `dict_as_list` though, because in Python `list(some_dict)` returns the list of keys. So maybe `dict_values`? (could even be `oc.dict.values` if we want to start a namespace hierarchy)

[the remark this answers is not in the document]
omry: yeah, I like dict_values better.
oc.dict.values is interesting, do you think we will need more dict functions?

omegaconf/base.py
odelalleau: > AnyNode with optional False does not make a whole lot of sense to me.

I was also surprised, but I noticed that it happened (there were AnyNodes created with optional=False when running the test suite), so I thought they had to be supported. But if we can drop them as you suggested, I totally agree it's a better fix. I'll give it a shot.

news/540.api_change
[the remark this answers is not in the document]
omry: lets do keys and values.
we can add items once we have a real use case.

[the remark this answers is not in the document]
odelalleau: > lets do keys and values.
> we can add items once we have a real use case.

Added to the list in #535 => resolving

omegaconf/nodes.py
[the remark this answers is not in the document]
odelalleau: > Can we sort out the changes to None validation in a separate PR before this one?

I created #592 for this. I will rebase against it once it is merged.

docs/source/usage.rst
```diff
@@ -472,6 +472,19 @@ simply use quotes to bypass character limitations in strings.
     'Hello, World'
 
 
+Custom resolvers can return lists or dictionaries, that are automatically converted into DictConfig and ListConfig:
```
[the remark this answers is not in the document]
odelalleau: > I am starting to think that we need a dedicated page for interpolations (not as a part of this PR).

Added to #535  => resolving

omegaconf/base.py
odelalleau: > Deducing it from type of the result type seems fragile.
> Maybe validate_and_convert should return a second boolean flag indicating if it converted or not?

Or, we impose the contract that it should return the same object if it is not converted -- back to my first version :)

> Alternatively, we can break it into two function

A potential concern with breaking into two functions is that it may do the same checks twice.

Let me know what you think makes most sense for this PR, and I'll go with it.

omry:
I had a comment I can't find now for some reason:
It basically said that AnyNode with optional False does not make any sense and that we should remove the optional flag from AnyNode and assume it's always True.

```python
# 1.py
from typing import *

x: Any = None # fine
y: Optional[Any] = None # also fine, but redundant.
```
```
$ mypy --strict 1.py 
Success: no issues found in 1 source file
```
Since AnyNode is modeling Any, I think it should not have an optional flag. This will eliminate some of the confiusion

omegaconf/base.py
omry: Oh, here is my comment.
Okay, so this can be fixed in the other PR addressing the optional cleanup.

omegaconf/base.py
omry: I think having the same type checks twice during assignment is minor.

Breaking the function into two will also play better when we try to deprecate the automatic conversion on assignment of primitive types. (as opposed to automatic conversion on merge, which I think we want to keep).

omegaconf/nodes.py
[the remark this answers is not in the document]
omry: Roger. resolving.

omegaconf/base.py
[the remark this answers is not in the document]
odelalleau: > This quoted string is a string, right?
> Why are you parsing it? I was expecting it to be a simple string parameter passed to test().

Interpolations in quoted strings are still resolved, similarly to string interpolations `{"greeting": "hello ${name}"}`

odelalleau:
> Since AnyNode is modeling Any, I think it should not have an optional flag. This will eliminate some of the confiusion

Agreed, I included this change in #592

omegaconf/base.py
odelalleau: Done in #592

omegaconf/base.py
odelalleau: Ok but how should I proceed for this PR? I think there are 4 options:
1. Keep current check based on type
2. Go back to previous version where the check was based on the object returned being the same
3. Have `validate_and_convert()` return a boolean flag indicating whether a change was made
4. Split `validate_and_convert()` in two functions

odelalleau:
I rebased on top of master and force-pushed, now that #592 has been merged.

I cleaned up a bit the commit history. I will reference the new meaningful commits in the corresponding comments.

omegaconf/base.py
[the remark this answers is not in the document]
odelalleau: Going to resolve this conversation to follow-up in https://github.com/omry/omegaconf/pull/578#discussion_r590975949

omegaconf/base.py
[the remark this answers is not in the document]
odelalleau: This has been discussed in #592 => resolving

news/488.api_change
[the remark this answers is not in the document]
odelalleau: Done in dca7e01

news/540.api_change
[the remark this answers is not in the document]
odelalleau: Done in 36167ba

omegaconf/base.py
```diff
@@ -413,15 +416,71 @@ def _resolve_interpolation_from_parse_tree(
...
-            # Result is a string: create a new StringNode for it.
-            return StringNode(
-                value=resolved,
-                key=key,
-                parent=parent,
-                is_optional=value._metadata.optional,
-            )
+        # If the output is not a Node already (e.g., because it is the output of a
+        # custom resolver), then we will need to wrap it within a Node.
+        must_wrap = not isinstance(resolved, Node)
```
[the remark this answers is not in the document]
odelalleau: Done in d09277d

omegaconf/base.py
odelalleau: Resolving as this was addressed in #592

omegaconf/errors.py
odelalleau: Done in 6159f3e
(whose commit message has the wrong exception name -- oops -- but shouldn't be an issue since we squash PRs)

omegaconf/base.py
odelalleau: Done in3d350db

tests/test_interpolation.py
```diff
@@ -753,3 +758,251 @@ def test_none_value_in_quoted_string(restore_resolvers: Any) -> None:
...
+            User(name="Bond", age=SI("${cast:int,'7'}")),
+            "age",
+            7,
+            IntegerNode,
+            id="expected_type",
+        ),
+        pytest.param(
+            # This example specifically test the case where intermediate resolver results
+            # are not of the same type as the key.
+            User(name="Bond", age=SI("${cast:int,${drop_last:${drop_last:7xx}}}")),
```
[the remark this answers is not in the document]
omry: Side note, this would have looked somewhat nicer if we had () syntax support. Maybe for 2.2 :).

```
${cast:int, ${identity:'7'}
vs
${cast(int, ${identity('7'))}
```
I would also say that for a single argument, `:` is better.
So:
```
${cast(int, ${identity:'7')}
```

tests/test_interpolation.py
```diff
@@ -753,3 +758,251 @@ def test_none_value_in_quoted_string(restore_resolvers: Any) -> None:
...
+            IntegerNode,
+            id="intermediate_type_mismatch_ok",
+        ),
+        pytest.param(
+            User(name="Bond", age=SI("${cast:str,'7'}")),
+            "age",
+            7,
+            IntegerNode,
+            id="convert_str_to_int",
+        ),
```
omry, line 788: This one is similar to the automatic conversion of ValueNodes I think we want to deprecate.

tests/test_interpolation.py
```diff
@@ -753,3 +758,251 @@ def test_none_value_in_quoted_string(restore_resolvers: Any) -> None:
...
+            MissingList(list=SI("${identity:[0, 1, 2]}")),
+            "list",
+            ["0", "1", "2"],
+            ListConfig,
+            id="list_int_to_str",
+        ),
+        pytest.param(
+            MissingDict(dict=SI("${identity:{a: b, c: d}}")),
+            "dict",
+            {"a": "b", "c": "d"},
```
omry, line 806: Maybe a bit clearer:
```suggestion
            MissingDict(dict=SI("${identity:{key1: val1, key2: val2}}")),
            "dict",
            {"key1": "val1", "key2": "val2"},
```

tests/test_interpolation.py
```diff
@@ -753,3 +758,251 @@ def test_none_value_in_quoted_string(restore_resolvers: Any) -> None:
...
+    key: str,
+    expected_value: Any,
+    expected_node_type: Any,
+    restore_resolvers: Any,
+) -> Any:
+    def cast(t: Any, v: Any) -> Any:
+        return {"str": str, "int": int}[t](v)  # cast `v` to type `t`
+
+    def drop_last(s: str) -> str:
+        return s[0:-1]  # drop last character from string `s`
```
omry, line 830: See if you can drop drop_last :)

tests/test_interpolation.py
```diff
@@ -753,3 +758,251 @@ def test_none_value_in_quoted_string(restore_resolvers: Any) -> None:
...
+    cfg: Any,
+    key: str,
+    expected_error: Any,
+    restore_resolvers: Any,
+) -> None:
+    def cast(t: Any, v: Any) -> Any:
+        return {"str": str, "int": int}[t](v)  # cast `v` to type `t`
+
+    OmegaConf.register_new_resolver("cast", cast)
+    OmegaConf.register_new_resolver("identity", lambda x: x)
```
omry, line 920: A good exercise is to use a fixture instead of duplicating those two functions (not insisting, but I think it's good to practice those).

tests/test_interpolation.py
```diff
@@ -753,3 +758,251 @@ def test_none_value_in_quoted_string(restore_resolvers: Any) -> None:
...
+    with pytest.raises(ReadonlyConfigError):
+        parent_node[sub_key] = -1
+
+
+def test_interpolation_readonly_node() -> None:
+    cfg = OmegaConf.structured(User(name="7", age=II("name")))
+    resolved = cfg._get_node("age")._dereference_node()
+    assert resolved == 7
+    with pytest.raises(ReadonlyConfigError):
+        resolved._set_value(8)
```
omry, line 964: I thought we decided to make generated (transient) nodes readonly, not real nodes.

Is this leaking out to users?

tests/test_interpolation.py
```diff
@@ -753,3 +758,251 @@ def test_none_value_in_quoted_string(restore_resolvers: Any) -> None:
...
+        ({"a": "${..y}"}, {"a": -1}),
+    ],
+)
+def test_resolver_output_dictconfig(
+    restore_resolvers: Any, cfg: Dict[str, Any], expected: Dict[str, Any]
+) -> None:
+    OmegaConf.register_new_resolver("dict", lambda: cfg)
+    c = OmegaConf.create({"x": "${dict:}", "y": -1})
+    assert isinstance(c.x, DictConfig)
+    assert c.x == expected
```
omry, line 989: is it the same object as well?
can you add `assert c.x is expected` or `assert c.x is not expected` and add a comment about the behavior?

tests/test_interpolation.py
```diff
@@ -753,3 +758,251 @@ def test_none_value_in_quoted_string(restore_resolvers: Any) -> None:
...
+        (["${..y}"], [-1]),
+    ],
+)
+def test_resolver_output_listconfig(
+    restore_resolvers: Any, cfg: List[Any], expected: List[Any]
+) -> None:
+    OmegaConf.register_new_resolver("list", lambda: cfg)
+    c = OmegaConf.create({"x": "${list:}", "y": -1})
+    assert isinstance(c.x, ListConfig)
+    assert c.x == expected
```
omry, line 1008: same question as the last one.

omry (changes requested):

tests/test_interpolation.py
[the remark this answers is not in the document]
odelalleau: > Can this be tested with identity instead? drop_last seems overly complex for what this test is doing.
> 
> ```python
>             User(name="Bond", age=SI("${cast:int,${identity:'7'}")),
> ```
> 
> Maybe even nest two identity calls (not sure what nesting twice is adding to the test).

I wanted to test explicitly the situation where there is an intermediate resolver output (by intermediate, I mean not the final result of the interpolation) that would be invalid if we tried to cast it to the node's type. The motivation was that my first implementation was doing the cast immediately after a resolver was called (even an intermediate one), which was incorrect (so I wanted a test to catch that).

If I just use `identity` then the string remains `"7"` which is actually valid when cast to the type of the node (integer). In that regard my comment motivating the test is not clear enough => I changed it in a99f776

I could also just remove that test if you feel like the extra complexity isn't worth it.

tests/test_interpolation.py
odelalleau: I guess we should have that discussion. The corresponding issue is #459, but right now it's not scheduled for 2.1.

One thing that happened since then is the decision for the `env` resolver to only return strings, where we said that one way to convert such strings to other types would be thanks to the automatic conversion of typed nodes from strings. As a result, I wasn't sure that we really wanted to go forward with #459.

Also, on a more personal note, I'd rather not address #459 myself -- at least not for 2.1 (got enough on my plate at this time).

tests/test_interpolation.py
odelalleau: Taking your suggestion => resolving

tests/test_interpolation.py
omry: I am not planning to do #459 for 2.1.
I think we just need to add a comment here saying that this relies on automatic conversion from the node.

tests/test_interpolation.py
odelalleau: No that's a good point, done in c2c1c03

tests/test_interpolation.py
odelalleau: Let's continue the discussion in https://github.com/omry/omegaconf/pull/578#discussion_r592688377 => resolving

tests/test_interpolation.py
odelalleau: Not sure what you mean, what wouldn't be a "real node"?

This specific test is using internal functions to test it on non-container nodes. I can't think of a way users using the public API can run into this, but I still wanted to test it for completeness.

tests/test_interpolation.py
odelalleau: Isn't it obvious? `c.x` is a DictConfig (asserted above) while `expected` is a regular Python dictionary. So they can't be the same objects. Also `expected` isn't used to compute `c.x` (only `cfg` is), so there is no way `c.x` can be the same object.

tests/test_interpolation.py
odelalleau: Same answer :)

tests/test_interpolation.py
odelalleau: > I think we just need to add a comment here saying that this relies on automatic conversion from the node.

Ah, got it! Done in 4af8eb5

tests/test_interpolation.py
omry: Okay.
I was mis interpreting what you were doing.
I thought your dict function actually returned a DictConfig, and then the question is if it's returned as is being used to initialize a new one.

tests/test_interpolation.py
omry: This is an interpolation from one node to another.
We have a use case we want to support:
```python
cfg = OmegaConf.create({
 "user":  {"age" : 7},
 "actor" : "${user}"
})

assert cfg.actor.age == 7
cfg.actor.age = 8
assert cfg.user.age = 8
```

This means that returned container nodes from interpolations are not read only.
Somehow, in this case the resolved value node is read-only.
Is it read-only because the validation generated a new node?

tests/test_interpolation.py
odelalleau: Got it => better test names in 804845f

tests/test_interpolation.py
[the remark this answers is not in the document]
omry: I think a combination of casts can do it then :)

```python
User(name="Bond", age=SI("${cast:int,${cast:str,${identity:7}}}")),
```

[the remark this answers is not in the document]
omry: > Side note, this would have looked somewhat nicer if we had () syntax support. Maybe for 2.2 :).
> 
> ```
> ${cast:int, ${identity:'7'}
> vs
> ${cast(int, ${identity('7'))}
> ```
> 
> I would also say that for a single argument, `:` is better.
> So:
> 
> ```
> ${cast(int, ${identity:'7')}
> ```

I guess the version I like most is this:

```
${cast(int, identity('7'))}
```

As in, once we are in an interpolation we no longer need the ${}, we are already in function mode.

tests/test_interpolation.py
odelalleau: > Is it read-only because the validation generated a new node?
Yes it's only read-only for nodes that are generated on-the-fly.

Also, the example you gave is for a container node (`user`): when we write `cfg.user.age = 8` we are actually writing (first) into `cfg.user`, not `cfg.user.age`. This means that if `cfg.user` is read-only, then this assignment will fail (even if we explicitly set the `"readonly"` flag to False on the `age` node).

tests/test_interpolation.py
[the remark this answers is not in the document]
odelalleau: > I think a combination of casts can do it then :)
> 
> ```python
> User(name="Bond", age=SI("${cast:int,${cast:str,${identity:7}}}")),
> ```

If I'm not mistaken the resolver outputs here are all either `7` (the integer) or `"7"` (the string) => both can be cast to int so that wouln't address the specific issue I wanted to pinpoint.

tests/test_interpolation.py
omry: 1. The example is wrong, the write is through actor. The assertion should be `assert cfg.user.age == 8`.
2. cfg.user is not read-only.
3. Assignment of into DictConfig is complicated. :)

Having said all that, I think we can can resolve this.

tests/test_interpolation.py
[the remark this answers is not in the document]
omry: Yeah, I guess I don't fully understand the error you are defending against.

tests/test_interpolation.py
odelalleau: Regarding point 2, yes, sorry that was a bit confusing. I wanted to emphasize that your example was testing the assignment into a DictConfig returned by an interpolation, and not into a ValueNode (while this test is specifically to test assignment into a ValueNode).

I can't think of a situation where we would try to assign into ValueNode generated by interpolation resolution (using the public API).

Feel free to resolve :)

tests/test_interpolation.py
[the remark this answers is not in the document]
odelalleau: This was an error with the approach from https://github.com/omry/omegaconf/pull/506/files

In this approach, the call to `validate_and_convert()` happened immediately after a resolver was called, so in a chain of resolvers *all* outputs were validated (and possibly converted) according to the node's type. But that's wrong: it's only the final result that needs to be checked.

tests/conftest.py
```diff
@@ -16,3 +17,13 @@ def restore_resolvers() -> Any:
...
+
+
+@pytest.fixture(scope="function")
+def register_identity(restore_resolvers: Any) -> Any:
+    """
+    A fixture to register the common `identity` resolver.
+    It depends on `restore_resolvers` to make it easier and safer to use.
+    """
+    OmegaConf.register_new_resolver("identity", lambda x: x)
+    yield
```
omry, line 29: does this play well with another fixture registering another function?

odelalleau: As far as I can think of, yes: assuming a similar pattern (depending on `restore_resolvers`, we will just copy & restore resolvers multiple times, but everything should work fine.

If we don't depend on `restore_resolvers` then users need to make sure to always put `restore_resolvers` first, which is more cumbersome and error prone.

odelalleau: > assuming a similar pattern (depending on `restore_resolvers`, we will just copy & restore resolvers multiple times, but everything should work fine.

Actually this is wrong, `restore_resolvers` is called only once (which is even better).

omry: Another thing you can try is to unregister after the yield. this way you don't need to depend on restore_resolvers.
Either way, this is good.

tests/test_interpolation.py
```diff
@@ -753,3 +756,250 @@ def test_none_value_in_quoted_string(restore_resolvers: Any) -> None:
...
+)
+def test_interpolation_type_validated_ok(
+    cfg: Any,
+    key: str,
+    expected_value: Any,
+    expected_node_type: Any,
+    register_identity: Any,
+) -> Any:
+    def cast(t: Any, v: Any) -> Any:
+        return {"str": str, "int": int}[t](v)  # cast `v` to type `t`
```
omry, line 827: cast is also a repeating function we use in many places. extract it as well?

tests/test_interpolation.py
```diff
@@ -753,3 +756,250 @@ def test_none_value_in_quoted_string(restore_resolvers: Any) -> None:
...
+    with pytest.raises(ReadonlyConfigError):
+        parent_node[sub_key] = -1
+
+
+def test_interpolation_readonly_node() -> None:
+    cfg = OmegaConf.structured(User(name="7", age=II("name")))
+    resolved = cfg._get_node("age")._dereference_node()
+    assert resolved == 7
+    with pytest.raises(ReadonlyConfigError):
+        resolved._set_value(8)
```
omry, line 961: Add a comment explaining why this is read only.
it's really not obvious looking at the function (In fact, you need to know the field types in User to even have a chance).

omry, line 961: Check the readonly flag directly here and in other places:
```suggestion
    assert resolved._get_flag("readonly") is True
```

tests/conftest.py
odelalleau: There's no unregister API right now => resolving

tests/test_interpolation.py
odelalleau: To keep tests simple I combined both `identity` and `cast` into the same `common_resolvers` feature

tests/test_interpolation.py
odelalleau: Done in b9a3197

odelalleau: Done in fefffc1 (without `is True` though since it seems redundant, considering that flags are `Optional[bool]`)

tests/test_interpolation.py
omry: sounds good.
You should also delete this inline cast function, right?

tests/test_interpolation.py
```diff
@@ -753,3 +755,244 @@ def test_none_value_in_quoted_string(restore_resolvers: Any) -> None:
...
+            "age",
+            pytest.raises(
+                InterpolationValidationError,
+                match=re.escape(
+                    dedent(
+                        """\
+                        Value 'Bond' could not be converted to Integer
+                            full_key: age
+                        """
+                    )
```
omry, line 873: I think for something like this we can omit testing the a the complete error. just go for the message.
```suggestion
                match=re.escape("'Bond' could not be converted to Integer")
```

tests/test_interpolation.py
```diff
@@ -753,3 +755,244 @@ def test_none_value_in_quoted_string(restore_resolvers: Any) -> None:
...
+        (["${..y}"], [-1]),
+    ],
+)
+def test_resolver_output_list_to_listconfig(
+    restore_resolvers: Any, cfg: List[Any], expected: List[Any]
+) -> None:
+    OmegaConf.register_new_resolver("list", lambda: cfg)
+    c = OmegaConf.create({"x": "${list:}", "y": -1})
+    assert isinstance(c.x, ListConfig)
+    assert c.x == expected
```
omry, line 998: You can also add a quick check to confirm those are readonly (here and in dict).
```suggestion
    assert c.x == expected
    assert x._get_node("x")._get_flag("readonly")
```

tests/test_interpolation.py
odelalleau: Taking suggestion => resolving

tests/test_interpolation.py
odelalleau: Done in 8245769
I ended up adding a utility function since it can't be a one-liner because of mypy

tests/test_interpolation.py
odelalleau: Oops, you're right, thanks, I had missed it! Done in 57ee921

omry (approved):
awesome.

omry:
Can you take a last look at the description of the PR and ensure it's still correct after all the changes? I will use it as is in the commit message.

odelalleau:
> Can you take a last look at the description of the PR and ensure it's still correct after all the changes? I will use it as is in the commit message.

Yes it looks good. I just added the fact that resolver-generated configs are read-only.

PR #585, opened 2021-03-08
`{Dict,List}Config.{get(),pop()}` now treat `None` like other valid values

### LINKED ISSUE #583: get() / pop() return their default value when the key exists but is None
**Describe the bug**
The following code currently runs without error, showing that `get()` and `pop()` fall back to their default value when the key exists and is set to `None`:

```python
cfg = OmegaConf.create({"x": None})
assert cfg.get("x", 0) == 0
assert cfg.pop("x", 0) == 0
assert cfg == {}  # `pop()` still removed the node even though it returned the default value!
```

This might actually be a (mostly) intended feature, but it is confusing because this is not how regular Python dictionaries work:
```python
cfg = {"x": None}
assert cfg.get("x", 0) is None
assert cfg.pop("x", 0) is None
assert cfg == {}  # as expected
```

**Expected behavior**
I suggest to change it to match the behavior of Python dictionaries.

Since this is a potentially dangerous breaking change, it could be deprecated as follows in 2.1:
* Add a new argument `allow_none: bool = False` to `get()` and `pop()`
* If the key exists and has a `None` value, then if `allow_none` is True we treat is as any other valid value. If `allow_none` is False, we display a deprecation warning and fall back to the current behavior.

This will force users to specify `allow_none=True` if they ever call `get()` / `pop()` on keys with `None` values. This flag could then be deprecated in 2.2 (changing its default value to `True`) and removed in 2.3.

Thoughts?

odelalleau:
Fixes #583

### DIFF
--- a/news/583.api_change
+++ b/news/583.api_change
@@ -1,1 +1,0 @@
-`DictConfig.{get(),pop()}` and `ListConfig.get()` now return `None` when the accessed key evaluates to `None`, instead of the specified default value.
--- a/news/583.api_change
+++ b/news/583.api_change
@@ -1,0 +1,1 @@
+`DictConfig.{get(),pop()}` and `ListConfig.get()` now return `None` when the accessed key evaluates to `None`, instead of the specified default value.
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -45,15 +45,12 @@ class BaseContainer(Container, ABC):
     def _resolve_with_default(
         self,
         key: Union[DictKeyType, int],
-        value: Any,
+        value: Node,
         default_value: Any = DEFAULT_VALUE_MARKER,
     ) -> Any:
         """returns the value with the specified key, like obj.key and obj['key']"""
-
-        has_default = default_value is not DEFAULT_VALUE_MARKER
-
         if _is_missing_value(value):
-            if has_default:
+            if default_value is not DEFAULT_VALUE_MARKER:
                 return default_value
             raise MissingMandatoryValue("Missing mandatory value: $FULL_KEY")
 
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -52,9 +52,6 @@ class BaseContainer(Container, ABC):
 
         has_default = default_value is not DEFAULT_VALUE_MARKER
 
-        if has_default and _get_value(value) is None:
-            return default_value
-
         if _is_missing_value(value):
             if has_default:
                 return default_value
@@ -66,7 +63,6 @@ class BaseContainer(Container, ABC):
             value=value,
             throw_on_resolution_failure=True,
         )
-        assert resolved_node is not None
 
         return _get_value(resolved_node)
 
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -420,9 +420,10 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
             node = self._get_node(key=key, throw_on_missing_key=True)
         except (ConfigAttributeError, ConfigKeyError):
             if default_value is not DEFAULT_VALUE_MARKER:
-                node = default_value
+                return default_value
             else:
                 raise
+        assert isinstance(node, Node)
         return self._resolve_with_default(
             key=key, value=node, default_value=default_value
         )
@@ -466,6 +467,7 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
             key = self._validate_and_normalize_key(key)
             node = self._get_node(key=key, validate_access=False)
             if node is not None:
+                assert isinstance(node, Node)
                 value = self._resolve_with_default(
                     key=key, value=node, default_value=default
                 )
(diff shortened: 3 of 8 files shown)


news/583.api_change
```diff
@@ -0,0 +1 @@
+`DictConfig.{get(),pop()}` and `ListConfig.get()` now return `None` when the accessed key evaluates to `None`, instead of the specified default value.
```
omry, line 1: I think this is a bug fix, not an API change.
Also, maybe split into two items, lumping DictConfig and ListConfig makes it hard because DictConfig motivation is to be consistent with plain dict and ListConfig motivation is to be consistent with DictConfig (plain list does not have a get method).

tests/test_basic_ops_dict.py
```diff
@@ -203,6 +200,30 @@ def test_dict_get_with_default(
         OmegaConf.set_struct(c, struct)
         assert c.get(key, default_val) == default_val
 
+    @pytest.mark.parametrize(
+        ("d", "select", "key", "expected"),
+        [
+            ({"hello": "world"}, "", "hello", "world"),
+            ({"hello": None}, "", "hello", None),
+            ({"hello": {"a": None}}, "hello", "a", None),
```
omry, line 208: hello is not very informative.
```suggestion
            ({"key": "value"}, "", "key", "value"),
            ({"key": None}, "", "key", None),
            ({"key": {"a": None}}, "key", "a", None),
```

etc.

omegaconf/basecontainer.py
```diff
@@ -52,9 +52,6 @@ def _resolve_with_default(
 
         has_default = default_value is not DEFAULT_VALUE_MARKER
 
-        if has_default and _get_value(value) is None:
-            return default_value
-
```
omry, line 57: it looks like the default value is only coming into play on a missing value now.
is it handled somewhere else as well?
why does this work?
```python
assert OmegaConf.create().get("foo", "bar") == "bar"
```

Under the assumption that the above case is handled at a higher level, maybe we should just handle the MISSING case at a higher level and change the function signature to reflect it?

news/583.api_change
odelalleau: Done in d2b0cf6

tests/test_basic_ops_dict.py
odelalleau: Done in eecac80

omegaconf/basecontainer.py
odelalleau: > it looks like the default value is only coming into play on a missing value now.

That's correct.

> is it handled somewhere else as well?

Not sure exactly what you mean here.

> why does this work?
> 
> ```python
> assert OmegaConf.create().get("foo", "bar") == "bar"
> ```

Because here we only call `_resolve_with_default()` if the node associated to `foo` exists. 

> Under the assumption that the above case is handled at a higher level, maybe we should just handle the MISSING case at a higher level and change the function signature to reflect it?

I actually considered it -- that would mean removing `_resolve_with_default()` and calling directly `_maybe_resolve_interpolation()` instead. But then there are ~4 places where we would need to implement a similar logic to handle missing values => it made sense to me to keep it there.

Looking more closely at it though, I was wondering about why `value` was of type `Any` instead of `Node`, and it turns out it is possible to make it `Node` (which makes it easier to understand what is going on) => I did that in fa6f68f

omegaconf/dictconfig.py
```diff
@@ -466,6 +467,7 @@ def pop(self, key: DictKeyType, default: Any = DEFAULT_VALUE_MARKER) -> Any:
             key = self._validate_and_normalize_key(key)
             node = self._get_node(key=key, validate_access=False)
             if node is not None:
+                assert isinstance(node, Node)
```
omry, line 470: If it's not None, what else can it be other than a Node?

tests/test_basic_ops_list.py
```diff
@@ -31,11 +31,24 @@ def test_list_of_dicts() -> None:
...
+)
+def test_list_get_with_default(
+    cfg: List[Any], key: int, expected: Any, default: Any
+) -> None:
+    c = OmegaConf.create(cfg)
+    val = c.get(key, default_value=default)
+    if expected == "__DEFAULT__":
+        assert val == default
+    else:
+        assert val == expected
```
omry, line 51: I know I have done it myself before, but please split the new test into two tests.
One when the expected value is the provided default and another where the expect value is given by the test parameterization.

omegaconf/dictconfig.py
odelalleau: Nothing, it's just the annoying signature of `get()`'s return type: `-> Union[Optional[Node], List[Optional[Node]]]` that makes mypy believe it may be a list...

I tried to change it but it causes other problems with mypy because this signature is actually defined in the parent `Container` class.

omry: ah, that.
annoying indeed.

tests/test_basic_ops_list.py
odelalleau: Done in 9f06325 (also added a few more test cases at the same time)

omry (approved):
Looking good.

PR #588, opened 2021-03-09
Introduced flags_root as an internal concept

omry:
config flags are currently defined as:

> the value of a config flag is the value set on a node, 
> or the value from the parent node (if there is a parent), otherwise None.

This diff changes the definition to:

> the value of a config flag is the value set on a node, or None if the node is a flags root.
> otherwise, the value from the parent node (if there is a parent), otherwise None.

This means we can now indicate that flag access will not query the parent node for chosen nodes.
For now I am keeping it as an internal API to be used by Hydra config composition:
the hydra node itself is a flags root, meaning it will not be affected by flags set on the user config root.

### DIFF
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -277,7 +277,9 @@ class Node(ABC):
         return self._metadata.flags_root
 
     def _set_flags_root(self, flags_root: bool) -> None:
-        self._metadata.flags_root = flags_root
+        if self._metadata.flags_root != flags_root:
+            self._metadata.flags_root = flags_root
+            self._invalidate_flags_cache()
 
 
 class Container(Node):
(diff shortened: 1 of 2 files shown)


omegaconf/base.py
```diff
@@ -265,6 +273,12 @@ def _key(self) -> Any:
     def _set_key(self, key: Any) -> None:
         self._metadata.key = key
 
+    def _is_flags_root(self) -> bool:
+        return self._metadata.flags_root
+
+    def _set_flags_root(self, flags_root: bool) -> None:
+        self._metadata.flags_root = flags_root
```
odelalleau, line 280: I think this should also invalidate the flags cache if `flags_root` takes a new value

odelalleau (changes requested):

omegaconf/base.py
omry: yeah, good call.

odelalleau (approved):
Go! :)

omry:
After some back, and forth, and back, and forth and ... I finally decided to backtrack from the issue that was the original motivation for this.
I still think this can be useful and it's a very small thing so I am keeping it.

PR #592, opened 2021-03-10
Improved handling of `None` values in node validation

odelalleau:
This PR extracts some of the changes from #578, related to the handling of `None` values.

Note that the test `test_none_value_in_quoted_string` added here is not directly related to these changes (since it worked before this PR too), but it is good to have this test in before #578.

### DIFF
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -292,9 +292,7 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
             return key  # type: ignore
         elif issubclass(key_type, Enum):
             try:
-                ret = EnumNode.validate_and_convert_to_enum(
-                    key_type, key, allow_none=False
-                )
+                ret = EnumNode.validate_and_convert_to_enum(key_type, key)
                 assert ret is not None
                 return ret
             except ValidationError:
--- a/omegaconf/nodes.py
+++ b/omegaconf/nodes.py
@@ -40,16 +40,22 @@ class ValueNode(Node):
         ):
             self._val = value
         else:
-            if not self._metadata.optional and value is None:
-                raise ValidationError("Non optional field cannot be assigned None")
             self._val = self.validate_and_convert(value)
 
     def validate_and_convert(self, value: Any) -> Any:
         """
         Validates input and converts to canonical form
         :param value: input value
-        :return:  converted value ("100" may be converted to 100 for example)
+        :return: converted value ("100" may be converted to 100 for example)
         """
+        if value is None:
+            if self._is_optional():
+                return None
+            raise ValidationError("Non optional field cannot be assigned None")
+        # Subclasses can assume that `value` is not None in `_validate_and_convert_impl()`.
+        return self._validate_and_convert_impl(value)
+
+    def _validate_and_convert_impl(self, value: Any) -> Any:
         return value
 
     def __str__(self) -> str:
@@ -123,7 +129,7 @@ class AnyNode(ValueNode):
             ),
         )
 
-    def validate_and_convert(self, value: Any) -> Any:
+    def _validate_and_convert_impl(self, value: Any) -> Any:
         from ._utils import is_primitive_type
 
         # allow_objects is internal and not an official API. use at your own risk.
@@ -159,12 +165,12 @@ class StringNode(ValueNode):
             ),
         )
 
-    def validate_and_convert(self, value: Any) -> Optional[str]:
+    def _validate_and_convert_impl(self, value: Any) -> Optional[str]:
         from omegaconf import OmegaConf
 
         if OmegaConf.is_config(value) or is_primitive_container(value):
             raise ValidationError("Cannot convert '$VALUE_TYPE' to string : '$VALUE'")
-        return str(value) if value is not None else None
+        return str(value)
 
     def __deepcopy__(self, memo: Dict[int, Any]) -> "StringNode":
         res = StringNode()
@@ -188,11 +194,9 @@ class IntegerNode(ValueNode):
             ),
         )
 
-    def validate_and_convert(self, value: Any) -> Optional[int]:
+    def _validate_and_convert_impl(self, value: Any) -> Optional[int]:
         try:
-            if value is None:
-                val = None
-            elif type(value) in (str, int):
+            if type(value) in (str, int):
                 val = int(value)
             else:
                 raise ValueError()
@@ -222,9 +226,7 @@ class FloatNode(ValueNode):
             ),
         )
 
-    def validate_and_convert(self, value: Any) -> Optional[float]:
-        if value is None:
-            return None
+    def _validate_and_convert_impl(self, value: Any) -> Optional[float]:
         try:
             if type(value) in (float, str, int):
                 return float(value)
@@ -273,16 +275,14 @@ class BooleanNode(ValueNode):
             ),
         )
 
-    def validate_and_convert(self, value: Any) -> Optional[bool]:
+    def _validate_and_convert_impl(self, value: Any) -> Optional[bool]:
         if isinstance(value, bool):
             return value
         if isinstance(value, int):
             return value != 0
-        elif value is None:
-            return None
         elif isinstance(value, str):
             try:
-                return self.validate_and_convert(int(value))
+                return self._validate_and_convert_impl(int(value))
             except ValueError as e:
                 if value.lower() in ("yes", "y", "on", "true"):
                     return True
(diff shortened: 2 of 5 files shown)


omegaconf/dictconfig.py
```diff
@@ -292,9 +292,7 @@ def _s_validate_and_normalize_key(self, key_type: Any, key: Any) -> DictKeyType:
             return key  # type: ignore
         elif issubclass(key_type, Enum):
             try:
-                ret = EnumNode.validate_and_convert_to_enum(
-                    key_type, key, allow_none=False
-                )
+                ret = EnumNode.validate_and_convert_to_enum(key_type, key)
```
omry, line 295: @Jasha10, can you double check to see that we are not currently missing a test? Not sure why we can just remove allow_none here without changing relevant tests.

omry (commented):
Overall looks ok.
A few inline comments.

Also, are there any known bugs this is fixing? If so, can you add corresponding tests?

omegaconf/dictconfig.py
odelalleau: This was discussed here: https://github.com/omry/omegaconf/pull/578#discussion_r587994126

omegaconf/omegaconf.py
```diff
@@ -937,7 +937,7 @@ def _node_wrap(
             element_type=element_type,
         )
     elif type_ == Any or type_ is None:
-        node = AnyNode(value=value, key=key, parent=parent, is_optional=is_optional)
+        node = AnyNode(value=value, key=key, parent=parent)
```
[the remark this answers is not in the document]
odelalleau: I actually did that before submitting the PR (though the correct assert is `assert is_optional`), the only failures are in a few tests that explicitly call `_node_wrap()` with `optional` set to False. It would be a bit annoying to fix because they are part of parameterized tests, so I let it go.

tests/test_interpolation.py
```diff
@@ -747,3 +747,9 @@ def fail_if_called(x: Any) -> None:
     x_node = cfg._get_node("x")
     assert isinstance(x_node, Node)
     assert x_node._dereference_node(throw_on_resolution_failure=False) is None
+
+
+def test_none_value_in_quoted_string(restore_resolvers: Any) -> None:
+    OmegaConf.register_new_resolver("test", lambda x: x)
+    cfg = OmegaConf.create({"x": "${test:'${missing}'}", "missing": None})
+    assert cfg.x == "None"
```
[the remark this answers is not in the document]
odelalleau: Yes. To clarify, there are two situations:
1. The interpolation is exactly surrounded by quotes (`"${...}"`) => we resolve it normally but cast the result to string.
2. There are other characters around the interpolation (`"hello ${...} and hi"`) => this is strictly equivalent to regular string interpolations

I think this is the most intuitive behavior.

odelalleau:
> Also, are there any known bugs this is fixing?

No user-facing bug that I know of. This is why it was originally part of #578 *after* the interpolation type validation (because it was fixing bugs there)

This PR still adds some tests of `validate_and_convert()` that would have failed before.

omegaconf/dictconfig.py
Jasha10: Yes, this is a clean change because `validate_and_convert_to_enum` function signature has changed too.

omegaconf/nodes.py
```diff
@@ -159,12 +162,12 @@ def __init__(
             ),
         )
 
-    def validate_and_convert(self, value: Any) -> Optional[str]:
+    def _validate_and_convert_impl(self, value: Any) -> Optional[str]:
```
Jasha10, line 165:
```suggestion
    def _validate_and_convert_impl(self, value: Any) -> str:
```

omegaconf/nodes.py
```diff
@@ -188,11 +191,9 @@ def __init__(
             ),
         )
 
-    def validate_and_convert(self, value: Any) -> Optional[int]:
+    def _validate_and_convert_impl(self, value: Any) -> Optional[int]:
```
Jasha10, line 194:
```suggestion
    def _validate_and_convert_impl(self, value: Any) -> int:
```

omegaconf/nodes.py
```diff
@@ -188,11 +191,9 @@ def __init__(
...
         try:
-            if value is None:
-                val = None
-            elif type(value) in (str, int):
+            if type(value) in (str, int):
                 val = int(value)
             else:
                 raise ValueError()
```
Jasha10, line 199: I would suggest to refactor this function's body like this:
```python
    def _validate_and_convert_impl(self, value: Any) -> int:
        if type(value) in (str, int):
            return int(value)
        else:
            raise ValidationError("Value '$VALUE' could not be converted to Integer")
```

omegaconf/nodes.py
```diff
@@ -222,9 +223,7 @@ def __init__(
             ),
         )
 
-    def validate_and_convert(self, value: Any) -> Optional[float]:
-        if value is None:
-            return None
+    def _validate_and_convert_impl(self, value: Any) -> Optional[float]:
```
Jasha10, line 226:
```suggestion
    def _validate_and_convert_impl(self, value: Any) -> float:
```

Jasha10: The same refactoring will work here:
```python
    def _validate_and_convert_impl(self, value: Any) -> float:
        if type(value) in (float, str, int):
            return float(value)
        else:
            raise ValidationError("Value '$VALUE' could not be converted to Float")
```

omegaconf/nodes.py
```diff
@@ -273,16 +272,14 @@ def __init__(
             ),
         )
 
-    def validate_and_convert(self, value: Any) -> Optional[bool]:
+    def _validate_and_convert_impl(self, value: Any) -> Optional[bool]:
```
Jasha10, line 275:
```suggestion
    def _validate_and_convert_impl(self, value: Any) -> bool:
```

omegaconf/nodes.py
```diff
@@ -335,16 +332,13 @@ def __init__(
...
 
-    def validate_and_convert(self, value: Any) -> Optional[Enum]:
+    def _validate_and_convert_impl(self, value: Any) -> Optional[Enum]:
         return self.validate_and_convert_to_enum(enum_type=self.enum_type, value=value)
 
     @staticmethod
     def validate_and_convert_to_enum(
-        enum_type: Type[Enum], value: Any, allow_none: bool = True
+        enum_type: Type[Enum], value: Any
     ) -> Optional[Enum]:
```
Jasha10, line 341: Return type should be `-> Enum` instead of `-> Optional[Enum]`.

omegaconf/nodes.py
```diff
@@ -335,16 +332,13 @@ def __init__(
             ),
         )
 
-    def validate_and_convert(self, value: Any) -> Optional[Enum]:
+    def _validate_and_convert_impl(self, value: Any) -> Optional[Enum]:
```
Jasha10, line 335:
```suggestion
    def _validate_and_convert_impl(self, value: Any) -> Enum:
```

Jasha10 (changes requested):
A few refactorings and type-signature changes:

omegaconf/nodes.py
```diff
@@ -40,16 +40,22 @@ def _set_value(self, value: Any, flags: Optional[Dict[str, bool]] = None) -> Non
...
         :param value: input value
-        :return:  converted value ("100" may be converted to 100 for example)
+        :return: converted value ("100" may be converted to 100 for example)
         """
+        if value is None:
+            if self._is_optional():
+                return None
+            raise ValidationError("Non optional field cannot be assigned None")
+        # Subclasses can assume that `value` is not None in `_validate_and_convert_impl()`.
+        return self._validate_and_convert_impl(value)
```
Jasha10, line 56: What do you think about adding `assert value is not None` to each of the `_validate_and_convert_impl` methods?
My motivation for this idea is to help catch future bugs as the OmegaConf internals evolve.

odelalleau: I don't mind either way -- let's have @omry tell us what he likes best :)

omegaconf/nodes.py
odelalleau: Fixed in 8e16ce4

omegaconf/nodes.py
odelalleau: Your suggestion wouldn't raise the intended ValidationError if `value` is a string that can't be converted to `int`

omegaconf/nodes.py
odelalleau: Same issue for invalid `str`

omegaconf/nodes.py
Jasha10: > Your suggestion wouldn't raise the intended ValidationError if `value` is a string that can't be converted to `int`

Oh, I see.

omegaconf/omegaconf.py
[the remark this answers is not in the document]
omry: roger.

omegaconf/nodes.py
Jasha10: Got it.

omegaconf/nodes.py
omry: I don't think it's too useful because _validate_and_convert_impl is only called from that one callsite and I am not expecting it to be called directly.

omry (approved):

PR #595, opened 2021-03-10
Deprecate `OmegaConf.is_none()`

### LINKED ISSUE #547: Deprecate `OmegaConf.is_none()`
**PROBLEM**

With the change to `OmegaConf.is_missing()` in #545, the current behavior of `OmegaConf.is_none()` in OmegaConf 2.0 can be confusing (because it resolves interpolations, while `OmegaConf.is_missing()` doesn't).

In addition, `OmegaConf.is_none()` has potentially non-intuitive behavior in situations where the key can't be resolved (see below for details).

**SOLUTION**
`OmegaConf.is_none()` is being deprecated in OmegaConf 2.1, and will be removed in OmegaConf 2.2

**MIGRATION INSTRUCTIONS**

*Please do not hesitate to open an issue if you run into problems that the instructions below do not address.*

__Scenario 1 -- Your code is checking a specific key:__
```python
OmegaConf.is_none(cfg, "foo")
OmegaConf.is_none(cfg, key)
```

It is recommended to change these to:
```python
cfg.foo is None
cfg[key] is None
```

Note however that when using this new syntax, an exception may be raised when accessing the key, while `OmegaConf.is_none()` used to silence such exceptions. If you want to keep the exact (but possibly unintuitive) behavior from before, you can implement your own `is_none()` function as follows:

```python
def is_none(obj: Container, key: Union[int, DictKeyType]) -> bool:
    try:
        return obj[key] is None
    except ConfigKeyError:
        return True  # `True` if the key does not exist in the config
    except (InterpolationResolutionError, MissingMandatoryValue):
        return False  # `False` if it is either missing ("???") or an invalid interpolation
```

__Scenario 2 -- Your code is not providing a key:__

```python
OmegaConf.is_none(obj)
```

There is currently no plan to keep supporting this usage in OmegaConf 2.2. If you need it, please open an issue explaining why.

**Additional context**
See discussion in https://github.com/omry/omegaconf/pull/545#discussion_r577239021

Dev note: at the same time, the internal `Node._is_none()` can be modified to *not* resolve interpolations.

odelalleau:
Notes to reviewer:
* Better review commits independently. The first two commits should be easy to review -- in particular read the commit message of the second one (c4cd6a3)
* There are a few places where I changed the order of checks in some conditions. This is purely to try and perform the fastest checks first (there is no behavior change)
* `OmegaConf.is_none()` was doing a lot of things, and handling several edge cases in a non-obvious way (see #547). I replaced its internal usage with a new `_is_none()` function in `_utils.py` that is more explicit (the caller has to specify whether interpolations should be resolved, and if yes, whether resolution errors should be silenced). I tried to call it with the settings that made most sense to me (and that passed the tests :) ) in each situation, but there might be weird (untested) situations where the behavior is not the same as before.

Fixes #547


odelalleau:
I just noticed though that Hydra is calling `OmegaConf.is_none(node)`. This is a use case that this PR is deprecating without a publicly available alternative.

As far as I can tell, these calls should be replaced by `_is_none(node, resolve=True)`. So I can see at least the two following options:
1. Have Hydra directly access OmegaConf's internal `_is_none()` function
2. Actually keep a public (non-deprecated) `OmegaConf.is_none(obj: Any, resolve: bool, throw_on_resolution_failure: bool=True)` where users would have to specify whether or not to resolve the node (in 2.1 the signature would be a bit different to preserve backward compatibility while going through a deprecation phase: `OmegaConf.is_none(obj: Any, key: Optional[Union[int, DictKeyType]] = None, resolve: Optional[bool] = None, throw_on_resolution_failure: Optional[bool] = None)`

@omry thoughts on these options / other ideas?

tests/test_omegaconf.py
```diff
@@ -328,6 +328,25 @@ def test_is_none(fac: Any, is_none: bool) -> None:
...
+    ],
+)
+def test_is_none_interpolation(cfg: Any, key: str, is_none: bool) -> None:
+    cfg = OmegaConf.create(cfg)
+    assert OmegaConf.is_none(cfg, key) == is_none
+
+
+def test_is_none_invalid_node() -> None:
+    cfg = OmegaConf.create({})
+    assert OmegaConf.is_none(cfg, "invalid")
```
omry: I would say this one is actually a bug.

tests/test_omegaconf.py
```diff
@@ -328,6 +328,25 @@ def test_is_none(fac: Any, is_none: bool) -> None:
     assert OmegaConf.is_none(cfg, "node") == is_none
 
 
+@pytest.mark.parametrize(
+    ("cfg", "key", "is_none"),
+    [
+        ({"foo": "${none}", "none": None}, "foo", True),
+        ({"foo": "${missing_node}"}, "foo", False),
+        ({"foo": "${missing_resolver:}"}, "foo", False),
+        ({"foo": "${oc.decode:'[1, '}"}, "foo", False),
```
omry, line 338: I am okay with a breaking change for OmegaConf.is_none() to make those all raise an exception.
1. It's very unlikely that anyone is depending on this.
2. This behavior is not intentional.

If we do that, I think we can get away without deprecating the public OmegaConf.is_none().

omry:
Let's go with your option 2.
2.1 will require throw_on_resolution failure (otherwise a deprecation warning), and 2.2 will make True the default.

odelalleau:
> Let's go with your option 2.
> 2.1 will require throw_on_resolution failure (otherwise a deprecation warning), and 2.2 will make True the default.

Sounds good. Just to be sure it's clear, I have in mind to force users to provide the `resolve` flag:
* will default to True with a deprecation warning in 2.1
* the default value will be removed in 2.2

This is to make sure there is no misunderstanding on the resolution of interpolations, considering that `OmegaConf.is_missing()` doesn't resolve them.

With this plan, there is no need to require `throw_on_resolution_failure`:
* If `resolve` is provided, then people are using the new API and we can default to `True` without a warning
* If `resolve` is not provided, then people are using the old API and there is already a deprecation warning (and we can thus default to `False` silently)

tests/test_omegaconf.py
odelalleau: Ok -- not going to fix it though since it's being deprecated

tests/test_omegaconf.py
odelalleau: Just to confirm, this will raise an exception by default with the deprecation plan I have in mind, but users will be able to silence the exception by specifying `throw_on_resolution_failure=False`

omry:
> > Let's go with your option 2.
> > 2.1 will require throw_on_resolution failure (otherwise a deprecation warning), and 2.2 will make True the default.
> 
> Sounds good. Just to be sure it's clear, I have in mind to force users to provide the `resolve` flag:
> 
> * will default to True with a deprecation warning in 2.1 (i.e., actual default value will be `None`, which will trigger the warning + changing it to `True`)
> * the default value will be removed in 2.2
> 
> This is to make sure there is no misunderstanding on the resolution of interpolations, considering that `OmegaConf.is_missing()` doesn't resolve them.
> 
> With this plan, there is no need to require `throw_on_resolution_failure`:
> 
> * If `resolve` is provided, then people are using the new API and we can default to `True` without a warning
> * If `resolve` is not provided, then people are using the old API and there is already a deprecation warning (and we can thus default to `False` silently)

Sounds good. This means that if resolve is True (or None), it implies that a resolution error will always throw.
If resolve is False, an exception is never raised and it tests if the node value is None directly.

odelalleau:
I just pushed 2fabf1b to simplify a bit a test

odelalleau:
Also, just mentioning (in case anyone else is watching) that after discussions with @omry we finally decided to go back to the orignal plan (= full deprecation of `OmegaConf.is_none()`)

tests/test_omegaconf.py
odelalleau: Resolving since we decided to deprecate it in the end

omry (commented):
Github is fucked up today.

omry (commented):
use stacklevel in warn().

omry (changes requested):
You probably missed my nit because of GitHub shitfuckery today.

odelalleau:
> use stacklevel in warn().

Oops, done in 2845aff

> You probably missed my nit because of GitHub shitfuckery today.

Yeah, looks like no notification...

omegaconf/listconfig.py
```diff
@@ -565,13 +566,13 @@ def _set_value(self, value: Any, flags: Optional[Dict[str, bool]] = None) -> Non
...
     ) -> None:
-        from omegaconf import MISSING, OmegaConf, flag_override
+        from omegaconf import MISSING, flag_override
 
         if flags is None:
             flags = {}
 
         vk = get_value_kind(value, strict_interpolation_validation=True)
-        if OmegaConf.is_none(value):
+        if _is_none(value, resolve=True):
```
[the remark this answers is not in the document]
odelalleau: Apparently DictConfig catches it in `_validate_set()` while ListConfig does it in `_set_value_impl()`... not sure what's the history here, but at least there doesn't seem to be a bug.

tests/test_matrix.py
```diff
@@ -104,7 +109,14 @@ def test_none_assignment_and_merging_in_dict(
             node = node_type(value=value, is_optional=False)
             data = {"node": node}
             cfg = OmegaConf.create(obj=data)
-            verify(cfg, "node", none=False, opt=False, missing=False, inter=False)
+            verify(
+                cfg,
+                "node",
+                none_internal=False,
```
omry, line 115: can you rename non_internal back to none and undo the formatting changes when not needed (by removing the trailing comma)?

omry (commented):
Overall looking good. Just one nit.

tests/test_matrix.py
odelalleau: Sure -- done in 44cd8af

Issue #370, opened 2020-09-11
Improve the API of register_resolver()


omry:
1. Currently, if a resolver is already registered there is an assertion error. this should be a regular exception (ValueError)
2. ~~If the same callable is registered twice with the same name (and the same cache settings), we can just ignore the call.~~ (dropped)
2. Allow overriding an existing resolver by registering it with `replace=True`
3. Currently, if one attempts to register a resolver that is not a callable, an assertion error is raised: it should be a TypeError
4. If an empty name is provided, a ValueError should be raised

omry:
Blocked on #445 which makes changes to this API.

@odelalleau, can you pick this one up later?

odelalleau:
> Blocked on #445 which makes changes to this API.
> 
> @odelalleau, can you pick this one up later?

Yes, I added it to #535 (it should be trivial)

Just to clarify, do you want to create a new exception for this, or just raise a `OmegaConfBaseException`?

omry:
I think even ValueError is fine here. assertion errors are not supposed to be used as user interface.
Since we are creating a new register function it's a chance to fix this.

PR #599, opened 2021-03-13
Support optional _parent_ and _root_ parameters in custom resolvers

omry:
Custom resolvers can now receive optional _parent_ and _root_ parameters.

This is the doc example:
```python

OmegaConf.register_new_resolver(
    "sum_friends",
    lambda a, b, *, _parent_: _parent_[a] + _parent_[b],
    use_cache=False,
)
cfg = OmegaConf.create(
    {
        "a": {
            "b": 1,
            "c": 2,
            "sum": "${sum_friends:b,c}",
        },
    }
)

assert cfg.a.sum == 3
```

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -531,6 +531,37 @@ inputs we always return the same value. This behavior may be disabled by setting
     >>> assert c.uncached != c.uncached
 
 
+
+Custom interpolations can also receive the following special parameters:
+
+- ``_parent_`` : the parent node of an interpolation.
+- ``_root_``: The config root.
+
+This can be achieved by adding ``_parent_`` and ``_root_`` to the resolver signature.
+Note that special parameters must be defined as named keywords (after the `*`):
+
+
+.. doctest::
+
+    >>> OmegaConf.register_new_resolver(
+    ...    "sum_friends",
+    ...    lambda a, b, *, _parent_: _parent_[a] + _parent_[b],
+    ...    use_cache=False,
+    ...)
+    >>>
+    >>> cfg = OmegaConf.create(
+    ...     {
+    ...         "a": {
+    ...             "b": 1,
+    ...             "c": 2,
+    ...             "sum": "${sum_friends:b,c}",
+    ...         },
+    ...     }
+    ... )
+    >>> cfg.a.sum
+    3
+
+
 Merging configurations
 ----------------------
 Merging configurations enables the creation of reusable configuration files for each logical component
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -540,15 +540,13 @@ This can be achieved by adding the special parameters to the resolver signature.
 Note that special parameters must be defined as named keywords (after the `*`):
 
 In this example, we use ``_parent_`` to implement a sum function that defaults to 0 if the node does not exist.
-(In contrast to the sum we defined earlier where accessing an invalid key, e.g. `"a_plus_z": ${sum:${a}, ${z}}` will result in an error).
+(In contrast to the sum we defined earlier where accessing an invalid key, e.g. ``"a_plus_z": ${sum:${a}, ${z}}`` will result in an error).
 
 .. doctest::
 
-    >>> OmegaConf.register_new_resolver(
-    ...     "sum2",
-    ...     lambda a, b, *, _parent_: _parent_.get(a, 0) + _parent_.get(b, 0),
-    ...     use_cache=False,
-    ... )
+    >>> def sum2(a, b, *, _parent_):
+    ...     return _parent_.get(a, 0) + _parent_.get(b, 0)
+    >>> OmegaConf.register_new_resolver("sum2", sum2, use_cache=False)
     >>> cfg = OmegaConf.create(
     ...     {
     ...         "node": {
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -544,11 +544,10 @@ Note that special parameters must be defined as named keywords (after the `*`):
 .. doctest::
 
     >>> OmegaConf.register_new_resolver(
-    ...    "sum_friends",
-    ...    lambda a, b, *, _parent_: _parent_[a] + _parent_[b],
-    ...    use_cache=False,
-    ...)
-    >>>
+    ...     "sum_friends", 
+    ...     lambda a, b, *, _parent_: _parent_[a] + _parent_[b], 
+    ...     use_cache=False
+    ... )
     >>> cfg = OmegaConf.create(
     ...     {
     ...         "a": {
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -489,13 +489,12 @@ You can take advantage of nested interpolations to perform custom operations ove
 
 .. doctest::
 
-    >>> OmegaConf.register_new_resolver("plus", lambda x, y: x + y)
+    >>> OmegaConf.register_new_resolver("sum", lambda x, y: x + y)
     >>> c = OmegaConf.create({"a": 1,
     ...                       "b": 2,
-    ...                       "a_plus_b": "${plus:${a},${b}}"})
+    ...                       "a_plus_b": "${sum:${a},${b}}"})
     >>> c.a_plus_b
     3
-
 
 More advanced resolver naming features include the ability to prefix a resolver name with a
 namespace, and to use interpolations in the name itself. The following example demonstrates both:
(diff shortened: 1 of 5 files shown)


omegaconf/omegaconf.py
```diff
@@ -508,7 +531,9 @@ def resolver_wrapper(
     @staticmethod
     def get_resolver(
         name: str,
```
omry, line 533: This should probably not be a public method.
it's exposing internal details.

omegaconf/omegaconf.py
```diff
@@ -483,8 +485,22 @@ def register_new_resolver(
             name not in BaseContainer._resolvers
         ), "resolver {} is already registered".format(name)
 
+        sig = inspect.signature(resolver)
+        pass_parent = "_parent_" in sig.parameters
+        pass_root = "_root_" in sig.parameters
+        if pass_parent and use_cache:
+            raise ValueError(
+                "use_cache=True is incompatible with functions that receive the _node_"
```
Jasha10, line 493:
```suggestion
                "use_cache=True is incompatible with functions that receive the _parent_"
```

omegaconf/omegaconf.py
odelalleau: Feel free to change it (this predates #445)

docs/source/usage.rst
```diff
@@ -518,6 +518,36 @@ inputs we always return the same value. This behavior may be disabled by setting
     >>> assert c.uncached != c.uncached
 
 
+
+Custom interpolations can also receive the following special parameters:
+
+- ``_parent_`` : the parent node of an interpolation.
+- ``_root_``: The config root.
+
+This can be achieved by adding ``_parent_`` and ``_root_`` to the resolver signature.
```
odelalleau, line 527:
```suggestion
This can be achieved by adding ``_parent_`` and/or ``_root_`` to the resolver signature.
```

docs/source/usage.rst
```diff
@@ -518,6 +518,36 @@ inputs we always return the same value. This behavior may be disabled by setting
...
+    >>> cfg = OmegaConf.create(
+    ...     {
+    ...         "a": {
+    ...             "b": 1,
+    ...             "c": 2,
+    ...             "sum": "${sum_friends:b,c}",
+    ...         },
+    ...     }
+    ... )
+    >>> cfg.a.sum
```
odelalleau, line 547: Just to make it easier to link the config to the resolver:
```suggestion
    ...         "x": {
    ...             "a": 1,
    ...             "b": 2,
    ...             "sum": "${sum_friends:a,b}",
    ...         },
    ...     }
    ... )
    >>> cfg.x.sum
```

docs/source/usage.rst
```diff
@@ -518,6 +518,36 @@ inputs we always return the same value. This behavior may be disabled by setting
...
+- ``_root_``: The config root.
+
+This can be achieved by adding ``_parent_`` and ``_root_`` to the resolver signature.
+Note that special parameters must be defined as named keywords (after the `*`):
+
+
+.. doctest::
+
+    >>> OmegaConf.register_new_resolver(
+    ...     "sum_friends",
```
odelalleau, line 534: Minor: rename to `sum_siblings`? (more standard graph terminology)

odelalleau (changes requested):
Just some minor doc tweaks (+ @Jasha10's comment)

docs/source/usage.rst
omry: I guess we can just make this sum, where the feature over the previous sum is that this one treats non existing keys as 0.

omegaconf/omegaconf.py
omry: To confuse the enemy, shhhh, the enemy is listening!

omegaconf/omegaconf.py
omry: It can be done in the pass to improve the resolver API (#370).
I think the intended use here is to check if a resolver exists.

omry:
I think I responded to all the comments. Take a second look (not always exactly as requested).

omegaconf/omegaconf.py
odelalleau: So is there a need to keep a public API to check the existence of a resolver?

omry: We can deprecate this one and ad a proper function to check that.

docs/source/usage.rst
```diff
@@ -524,28 +523,33 @@ Custom interpolations can also receive the following special parameters:
...
+(In contrast to the sum we defined earlier where accessing an invalid key, e.g. `"a_plus_z": ${sum:${a}, ${z}}` will result in an error).
 
 .. doctest::
 
     >>> OmegaConf.register_new_resolver(
-    ...     "sum_friends", 
-    ...     lambda a, b, *, _parent_: _parent_[a] + _parent_[b], 
-    ...     use_cache=False
+    ...     "sum2",
+    ...     lambda a, b, *, _parent_: _parent_.get(a, 0) + _parent_.get(b, 0),
```
odelalleau, line 536: Unfortunately this line just slightly too long to fit in the doc code box.
Suggestion:
```python
    >>> def sum2(a, b, *, _parent_):
    ...     return _parent_.get(a, 0) + _parent_.get(b, 0)
    >>> OmegaConf.register_new_resolver(
    ...     "sum2",
    ...     sum2,
    ...     use_cache=False,
    ... )
```
(when I switch the default value of `use_cache` to False it will look better)

odelalleau (changes requested):
Just a minor doc formatting issue

docs/source/usage.rst
omry: Good catch. did you check out the PR and built the doc to check?

odelalleau: Yes -- this got me a few times before so it caught my eye

odelalleau (approved):
Great!

omry:
The last PR I just merged is breaking one of the tests.
It's hard to debug because it gets into an infinite loop, but I suspect the problem is that the tuple(cfg, cfg) is being converted to a ListConfig which in turns causes some issues.

omry:
I worked around it by not returning a tuple, but this is something we did not anticipate.

omry:
It might make sense to allow control over the generation of config nodes from dicts, list and tuples returned by custom resolvers.
to avoid surprises like this one we might want to default to not generate.

Side note, this exposed a bug: the code before my fix got into an infinite loop. Not sure what this is about, worth digging further.

odelalleau:
> It might make sense to allow control over the generation of config nodes from dicts, list and tuples returned by custom resolvers.
> to avoid surprises like this one we might want to default to not generate.
> 
> Side note, this exposed a bug: the code before my fix got into an infinite loop. Not sure what this is about, worth digging further.

That's interesting one. It's caused by a cycle in the parent -> child dependency graph. Minimum repro:
```python
c1 = OmegaConf.create({"x": []})
c1.x.append(c1)
c1.x
```

I'm working on a PR to detect such cycles in typical situations, to at least get an explicit error instead of an infinite loop. Regarding an actual fix, not sure if it'd be possible without major changes (like allowing multiple parents), but a couple of ideas (which I haven't fully thought through):
- When appending a node to a ListConfig (and possibly in other similar situations), don't set the parent if the node is already an ancestor of the ListConfig
- Or, create a wrapper node that would behave a bit like current interpolated nodes (could even treat them as another kind of interpoolation). The wrapper node would have the ListConfig as parent, but the underlying node's parent wouldn't be modified

omry:
Yeah, I dug a bit as well and I think it's a pretty unique circumstances.
Essentially the object contains cycles which means it cannot be traversed. Normal traversal would cause a stack-overflow which is not that bad, but somehow this results in an infinite loop.

I didn't dug enough to see why this is happening, but my intuition matches what are describing.
Ideally this would result in an exception, but I think it's not a huge problem.

I have a handling for this situation [here](https://github.com/omry/omegaconf/blob/8740730c72ef9c2cac638f436ac0f7a98f9a6281/omegaconf/basecontainer.py#L498-L500) for some scenarios when adding a node that belongs to the same config.
I think a fix here can be similar but to check if a added node is in the parent chain of the node it's being added to.

Supporting two parents would break interpolation.

parent1:
```
node:
  foo: 10
```
parent2:
```
node:
  foo: 20
```
```python
cfg1 = OmegaConf.create(parent1)
cfg2 = OmegaConf.create(parent2)
cfg3 = OmegaConf.create({"bar" : "${a.node.foo}"})
cfg1.a = cfg3 # cfg1 is now a parent
cfg2.a = cfg3 # cfg2 is now a parent too (?)

# At this point, what is the value of ${a.node.foo} ?
```

odelalleau:
> Ideally this would result in an exception, but I think it's not a huge problem.

Good :) Because I gave up on my PR, it turns out there are too many intertwined functionalities that break when there is a cycle in the parents. The simple approach I started with is too limited, and other potential solutions I can think of would be trickier to implement, so I'm not going to address this now.

> I have a handling for this situation [here](https://github.com/omry/omegaconf/blob/8740730c72ef9c2cac638f436ac0f7a98f9a6281/omegaconf/basecontainer.py#L498-L500) for some scenarios when adding a node that belongs to the same config.
> I think a fix here can be similar but to check if a added node is in the parent chain of the node it's being added to.

The specific line you pointed too isn't super relevant though (I think), but what's around it is: the important point is that currently the node is deepcopied in `_set_item_impl()` by default. This happens in a ListConfig when calling `c[idx] = node`, but not in `c.append(node)`. Is that intended?

> Supporting two parents would break interpolation.

I'm sure it would break a lot of things :) Not going to suggest that we should get into that.

omry:
> The specific line you pointed too isn't super relevant though (I think), but what's around it is: the important point is that currently the node is deepcopied in _set_item_impl() by default. This happens in a ListConfig when calling c[idx] = node, but not in c.append(node). Is that intended?

Not intentional.
I think we can deepcopy an input Node (unless that flag I pointed to is set).
That might actually solve this issue.

odelalleau:
> That might actually solve this issue (is append on the path here?)

Yes. The path is _node_wrap() -> ListConfig.__init__() -> _set_value() -> _set_value_impl() -> append()

I created an issue to keep track of it: #601

PR #602, opened 2021-03-14
Avoid infinite loop in `_get_full_key()`

odelalleau:
When there was a cycle when iterating over parents of a node,
`_get_full_key()` would get into an infinite loop.

This commit detects this situation to raise an explicit exception
instead.

In addition, `format_and_raise()` is modified to *not* raise exceptions
when trying to obtain the full key, so as to avoid hiding the original
exception with a different error. Instead, any exception raised when
trying to obtain the full key is displayed on the "full key" line of the
error message.

### DIFF
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -615,7 +615,7 @@ def format_and_raise(
         except Exception as exc:
             # Since we are handling an exception, raising a different one here would
             # be misleading. Instead, we display it in the key.
-            full_key = f"<unresolvable due to {type(exc).__name__}: {exc}"
+            full_key = f"<unresolvable due to {type(exc).__name__}: {exc}>"
 
         object_type = OmegaConf.get_type(node)
         object_type_str = type_str(object_type)
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -610,7 +610,12 @@ def format_and_raise(
         if key is not None and not node._is_none():
             child_node = node._get_node(key, validate_access=False)
 
-        full_key = node._get_full_key(key=key)
+        try:
+            full_key = node._get_full_key(key=key)
+        except Exception as exc:
+            # Since we are handling an exception, raising a different one here would
+            # be misleading. Instead, we display it in the key.
+            full_key = f"<unresolvable due to {type(exc).__name__}: {exc}"
 
         object_type = OmegaConf.get_type(node)
         object_type_str = type_str(object_type)
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -28,8 +28,8 @@ from ._utils import (
 )
 from .base import Container, ContainerMetadata, DictKeyType, Node, SCMode
 from .errors import (
+    ConfigCycleDetectedException,
     MissingMandatoryValue,
-    OmegaConfBaseException,
     ReadonlyConfigError,
     ValidationError,
 )
@@ -732,7 +732,7 @@ class BaseContainer(Container, ABC):
         while cur._get_parent() is not None:
             cur = cur._get_parent()
             if id(cur) in memo:
-                raise OmegaConfBaseException(
+                raise ConfigCycleDetectedException(
                     f"Cycle when iterating over parents of key `{key}`"
                 )
             memo.add(id(cur))
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -27,7 +27,12 @@ from ._utils import (
     is_structured_config,
 )
 from .base import Container, ContainerMetadata, DictKeyType, Node, SCMode
-from .errors import MissingMandatoryValue, ReadonlyConfigError, ValidationError
+from .errors import (
+    MissingMandatoryValue,
+    OmegaConfBaseException,
+    ReadonlyConfigError,
+    ValidationError,
+)
 
 if TYPE_CHECKING:
     from .dictconfig import DictConfig  # pragma: no cover
@@ -723,11 +728,16 @@ class BaseContainer(Container, ABC):
             full_key = self._key()
 
         assert cur is not None
+        memo = {id(cur)}  # remember already visited nodes so as to detect cycles
         while cur._get_parent() is not None:
             cur = cur._get_parent()
+            if id(cur) in memo:
+                raise OmegaConfBaseException(
+                    f"Cycle when iterating over parents of key `{key}`"
+                )
+            memo.add(id(cur))
             assert cur is not None
-            key = cur._key()
-            if key is not None:
+            if cur._key() is not None:
                 full_key = prepand(
                     full_key, type(cur._get_parent()), type(cur), cur._key()
                 )
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -1,4 +1,5 @@
 import re
+import sys
 from dataclasses import dataclass
 from enum import Enum
 from textwrap import dedent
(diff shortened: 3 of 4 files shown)


Jasha10:
Usually we will want our configs to be a directed acyclic graph, right? I've been thinking about adding a note in the docs to say that non-acyclic configs are not well supported...

odelalleau:
> Usually we will want our configs to be a directed acyclic graph, right? I've been thinking about adding a note in the docs to say that non-acyclic configs are not well supported...

Yes definitely, cycles are bad and we don't want them. The purpose of this PR is to have a clear error if a cycle still happens by accident (to make debugging easier -- infinite loops aren't great for this)

tests/test_errors.py
```diff
@@ -1333,3 +1334,44 @@ def test_parse_error_on_creation(create_func: Any, arg: Any) -> None:
...
+    assert isinstance(x_node, DictConfig)
+    # We create a cycle in the parent relationship that will trigger a RecursionError
+    # when trying to access `c.x`. This test verifies that this RecursionError is properly
+    # raised even if another exception occurs in `format_and_raise()` when trying to
+    # obtain the full key.
+    c._set_parent(x_node)
+
+    # For some reason in Python 3.6 the error message may vary depending on how this
+    # test is run (`nox -s coverage-3.6` vs `nox -s omegaconf-3.6`). As a result, we
+    # do not check the message.
```
omry, line 1363: I remember having some bad time trying to test recursion in previous versions. iirc in 2.7 Python the whole process was terminating sometimes without a proper error.
I am okay with limiting this testing to 3.7+ if it's stable there.

tests/test_errors.py
```diff
@@ -1333,3 +1334,44 @@ def test_parse_error_on_creation(create_func: Any, arg: Any) -> None:
...
+            dedent(
+                """\
+                maximum recursion depth exceeded in comparison
+                    full_key: <unresolvable due to OmegaConfBaseException: Cycle when iterating over parents of key `x`
+                """
+            )
+        )
+
+    with pytest.raises(RecursionError, match=match):
+        c.x
```
omry, line 1377: I am curious, where in that flow are we comparing objects (based on the error message, we are).

tests/test_errors.py
odelalleau: Yes seems stable in 3.7+.
The test is also always raising a `RecursionError` in 3.6, but it doesn't always have the same error message... so I'm just skipping the test of the message for 3.6.

tests/test_errors.py
odelalleau: Good question, I don't really understand why, since here the recursion is in `_get_flag()` and I see no comparison...
I thought it was just the standard recursion error message, but trying to repro on a simpler example, the "in comparison" doesn't show.
So I'm afraid I don't have a good answer for this. Not sure it's worth digging too deep into it...

tests/test_errors.py
omry: As I said, In the past I saw that RecursionError was sometimes really crashing the Python process so I am a bit concerned about testing it at all.
If this is stable we can go with it though.

tests/test_errors.py
omry: Maybe put a breakpoint in the `___eq__` override and see if you hit it?

tests/test_errors.py
odelalleau: > If this is stable we can go with it though.

It's been stable in my tests so far. If it turns out not to be, it's easy to disable this test until a replacement is found.

tests/test_errors.py
odelalleau: So actually it's because of the `assert isinstance(parent, Container)` in `Node._get_parent()`. It's a bit messed up (as far as I can tell the `isinstance()` call is *not* part of the recursion stack trace, except when Python throws), but that's how it is right now...
If this breaks in the future I'll just handcraft a different error for the purpose of the test.

omry: I don't follow:
Why does `assert parent is None or isinstance(parent, Container)` trigger an equality check?

odelalleau: There's no equality check, my (admittedly superficial) understanding is that `isinstance()` is considered a comparison, and somehow when the recursion error triggers it is linked to this check. But it seems to be an implementation detail mostly -- messing around with the code I was able to get it to have / not have the "in comparison" part of the error just with irrelevant print statements.

I changed the test in f387495 to just not test the specific exeption message coming from Python, since it doesn't seem reliable.

omry: Okay, I don't think instancecheck is comparing the object.
it should be an operating on the type of the object.

omegaconf/basecontainer.py
```diff
@@ -723,11 +728,16 @@ def prepand(full_key: str, parent_type: Any, cur_type: Any, key: Any) -> str:
             full_key = self._key()
 
         assert cur is not None
+        memo = {id(cur)}  # remember already visited nodes so as to detect cycles
         while cur._get_parent() is not None:
             cur = cur._get_parent()
+            if id(cur) in memo:
+                raise OmegaConfBaseException(
```
omry, line 735: I didn't really intend OmegaConfBaseException to be used directly.

We can add a user facing generic exception type `OmegaConfException(OmegaConfBaseException)`, or just use one of the built in exception types here (I can't think of any that fits well though).

omegaconf/_utils.py
```diff
@@ -610,7 +610,12 @@ def format_and_raise(
         if key is not None and not node._is_none():
             child_node = node._get_node(key, validate_access=False)
 
-        full_key = node._get_full_key(key=key)
+        try:
+            full_key = node._get_full_key(key=key)
+        except Exception as exc:
+            # Since we are handling an exception, raising a different one here would
+            # be misleading. Instead, we display it in the key.
+            full_key = f"<unresolvable due to {type(exc).__name__}: {exc}"
```
omry, line 618:
```suggestion
            full_key = f"<unresolvable due to {type(exc).__name__}: {exc}>"
```

omegaconf/basecontainer.py
odelalleau: The generic builtin type is `RuntimeError` => I could just raise that if we don't really care about it inheriting from `OmegaConfBaseException` (another option could be `RecursionError`, which is thematically fit but technically wrong since there is no actual recursion).

Two other options would be:
- `OmegaConfException(OmegaConfBaseException)` as you suggested
- A more specific `ConfigDependencyError(OmegaConfBaseException)` which could be re-used in the future in any other situation where we detect a similar problem with the dependency graph between nodes.

I don't have a strong opinion on any of these, so let me know what you prefer.

omegaconf/_utils.py
odelalleau: Taking suggestion => resolving

tests/test_errors.py
odelalleau: > Okay, I don't think instancecheck is comparing the object.

It doesn't -- at least not in the sense of calling `__eq__()`. However, for the purpose of Python's max recursion error message, it seems to be considered a comparison.

omegaconf/basecontainer.py
omry: I guess with could introduce a `ConfigCycleDetectedException(OmegaConfBaseException)` in the spirit of your last suggestion.
seems generic enough and we may reuse it in other similar scenarios.

odelalleau: Sounds good, done in 650e710

omry (approved):
awesome.

PR #603, opened 2021-03-15
pydevd debugger plugin

### LINKED ISSUE #214: Consider implementing a pydev resolver plugin
https://github.com/fabioz/PyDev.Debugger/tree/master/pydevd_plugins/extensions

omry:
Closes #214

This is more or less ready for initial usage.
Left to do:
- [ ] Document (?)
- [ ] Bug fixes (Just looking at the screen shots I am already seeing a few issues).

This is implementing a pydevd resolver for OmegaConf.
pydev is used in PyCharm and vscode as the Python debugger.
The resolver is aware of the special semantics of OmegaConf containers and handles missing, interpolations etc in a better way.
A future enhancement can be to expose the underlying Structured Config type in the debugger.

image

image

### DIFF
--- a/news/214.feature
+++ b/news/214.feature
@@ -1,0 +1,1 @@
+New pydevd resolver plugin for easier debugging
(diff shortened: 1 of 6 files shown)


omry:
I need to revisit this, it doesn't quiet do what I want. Closing for now but will keep trying.

Issue #230, opened 2020-04-23
null as default value for env interpolation


apsdehal:
If I pass null as default value to env interpolation, as with yaml, I expect to get the value as None, but I get the string 'null'.

```py
from omegaconf import OmegaConf
a = OmegaConf.create({"a": "${env:a, null}"})
print(a.a) # prints "null"
```

If this is a not a bug, how can I specify None as the default value?

omry:
smells like a bug to me.

omry:
PR welcome :)

omry:
The [docs](https://omegaconf.readthedocs.io/en/latest/usage.html#custom-interpolations) does not specify the type of arguments and currently they are strings.
```python
OmegaConf.register_resolver("concat", lambda x,y: print(type(x), type(y), x+y))
cfg = OmegaConf.create({"foo" : "${concat:10,foo}"})
<class 'str'> <class 'str'> 10foo
cfg.foo
```
There is a case to make for converting the strings to a primitive or even with yaml_decode but the parsing now is very primitive and I really don't want to support something as sophisticated as yaml there.

Let me think about it, I am leaning toward keeping things as is. giving arbitrary types may not be simpler for the receiving code. for example the registered resolver will need to handle a plathora of types based on what the user is passing in which will probably make most cases more complicated, only to simplify one case (yours).

apsdehal:
A current workaround for this is to pass empty string as default parameter. This will give empty string back. Then `if` will evaluate it directly to false.

omry:
default value where? code? config?

apsdehal:
`OmegaConf.create({"a": "${env:a,}"})`

omry:
Okay, I can consider adding a flag to register resolver that will cause the variables to be converted to primitives for 2.1.

omry:
blocked on #445

PR #606, opened 2021-03-15
Introduce new `oc.env` and `oc.decode` resolvers

### LINKED ISSUE #383: Add env_str resolver built-in resolver
In some cases the automatic conversion of environment variable does not work well or is not what the user wants.
A simple fix is to add another resolver named "env_str" that would return the env string as is.

### LINKED ISSUE #573: Deprecation of the `env` resolver for variable environment interpolations
In OmegaConf 2.1, the `env` resolver is being deprecated in favor of a new `oc.env` resolver.

To migrate to this new resolver:
* Replace `${env:MYVAR}` with `${oc.env:MYVAR}`
* Consider whether your config or code may need to be updated due to the following three main differences with `oc.env`:
  1. Contrary to `env`, `oc.env` does not use the resolver cache: this means that if `MYVAR` is updated within the program's process, the value of `${oc.env:MYVAR}` will change accordingly.
  2. Contrary to `env`, `oc.env` does not attempt to convert booleans, integers and floats: instead, the string representation of the environment variable is always returned. If you wish to convert from string to a different type, you may either take advantage of automatic conversion for typed nodes (see https://github.com/omry/omegaconf/issues/488#issuecomment-774490332), or use the new `oc.decode` resolver (see #574)
  3. The default value of `oc.env` can only be a string or `null`. For instance, `${env:MISSING,123}` should be replaced with `${oc.env:MISSING,"123"}` (and if the `MISSING` environment variable does not exist then the string `"123"` will be returned instead of the integer `123`)

### LINKED ISSUE #574: Add an `oc.decode` resolver to parse and evaluate strings

With `env` being deprecated in favor of `oc.env` (#573), which always returns environment variables as strings, we need an easy way to convert strings to common types like bool, int, float, etc.

Introduce a new resolver `oc.decode` that converts a string automatically, using the common OmegaConf grammar.
For instance:

```python
cfg = OmegaConf.create(
    {
        "database": {
            "port": '${oc.decode:${oc.env:DB_PORT}}',
            "nodes": '${oc.decode:${oc.env:DB_NODES}}',
        }
    }
)

os.environ["DB_PORT"] = '3308'  # integer
os.environ["DB_NODES"] = '[host1, host2, host3]'  # list

print(repr(cfg.database.port))
print(repr(cfg.database.nodes))
```
should print:
```
3308
['host1', 'host2', 'host3']
```

In addition, `${oc.decode:null}` is valid and returns `None`, in order to support for instance `${oc.decode:${oc.env:DB_PORT,null}}`

odelalleau:
* Restore and deprecate the old `env` resolver for backward
  compatibility with OmegaConf 2.0

* The new `oc.env` resolver keeps the string representation of
  environment variables, and does not use the cache

* The new `oc.decode` resolver can be used to parse and evaluate strings
  according to the OmegaConf grammar

Fixes #383
Fixes #573
Fixes #574

Notes:
* the diff for the notebook can be seen at https://odelalleau.github.io/diff/diff_notebook_decode.html
* the diff of test_interpolation.py is hard to read because there was an indentation change => I created a dummy diff at https://github.com/odelalleau/omegaconf/pull/2/files that doesn't have the indentation change for better readability

### DIFF
--- a/docs/notebook/Tutorial.ipynb
+++ b/docs/notebook/Tutorial.ipynb
@@ -710,8 +710,7 @@
    "source": [
     "# Let's set up the environment first (only needed for this demonstration)\n",
     "import os\n",
-    "os.environ['USER'] = 'omry'\n",
-    "os.environ['USERID'] = '123456'"
+    "os.environ['USER'] = 'omry'"
    ]
   },
   {
@@ -737,7 +736,6 @@
       "user:\n",
       "  name: ${oc.env:USER}\n",
       "  home: /home/${oc.env:USER}\n",
-      "  id: ${oc.env:USERID}\n",
       "\n"
      ]
     }
@@ -763,7 +761,6 @@
       "user:\n",
       "  name: omry\n",
       "  home: /home/omry\n",
-      "  id: '123456'\n",
       "\n"
      ]
     }
--- a/docs/notebook/Tutorial.ipynb
+++ b/docs/notebook/Tutorial.ipynb
@@ -775,9 +775,9 @@
    "metadata": {},
    "source": [
     "You can specify a default value to use in case the environment variable is not defined.\n",
-    "This default value must be a string (with the exception of `${oc.env:VAR,null}` that returns `None` if `VAR` is not defined).\n",
+    "This default value can be a string or ``null`` (representing Python ``None``). Passing a default with a different type will result in an error.\n",
     "\n",
-    "The following example sets `abc123` as the the default value when `DB_PASSWORD` is not defined:"
+    "The following example sets default database passwords when ``DB_PASSWORD`` is not defined:"
    ]
   },
   {
@@ -786,20 +786,26 @@
    "metadata": {},
    "outputs": [
     {
-     "data": {
-      "text/plain": [
-       "'abc123'"
-      ]
-     },
-     "execution_count": 25,
-     "metadata": {},
-     "output_type": "execute_result"
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "'abc123'\n",
+      "'12345'\n"
+     ]
     }
    ],
    "source": [
     "os.environ.pop('DB_PASSWORD', None)  # ensure env variable does not exist\n",
-    "cfg = OmegaConf.create({'database': {'password': '${oc.env:DB_PASSWORD,abc123}'}})\n",
-    "cfg.database.password"
+    "cfg = OmegaConf.create(\n",
+    "    {\n",
+    "        \"database\": {\n",
+    "            \"password1\": \"${oc.env:DB_PASSWORD,abc123}\",  # the string 'abc123'\n",
+    "            \"password2\": \"${oc.env:DB_PASSWORD,'12345'}\",  # the string '12345'\n",
+    "        },\n",
+    "    }\n",
+    ")\n",
+    "print(repr(cfg.database.password1))\n",
+    "print(repr(cfg.database.password2))"
    ]
   },
   {
--- a/docs/notebook/Tutorial.ipynb
+++ b/docs/notebook/Tutorial.ipynb
@@ -698,7 +698,8 @@
    "source": [
     "## Environment variable interpolation\n",
     "\n",
-    "Environment variable interpolation is also supported."
+    "Environment variable interpolation is also supported.\n",
+    "An environment variable is always returned as a string."
    ]
   },
   {
@@ -709,14 +710,15 @@
    "source": [
     "# Let's set up the environment first (only needed for this demonstration)\n",
     "import os\n",
-    "os.environ['USER'] = 'omry'"
+    "os.environ['USER'] = 'omry'\n",
+    "os.environ['USERID'] = '123456'"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Here is an example config file interpolates with the USER environment variable:"
+    "Here is an example config file interpolating with the USER and USERID environment variables:"
    ]
   },
   {
@@ -733,8 +735,9 @@
      "output_type": "stream",
      "text": [
       "user:\n",
-      "  name: ${env:USER}\n",
-      "  home: /home/${env:USER}\n",
+      "  name: ${oc.env:USER}\n",
+      "  home: /home/${oc.env:USER}\n",
+      "  id: ${oc.env:USERID}\n",
       "\n"
      ]
     }
@@ -760,6 +763,7 @@
       "user:\n",
       "  name: omry\n",
       "  home: /home/omry\n",
+      "  id: '123456'\n",
       "\n"
      ]
     }
@@ -773,7 +777,10 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "You can specify a default value to use in case the environment variable is not defined. The following example sets `abc123` as the the default value when `DB_PASSWORD` is not defined."
+    "You can specify a default value to use in case the environment variable is not defined.\n",
+    "This default value must be a string (with the exception of `${oc.env:VAR,null}` that returns `None` if `VAR` is not defined).\n",
+    "\n",
+    "The following example sets `abc123` as the the default value when `DB_PASSWORD` is not defined:"
    ]
   },
   {
@@ -794,7 +801,7 @@
    ],
    "source": [
     "os.environ.pop('DB_PASSWORD', None)  # ensure env variable does not exist\n",
-    "cfg = OmegaConf.create({'database': {'password': '${env:DB_PASSWORD,abc123}'}})\n",
+    "cfg = OmegaConf.create({'database': {'password': '${oc.env:DB_PASSWORD,abc123}'}})\n",
     "cfg.database.password"
    ]
   },
@@ -802,7 +809,17 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Environment variables are parsed when they are recognized as valid quantities that may be evaluated (e.g., int, float, dict, list):"
+    "## Decoding strings with interpolations"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "You can automatically convert a string to its corresponding type (e.g., bool, int, float, dict, list) using `oc.decode`.\n",
+    "This resolver also accepts ``None`` as input, in which case it returns ``None``.\n",
+    "\n",
+    "This can be useful for instance to parse environment variables:"
    ]
   },
   {
@@ -814,25 +831,30 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "3308\n",
-      "['host1', 'host2', 'host3']\n",
-      "'a%#@~{}$*&^?/<'\n"
+      "port (int): 3308\n",
+      "nodes (list): ['host1', 'host2', 'host3']\n",
+      "nodes (None due to missing variable): None\n"
      ]
     }
    ],
    "source": [
-    "cfg = OmegaConf.create({'database': {'password': '${env:DB_PASSWORD,abc123}',\n",
-    "                                    'user': 'someuser',\n",
-    "                                    'port': '${env:DB_PORT,3306}',\n",
-    "                                    'nodes': '${env:DB_NODES,[]}'}})\n",
+    "cfg = OmegaConf.create(\n",
+    "    {\n",
+    "        \"database\": {\n",
+    "            \"port\": '${oc.decode:${oc.env:DB_PORT}}',\n",
+    "            \"nodes\": '${oc.decode:${oc.env:DB_NODES,null}}',\n",
+    "        }\n",
+    "    }\n",
+    ")\n",
     "\n",
-    "os.environ[\"DB_PORT\"] = '3308'  # integer\n",
-    "os.environ[\"DB_NODES\"] = '[host1, host2, host3]'  # list\n",
-    "os.environ[\"DB_PASSWORD\"] = 'a%#@~{}$*&^?/<'  # string\n",
+    "os.environ[\"DB_PORT\"] = \"3308\"  # integer\n",
+    "os.environ[\"DB_NODES\"] = \"[host1, host2, host3]\"  # list\n",
     "\n",
-    "print(repr(cfg.database.port))\n",
-    "print(repr(cfg.database.nodes))\n",
-    "print(repr(cfg.database.password))"
+    "print(\"port (int):\", repr(cfg.database.port))\n",
+    "print(\"nodes (list):\", repr(cfg.database.nodes))\n",
+    "\n",
+    "del os.environ[\"DB_NODES\"]\n",
+    "print(\"nodes (None due to missing variable):\", repr(cfg.database.nodes))"
    ]
   },
   {
--- a/docs/notebook/Tutorial.ipynb
+++ b/docs/notebook/Tutorial.ipynb
@@ -717,7 +717,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Here is an example config file interpolating with the USER and USERID environment variables:"
+    "Here is an example config file interpolates with the USER environment variable:"
    ]
   },
   {
--- a/docs/notebook/Tutorial.ipynb
+++ b/docs/notebook/Tutorial.ipynb
@@ -816,7 +816,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "You can automatically convert a string to its corresponding type (e.g., bool, int, float, dict, list) using `oc.decode`.\n",
+    "You can automatically convert a string to its corresponding type (e.g., bool, int, float, dict, list) using `oc.decode` (which can even resolve interpolations).\n",
     "This resolver also accepts ``None`` as input, in which case it returns ``None``.\n",
     "\n",
     "This can be useful for instance to parse environment variables:"
@@ -833,7 +833,8 @@
      "text": [
       "port (int): 3308\n",
       "nodes (list): ['host1', 'host2', 'host3']\n",
-      "nodes (None due to missing variable): None\n"
+      "timeout (missing variable): None\n",
+      "timeout (interpolation): 3308\n"
      ]
     }
    ],
@@ -843,18 +844,21 @@
     "        \"database\": {\n",
     "            \"port\": '${oc.decode:${oc.env:DB_PORT}}',\n",
     "            \"nodes\": '${oc.decode:${oc.env:DB_NODES,null}}',\n",
+    "            \"timeout\": '${oc.decode:${oc.env:DB_TIMEOUT,null}}',\n",
     "        }\n",
     "    }\n",
     ")\n",
     "\n",
     "os.environ[\"DB_PORT\"] = \"3308\"  # integer\n",
     "os.environ[\"DB_NODES\"] = \"[host1, host2, host3]\"  # list\n",
+    "os.environ.pop(\"DB_TIMEOUT\", None)  # unset variable\n",
     "\n",
     "print(\"port (int):\", repr(cfg.database.port))\n",
     "print(\"nodes (list):\", repr(cfg.database.nodes))\n",
+    "print(\"timeout (missing variable):\", repr(cfg.database.timeout))\n",
     "\n",
-    "del os.environ[\"DB_NODES\"]\n",
-    "print(\"nodes (None due to missing variable):\", repr(cfg.database.nodes))"
+    "os.environ[\"DB_TIMEOUT\"] = \"${.port}\"\n",
+    "print(\"timeout (interpolation):\", repr(cfg.database.timeout))"
    ]
   },
   {
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -340,13 +340,9 @@ Example:
     >>> # Primitive interpolation types are inherited from the reference
     >>> show(conf.client.server_port)
     type: int, value: 80
+    >>> # String interpolations concatenate fragments into a string
     >>> show(conf.client.description)
     type: str, value: 'Client of http://localhost:80/'
-
-    >>> # Composite interpolation types are always string
-    >>> show(conf.client.url)
-    type: str, value: 'http://localhost:80/'
-
 
 Interpolations may be nested, enabling more advanced behavior like dynamically selecting a sub-config:
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -422,9 +422,13 @@ The following example sets ``abc123`` as the default value when ``DB_PASSWORD`` 
 Decoding strings with interpolations
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-You can automatically convert a string to its corresponding type (e.g., bool, int, float, dict, list) using ``oc.decode``
-(which can even resolve interpolations).
-This resolver also accepts ``None`` as input, in which case it returns ``None``.
+Strings may be converted using ``oc.decode``:
+
+- Primitive values (e.g., ``"true"``, ``"1"``, ``"1e-3"``) are automatically converted to their corresponding type (bool, int, float)
+- Dictionaries and lists (e.g., ``"{a: b}"``, ``"[a, b, c]"``)  are returned as transient config nodes (DictConfig and ListConfig)
+- Interpolations (e.g., ``"${foo}"``) are automatically resolved
+- ``None`` is the only valid non-string input to ``oc.decode`` (returning ``None`` in that case)
+
 This can be useful for instance to parse environment variables:
 
 .. doctest::
@@ -438,15 +442,19 @@ This can be useful for instance to parse environment variables:
     ...         }
     ...     }
     ... )
+    >>> def show(x):
+    ...     print(f"type: {type(x).__name__}, value: {x}")
     >>> os.environ["DB_PORT"] = "3308"
-    >>> cfg.database.port  # converted to int
-    3308
+    >>> show(cfg.database.port)  # converted to int
+    type: int, value: 3308
     >>> os.environ["DB_NODES"] = "[host1, host2, host3]"
-    >>> cfg.database.nodes  # converted to list
-    ['host1', 'host2', 'host3']
-    >>> assert cfg.database.timeout is None  # keeping `None` as is
+    >>> show(cfg.database.nodes)  # converted to a ListConfig
+    type: ListConfig, value: ['host1', 'host2', 'host3']
+    >>> show(cfg.database.timeout)  # keeping `None` as is
+    type: NoneType, value: None
     >>> os.environ["DB_TIMEOUT"] = "${.port}"
-    >>> assert cfg.database.timeout == 3308  # resolving interpolation
+    >>> show(cfg.database.timeout)  # resolving interpolation
+    type: int, value: 3308
 
 
 Custom interpolations
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -6,7 +6,6 @@
     import tempfile
     import pickle
     os.environ['USER'] = 'omry'
-    os.environ['USERID'] = '123456'
     os.environ.pop('DB_TIMEOUT', None)
 
 .. testsetup:: loaded
@@ -403,8 +402,6 @@ Input YAML file:
     'omry'
     >>> conf.user.home
     '/home/omry'
-    >>> conf.user.id  # returned as string, even if it is a number
-    '123456'
 
 You can specify a default value to use in case the environment variable is not defined.
 This default value can be a string or `null` (representing Python `None`). Passing a default with a different type will result in an error.
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -404,16 +404,23 @@ Input YAML file:
     '/home/omry'
 
 You can specify a default value to use in case the environment variable is not defined.
-This default value can be a string or `null` (representing Python `None`). Passing a default with a different type will result in an error.
-The following example sets ``abc123`` as the default value when ``DB_PASSWORD`` is not defined.
+This default value can be a string or ``null`` (representing Python ``None``). Passing a default with a different type will result in an error.
+The following example sets default database passwords when ``DB_PASSWORD`` is not defined:
 
 .. doctest::
 
-    >>> cfg = OmegaConf.create({
-    ...       'database': {'password': '${oc.env:DB_PASSWORD,abc123}'}
-    ... })
-    >>> cfg.database.password
+    >>> cfg = OmegaConf.create(
+    ...     {
+    ...         "database": {
+    ...             "password1": "${oc.env:DB_PASSWORD,abc123}",
+    ...             "password2": "${oc.env:DB_PASSWORD,'12345'}",
+    ...         },
+    ...     }
+    ... )
+    >>> cfg.database.password1  # the string 'abc123'
     'abc123'
+    >>> cfg.database.password2  # the string '12345'
+    '12345'
 
 
 Decoding strings with interpolations
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -385,8 +385,7 @@ Interpolated nodes can be any node in the config, not just leaf nodes:
 Environment variable interpolation
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Environment variable interpolation is also supported.
-An environment variable is always returned as a string.
+Access to environment variables is supported using ``oc.env``:
 
 Input YAML file:
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -402,7 +402,7 @@ Input YAML file:
 
 You can specify a default value to use in case the environment variable is not defined.
 This default value can be a string or ``null`` (representing Python ``None``). Passing a default with a different type will result in an error.
-The following example sets default database passwords when ``DB_PASSWORD`` is not defined:
+The following example falls back to default passwords when ``DB_PASSWORD`` is not defined:
 
 .. doctest::
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -407,7 +407,7 @@ Input YAML file:
     '123456'
 
 You can specify a default value to use in case the environment variable is not defined.
-This default value must be a string (with the exception of ``${oc.env:VAR,null}`` that returns ``None`` if ``VAR`` is not defined).
+This default value can be a string or `null` (representing Python `None`). Passing a default with a different type will result in an error.
 The following example sets ``abc123`` as the default value when ``DB_PASSWORD`` is not defined.
 
 .. doctest::
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -341,6 +341,9 @@ Example:
     >>> show(conf.client.server_port)
     type: int, value: 80
     >>> # String interpolations concatenate fragments into a string
+    >>> show(conf.client.url)
+    type: str, value: 'http://localhost:80/'
+    >>> # Relative interpolation example
     >>> show(conf.client.description)
     type: str, value: 'Client of http://localhost:80/'
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -6,6 +6,7 @@
     import tempfile
     import pickle
     os.environ['USER'] = 'omry'
+    os.environ['USERID'] = '123456'
 
 .. testsetup:: loaded
 
@@ -387,6 +388,7 @@ Environment variable interpolation
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Environment variable interpolation is also supported.
+An environment variable is always returned as a string.
 
 Input YAML file:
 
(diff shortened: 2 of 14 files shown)


odelalleau:
@omry in this first version `oc.decode` can't resolve interpolations. I want to take advantage of the new `_parent_` feature you added recently to make it possible, however in order to properly resolve relative interpolations I think I'll also need a new special `_key_` argument. Any objection to me adding it?

odelalleau:
Marked as draft as I want to change `oc.decode` to work with interpolations too. Initially I thought I needed something new, but actually the `_parent_` keyword seems enough => I will update the PR later and request review when ready.

omry:
Didn't look at the code yet, but a question:
if you decode a list or a dict, are they convert to ListConfig and DictConfig now?
If so, is this the intention?

odelalleau:
> Didn't look at the code yet, but a question:
> if you decode a list or a dict, are they convert to ListConfig and DictConfig now?
> If so, is this the intention?

Oh, yes, that's a good point. As far as I am concerned this is intended, however this is a bit tricky to explain in the documentation because currently the fact that dict/list outputs are converted to DictConfig/ListConfig is documented *after* the default resolvers.

Any objection to moving things around? (i.e, first explain what resolvers are and how they behave, then talk about `oc.env` and `oc.decode`)

omry:
Two points:
1. I can imagine scenarios where certain resolvers would need to opt out of of the automatic conversion. in this case it's fine but imagine a resolver that wanted to return an object that is incompatible with OmegaConf inside a list?

2. I think we should split the documentation of the built in resolvers from the documentation how you need to know when you write new resolvers. There are two audiences here. People using the resolvers is a superset of people writing resolvers. I think we can document each resolver based on it's behavior without linking to how this is achieved.
In this particular case, we should just say that the oc.decode returns a transient config node when it sees list and dict. you can also demonstrate the usefulness by showing a usecase with interpolation in the decoded object.

odelalleau:
I requested review but that's for the code only -- doc still needs updating according to ongoing discussion

odelalleau:
> 1. I can imagine scenarios where certain resolvers would need to opt out of of the automatic conversion. in this case it's fine but imagine a resolver that wanted to return an object that is incompatible with OmegaConf inside a list?

If I'm not mistaken, right now the only way for someone to have an object in the config that isn't compatible with OmegaConf is by setting the `allow_objects` flag. 
If this is correct and we are ok to keep it that way, then it can be used for this purpose, for instance:
```python
class A:
    ...
    
OmegaConf.clear_resolvers()
OmegaConf.register_resolver("test", lambda: [A()])
c = OmegaConf.create({"x": "${test:}"})
c._set_flag("allow_objects", True)
assert isinstance(c.x[0], A)
```

It is true that in OmegaConf 2.0 a resolver could output anything which would get wrapped in a `ValueNode`. I'm not sure to which extent this use case was meant to be supported though. It seems a bit dangerous since it lets you do things that aren't meant to be possible without interpolations.

>    In this particular case, we should just say that the oc.decode returns a transient config node when it sees list and dict. you can also demonstrate the usefulness by showing a usecase with interpolation in the decoded object.

Ok, though technically it's hiding what's really happening in practice: `oc.decode` returns a list or dict, which then gets converted to a ListConfig / DictConfig at the last step of the interpolation resolution. The distinction may matter in more advanced use cases where `oc.decode` would be used as an intermediate resolver in nested interpolations, e.g. `${my_resolver:${oc.decode:${oc.env:VAR}}}`.
It's probably fine (and better) not to go into these details though, so I'll go with the simpler explanation you suggested.

odelalleau:
> It's probably fine (and better) not to go into these details though, so I'll go with the simpler explanation you suggested.

I gave it a shot in d1229d6 (doc update). The notebook will also need to be updated but I'll wait until we settle on the documentation first.

odelalleau:
> What I have in mind here is to add a flag when registering the resolver that would qualify it's behavior as converting or not converting the resulting containers to corresponding OmegaConf containers

Just to be clear, in that case, would the resulting node (wrapping the output) be a `ValueNode` or an `AnyNode` with `allow_objects` set to True? (both can work, but personally I was seeing `ValueNode` as an abstract class -- actually made it so recently, but this can be undone easily if needed)

> The distinction you are drawing about an intermediate resolver is something I want some clarification about:
> I think you are combining two features here:
> 
> 1. validate against the underlying annotate type (which potentially means create a new object).
> 2. convert a returned dict/list/tuple to a corresponding OmegaConf container. this is not connected to the annotation.
> 
> In my view, the second should be controlled by a flag when registering the resolver.

What I mean is that (2) (and actually also (1), but let's ignore it) happens only at the very last step of interpolation resolution: we look at the final result, and if it's a dict/list, then we convert it into an OmegaConf container. But the output of an intermediate resolver is fed unchanged as input to the next resolver in the chain. So in my example `${my_resolver:${oc.decode:${oc.env:VAR}}}` => if `VAR` is set to `"[1, 2, 3]"` then `my_resolver` will get a plain list as input (not a ListConfig).

odelalleau:
> the output of an intermediate resolver is fed unchanged (*) as input to the next resolver in the chain. So in my example `${my_resolver:${oc.decode:${oc.env:VAR}}}` => if `VAR` is set to `"[1, 2, 3]"` then `my_resolver` will get a plain list as input (not a ListConfig).

Thinking more about it, maybe it's better to change this behavior and systematically convert the ouptut of a resolver from dict/list to DictConfig/ListConfig, even for intermediate computations. The main advantage I see is that it is simpler to understand. In that case having a flag to control whether or not this conversion occurs is even more important.

So here is a suggestion, @omry please let me know if you'd prefer something different:
* Add a flag `convert_container_output: bool = True` to `register_new_resolver()` to control this behavior
* `oc.decode` would use this flag set to `True` (same thing for the upcoming `oc.dict.values`)
* If the final result of an interpolation is a dict/list (which can only happen if it is `${foo:...}` where `foo` was registered with `convert_container_output=False`), then wrap it within an `AnyNode` with `allow_objects` set to `True` (and also do this for any other unsupported type of resolver output)

omry:
Wait with any changes related to it, have some other ideas which might change the direction of this discussion. will comment later.

omry:
I realized that now that we support passing the `_parent_` node, the function itself can do any conversions explicitly.
If a function wants to return a DictConfig, it can do it:
```python
OmegaConf.register_resolver("dict", lambda: {"a": 10})
OmegaConf.register_resolver("dictconfig", lambda _parent_: DictConfig({"a": 10}, parent=_parent_))

cfg = OmegaConf.create({
  "d" : "${dict:}",
  "dc" : "${dictconfig:}"
})
assert type(cfg.d) is dict
assert type(cfg.dc) is DictConfig
```

I did not try to use `_parent_` like that, but if we do that things becomes significantly simpler as there are no more automatic conversions.

We still need to think what this means for type based validation/conversion:
```
@dataclass
class Foo:
   a : str = "10"
   b : int = "${a}"
```
`oc.env` can do the conversion based on the ref_type of _parent_, but this might be useful as a generic behavior for all resolvers.

odelalleau:
> I realized that now that we support passing the `_parent_` node, the function itself can do any conversions explicitly.

True (I kinda liked the automatic conversion though, it seemed natural and more straightforward -- but maybe explicit is still better)

Assuming we go with this option, we still need to decide what type of node should wrap a plain dict/list (or any other non-standard object). Ok with AnyNode with `allow_objects` set to True, or do you prefer to get back to the old `ValueNode`?

> We still need to think what this means for type based validation/conversion:
> 
> ```
> @dataclass
> class Foo:
>    a : str = "10"
>    b : int = "${a}"
> ```
> 
> `oc.env` can do the conversion based on the ref_type of _parent_, but this might be useful as a generic behavior for all resolvers.

I don't think we should change the current type based validation/conversion: this mechanism has to happen as the very last step (it doesn't make sense to convert the ouptut of an intermediate resolver based on the node type), and it applies both to resolver interpolations *and* node interpolations (so I don't think resolver should worry about it).

omry:
> > I realized that now that we support passing the `_parent_` node, the function itself can do any conversions explicitly.
> 
> True (I kinda liked the automatic conversion though, it seemed natural and more straightforward -- but maybe explicit is still better)

We can introduce a flag when registering the resolver that would do it automatically on demand if this turns out to be a common pattern. I think being explicit is better because it reduces surprises.
 
> Assuming we go with this option, we still need to decide what type of node should wrap a plain dict/list (or any other non-standard object). Ok with AnyNode with `allow_objects` set to True, or do you prefer to get back to the old `ValueNode`?

Why do you need to wrap it at all?
 
> > We still need to think what this means for type based validation/conversion:
> > ```
> > @dataclass
> > class Foo:
> >    a : str = "10"
> >    b : int = "${a}"
> > ```
> > 
> > `oc.env` can do the conversion based on the ref_type of _parent_, but this might be useful as a generic behavior for all resolvers.
> 
> I don't think we should change the current type based validation/conversion: this mechanism has to happen as the very last step (it doesn't make sense to convert the ouptut of an intermediate resolver based on the node type), and it applies both to resolver interpolations _and_ node interpolations (so I don't think resolver should worry about it).

Yes, I tend to agree.
I just wanted to mention that option. Those two topics are related.

omry:
> Why do you need to wrap it at all?

I guess for compatibility with regular interpolations that are returning a Node.
We can introduce a new DummyNode(ValueNode) for that purpose.
This would be more obvious if someone gets it when they think they got a real thing.

odelalleau:
> > Why do you need to wrap it at all?
> 
> I guess for compatibility with regular interpolations that are returning a Node.
> We can introduce a new DummyNode(ValueNode) for that purpose.
> This would be more obvious if someone gets it when they think they got a real thing.

Generally speaking, `_dereference_node()` must return a Node.
I'd rather avoid creating a new type of `ValueNode` if we can re-use one...

omry:
AnyNode with allow_objects=True works fine then.

odelalleau:
> AnyNode with allow_objects=True works fine then.

Ok sounds good. I'm planning to do this in a follow-up PR that will:
1. Roll back the automatic conversion of dict / list outputs into DictConfig / ListConfig
2. Add a new resolver `oc.to_config` (let me know if you prefer another name) that takes a dict/list as input and converts it into a DictConfig / ListConfig

omry:
> > AnyNode with allow_objects=True works fine then.
> 
> Ok sounds good. I'm planning to do this in a follow-up PR that will:
> 
> 1. Roll back the automatic conversion of dict / list outputs into DictConfig / ListConfig
> 2. Add a new resolver `oc.to_config` (let me know if you prefer another name) that takes a dict/list as input and converts it into a DictConfig / ListConfig

I think oc.create could make sense. this is a parallel to OmegaConf.create().
In fact, we can use it directly there, this will also enable support for creating from a yaml string.

odelalleau:
> I think oc.create could make sense. this is a parallel to OmegaConf.create().

Sounds like a plan!

news/573.api_change
```diff
@@ -0,0 +1 @@
+The `env` resolver is deprecated in favor of `oc.env`, which keeps the string representation of environment variables.
```
omry, line 1:
```suggestion
The `env` resolver is deprecated in favor of `oc.env`, which keeps the string representation of environment variables and does not cache the resulting value.
```

omegaconf/_utils.py
```diff
@@ -404,6 +404,11 @@ def get_value_kind(
         return ValueKind.VALUE
 
 
+def is_bool(st: str) -> bool:
```
omry, line 407: flag functions introduced here and are scheduled to be removed in 2.2.
```suggestion
# DEPRECATED: remove in 2.2
def is_bool(st: str) -> bool:
```

omegaconf/_utils.py
```diff
@@ -420,6 +425,19 @@ def is_int(st: str) -> bool:
         return False
 
 
+def decode_primitive(s: str) -> Any:
```
omry, line 428:
```suggestion
# DEPRECATED: remove in 2.2
def decode_primitive(s: str) -> Any:
```

omegaconf/omegaconf.py
```diff
@@ -101,25 +101,55 @@ def SI(interpolation: str) -> Any:
 
 
 def register_default_resolvers() -> None:
-    def env(key: str, default: Any = _EMPTY_MARKER_) -> Any:
+    # # DEPRECATED: remove in 2.2
```
omry, line 103: extra #

omegaconf/omegaconf.py
```diff
@@ -101,25 +101,55 @@ def SI(interpolation: str) -> Any:
...
         try:
-            val_str = os.environ[key]
+            return os.environ[key]
         except KeyError:
             if default is not _EMPTY_MARKER_:
+                if default is not None and not isinstance(default, str):
+                    raise ValidationError(
+                        f"The default value of the `oc.env` resolver must be a string or "
+                        f"None, but `{default}` is of type {type(default).__name__}"
+                    )
```
omry, line 131: I think we want this validation to happen on usage of the function.
The motivation is to enable this error in usage to show up immediately and not wait for the environment to also not contain the key.

omegaconf/omegaconf.py
```diff
@@ -101,25 +101,55 @@ def SI(interpolation: str) -> Any:
...
             else:
                 raise ValidationError(f"Environment variable '{key}' not found")
 
-        # We obtained a string from the environment variable: we try to parse it
-        # using the grammar (as if it was a resolver argument), so that expressions
-        # like numbers, booleans, lists and dictionaries can be properly evaluated.
-        try:
-            parse_tree = parse(
-                val_str, parser_rule="singleElement", lexer_mode="VALUE_MODE"
+    def decode(expr: Optional[str]) -> Any:
```
omry: I think it's time to create a module for built in resolvers.
Maybe:
```
omegaconf/built_in_resolvers.py
```

Also, to make the built in resolvers first class functions in that module and not hidden in a function.

omegaconf/omegaconf.py
```diff
@@ -101,25 +101,55 @@ def SI(interpolation: str) -> Any:
...
+        If `expr` is `None`, then return `None`.
+        """
+        if expr is None:
+            return None
+
+        if not isinstance(expr, str):
+            raise ValidationError(
+                f"`oc.decode` can only take strings or None as input, "
+                f"but `{expr}` if of type {type(expr).__name__}"
             )
```
omry: This reminds me of Hydra [Functions class](https://github.com/facebookresearch/hydra/blob/44b843629e7f01e40b853bcdf66d7fa2645365fd/hydra/_internal/grammar/functions.py). It supports registration of functions and it uses the arguments types for automatic validation.
It's something we can think about as a future enhancement.

Hydra got a small set of function it supports, but at the same time we also support a limited set of input types here, so I think it's doable.

omegaconf/omegaconf.py
```diff
@@ -128,16 +158,16 @@ def env(key: str, default: Any = _EMPTY_MARKER_) -> Any:
             val = empty_config.resolve_parse_tree(parse_tree)
         except InterpolationKeyError as exc:
             raise InterpolationKeyError(
-                f"When attempting to resolve env variable '{key}', a node interpolation "
-                f"caused the following exception: {exc}. Node interpolations are not "
-                f"supported in environment variables: either remove them, or escape "
-                f"them to keep them as a strings."
+                f"When attempting to resolve expression `{expr}` in `oc.decode`, "
+                f"a node interpolation caused the following exception: {exc}. "
+                f"Node interpolations are not supported by `oc.decode`."
```
omry: Does it make sense to keep interpolations as strings instead of trying to resolve them here?

tests/test_interpolation.py
```diff
@@ -211,71 +211,203 @@ def test_type_inherit_type(cfg: Any) -> None:
...
+
+        cfg["env_func"] = env_func  # allows choosing which env resolver to use
+        cfg = OmegaConf.create(cfg)
+
+        # The legacy env resolver triggers a deprecation warning.
+        if env_func == "env":
+            with pytest.warns(UserWarning):
+                assert OmegaConf.select(cfg, key) == expected
+        else:
+            assert OmegaConf.select(cfg, key) == expected
```
omry, line 266: You can add recwarn to the function signature to ignore all deprecation warnings from this function.

tests/test_interpolation.py
```diff
@@ -211,71 +211,203 @@ def test_type_inherit_type(cfg: Any) -> None:
...
+        self,
+        env_func: str,
+        cfg: Any,
+        key: str,
+        expected: Any,
+    ) -> None:
+        cfg["env_func"] = env_func  # allows choosing which env resolver to use
+        cfg = _ensure_container(cfg)
+
+        # The legacy env resolver triggers a deprecation warning.
```
omry, line 292: ditto. add one test for directly testing the deprecation warning and ignore the warnings everywhere else.

omry: side note, we should probably split test_interpolation into something like test_simple_interpolations.py and test_custom_resolvers.py.

docs/source/usage.rst
```diff
@@ -421,7 +422,8 @@ The following example sets ``abc123`` as the default value when ``DB_PASSWORD``
 Decoding strings with interpolations
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-You can automatically convert a string to its corresponding type (e.g., bool, int, float, dict, list) using ``oc.decode``.
+You can automatically convert a string to its corresponding type (e.g., bool, int, float, dict, list) using ``oc.decode``
```
omry: It's not automatic, you are using the oc.decode resolver :).

omegaconf/omegaconf.py
```diff
@@ -150,19 +149,7 @@ def decode(expr: Optional[str]) -> Any:
...
-        try:
-            val = empty_config.resolve_parse_tree(parse_tree)
-        except InterpolationKeyError as exc:
-            raise InterpolationKeyError(
-                f"When attempting to resolve expression `{expr}` in `oc.decode`, "
-                f"a node interpolation caused the following exception: {exc}. "
-                f"Node interpolations are not supported by `oc.decode`."
-            ).with_traceback(sys.exc_info()[2])
-
+        val = _parent_.resolve_parse_tree(parse_tree)
```
omry, line 152: noice.

docs/source/usage.rst
```diff
@@ -422,9 +422,13 @@ The following example sets ``abc123`` as the default value when ``DB_PASSWORD``
...
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-You can automatically convert a string to its corresponding type (e.g., bool, int, float, dict, list) using ``oc.decode``
-(which can even resolve interpolations).
-This resolver also accepts ``None`` as input, in which case it returns ``None``.
+Strings may be converted using ``oc.decode``:
+
+- Primitive values (e.g., ``"true"``, ``"1"``, ``"1e-3"``) are automatically converted to their corresponding type (bool, int, float)
+- Dictionaries and lists (e.g., ``"{a: b}"``, ``"[a, b, c]"``)  are returned as transient config nodes (DictConfig and ListConfig)
+- Interpolations (e.g., ``"${foo}"``) are automatically resolved
```
omry, line 429: Are interpolations resolved on access (lazily) of the interpolation field or eagerly on access of the oc.decode resolver?

docs/source/usage.rst
```diff
@@ -438,15 +442,19 @@ This can be useful for instance to parse environment variables:
     ...         }
     ...     }
     ... )
+    >>> def show(x):
+    ...     print(f"type: {type(x).__name__}, value: {x}")
```
omry, line 446: We can probably add this function to the top of the testsetup or somewhere else early on.

omry:
I added a bunch of comments on intermediate diffs. be sure to check them as well.

docs/source/usage.rst
```diff
@@ -400,38 +402,49 @@ Input YAML file:
     'omry'
     >>> conf.user.home
     '/home/omry'
+    >>> conf.user.id  # returned as string, even if it is a number
+    '123456'
 
 You can specify a default value to use in case the environment variable is not defined.
-The following example sets `abc123` as the the default value when `DB_PASSWORD` is not defined.
+This default value must be a string (with the exception of ``${oc.env:VAR,null}`` that returns ``None`` if ``VAR`` is not defined).
```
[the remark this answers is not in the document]
odelalleau: Taking suggestion => resolving

news/573.api_change
odelalleau: Taking suggestion => resolving

omegaconf/omegaconf.py
odelalleau: Yeah, I already have that in a follow-up PR (WIP), so going to resolve it for now

omegaconf/omegaconf.py
odelalleau: Sounds like a good idea => created #612 to keep track of it

omegaconf/omegaconf.py
odelalleau: Just to be clear, this specific quoted piece of code is outdated, because `oc.decode` can actually resolve interpolations (sorry about that, I should have pointed it out in the review notes).
See the second commit b29a019b9fa49e574652a8504b4b3fbaaea3f0df

docs/source/usage.rst
odelalleau: They are resolved when the `oc.decode` resolver is called. I'm not sure what would be the "lazy" alternative you mention though, since if `oc.decode` is called, it means we are accessing the field (so we have to resolve it). Can you clarify?

docs/source/usage.rst
```diff
@@ -6,6 +6,7 @@
     import tempfile
     import pickle
     os.environ['USER'] = 'omry'
+    os.environ['USERID'] = '123456'
```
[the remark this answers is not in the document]
odelalleau: Removed in c6a7c86

docs/source/usage.rst
```diff
@@ -400,38 +402,49 @@ Input YAML file:
...
-The following example sets `abc123` as the the default value when `DB_PASSWORD` is not defined.
+This default value must be a string (with the exception of ``${oc.env:VAR,null}`` that returns ``None`` if ``VAR`` is not defined).
+The following example sets ``abc123`` as the default value when ``DB_PASSWORD`` is not defined.
 
 .. doctest::
 
     >>> cfg = OmegaConf.create({
-    ...       'database': {'password': '${env:DB_PASSWORD,abc123}'}
+    ...       'database': {'password': '${oc.env:DB_PASSWORD,abc123}'}
     ... })
```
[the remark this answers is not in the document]
odelalleau: Done in e63e669

omegaconf/omegaconf.py
odelalleau: Removed in 4cb1c9e

omegaconf/omegaconf.py
odelalleau: Good idea, done in 2f95908

tests/test_interpolation.py
odelalleau: Oh, nice! Done in a973d21

tests/test_interpolation.py
odelalleau: > ditto. add one test for directly testing the deprecation warning and ignore the warnings everywhere else.

Done in a973d21 (I didn't add a new test since there are a couple of tests specific to the legacy env that still explicitly catch that warning, doesn't seem worth doing more refactoring since they will all go away in 2.2)

omry:
request review when ready.

tests/test_interpolation.py
odelalleau: > side note, we should probably split test_interpolation into something like test_simple_interpolations.py and test_custom_resolvers.py.

Noted, will do in a follow-up PR

docs/source/usage.rst
odelalleau: It's automatic in the sense that you don't have to specify the target type.
Anyway, this was rephrased when I added the resolution of interpolations => resolving.

docs/source/usage.rst
odelalleau: I moved it up and used it in more places in e090055
I didn't put it in testsetup because it doesn't appear in the doc and people may wonder exactly what is this function.

odelalleau:
> I added a bunch of comments on intermediate diffs. be sure to check them as well.

They appeared in the review as far as I can tell.

tests/test_interpolation.py
```diff
@@ -343,14 +336,25 @@ def test_env_is_not_cached(monkeypatch: Any) -> None:
...
     ],
 )
 def test_decode(monkeypatch: Any, value: Optional[str], expected: Any) -> None:
     monkeypatch.setenv("MYKEY", "456")
-    c = OmegaConf.create({"x": f"${{oc.decode:'{value}'}}"})
-    assert c.x == expected
+    c = OmegaConf.create(
+        {
+            # The node of interest is "node" (others are used to test interpolations).
+            "parent": {"node": f"${{oc.decode:'{value}'}}", "sibling": 1},
```
omry, line 353: add a trailing comma after 1 multiline formatting (will be easier to read).

tests/test_interpolation.py
```diff
@@ -450,19 +454,21 @@ def test_legacy_env_values_are_typed(
         assert c.my_key == expected
 
 
+# DEPRECATED: remove `recwarn` in 2.2 with the legacy env resolver
 @pytest.mark.parametrize("env_func", ["env", "oc.env"])
-def test_env_default_none(monkeypatch: Any, env_func: str) -> None:
+def test_env_default_none(monkeypatch: Any, recwarn: Any, env_func: str) -> None:
```
omry, line 459: If you add a trailing comma, black will format it as a list, which will allow you to place the comment in the right place.
```suggestion
def test_env_default_none(monkeypatch: Any, recwarn: Any, env_func: str,) -> None:
```

docs/source/usage.rst
```diff
@@ -335,19 +335,17 @@ Example:
...
+    >>> show(conf.client.description)
+    type: str, value: 'Client of http://localhost:80/'
 
     >>> # Composite interpolation types are always string
-    >>> conf.client.url
-    'http://localhost:80/'
-    >>> type(conf.client.url).__name__
-    'str'
+    >>> show(conf.client.url)
+    type: str, value: 'http://localhost:80/'
```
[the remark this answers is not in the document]
odelalleau: I gave it a shot in c5fc734

docs/source/usage.rst
odelalleau: Ok thanks, I think I see what you mean. It's going to depend on what `oc.decode` gets as input:
* If it is the string `${foo}` then it will resolve it and return the corresponding value (that may be of any type)
* If it is the string `\${foo}` then it will return the string `${foo}`, that may get resolved later dynamically as an interpolation depending on what is being done with it

So, the bottom line is that both are possible, depending on the escaping in the string that `oc.decode` is processing.

While messing with it I noticed that there are a couple of problems with the escaping in quoted strings => I'll open separate issues for these.

tests/test_interpolation.py
odelalleau: Done in 9090368

tests/test_interpolation.py
odelalleau: Done in a6094e0

docs/source/usage.rst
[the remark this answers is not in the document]
odelalleau: Pointing to follow-up discussion here: https://github.com/omry/omegaconf/commit/c5fc7342ca415dbe6033df297812fb28819339d3#r48366885

omegaconf/omegaconf.py
```diff
@@ -101,43 +100,67 @@ def SI(interpolation: str) -> Any:
...
         except KeyError:
-            if default is not _EMPTY_MARKER_:
-                return default
+            if default is not None:
+                # We handle the "null" string in a special way so as to preserve
+                # the old behavior of `env` while keeping the fix from #230.
+                if default.lower() == "null":
+                    return None
+                else:
+                    return decode_primitive(default)
```
omry, line 118: OmegaConf 2.0 does not treat null in a special way:
```python

In [1]: from omegaconf import *

In [2]: __version__
Out[2]: '2.0.6'

In [3]: OmegaConf.create({"a": "${env:ABC,null}"}).a
Out[3]: 'null'
```

#230 is for 2.1, there is no need for the legacy resolver to change behavior.

omry (changes requested):

docs/notebook/Tutorial.ipynb
```diff
@@ -716,7 +717,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Here is an example config file interpolates with the USER environment variable:"
+    "Here is an example config file interpolating with the USER and USERID environment variables:"
```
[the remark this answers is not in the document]
odelalleau: Oops, thanks, grabbing suggestion

docs/source/usage.rst
```diff
@@ -387,6 +382,7 @@ Environment variable interpolation
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Environment variable interpolation is also supported.
+An environment variable is always returned as a string.
```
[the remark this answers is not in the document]
odelalleau: Alright, taking suggestion, but previous users may be surprised since this is a change compared to how it used to work.

omry:
I an unsubscribing to reduce noise, request review when ready.

docs/source/usage.rst
```diff
@@ -402,36 +398,59 @@ Input YAML file:
     '/home/omry'
 
 You can specify a default value to use in case the environment variable is not defined.
-The following example sets `abc123` as the the default value when `DB_PASSWORD` is not defined.
+This default value can be a string or ``null`` (representing Python ``None``). Passing a default with a different type will result in an error.
+The following example sets default database passwords when ``DB_PASSWORD`` is not defined:
```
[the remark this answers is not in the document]
odelalleau: Done in 066314d (slightly different version)

docs/source/usage.rst
odelalleau: > The second case is also possible without escaping if the interpolation is a return of a custom resolver (like oc.env), right?
> 
> ```python
> os.environ["FOO"] = "${a}"
> cfg = OmegaConf.create({ "a" : 10, "b": "${oc.decode:{'foo:${env:FOO}'}" })
> ```
> 
> I am expecting:
> 
> ```
> b = cfg.b
> # b == DictConfig({"foo" : "${a}"}, parent=cfg)
> ```

No your example won't work. First, here's a version that runs (yours has some syntax issues) but doesn't give the outome you were hoping for:

```python
os.environ["FOO"] = "${a}"
cfg = OmegaConf.create({ "a" : 10, "b": "${oc.decode:'{foo:${oc.env:FOO}}'}" })
cfg.b  # {'foo': 10}
```
The reason is that the string the `decode()` function sees is `"{foo: ${a}}"` (see my first bullet point above).

What you want to achieve can be done with the following, which feeds `decode()` with the string `"{foo: '\${a}'}"` (my second bullet point above):
```python
os.environ["FOO"] = "'\\${a}'"
cfg = OmegaConf.create({ "a" : 10, "b": "${oc.decode:'{foo:${oc.env:FOO}}'}" })
cfg.b  # {'foo': '${a}'}
```

docs/source/usage.rst
```diff
@@ -780,9 +798,8 @@ as DictConfig, allowing attribute style access on the resulting node.
     >>> from omegaconf import SCMode
     >>> conf = OmegaConf.create({"structured_config": MyConfig})
     >>> container = OmegaConf.to_container(conf, structured_config_mode=SCMode.DICT_CONFIG)
-    >>> print(container)
-    {'structured_config': {'port': 80, 'host': 'localhost'}}
-    >>> assert type(container) is dict
+    >>> show(container)
+    type: dict, value: {'structured_config': {'port': 80, 'host': 'localhost'}}
     >>> assert type(container["structured_config"]) is DictConfig
     >>> assert container["structured_config"].port == 80
```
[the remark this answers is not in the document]
odelalleau: Almost (`container` is a plain dict) => done in 658eebb

omegaconf/omegaconf.py
odelalleau: In the meantime, I removed this feature in 459a734

odelalleau: This actually also requires updating news items: done in e6342df

omegaconf/omegaconf.py
```diff
@@ -101,43 +100,62 @@ def SI(interpolation: str) -> Any:
...
             else:
                 raise ValidationError(f"Environment variable '{key}' not found")
 
-        # We obtained a string from the environment variable: we try to parse it
-        # using the grammar (as if it was a resolver argument), so that expressions
-        # like numbers, booleans, lists and dictionaries can be properly evaluated.
-        try:
-            parse_tree = parse(
-                val_str, parser_rule="singleElement", lexer_mode="VALUE_MODE"
+    def env(key: str, default: Any = _EMPTY_MARKER_) -> Optional[str]:
```
omry, line 117:
```suggestion
    def env(key: str, default: Optional[str] = _EMPTY_MARKER_) -> Optional[str]:
```

omegaconf/omegaconf.py
```diff
@@ -101,43 +100,62 @@ def SI(interpolation: str) -> Any:
...
-            raise InterpolationKeyError(
-                f"When attempting to resolve env variable '{key}', a node interpolation "
-                f"caused the following exception: {exc}. Node interpolations are not "
-                f"supported in environment variables: either remove them, or escape "
-                f"them to keep them as a strings."
-            ).with_traceback(sys.exc_info()[2])
+            return os.environ[key]
+        except KeyError:
+            if default is not _EMPTY_MARKER_:
+                assert default is None or isinstance(default, str)  # redundant for mypy
```
omry, line 132: you can probably get rid of this now.
In the future, we may use Optional[str] in the signature for type validation.

omegaconf/omegaconf.py
```diff
@@ -101,43 +100,62 @@ def SI(interpolation: str) -> Any:
...
+        If `expr` is `None`, then return `None`.
+        """
+        if expr is None:
+            return None
+
+        if not isinstance(expr, str):
+            raise ValidationError(
+                f"`oc.decode` can only take strings or None as input, "
+                f"but `{expr}` if of type {type(expr).__name__}"
+            )
```
omry, line 150: I think ValueError is probably more appropriate here.

omry (commented):
A few more nits. once you address those I will look at the tests.

omegaconf/omegaconf.py
odelalleau: Done in dfcbf44
(had to make `_EMPTY_MARKER_` a string -- there's a neligible risk of someone accidentally using that same string which I think should be fine -- if you don't like it I can either create an instance of a custom subclass of `str` or make `default` a `Union`)

omegaconf/omegaconf.py
odelalleau: Done in d79a4c7

omegaconf/omegaconf.py
odelalleau: Done in a5bdd7a (actually using `TypeError` which I believe is an even better fit: https://docs.python.org/3/library/exceptions.html#TypeError)

omegaconf/omegaconf.py
```diff
@@ -77,7 +76,7 @@
 # -  in OmegaConf.create() to differentiate between creating an empty {} DictConfig
 #    and creating a DictConfig with None content
 # - in env() to detect between no default value vs a default value set to None
-_EMPTY_MARKER_ = object()
+_EMPTY_MARKER_ = "_OMEGACONF_EMPTY_MARKER_"
```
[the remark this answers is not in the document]
odelalleau: Yeah, I wasn't sure if you'd care about someone using the string `"_OMEGACONF_EMPTY_MARKER_"` in their own code. I had in mind something a bit different though regarding a potential fix:
```python
class StringMarker(str):
    ...
```
then
```python
_EMPTY_MARKER_ = StringMarker()
```
Motivation:
* I don't really see the point of having these markers hold data (their name should be enough)
* Inheriting from `str` avoids typing problems with `Optional[str] = _EMPTY_MARKER_` (it also means we may need other markers for other types). The alternative is to have a single `Marker` class and use `Union[Optional[str], Marker]` instead, but it makes things a bit ugly IMO.

tests/test_interpolation.py
```diff
@@ -211,71 +211,210 @@ def test_type_inherit_type(cfg: Any) -> None:
...
+        cfg: Any,
+        env_name: Optional[str],
+        env_val: str,
+        key: str,
+        expected: Any,
+    ) -> None:
+        if env_name is not None:
+            monkeypatch.setenv(env_name, env_val)
+
+        cfg["env_func"] = env_func  # allows choosing which env resolver to use
```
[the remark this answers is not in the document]
odelalleau: And I thought I had found a good use case for dynamic resolver names :'(

omegaconf/omegaconf.py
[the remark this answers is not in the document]
omry: > I don't really see the point of having these markers hold data (their name should be enough)

Makes it easier to debug. (The string as opposed to some unknown object).

> Inheriting from str avoids typing problems with Optional[str] = _EMPTY_MARKER_

But does not work for any other type.
for example, `dict.pop(key, default)` should work for any default type. 

I think a marker should not be typed, it's just a sentry. Annotating it as Any is a nice workaround.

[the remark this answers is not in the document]
odelalleau: Changed in 9763b66986b6c4e966f230594d741296feff500e:
* I created a new string marker since it made more sense to me to use typed markers only when needed
* I also took the opportunity to remove another marker based on a constant string (which could have led to similar issues)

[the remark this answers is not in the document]
odelalleau: > I think a marker should not be typed, it's just a sentry. Annotating it as Any is a nice workaround.

So, before I do another pass, do you prefer that I use `Union[Optional[str], Marker]` or just `# type: ignore`?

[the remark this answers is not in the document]
odelalleau: > So, before I do another pass, do you prefer that I use `Union[Optional[str], Marker]` or just `# type: ignore`?

Well, my bad, it turns out that typing the marker with `Any` is enough to make mypy happy. Somehow I thought I had tried it but I guess I was mistaken...

8a44cc9 should be good to go *crossing fingers*

docs/source/usage.rst
```diff
@@ -386,7 +385,7 @@ Interpolated nodes can be any node in the config, not just leaf nodes:
 Environment variable interpolation
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Environment variable interpolation is also supported.
+Access to environment variables is supported using ``oc.env``:
```
omry, line 388: At this stage we might be able to document it like a regular function and copy the doc here as is. Example:
```python
def env(key: str, default: Optional[str] = _EMPTY_MARKER_) -> Optional[str]:
   """
   :param key: environment variable to return 
   :param default: Default value to use if environment variable is not present (use the literal 'null' in the config to indicate None. If a default is not provided and the environment variable is not defined an exception is raised.
   :return: The environment variable string or the default value
   """
```

This along with the example looks like a good structure to document this and other built-in resolvers (Especially once we start to annotate the types in the resolvers).

omegaconf/base.py
```diff
@@ -32,7 +33,7 @@
 
 DictKeyType = Union[str, int, Enum, float, bool]
 
-_MARKER_ = object()
+_MARKER_: Any = Marker("_MARKER_")
```
omry, line 36: Looks like all markers are serving the exact same purpose (act as a sort of NO_DEFAULT_PROVIDED indicator).
We can have a pass later to consolidate them somehow and use a single one.

omegaconf/omegaconf.py
```diff
@@ -101,43 +101,61 @@ def SI(interpolation: str) -> Any:
...
-        try:
-            parse_tree = parse(
-                val_str, parser_rule="singleElement", lexer_mode="VALUE_MODE"
+    def env(key: str, default: Optional[str] = _EMPTY_MARKER_) -> Optional[str]:
+        if (
+            default is not _EMPTY_MARKER_
+            and default is not None
+            and not isinstance(default, str)
+        ):
+            raise ValidationError(
```
omry, line 124: TypeError seem more appropriate here as well.

omry (commented):
Almost there.

Feel free to bounce documentation changes to an issue. I can do it too.

omegaconf/omegaconf.py
odelalleau: Done in a7e831a

omegaconf/base.py
odelalleau: Oh well, I thought why not do it at the same time, one less thing to worry about: see 8862343

docs/source/usage.rst
odelalleau: Created new issue for this: #619

omry:
I think this is a go, except the merge conflict :)

odelalleau:
> I think this is a go, except the merge conflict :)

Funny, I just rebased on top of master without conflict... Just force-pushed.

omry (approved):
Another one bites the dust!

PR #607, opened 2021-03-16
Pydev.Debugger extension

### LINKED ISSUE #214: Consider implementing a pydev resolver plugin

omry:
Closes #214

See included doc for user facing description.

Example:
For this config:
```yaml
simple = OmegaConf.create(
    {
        "a": "bar",
        "inter": "${a}",
        "inter2": "foo_${a}",
        "inter_error1": "${zzz}",
        "inter_error2": "foo_${zzz}",
        "missing": "???",
    },
)
```

Object inspection with USER flavor:
image

Object inspection with DEV flavor:
image

Object inspection with the resolver disabled (default behavior prior to extension):
image

### DIFF
--- a/pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
+++ b/pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
@@ -53,14 +53,10 @@ class OmegaConfUserResolver(StrPresentationProvider):  # type: ignore
         ListConfig = find_mod_attr("omegaconf", "ListConfig")
         DictConfig = find_mod_attr("omegaconf", "DictConfig")
         Node = find_mod_attr("omegaconf", "Node")
-        IRE = find_mod_attr("omegaconf.errors", "InterpolationResolutionError")
 
         if isinstance(obj, Node):
-            try:
-                obj = obj._dereference_node(throw_on_resolution_failure=False)
-                if obj._is_none() or obj._is_missing():
-                    return {}
-            except IRE:
+            obj = obj._dereference_node(throw_on_resolution_failure=False)
+            if obj is None or obj._is_none() or obj._is_missing():
                 return {}
 
         if isinstance(obj, DictConfig):
--- a/pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
+++ b/pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
@@ -1,0 +1,114 @@
+# based on https://github.com/fabioz/PyDev.Debugger/tree/main/pydevd_plugins/extensions
+import os
+import sys
+from functools import lru_cache
+from typing import Any, Dict, Sequence
+
+from _pydevd_bundle.pydevd_extension_api import (  # type: ignore
+    StrPresentationProvider,
+    TypeResolveProvider,
+)
+
+
+@lru_cache(maxsize=128)
+def find_mod_attr(mod_name: str, attr: str) -> Any:
+    mod = sys.modules.get(mod_name)
+    return getattr(mod, attr, None)
+
+
+class OmegaConfDeveloperResolver(object):
+    def can_provide(self, type_object: Any, type_name: str) -> bool:
+        Node = find_mod_attr("omegaconf", "Node")
+        return Node is not None and issubclass(type_object, Node)
+
+    def resolve(self, obj: Any, attribute: str) -> Any:
+        return getattr(obj, attribute)
+
+    def get_dictionary(self, obj: Any) -> Any:
+        return obj.__dict__
+
+
+class OmegaConfUserResolver(StrPresentationProvider):  # type: ignore
+    def can_provide(self, type_object: Any, type_name: str) -> bool:
+        Node = find_mod_attr("omegaconf", "Node")
+        return Node is not None and issubclass(type_object, Node)
+
+    def resolve(self, obj: Any, attribute: Any) -> Any:
+        if isinstance(obj, Sequence):
+            if isinstance(attribute, str):
+                attribute = int(attribute)
+            val = obj.__dict__["_content"][attribute]
+        else:
+            val = obj.__dict__["_content"][attribute]
+
+        return val
+
+    def _is_simple_value(self, val: Any) -> bool:
+        ValueNode = find_mod_attr("omegaconf", "ValueNode")
+        return (
+            isinstance(val, ValueNode)
+            and not val._is_missing()
+            and not val._is_interpolation()
+            and not val._is_none()
+        )
+
+    def get_dictionary(self, obj: Any) -> Dict[str, Any]:
+        ListConfig = find_mod_attr("omegaconf", "ListConfig")
+        DictConfig = find_mod_attr("omegaconf", "DictConfig")
+        Node = find_mod_attr("omegaconf", "Node")
+        IRE = find_mod_attr("omegaconf.errors", "InterpolationResolutionError")
+
+        if isinstance(obj, Node):
+            try:
+                obj = obj._dereference_node()
+                if obj._is_none() or obj._is_missing():
+                    return {}
+            except IRE:
+                obj = None
+
+        if isinstance(obj, DictConfig):
+            d = {}
+            for k, v in obj.__dict__["_content"].items():
+                if self._is_simple_value(v):
+                    v = v._value()
+                d[k] = v
+        elif isinstance(obj, ListConfig):
+            d = {}
+            for idx, v in enumerate(obj.__dict__["_content"]):
+                if self._is_simple_value(v):
+                    v = v._value()
+                d[str(idx)] = v
+        else:
+            d = {}
+
+        return d
+
+    def get_str(self, val: Any) -> str:
+        IRE = find_mod_attr("omegaconf.errors", "InterpolationResolutionError")
+
+        if val._is_missing():
+            return "??? <MISSING>"
+        if val._is_interpolation():
+            try:
+                dr = val._dereference_node()
+            except IRE as e:
+                dr = f"ERR: {e}"
+            return f"{val._value()} -> {dr}"
+        else:
+            return f"{val}"
+
+
+# OC_PYDEVD_RESOLVER env can take:
+#  DISABLE: Do not install a pydevd resolver
+#  USER: Install a resolver for OmegaConf users (default)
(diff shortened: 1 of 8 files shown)


pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
```diff
@@ -0,0 +1,114 @@
...
+        Node = find_mod_attr("omegaconf", "Node")
+        return Node is not None and issubclass(type_object, Node)
+
+    def resolve(self, obj: Any, attribute: Any) -> Any:
+        if isinstance(obj, Sequence):
+            if isinstance(attribute, str):
+                attribute = int(attribute)
+            val = obj.__dict__["_content"][attribute]
+        else:
+            val = obj.__dict__["_content"][attribute]
```
odelalleau, line 42: Small simplification:
```suggestion
        if isinstance(obj, Sequence) and isinstance(attribute, str):
            attribute = int(attribute)
        val = obj.__dict__["_content"][attribute]
```

pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
```diff
@@ -0,0 +1,114 @@
...
+
+        return val
+
+    def _is_simple_value(self, val: Any) -> bool:
+        ValueNode = find_mod_attr("omegaconf", "ValueNode")
+        return (
+            isinstance(val, ValueNode)
+            and not val._is_missing()
+            and not val._is_interpolation()
+            and not val._is_none()
```
odelalleau, line 52: Unlikely to matter much, but testing `_is_none()` first is more efficient
```suggestion
            and not val._is_none()
            and not val._is_missing()
            and not val._is_interpolation()
```

pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
```diff
@@ -0,0 +1,114 @@
...
+@lru_cache(maxsize=128)
+def find_mod_attr(mod_name: str, attr: str) -> Any:
+    mod = sys.modules.get(mod_name)
+    return getattr(mod, attr, None)
+
+
+class OmegaConfDeveloperResolver(object):
+    def can_provide(self, type_object: Any, type_name: str) -> bool:
+        Node = find_mod_attr("omegaconf", "Node")
+        return Node is not None and issubclass(type_object, Node)
```
odelalleau, line 22: Why check for `None`? (other places where `find_mod_attr()` is used don't seem to do such check)

pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
```diff
@@ -0,0 +1,114 @@
...
+        Node = find_mod_attr("omegaconf", "Node")
+        IRE = find_mod_attr("omegaconf.errors", "InterpolationResolutionError")
+
+        if isinstance(obj, Node):
+            try:
+                obj = obj._dereference_node()
+                if obj._is_none() or obj._is_missing():
+                    return {}
+            except IRE:
+                obj = None
```
odelalleau, line 67: More explicit:
```suggestion
                return {}
```

pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
```diff
@@ -0,0 +1,114 @@
...
+
+        if isinstance(obj, DictConfig):
+            d = {}
+            for k, v in obj.__dict__["_content"].items():
+                if self._is_simple_value(v):
+                    v = v._value()
```
odelalleau, line 73: What's the motivation for excluding `None` from being a simple value? (now `node._is_none()` only returns True if `node._value() is None`, so it seems to me that it should work fine (unless `None` breaks pydevd somehow?)

Is it because you want to show `DictConfig(None)` differently from a plain `None`?

tests/test_pydev_resolver_plugin.py
```diff
@@ -0,0 +1,254 @@
...
+        # nodes
+        param(AnyNode(10), {}, id="any:10"),
+        param(StringNode("foo"), {}, id="str:foo"),
+        param(IntegerNode(10), {}, id="int:10"),
+        param(FloatNode(3.14), {}, id="float:3.14"),
+        param(BooleanNode(True), {}, id="bool:True"),
+        param(EnumNode(enum_type=Color, value=Color.RED), {}, id="Color:Color.RED"),
+        # nodes are never returning a dictionary
+        param(AnyNode("${foo}", parent=DictConfig({"foo": 10})), {}, id="any:10"),
+        param(AnyNode("${foo}", parent=DictConfig({"foo": 10})), {}, id="any:10"),
```
odelalleau, line 45: Duplicated lines + id:
```suggestion
        param(AnyNode("${foo}", parent=DictConfig({"foo": 10})), {}, id="any:inter_10"),
```

tests/test_pydev_resolver_plugin.py
```diff
@@ -0,0 +1,254 @@
...
+        param(DictConfig({"a": 10}), "a", AnyNode(10), id="dict"),
+        param(
+            DictConfig({"a": DictConfig(None)}),
+            "a",
+            DictConfig(None),
+            id="dict:none",
+        ),
+        # listconfig
+        param(ListConfig([10]), 0, AnyNode(10), id="list"),
+        param(ListConfig(["???"]), 0, AnyNode("???"), id="list"),
```
odelalleau, line 94:
```suggestion
        param(ListConfig(["???"]), 0, AnyNode("???"), id="list:missing_item"),
```

tests/test_pydev_resolver_plugin.py
```diff
@@ -0,0 +1,254 @@
...
+            "missing",
+            {},
+            id="dict:missing_dictconfig_value",
+        ),
+        param(
+            OmegaConf.create({"missing": DictConfig("???")}),
+            "missing",
+            {},
+            id="dict:missing_dictconfig_value",
+        ),
```
odelalleau, line 143: Duplicated
```suggestion
```

tests/test_pydev_resolver_plugin.py
```diff
@@ -0,0 +1,254 @@
...
+            OmegaConf.create({"a": ListConfig("???")}),
+            "a",
+            {},
+            id="list:missing_listconfig_value",
+        ),
+        param(
+            OmegaConf.create({"a": [1, 2], "b": ListConfig("${a}")}),
+            "b",
+            {"0": 1, "1": 2},
+            id="list:interpolationn_listconfig_value",
```
odelalleau, line 184: Typo
```suggestion
            id="list:interpolation_listconfig_value",
```

tests/test_pydev_resolver_plugin.py
```diff
@@ -0,0 +1,254 @@
...
+        (FloatNode, True),
+        (StringNode, True),
+        (BooleanNode, True),
+        # not covering some other things.
+        (builtins.int, False),
+        (dict, False),
+        (list, False),
+    ],
+)
+def test_can_provide(resolver: Any, type_: Any, expected: bool) -> None:
```
odelalleau, line 222: What about enums?

odelalleau (changes requested):
I don't know much about Pydev.Debugger but overall looks good, just a few small comments

pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
omry: To be safe: if the extension ends up running without OmegaConf installed, I don't want it to break the debugger.
Once we decided that we can support an object, it means OmegaConf must be available.
In addition, further interactions are limited to a specific object.

omry:
@fabioz, thanks for the plugin API!
It would be awesome if you can take a quick look and provide feedback.

pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
omry: Sure.
By the way, using `throw_on_resolution_failure=False` is not sufficient. There is still a raised exception if there is no parent (recent change).

pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
omry: Yes.
This would make more sense if and when we show more type information, maybe:
```python
DictConfig[User](None)
# or
DictConfig[int, str](None)
```

setup.py
```diff
@@ -50,7 +50,13 @@
         tests_require=["pytest"],
         url="https://github.com/omry/omegaconf",
         keywords="yaml configuration config",
-        packages=["omegaconf", "omegaconf.grammar", "omegaconf.grammar.gen"],
+        packages=[
+            "omegaconf",
+            "omegaconf.grammar",
+            "omegaconf.grammar.gen",
+            "pydevd_plugins",
```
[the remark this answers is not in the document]
omry: Yes, this is supposed to work correctly as long as everyone is playing ball and includes the exact same boilerplate `__init__.py` files (which are intentionally overwritten by every plugin).

You can learn more about it [here](https://packaging.python.org/guides/packaging-namespace-packages/#pkg-resources-style-namespace-packages).

@fabioz, you might want to change the style in the example to something more modern:
image

tests/test_pydev_resolver_plugin.py
omry: duplicate id is okay (they get an automatic counter appended). 
Duplicate line, not so much.

tests/test_pydev_resolver_plugin.py
omry: good point.

setup.py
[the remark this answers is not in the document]
odelalleau: Interesting, thanks :) If I ever have to do something similar I'll know where to look then...
=> resolving

pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
odelalleau: Oops, I'll fix that

odelalleau: #611 => resolving

odelalleau (approved):
Looks good to VSCode debugger!

PR #609, opened 2021-03-16
Deprecate `get_resolver()` and add new `has_resolver()`

### LINKED ISSUE #608: Deprecation of `OmegaConf.get_resolver()`
**PROBLEM**

`OmegaConf.get_resolver()` is exposing OmegaConf internals that are not meant to be public-facing (they may change arbitrarily).

**DEPRECATION PLAN**

* `OmegaConf.get_resolver()` is being deprecated in OmegaConf 2.1 and will be removed in OmegaConf 2.2.
* A new function `OmegaConf.has_resolver(name: str) -> bool` is being added in OmegaConf 2.1 to make it possible to check for the existence of a resolver. If this new function is not enough for your use case, please open an issue explaining why.

**Additional context**
See discussion in https://github.com/omry/omegaconf/pull/599#discussion_r593540009

odelalleau:
Fixes #608

Best review both commits separately (second one is just moving functions around)

### DIFF
--- a/news/608.api_change
+++ b/news/608.api_change
@@ -1,0 +1,1 @@
+`OmegaConf.get_resolver()` is deprecated: use the new `OmegaConf.has_resolver()` to check for the existence of a resolver.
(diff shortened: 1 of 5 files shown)


omry (approved):
Looking good.

PR #611, opened 2021-03-16
Do not throw when throw_on_resolution_failure is False

odelalleau:
Trying to dereference a node without a parent was raising an
`InterpolationResolutionError` even with `throw_on_resolution_failure`
set to False.

This bug was pointed out in https://github.com/omry/omegaconf/pull/607#discussion_r595485365 and is related to the recent #605.


omry (approved):
Thanks. merging.

PR #616, opened 2021-03-17
Simplify the escaping of backslashes in quoted strings

### LINKED ISSUE #615: Escaping in quoted strings (within interpolations) is too complex
**Describe the bug**

Technically the following is working as intended, but it is hard to understand and error prone. To explain the issue I have to show increasingly more complex scenarios:

(1) Consider the following code (note that I'm using a raw string `r'...'` to avoid having to double all `\` due to Python string processing):
```python
cfg = OmegaConf.create({
  "msg": r'${identity:"The root drive is: ${drive}:\\"}',
  "drive": "C",
})
print(cfg.msg)  # The root drive is: C:\
```

Importantly, the escape at the end of the quoted string is doubled so that it is not mistaken for an escaped quote by the grammar (side note: there is a bug related to this, but I'll discuss it in another issue, we can ignore it here). So far, so good.

(2) Now let's say you want to print instead `The root drive is: ${drive}:\`, i.e., the interpolation shouldn't be resolved. Here is how to achieve it:

```python
cfg = OmegaConf.create({
  "msg": r'${identity:"The root drive is: \${drive}:\\"}',
  "drive": "C",
})
print(cfg.msg)  # The root drive is: ${drive}:\
```
Still looks good :)

(3) What if instead you want to print: `The root drive is: \C:\`? (i.e., the interpolation should be resolved, but with a backslash in front). The "naive" way of adding an extra backslash (to have an escaped backslash instead of an escaped interpolation) won't cut it:

```python
cfg = OmegaConf.create({
  "msg": r'${identity:"The root drive is: \\${drive}:\\"}',
  "drive": "C",
})
print(cfg.msg)  # The root drive is: ${drive}:\
```

Instead, you need to use four backslashes (or three, it gives the same result):

```python
cfg = OmegaConf.create({
  "msg": r'${identity:"The root drive is: \\\\${drive}:\\"}',
  "drive": "C",
})
print(cfg.msg)  # The root drive is: \C:\
```

Now this is getting confusing! The reason is essentially that escaped backslashes get un-escaped twice:
* First, to un-escape `\\`, that may be used as `\\"` to avoid confusion with an escaped quoted `\"`. This step happens before we feed the resulting string to `quoted_string_callback()`
* Second, to un-escape `\\`, that may be used as `\\${..}` to avoid confusion with an escaped interpolation `\${..}`. This step happens within `quoted_string_callback()`

**Expected behavior**

Scenarios (1) and (2) should still work the same. However, in scenario (3) the first example should work, i.e.:

```python
cfg = OmegaConf.create({
  "msg": r'${identity:"The root drive is: \\${drive}:\\"}',
  "drive": "C",
})
print(cfg.msg)  # The root drive is: \C:\
```

This is doable by skipping the first un-escaping step described above, and rely instead on the second one to un-escape `\\`.

I don't see any major drawback to it. A minor one is that it means we will need to process the content of *all* quoted strings as if they contained interpolations, so as to properly un-escape `\\`, which incurs some extra overhead (but since we are already in the middle of resolving an interpolation it shouldn't make a big difference).

**Additional context**
Related discussion: https://github.com/omry/omegaconf/pull/606#discussion_r596009899
(in the example discussed there, proper escaping proved very confusing as I was trying to achieve the desired result)

odelalleau:
Fixes #615

Highlighting the important points of this PR:

* The main goal is to remove the "double un-escaping" of backslashes when processing quoted strings in the grammar (see #615)
* A side effect is the need to process the content of all quoted strings through the grammar. This includes the empty string, which was not allowed before (so this PR allows it as well, and adds some explicit tests for it)
* Another side effect is that grammar tests that used a dummy quoted string callback (returning the string "as is" since it doesn't contain interpolations) aren't valid anymore (since backslashes must now be properly un-escaped by this callback). So I refactored a bit the tests accordingly, splitting them by "needs callback" vs "doesn't need callback", instead of "has interpolation" vs "doesn't have interpolation" (the question of a deeper refactoring of these tests will be addressed later)

### DIFF
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -610,10 +610,13 @@ class Container(Node):
         key: Any,
         value: "Node",
         throw_on_resolution_failure: bool,
+        force_resolution: bool = False,
     ) -> Any:
-        value_kind = get_value_kind(value)
-        if value_kind != ValueKind.INTERPOLATION:
-            return value
+        if not force_resolution:
+            # Skip resolution if not an interpolation.
+            value_kind = get_value_kind(value)
+            if value_kind != ValueKind.INTERPOLATION:
+                return value
 
         parse_tree = parse(_get_value(value))
         return self._resolve_interpolation_from_parse_tree(
@@ -662,6 +665,9 @@ class Container(Node):
                     is_optional=True,
                 ),
                 throw_on_resolution_failure=True,
+                # We must always process the string as if it contained an
+                # interpolation, for proper un-escaping of backslashes (\\).
+                force_resolution=True,
             )
             return str(quoted_val)
 
--- a/omegaconf/grammar/OmegaConfGrammarParser.g4
+++ b/omegaconf/grammar/OmegaConfGrammarParser.g4
@@ -17,7 +17,7 @@ options {tokenVocab = OmegaConfGrammarLexer;}
 
 // Main rules used to parse OmegaConf strings.
 
-configValue: (toplevelStr | (toplevelStr? (interpolation toplevelStr?)+)) EOF;
+configValue: (toplevelStr | (toplevelStr? (interpolation toplevelStr?)+))? EOF;
 singleElement: element EOF;
 
 // Top-level string (that does not need to be parsed).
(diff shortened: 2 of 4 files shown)


odelalleau:
// Converting to draft, ran into an issue while looking at another problem related to escaped quotes in quoted strings, whose fix doesn't play well with this PR

odelalleau:
> I'll update after looking at it more closely

So, long story short, I don't think the approach from this PR is right, I'll close it for now to fix #617 first.

odelalleau:
> So, long story short, I don't think the approach from this PR is right, I'll close it for now to fix #617 first.

I have another approach in mind that I hope can solve both #615 and #617 at the same time. Will try it out tomorrow.

PR #618, opened 2021-03-18
optimized list iterator some more

### LINKED ISSUE #613: ListConfig iteration is much slower than plain list iteration
See benchmark and discussion [here](https://github.com/omry/omegaconf/pull/530/files#diff-831239367e673d4efaa031d85d3a08ad4309e0ef462f0331865e3922a3f6be3aR116).

Tentative for 2.1. Not sure where this is coming from.

omry:
Closes #613

Second optimization pass.
This is more impressive than the 7.7 -> 8.8x recorded improvement since there seems to be a regression compares to the measurements in #533.
Speed is still not great, but I this is as close as it gets given that we are probably losing a bunch of low level optimization done on native list because we need to unbox primitive items when we iterate.

Comparing `test_list_iter[small_listconfig]` by itself to what is in master shows about 50% speedup:
image

When it's pitched against the native list iteration it's "only" 130 times slower now, compares to 180 times slower before:
image

By the way: here is why this is important:
https://twitter.com/jackyliang42/status/1371530444529881094

### DIFF
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -532,12 +532,14 @@ def _get_value(value: Any) -> Any:
     from .base import Container
     from .nodes import ValueNode
 
-    if isinstance(value, Container) and (
-        value._is_none() or value._is_missing() or value._is_interpolation()
-    ):
-        return value._value()
     if isinstance(value, ValueNode):
-        value = value._value()
+        return value._value()
+    elif isinstance(value, Container):
+        boxed = value._value()
+        if boxed is None or _is_missing_literal(boxed) or _is_interpolation(boxed):
+            return boxed
+
+    # return primitives and regular OmegaConf Containers as is
     return value
 
 
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -124,7 +124,10 @@ class BaseContainer(Container, ABC):
         ...
 
     def __len__(self) -> int:
-        return self.__dict__["_content"].__len__()  # type: ignore
+        if self._is_none() or self._is_missing() or self._is_interpolation():
+            return 0
+        content = self.__dict__["_content"]
+        return len(content)
 
     def merge_with_cli(self) -> None:
         args_list = sys.argv[1:]
--- a/omegaconf/listconfig.py
+++ b/omegaconf/listconfig.py
@@ -16,7 +16,6 @@ from typing import (
 
 from ._utils import (
     ValueKind,
-    _get_value,
     _is_none,
     format_and_raise,
     get_value_kind,
@@ -152,14 +151,6 @@ class ListConfig(BaseContainer, MutableSequence[Any]):
         if self._is_missing() or self._is_none():
             return []
         return [str(x) for x in range(0, len(self))]
-
-    def __len__(self) -> int:
-        if self._is_none():
-            return 0
-        if self._is_missing():
-            return 0
-        assert isinstance(self.__dict__["_content"], list)
-        return len(self.__dict__["_content"])
 
     def __setattr__(self, key: str, value: Any) -> None:
         self._format_and_raise(
@@ -500,20 +491,32 @@ class ListConfig(BaseContainer, MutableSequence[Any]):
 
     class ListIterator(Iterator[Any]):
         def __init__(self, lst: Any, resolve: bool) -> None:
-            self.iter = iter(lst.__dict__["_content"])
             self.resolve = resolve
+            self.iterator = iter(lst.__dict__["_content"])
             self.index = 0
+            from .nodes import ValueNode
+
+            self.ValueNode = ValueNode
 
         def __next__(self) -> Any:
-            v = next(self.iter)
 
+            x = next(self.iterator)
             if self.resolve:
-                v = v._dereference_node()
+                x = x._dereference_node()
+                if x._is_missing():
+                    raise MissingMandatoryValue(f"Missing value at index {self.index}")
 
-            if v._is_missing():
-                raise MissingMandatoryValue(f"Missing value at index {self.index}")
             self.index = self.index + 1
-            return _get_value(v)
+            if isinstance(x, self.ValueNode):
+                return x._value()
+            else:
+                # Must be omegaconf.Container. not checking for perf reasons.
+                if x._is_none():
+                    return None
+                return x
+
+        def __repr__(self) -> str:  # pragma: no cover
+            return f"ListConfig.ListIterator(resolve={self.resolve})"
 
     def _iter_ex(self, resolve: bool) -> Iterator[Any]:
         try:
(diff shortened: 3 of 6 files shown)


omegaconf/listconfig.py
```diff
@@ -500,20 +499,29 @@ def __iter__(self) -> Iterator[Any]:
...
         def __init__(self, lst: Any, resolve: bool) -> None:
-            self.iter = iter(lst.__dict__["_content"])
             self.resolve = resolve
+            self.iterator = iter(lst.__dict__["_content"])
             self.index = 0
+            from .base import Container
+            from .nodes import ValueNode
+
+            self.Container = Container
+            self.ValueNode = ValueNode
```
omry, line 509: The fact that this made a big difference in the perf is telling. We are importing in functions in a million places and it can actually be hurting performance in hotspots.

omegaconf/_utils.py
```diff
@@ -532,12 +532,15 @@ def _get_value(value: Any) -> Any:
...
     from .nodes import ValueNode
 
-    if isinstance(value, Container) and (
-        value._is_none() or value._is_missing() or value._is_interpolation()
-    ):
-        return value._value()
     if isinstance(value, ValueNode):
-        value = value._value()
+        boxed = value._value()
+        return boxed
```
odelalleau, line 537: Saving one line:
```suggestion
        return value._value()
```

omegaconf/listconfig.py
```diff
@@ -500,20 +499,26 @@ def __iter__(self) -> Iterator[Any]:
...
+                x = x._dereference_node()
 
-            if v._is_missing():
+            if x._is_missing():
                 raise MissingMandatoryValue(f"Missing value at index {self.index}")
+
             self.index = self.index + 1
-            return _get_value(v)
+            if isinstance(x, self.ValueNode):
+                return x._value()
```
odelalleau, line 520: This seems like a change in behavior in case `x` is a container that would be None / Missing / an interpolation. Is that intended? (and I guess we need a test since nothing broke)

odelalleau (commented):
Just one meaningful concern

(on the speedup from moving the in-line import => that's interesting, I've always wondered if there was a performance risk with these... good to know!)

omegaconf/listconfig.py
omry: it is intended. I mentioned it in one of the comments (tested on Hydra as well).
will add tests covering it.

omegaconf/_utils.py
omry: done.

omegaconf/listconfig.py
omry: 4d3053773790bb95aef113ba1e1781f55e8fdf34, this was a little pandora's box. see commit comment.

omegaconf/basecontainer.py
```diff
@@ -123,8 +123,9 @@ def __setstate__(self, d: Dict[str, Any]) -> None:
     def __delitem__(self, key: Any) -> None:
         ...
 
+    @abstractmethod
     def __len__(self) -> int:
-        return self.__dict__["_content"].__len__()  # type: ignore
+        ...
```
odelalleau, line 128: It seems that you ended up implementing the same method in both subclasses (up to the assert on type) => wouldn't it actually make sense to keep its implementation in BaseContainer?

omegaconf/listconfig.py
odelalleau: Regarding the new commit: ok

Regarding the change in behavior, looking into it more closely, I'm not super comfortable with it for two reasons:
* It now gives a way for users to end up getting access to a container that is None / missing through the public API (which wasn't really the case until now, except by explicitly instantiating one)
* my_listconfig[0] doesn't give the same object type as iter(next(my_listconfig)) (and this is inconsistent with the behavior of DictConfig)

I may have missed your comment on this topic => is there a strong motivation for this?

odelalleau (commented):
Some questions

omegaconf/basecontainer.py
omry: I had a comment pending explaining it exactly on this line.

omegaconf/listconfig.py
omry: > Regarding the new commit: ok
> 
> Regarding the change in behavior, looking into it more closely, I'm not super comfortable with it for two reasons:
> 
> * It now gives a way for users to end up getting access to a container that is None / missing through the public API (which wasn't really the case until now, except by explicitly instantiating one)

It does not.

> * my_listconfig[0] doesn't give the same object type as iter(next(my_listconfig)) (and this is inconsistent with the behavior of DictConfig)
> 

It does.

```python
In [17]: lst = ListConfig([None, ListConfig(None)])

In [18]: itr = iter(lst)

In [19]: repr(next(itr))
Out[19]: 'None'

In [20]: repr(next(itr))
Out[20]: 'None'

In [21]: repr(next(itr))
StopIteration                             Traceback (most recent call last)
<ipython-input-21-b5146a100c72> in <module>
----> 1 repr(next(itr))

~/dev/omegaconf/omegaconf/listconfig.py in __next__(self)
    508         def __next__(self) -> Any:
    509 
--> 510             x = next(self.iterator)
    511             if self.resolve:
    512                 x = x._dereference_node()

StopIteration: 
```

You are confusing `__iter__()` with the non public _iter_ex(resolve=False).

omry: As for DictConfig, I am looking at it now.

omegaconf/basecontainer.py
```diff
@@ -124,7 +124,10 @@ def __delitem__(self, key: Any) -> None:
         ...
 
     def __len__(self) -> int:
-        return self.__dict__["_content"].__len__()  # type: ignore
+        if self._is_none() or self._is_missing() or self._is_interpolation():
+            return 0
+        content = self.__dict__["_content"]
+        return len(content)
```
omry, line 130: Moved it here without the assert, I think having two different asserts is not worth the near duplication.

omegaconf/listconfig.py
odelalleau: > > Regarding the new commit: ok
> > Regarding the change in behavior, looking into it more closely, I'm not super comfortable with it for two reasons:
> > 
> > * It now gives a way for users to end up getting access to a container that is None / missing through the public API (which wasn't really the case until now, except by explicitly instantiating one)
> 
> It does not.

```python
lst = ListConfig([ListConfig(None)])
item = next(iter(lst))
assert isinstance(item, ListConfig) and item._value() is None
```

> 
> > * my_listconfig[0] doesn't give the same object type as iter(next(my_listconfig)) (and this is inconsistent with the behavior of DictConfig)
> 
> It does.
> 
> ```python
> In [17]: lst = ListConfig([None, ListConfig(None)])
> 
> In [18]: itr = iter(lst)
> 
> In [19]: repr(next(itr))
> Out[19]: 'None'
> 
> In [20]: repr(next(itr))
> Out[20]: 'None'

You're only looking at the string representation, but:

```python
lst = ListConfig([None, ListConfig(None)])
itr = iter(lst)
print(next(itr) is None)  # True
print(next(itr) is None)  # False
```

omegaconf/basecontainer.py
odelalleau: Agreed! => resolving

omegaconf/listconfig.py
omry: Yeah, just saw that too.
will take a closer look.

omry:
@odelalleau I am merging this, let me know if you have any other feedback.

omegaconf/listconfig.py
odelalleau: Sounds good

PR #623, opened 2021-03-18
Further speedup of ListConfig iteration

omry:
Another 20% on top of master by removing a type assertion from _is_missing_literal.

image

### DIFF
--- a/news/532.misc
+++ b/news/532.misc
@@ -1,1 +1,1 @@
-Optimized ListConfig iteration by 8.8x in a benchmark
+Optimized ListConfig iteration by 10.25x in a benchmark
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -349,9 +349,8 @@ def _is_missing_value(value: Any) -> bool:
 
 
 def _is_missing_literal(value: Any) -> bool:
-    from omegaconf import MISSING, Node
+    from omegaconf import MISSING
 
-    assert not isinstance(value, Node)
     ret = isinstance(value, str) and value == MISSING
     assert isinstance(ret, bool)
     return ret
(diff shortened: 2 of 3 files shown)


news/532.misc
```diff
@@ -1 +1 @@
-Optimized ListConfig iteration by 8.8x in a benchmark
+Optimized ListConfig iteration by 10.25x in a benchmark
```
omry, line 1: The baseline here is what I got before the first optimization on 2.1.
#533

omegaconf/_utils.py
```diff
@@ -349,9 +349,8 @@ def _is_missing_value(value: Any) -> bool:
 
 
 def _is_missing_literal(value: Any) -> bool:
-    from omegaconf import MISSING, Node
+    from omegaconf import MISSING
```
odelalleau, line 352: Here's an alternate version that squeezes even more juice out of it:

```python
MISSING = None
def _is_missing_literal(value: Any) -> bool:
    global MISSING
    if MISSING is None:
        from omegaconf import MISSING

    return isinstance(value, str) and value == MISSING
```

odelalleau (commented):
Suggestion for an additional optimization

omegaconf/_utils.py
odelalleau: Actually ~twice as fast:
```python
MISSING = "???"
def _is_missing_literal(value: Any) -> bool:
    return isinstance(value, str) and value == MISSING
```

=> I suggest to define MISSING in a different file that can be imported without circular dependencies

omry: nice.
Or actually switch back to using literal "???" in hotspots.
MISSING was defined to be used with Structured Configs.

omry: Using literal is 8x faster for this function (compares to master):

image

odelalleau: Yeah `"???"` is the shortcut solution -- FYI it's the same speed as defining `MISSING = "???"` so if we can import it without circular dependencies it should be equivalent (and it's a bit cleaner). But either works.

omry: See followup diff.

PR #624, opened 2021-03-19
DictConfig.items() handles DictConfig(None) and DictConfig("???") consistently with ListConfig

omry:
- Diff 1 : No functional changes. fixing the pytest imports for all tests. Expect minor conflicts because of this for outstanding diffs. 
- Diff 2: Adding better testing to DictConfig items
- calling items() on DictConfig("???") and DictConfig(None) now results in exception similar to what you get when iterating on corresponding ListConfigs. the rest of the changes are tests and updating other code in response to this change.

I think this is not user facing for people using the public API so there is no need for a news fragment or doc change.

### DIFF
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -537,9 +537,19 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
         self, resolve: bool = True, keys: Optional[Sequence[DictKeyType]] = None
     ) -> AbstractSet[Tuple[DictKeyType, Any]]:
         items: List[Tuple[DictKeyType, Any]] = []
+
+        if self._is_none():
+            self._format_and_raise(
+                key=None,
+                value=None,
+                cause=TypeError("Cannot iterate a DictConfig object representing None"),
+            )
+        if self._is_missing():
+            raise MissingMandatoryValue("Cannot iterate a missing DictConfig")
+
         for key in self.keys():
             if resolve:
-                value = self.get(key)
+                value = self[key]
             else:
                 value = self.__dict__["_content"][key]
                 if isinstance(value, ValueNode):
@@ -662,6 +672,9 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
         assert isinstance(d2, DictConfig)
         if len(d1) != len(d2):
             return False
+        if d1._is_missing() or d2._is_missing():
+            return d1._is_missing() is d2._is_missing()
+
         for k, v in d1.items_ex(resolve=False):
             if k not in d2.__dict__["_content"]:
                 return False
--- a/tests/test_basic_ops_dict.py
+++ b/tests/test_basic_ops_dict.py
@@ -4,7 +4,7 @@ import re
 import tempfile
 from typing import Any, Dict, List, Optional, Union
 
-import pytest
+from pytest import mark, param, raises
 
 from omegaconf import (
     MISSING,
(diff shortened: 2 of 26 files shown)


tests/test_basic_ops_dict.py
```diff
@@ -255,10 +255,13 @@ def foo(a: int, b: int) -> int:
     assert 12 == foo(**c)
 
 
-def test_items() -> None:
+def test_items_iterator_behavior() -> None:
     c = OmegaConf.create({"a": 2, "b": 10})
-    assert sorted([("a", 2), ("b", 10)]) == sorted(list(c.items()))
+    assert list(c.items()) == [("a", 2), ("b", 10)]
```
odelalleau, line 260: I wouldn't worry about it myself, but FYI the deterministic ordering of Python dictionaries is considered an implementation detail of CPython in 3.6.

tests/test_basic_ops_dict.py
```diff
@@ -255,10 +255,13 @@ def foo(a: int, b: int) -> int:
...
 
+    # This is actually not compatible with native dict:
+    # Due to implementation considerations, DictConfig items() returns a list.
+    # If that can be fixed, feel free to remove this block
```
odelalleau, line 264: There's no reason why it couldn't be changed, is there?

tests/test_basic_ops_dict.py
```diff
@@ -254,27 +255,40 @@ def foo(a: int, b: int) -> int:
...
+        }
+    )
     for k, v in c.items():
         v.v = 2
 
-    assert c == {"a": {"v": 2}, "b": {"v": 2}}
+    assert c == {
+        "a": {"v": 2},
+        "b": {
+            "v": 2,
```
odelalleau, line 289: Can you remove that trailing comma to have a more consistent formatting?

tests/test_basic_ops_dict.py
```diff
@@ -287,6 +300,60 @@ def test_items_with_interpolation() -> None:
...
+        ),
+    ],
+)
+def test_items(cfg: Any, expected: Any, expected_no_resolve: Any) -> None:
+    cfg = _ensure_container(cfg)
+
+    if isinstance(expected, list):
+        assert list(cfg.items()) == expected
+    else:
+        with expected:
```
odelalleau, line 343: What happened to "better handle error situations in a different test"? :P
(not asking you to change it, but curious what is the reasoning you apply when deciding to split or not)

omegaconf/dictconfig.py
```diff
@@ -662,6 +672,9 @@ def _dict_conf_eq(d1: "DictConfig", d2: "DictConfig") -> bool:
         assert isinstance(d2, DictConfig)
         if len(d1) != len(d2):
             return False
+        if d1._is_missing() or d2._is_missing():
+            return d1._is_missing() == d2._is_missing()
```
odelalleau, line 676: Very minor optimization (I doubt we care, but since we've been hunting down bottlenecks...)
```suggestion
            return d1._is_missing() and d2._is_missing()
```

odelalleau (approved):
I have a few minor comments but feel free to merge

tests/test_basic_ops_dict.py
omry: We have CI coverage so I am not worried.

tests/test_basic_ops_dict.py
omry: It was much easier to understand what this test was doing this way.

tests/test_basic_ops_dict.py
omry: What couldn't be changed?

tests/test_basic_ops_dict.py
omry: Don't be annoying :P.

> not asking you to change it, but curious what is the reasoning you apply when deciding to split or not

In this case, the test is testing one input against two outputs. one or both of which can fail.
The overall complexity of the test code is still low.
I think splitting into error cases would result in 3 times more test code without any meaningful gain in test simplicity.

tests/test_basic_ops_dict.py
omry: I will change it now that the test has a more meaningful name.

tests/test_basic_ops_dict.py
omry: If I understand you correctly:
The point of this comment is to say you fixing items to be more consistent is fine. this would require a change to this test.

odelalleau: Ok, the way I read it ("If ithat can be fixed"), it sounded like something that maybe would be difficult to change, and I couldn't see why. Anyway, not a big deal => resolving.

tests/test_basic_ops_dict.py
odelalleau: I suspect you misunderstood which comma I was talking about, but anyway, resolving since not relevant anymore

omegaconf/dictconfig.py
```diff
@@ -662,6 +672,9 @@ def _dict_conf_eq(d1: "DictConfig", d2: "DictConfig") -> bool:
         assert isinstance(d2, DictConfig)
         if len(d1) != len(d2):
             return False
+        if d1._is_missing() or d2._is_missing():
+            return d1._is_missing() is d2._is_missing()
```
odelalleau, line 676: So, to clarify, my suggestion to use `and` wasn't related to the speed of `==`: it was a cheap way to avoid computing `d2._is_missing()` if `d1._is_missing()` was False.
(ideally both would be called only once, but until we can use the new `:=` operator it's probably not worth it as it would make the code more complex for very little gain)

PR #626, opened 2021-03-19
optimized _is_missing_literal

omry:
Further speedup of ListConfig iteration.
```
                                        master     now     speedup
test__is_missing_literal                439.6 ns   96.0 ns    4.5x
test__list_iter[small_listconfig]       11.04 us   9.11 us    1.2x
```

### DIFF
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -349,11 +349,8 @@ def _is_missing_value(value: Any) -> bool:
 
 
 def _is_missing_literal(value: Any) -> bool:
-    from omegaconf import MISSING
-
-    ret = isinstance(value, str) and value == MISSING
-    assert isinstance(ret, bool)
-    return ret
+    # Uses literal '???' instead of the MISSING const for performance reasons.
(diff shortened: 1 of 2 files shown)


omegaconf/_utils.py
```diff
@@ -349,11 +349,8 @@ def _is_missing_value(value: Any) -> bool:
 
 
 def _is_missing_literal(value: Any) -> bool:
-    from omegaconf import MISSING
-
-    ret = isinstance(value, str) and value == MISSING
-    assert isinstance(ret, bool)
-    return ret
+    # Uses literal '???' instead of the MISSING const for performance reasons.
+    return isinstance(value, str) and value == "???"
```
omry, line 353: We can make it 25% faster (on top of this function) by throwing away the instance check, but that's conflicting with the dataframe issue (a test will fail).

At this point, we can also consider caching the value type on the node, to avoid recalculating it more than once.

odelalleau (approved):
Wooooooooosh!

PR #627, opened 2021-03-19
Fixed container inspection through interpolating

omry:
Example:
```python
foo = OmegaConf.create({
        "list": ListConfig([{"a": 10}]),
        "inter": ListConfig("${list}"),
    },
)
```
Expanding `foo.inter`, the first element in the list was incorrect.

### DIFF
--- a/pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
+++ b/pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
@@ -1,7 +1,7 @@
 # based on https://github.com/fabioz/PyDev.Debugger/tree/main/pydevd_plugins/extensions
 import os
 import sys
-from typing import Any, Dict, Sequence
+from typing import Any, Dict
 
 from _pydevd_bundle.pydevd_extension_api import (  # type: ignore
     StrPresentationProvider,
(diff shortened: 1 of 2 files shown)


pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
```diff
@@ -29,43 +34,59 @@ def get_dictionary(self, obj: Any) -> Any:
...
     def can_provide(self, type_object: Any, type_name: str) -> bool:
-        Node = find_mod_attr("omegaconf", "Node")
-        return Node is not None and issubclass(type_object, Node)
+        return self.Node is not None and issubclass(type_object, self.Node)
 
     def resolve(self, obj: Any, attribute: Any) -> Any:
         if isinstance(obj, Sequence) and isinstance(attribute, str):
             attribute = int(attribute)
+
+        if isinstance(obj, self.Node):
```
odelalleau, line 53: Minor:
```suggestion
        elif isinstance(obj, self.Node):
```

pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
```diff
@@ -29,43 +34,59 @@ def get_dictionary(self, obj: Any) -> Any:
...
+            obj = obj._dereference_node(throw_on_resolution_failure=False)
```
odelalleau, line 54: `obj` may now be `None` which would cause a crash on the next line (test needed?)

odelalleau (changes requested):

pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
omry: no. the second if should happen regardless of the first.

pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
omry: Yeah, maybe. will check.

pydevd_plugins/extensions/pydevd_plugin_omegaconf.py
odelalleau: Oh, right, my bad.

odelalleau (approved):
Go!

Issue #526, opened 2021-02-08
[Discussion] Resolver cache behavior


odelalleau:
Creating this issue to discuss potential changes to the cache behavior, following discussions in #445 

Open questions:

* Should resolvers be cached or not by default?
* How should the cache key be computed?

For the first question, the current behavior is to cache by default, but maybe this mostly makes sense for "non-deterministic" resolvers like random / time, to avoid config options to change on each access (while it may not be needed for most resolvers). Also, the answer may depend on the next question as well.

For the second question, we discussed several options (in the context of nested interpolations):

1. Current behavior in #445  as of 2021-02-08: use the inputs to the resolver functions as cache key. This may require caching arbitrary data (ex: outputs of other resolvers).
2. Use the string literal as cache key (ex: if `a` is equal to 0, the cache key for `${f:${a}}` would be `"${a}"`and not `0`). This makes the cache simpler, while losing the ability to dynamically depend on the value of other variables.
3. Also account for the node itself in the cache key. For instance, with a config `{"a": "${random:}", "b": "${random:}"}` someone is likely to want `a` and `b` to be two different numbers, while remaining fixed on each access.
4. (Not discussed so far, new idea I'm proposing) Let the user decide whether to cache or not (default: not cached), and which key to use. It would be a bit more verbose but more explicit and flexible. One possible syntax could be to use `[X]` to say that the resolver's output should be stored under key `X`, with `[]` implicitly using the current node as cache key. My previous example would become `{"a": "${random[]:}", "b": "${random[]:}"}`. And if for any reason we wanted both a and b to use the same random sample, it would be instead `{"a": "${random[sample]:}", "b": "${random[sample]:}"}`. The kind of flexible use that it would enable is something like `{"a": "${f[${b}, ${c}]: ${b}, ${c}, ${d}}"}` where the cache key depends on b and c but not on d.

I haven't thought through all these options in depth yet, just throwing them out there to serve as a basis for dicussion.

omry:
Regardless of the default value for caching (I can see pros and cons in either choice), I am leaning toward making things as simple as they can be.
The motivation for introducing the cache was non-deterministic functions (specifically `time()`, not the performance aspect).

I think you are rightfully focusing on the cache key here. it's the bigger issue.
using arbitrary data as cache key seems dangerous to me. It's both a performance risk but also instability risk (a function that suddenly return an un-hashable value will cause issues in some scenarios).
[...]

PR #628, opened 2021-03-19
New resolver cache behavior

odelalleau:
* The cache is now disabled by default in `register_new_resolver()`
* The cache is now based on the string literals representing the inputs,
  instead of the inputs themselves

See #526 for the corresponding discussion.


docs/source/usage.rst
```diff
@@ -534,22 +534,33 @@ namespace, and to use interpolations in the name itself. The following example d
...
+    ...         "cached_2": "${cached:0,${uncached}}",
+    ...     }
+    ... )
     >>> # same value on repeated access thanks to the cache
-    >>> assert c.cached == c.cached == 7220
+    >>> assert c.cached_1 == c.cached_1 == 7220
     >>> # not the same since the cache is disabled
     >>> assert c.uncached != c.uncached
+    >>> # same string literal "${uncached}" => same value
+    >>> assert c.cached_2 == c.cached_2 == 1192
```
omry, line 563: I think we should also make the point that the caching is at the function call level and not at the node level:

```
"cached_1": "${cached:0,10000}",
"cached_3": "${cached:0,10000}",
```

cached_1 and cached_3 will always output the same value.

omegaconf/_utils.py
```diff
@@ -794,64 +789,3 @@ def is_generic_dict(type_: Any) -> bool:
...
-            raise  # pragma: no cover
-        return -1 if idx_x < idx_y else 1
-
-
-_safe_key = cmp_to_key(_safe_cmp)
-
-
-def _safe_items_sort_key(kv: Tuple[Any, Any]) -> Any:
-    """Safe function to use as sort key when sorting items in a dictionary"""
-    return _safe_key(kv[0])
```
omry, line 857: This makes me happy :)

omegaconf/omegaconf.py
```diff
@@ -411,7 +410,7 @@ def resolver_wrapper(
     def register_new_resolver(
         name: str,
         resolver: Resolver,
-        use_cache: Optional[bool] = True,
+        use_cache: Optional[bool] = False,
     ) -> None:
```
omry, line 414: make use_cache keyword only:
```suggestion
        *,
        use_cache: Optional[bool] = False,
    ) -> None:
```

I suspect we will want to add support for a more flexible caching policy later and this will make it easier.

omegaconf/omegaconf.py
```diff
@@ -399,7 +398,7 @@ def resolver_wrapper(
                     f"`register_new_resolver()` instead (see "
                     f"https://github.com/omry/omegaconf/issues/426 for migration instructions)."
                 )
-            key = args
+            key = args_str
```
omry, line 401: is this sensitive to whitespace and other insignificant changes:
```
c1 : ${cached:a, b}
c2 : ${cached:a,b}
```
I think those will have the same key. If that's true:
I think a way around it is to use string representation of the parsed input as key:
```
key = "||".join(str(x) for x in args)
```

omry (changes requested):

docs/source/usage.rst
odelalleau: Indeed, done in 71b5536 (I made it `cache_2` as I felt like it flowed better that way)

omegaconf/_utils.py
odelalleau: I knew it!

omegaconf/omegaconf.py
odelalleau: Good idea => done in 3a76edc

docs/source/usage.rst
omry: > Indeed, done in [71b5536](https://github.com/omry/omegaconf/commit/71b553695cc4adeca48d3940817b4d835a3fc190) (I made it `cache_2` as I felt like it flowed better that way)

Yeah, I didn't really suggest that you use cached_3 for this one. cached_2 makes more sense :)

omegaconf/omegaconf.py
odelalleau: Not sure I understand. This is *not* sensitive to whitespaces like the one you showed (in fact, I intentionally introduced an extra whitespace in the `cached_2` doc example to show it implicitly).

It will however be sensitive to other kinds of whitespaces. For instance `${foo}` and `${ foo }` both evaluate to the same value but have different string literal representations.

We can't use `args` unless we do another 180 to base the key on the arguments and not the string literals (for instance, for an interpolation we could have `10` in `args` and `"${foo}"` in `args_str`)

omry: Oh, okay: args_str is a Tuple of strings. I thought it was a single string.
Why do we even want to support `${  foo  }` as an alternative to `${foo}` ?

We can either be more restrictive or maybe cannonize the passed input (which sounds a bit hard to me).

odelalleau: > Why do we even want to support `${ foo }` as an alternative to `${foo}` ?

I think that back when we were discussing the grammar you mentioned ignoring whitespaces around braces. I can imagine situations where it could make things a bit more readable, e.g. `${ ${foo:bar} }`.
That being said, no objection if you prefer to remove this ability (it doesn't seem like it was allowed in 2.0).

> We can either be more restrictive or maybe cannonize the passed input (which sounds a bit hard to me).

I wouldn't canonize it (too hard indeed). The example I gave is just one case, there are plenty more (ex: `1000` vs `1_000`, `"foo"` vs `foo`, `{a: b}` vs `{a:b}`, etc.)
I don't think we can ever hope the string literal representation to be a unique identifier of the value that is being passed to the resolver. If you prefer, we can switch to the string representation of the input value as you suggested (though I would use `repr()` instead of `str()`). It's closer to what I had before, but without the added complexity of handling different data types.
Let me know what seems best to you.

omry: > I think that back when we were discussing the grammar you mentioned ignoring whitespaces around braces. I can imagine situations where it could make things a bit more readable, e.g. ${ ${foo:bar} }.
That being said, no objection if you prefer to remove this ability (it doesn't seem like it was allowed in 2.0).

I think if we can find a way to make it work in a more unique way we can keep this flexibility.


> I wouldn't canonize it (too hard indeed). The example I gave is just one case, there are plenty more (ex: 1000 vs 1_000, "foo" vs foo, {a: b} vs {a:b}, etc.)
I don't think we can ever hope the string literal representation to be a unique identifier of the value that is being passed to the resolver. 

Okay, let's drop this idea.

> If you prefer, we can switch to the string representation of the input value as you suggested (though I would use repr() instead of str()). It's closer to what I had before, but without the added complexity of handling different data types.
Let me know what seems best to you.

I still want this to be static, so it should not resolve interpolations (including custom resolvers).
Maybe something like this:
```
1 -> 1 - > "1"
1_0  -> 10 - > "10"
${foo} -> ${foo} -> ${foo}
${foo:1,2} -> ${foo, [1,2]} -> "${foo, [1,2]}"
${foo:1,"2 "} -> ${foo, [1,"2 "]} ->  "${foo, ['1','2 ']}"
```
But I suspect this would actually require converting the parse tree to an AST as an intermediate step.
Going through an AST would make things easier to test as well, but it's a pretty big change.

omry: Do we have an intermediate representation of an interpolation as something with the data of:
```
${foo} : Interpolation(key="foo")
${bar:a,b,c} : CustomResolver(name="bar", args=["a","b","c"]) 
${bar:${z:1,2}} : CustomResolver(name="bar", args=[CustomResolver(name="z", args=[1,2])]) 
```
?

odelalleau: > Do we have an intermediate representation of an interpolation as something with the data of:
> (...)

At the moment we don't keep track of these.

I don't think it would be too hard to achieve what you have in mind, but I expect it to be a bit annoying (mostly due to quoted strings -- also should make sure to only do it for resolvers with cache enabled to avoid adding overhead everywhere).

It's something I'd rather not get into for 2.1, considering that IMO it adds little benefit over the current basic approach based on raw string literals (it's only going to be useful in niche cases).

omry: This is more similar to what I did in Hydra.
The visitor is returning an AST, which is easier to deal with:
1. It's easier to test ([example](https://github.com/facebookresearch/hydra/blob/master/tests/test_overrides_parser.py#L209-L234)).
2. It's easier to reason about.
3. It allows canonization

In Hydra, primitive rules are returning primitive results, except quoted strings  which returns QuotedString object, which allows reconstruction of the original input if needed. ([link](https://github.com/facebookresearch/hydra/blob/master/tests/test_overrides_parser.py#L495-L643)).

Number 2 is important:
For example, what would this return?
```
x: 'foo ${bar}'
```

I hope that would be quoted string `foo ${bar}` that is later parsed (like any other string) for interpolations in it, something like `QuotedString(quote="'", text="foo ${bar}")`.

Note that this will allow the eliminations of the callbacks, disentangling the tests of the parsing from the test of interpolation behavior.

docs/source/usage.rst
```diff
@@ -534,23 +534,36 @@ namespace, and to use interpolations in the name itself. The following example d
...
-    ...                       "uncached": "${uncached:0,10000}"})
+    ...    "cached", random.randint, use_cache=True
+    ... )
+    >>> OmegaConf.register_new_resolver("uncached", random.randint)
+    >>> c = OmegaConf.create(
+    ...     {
+    ...         "cached_1": "${cached:0,10000}",
+    ...         "cached_2": "${cached:0, 10000}",
+    ...         "uncached": "${uncached:0,10000}",
+    ...         "cached_3": "${cached:0,${uncached}}",
```
omry, line 556: This looks like a better order, first demonstrate the problem that ocurres, then the solution, then the limitations.
```suggestion
    ...         "uncached": "${uncached:0,10000}",
    ...         "cached_1": "${cached:0,10000}",
    ...         "cached_2": "${cached:0, 10000}",
    ...         "cached_3": "${cached:0,${uncached}}",
```

omry (approved):
Overall I think this is good.
We have the problem I pointed to with the cache keys in some scenarios but I think it's a rare problem (if surprising).

The proposed change about using an AST is a big one, we may still want to do it for 2.1 but we can separate the discussion from this diff (but remember that this caching issue is one of the motivations for it).

omegaconf/omegaconf.py
odelalleau: > The visitor is returning an AST

The way I see it, the tree that the visitor is going through is already a kind of AST, isn't it? If we return yet another tree then we'll need to visit it anyway at some point in order to resolve it and get the actual value we care about.
(not saying there couldn't be benefits in "transforming" the tree provided by ANTLR into a different one that would be more convenient for our needs though)

> Number 2 is important:
> For example, what would this return?
> 
> ```
> x: 'foo ${bar}'
> ```

If this is an excerpt from a YAML file, then this is something like a `ConfigValue([TopLevelStr("foo "), NodeInterpolation("bar")])`

If this is a key/value pair while parsing a dict, it would be something like `dictKeyValuePair(dictKey=ID("x"), element=QuotedString(quote="'", ...)`. I can see two options for `...`:
- Either it's something like you wrote: `text="foo ${bar}"`
- Or we actually parse it: `content=ConfigValue([TopLevelStr("foo "), NodeInterpolation("${bar}")])`

The advantage of the second option, from a cache perspective, is that it would make it more robust to irrelevant changes in the quoted content (ex: using `${ bar }` instead of `${bar}`).

This is why I mentioned quoted strings as making things a bit annoying. Also, ideally I'd want `${foo:bar}` and `${foo:"bar"}` to use the same cache key (but that shouldn't be too hard considering that unquoted strings are quite limited in what they support).

> Note that this will allow the eliminations of the callbacks, disentangling the tests of the parsing from the test of interpolation behavior.

Right now we could easily test that the parsing runs without error separately from any callback, if we wanted to.
It's true that we don't have tests for the correctness of the parsing though. Maybe I'm lazy but given the number of tests, I find it easier to provide the final expected value than the expected parse tree ;)

> I think the overhead is limited to interpolations in general. so we are already protected from significant slowdown.

Right, may not be a meaningful concern.

Bottom line: I agree that we could build our own tree to replace the one from ANTLR. I'd rather not do that now myself though.

docs/source/usage.rst
odelalleau: Done in a84e17d
(also rebased on top of master)

odelalleau:
> We have the problem I pointed to with the cache keys in some scenarios but I think it's a rare problem (if surprising).

If we don't care about the overhead + the parsing of quoted strings, it should actually be pretty easy to build the cache key when visiting the tree in order to make it a bit more robust. Though if the plan is to do a bigger refactor where we create a different tree (the AST you mentioned), it's probably not worth it.

omry:
> If we don't care about the overhead + the parsing of quoted strings, it should actually be pretty easy to build the cache key when visiting the tree in order to make it a bit more robust. Though if the plan is to do a bigger refactor where we create a different tree (the AST you mentioned), it's probably not worth it.

I am not sure I will have the time to spend to build that AST before OmegaConf 2.1.
I think this is a corner case unlikely to bite many people (especially given that cache is no longer the default).
At this point I think we can leave it as it and consider your proposal as a bugfix if it comes up as a significant issue.

odelalleau:
> I think this is a corner case unlikely to bite many people (especially given that cache is no longer the default).
> At this point I think we can leave it as it and consider your proposal as a bugfix if it comes up as a significant issue.

Agreed, sounds good to me!

PR #629, opened 2021-03-19
Avoid crash when registering a non-inspectable resolver

odelalleau:
Some builtin functions like `print()` cannot be inspected.

### DIFF
--- a/tests/test_interpolation.py
+++ b/tests/test_interpolation.py
@@ -1,5 +1,4 @@
 import copy
-import inspect
 import random
 import re
 from textwrap import dedent
@@ -7,7 +6,6 @@ from typing import Any, Dict, List, Optional, Tuple
 
 import pytest
 from _pytest.python_api import RaisesContext
-from pytest import mark
 
 from omegaconf import (
     II,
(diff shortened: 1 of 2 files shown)


Jasha10 (approved):
It's also possible for `inspect.signature` to raise a `TypeError` for unsupported objects, e.g. `inspect.signature(123)`, but most of those cases should get caught by the `assert callable(resolver)` on line 427.

odelalleau:
> It's also possible for `inspect.signature` to raise a `TypeError` for unsupported objects, e.g. `inspect.signature(123)`, but most of those cases should get caught by the `assert callable(resolver)` on line 427.

Yes, that was also my assumption (so I preferred not to silence it -- so that we know if it actually comes up)

tests/test_interpolation.py
```diff
@@ -499,6 +501,20 @@ def foo() -> int:
         OmegaConf.register_new_resolver("foo", lambda: 10)
 
 
+try:
+    inspect.signature(print)
+    print_is_inspectable = False
+except ValueError:
+    print_is_inspectable = False
```
omry, line 508: If we find an example of a function that is always un-inspectable, let's use it.
otherwise, good to go.

omry (approved):

tests/test_interpolation.py
odelalleau: I looked a bit (at the doc of `inspect.signature()`, at a way to create a non-inspectable object in Python, and at the content of a function to see if I could delete some attributes to fool `inspect`), but couldn't easily find something. So that's the best I could come up with.

omry: This looks like a good use case for mocking the function to raise the expected exception.

odelalleau: Good idea, let me give it a shot

odelalleau: Done in 8c67b8c

omry (approved):
awesome.

PR #630, opened 2021-03-19
Fix edge cases in the parsing of quoted values

### LINKED ISSUE #617: Quoted strings are not always properly parsed by the grammar

odelalleau:
Fixes #617

Let's agree on this fix before continuing the discussion in #621.


tests/test_grammar.py
```diff
@@ -309,6 +308,7 @@
     ("str_top_middle_escapes", r"abc\\\\\${str}", r"abc\\${str}"),
     ("str_top_trailing_escapes", "${str}" + "\\" * 5, "hi" + "\\" * 3),
     ("str_top_concat_interpolations", "${null}${float}", "None1.2"),
+    ("str_top_issue_617", r""" ${test: "hi\\" }"} """, ' hi\\"} '),
```
omry, line 311: I would like to see many more less confusing tests around this issue to convince me this is working in an intuitive way.

Ideally, those tests would be in a standalone test function and not overload the already big config object.

omry (changes requested):

omegaconf/grammar/OmegaConfGrammarLexer.g4
```diff
@@ -60,9 +60,21 @@ ESC: (ESC_BACKSLASH | '\\(' | '\\)' | '\\[' | '\\]' | '\\{' | '\\}' |
...
+//    ""                    -> <empty>
+//    '\\'                  -> \
+//    "\\\\"                -> \\
+//    'abc\\'               -> abc\
+//    "abc\"def\\\'ghi\\\\" -> abc"def\\\'ghi\\
 QUOTED_VALUE:
-      '\'' ('\\\''|.)*? '\'' // Single quotes, can contain escaped single quote : /'
-    | '"' ('\\"'|.)*? '"' ;  // Double quotes, can contain escaped double quote : /"
+      '"' (('\\\\')* | (.)*? ~[\\] ('\\\\')*) '"'     // double quotes
+    | '\'' (('\\\\')* | (.)*? ~[\\] ('\\\\')*) '\'';  // single quotes
```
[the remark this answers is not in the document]
odelalleau: The comments above are meant to explain it => I guess not clear enough?

Maybe it's missing an explanation of the main idea, which is that we want the string to finish by a quote optionally preceded by an even number of `\`. Would that help? Otherwise, suggestion to make it clearer?

tests/test_grammar.py
odelalleau: Agreed on more tests, I realized it afterwards and thought I'd bring in some tests from #621 before merging. I thought I'd wait until I'd get some feedback on the issue + fix first. I guess now is the time :)

There's no need to change the config itself since it's just about testing quoted strings => done in cd17736
If you're curious, here are the tests that would fail before this PR:
```
=========================================================================== short test summary info ===========================================================================
FAILED tests/test_grammar.py::TestOmegaConfGrammar::test_single_element_no_interpolation[str_quoted_bad_1] - Failed: DID NOT RAISE <class 'omegaconf.errors.GrammarParseError'>
FAILED tests/test_grammar.py::TestOmegaConfGrammar::test_single_element_no_interpolation[str_quoted_bad_2] - Failed: DID NOT RAISE <class 'omegaconf.errors.GrammarParseError'>
FAILED tests/test_grammar.py::TestOmegaConfGrammar::test_single_element_no_interpolation[str_quoted_esc_quote_single_bad] - Failed: DID NOT RAISE <class 'omegaconf.errors.G...
FAILED tests/test_grammar.py::TestOmegaConfGrammar::test_single_element_no_interpolation[str_quoted_esc_quote_double_bad] - Failed: DID NOT RAISE <class 'omegaconf.errors.G...
FAILED tests/test_grammar.py::TestOmegaConfGrammar::test_config_value[str_top_issue_617] - assert ' hi\\" } ' == ' hi\\"} '
======================================================================== 5 failed, 248 passed in 0.78s ========================================================================
```

omegaconf/grammar/OmegaConfGrammarLexer.g4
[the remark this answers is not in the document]
omry: The comment is explaining the behavior.
I am asking about the token definition.
Usually I can read tokens/parser rules and get a good about what to expect. this is not the case here.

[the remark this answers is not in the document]
omry: (I am not asking you to change anything in the comment)

tests/test_grammar.py
```diff
@@ -137,6 +137,23 @@
     ("str_quoted_too_many_1", "''a'", GrammarParseError),
     ("str_quoted_too_many_2", "'a''", GrammarParseError),
     ("str_quoted_too_many_3", "''a''", GrammarParseError),
+    ("str_quoted_trailing_esc_1", "'abc\\\\'", "abc\\"),
```
omry, line 140: This seems like the obvious first test for trailing backslash. do you have it somewhere else?
```
    ("str_quoted_trailing_backslash", "'abc\\'", "abc\"),
```

tests/test_grammar.py
```diff
@@ -137,6 +137,23 @@
...
+    ("str_quoted_trailing_esc_2", "'abc\\\\\\\\'", "abc\\\\"),
+    ("str_quoted_no_esc_1", '"abc\\def"', "abc\\def"),
```
omry, line 142: As far as I understand, this is interpreted as `abc\def` in the current grammar, right?
I understand this can simplify things, but I think there is value in simple and consistent handling of backslash escaping.

omegaconf/grammar/OmegaConfGrammarLexer.g4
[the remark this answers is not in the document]
odelalleau: Here's the full rule as a reminder:
```
QUOTED_VALUE:
      '"' (('\\\\')* | (.)*? ~[\\] ('\\\\')*) '"'     // double quotes
    | '\'' (('\\\\')* | (.)*? ~[\\] ('\\\\')*) '\'';  // single quotes
```

The comments above explain the different chunks in order. Let's go through them:

> Quoted values for both types of quotes.

This is why there are two lines (one for double quotes and one for single quotes).
In the following, let's look only at the first line.

> A quoted value is made of the enclosing quotes...

These are the first and last elements in the expression: `'"'` ... `'"'`

> ... and either:
> - nothing else
> - an even number of backslashes (meaning they are escaped)

These are captured by the first option for the content of the string, matched by: `('\\\\')*`

> - a sequence of any character, followed by any non-backslash character, and optionally an even number of backslashes (i.e., also escaped)

This is captured by the second option for the content of the string, matched by: `(.)*? ~[\\] ('\\\\')*`. We have respectively:
* a sequence of any character: `(.)*?`
* any non-backslash character: `~[\\]`
* optionally an even number of backslashes: `('\\\\')*`

Let me know if that helps. Also if you can think of a simpler way to express the same logic, happy to take it!

tests/test_grammar.py
odelalleau: Nope, I realized such tests were missing when working on #621 (I agree this one should have been there earlier, and some of the other ones as well)

tests/test_grammar.py
odelalleau: Are you saying that *all* backslashes should be escaped? We could do that, but combined with the escaping in Python strings and the current "double un-escaping" issue (#615), this could make things quite ugly:
```python
OmegaConf.register_new_resolver("listdir", os.listdir)
cfg = OmegaConf.create({
    "all_files": "${listdir: '${drive}:\\\\\\\\path\\\\\\\\to\\\\\\\\folder'}",
    "drive": "C",
})
cfg.all_files  # will list content of C:\path\to\folder
```

omry: Isn't that the case now in Hydra?

The escaping in Python strings is not our concern. A user can use raw string to make things cleaner.
The double unescaping is an implementation detail, I am not sure it's really needed to make things work.

With those two statements, the above should become which seems reasonable to me.
```python
cfg = OmegaConf.create({
    "all_files": "${listdir: r'${drive}:\\path\\to\\folder'}",
    "drive": "C",
})
```

odelalleau: > Isn't that the case now in Hydra? (and for that matter, in OmegaConf as well?)

Not sure about Hydra. But in OmegaConf a single backslash is ok: the fact that we un-escape double `\\` into `\` doesn't mean we have to touch single `\`.

> The escaping in Python strings is not our concern. A user can use raw string to make things cleaner.

Maybe. There are a few places where raw strings don't work though, and there's also shell escaping (e.g. with Hydra).

> The double unescaping is an implementation detail, I am not sure it's really needed to make things work.

I'm not sure either but at this point I've given up on fixing it (besides going the other way that you're suggesting, i.e. what I did in #621 where we *only* escape backslashes before quotes or interpolations).
If you have an idea that you believe may work, let me know what it is and I can try to either implement it or show you why it doesn't work :)

> With those two statements, the above should become which seems reasonable to me. (...)

Just a note that I fixed your example (the `r` was misplaced) -- it doesn't change your point.

I think that forcing double `\\` should be part of a different PR, if that's what we want to do.

omry: > > Isn't that the case now in Hydra? (and for that matter, in OmegaConf as well?)
> 
> Not sure about Hydra. But in OmegaConf a single backslash is ok: the fact that we un-escape double `\\` into `\` doesn't mean we have to touch single `\`.

Ok.
 
> > The escaping in Python strings is not our concern. A user can use raw string to make things cleaner.
> 
> Maybe. There are a few places where raw strings don't work though, and there's also shell escaping (e.g. with Hydra).

Escaping is not a thing in bash, They do have a syntax for literal strings though:
https://unix.stackexchange.com/questions/30903/how-to-escape-quotes-in-shell

> > The double unescaping is an implementation detail, I am not sure it's really needed to make things work.
> 
> I'm not sure either but at this point I've given up on fixing it (besides going the other way that you're suggesting, i.e. what I did in #621 where we _only_ escape backslashes before quotes or interpolations).
> If you have an idea that you believe may work, let me know what it is and I can try to either implement it or show you why it doesn't work :)

Can you create a small example and explain how it leads to double unescaping?
This will help me play with it.

> > With those two statements, the above should become which seems reasonable to me. (...)
> 
> Just a note that I fixed your example (the `r` was misplaced) -- it doesn't change your point.
> 
> I think that forcing double `\\` should be part of a different PR, if that's what we want to do.
> 
> Edit: as a minor note, this would be a breaking change to existing string interpolations in 2.0: `{"path": "${drive}:\test"}`, as I'm assuming that if we consistently enforce `\\` in quoted strings, we should also do it in top-level strings (i.e. outside of interpolations)

I don't think we should make big changes (if any) outside of quoted strings and I think I am convinced we should not require escaping of free backslash if we can make it work and be consistent.

I introduced escaping in the Hydra grammar to support single quote in a single quoted string and double quote in a double quoted string.
The introduction of nested interpolations added more use cases to escaping:

1. Nesting of quoted strings for the purposes of nested interpolations.
This is a pretty rare one. and in most cases using a different quote style should be enough.
```
'hello ${identity:"James Bond"}'
# however this turns weird if you have another level:
'hello ${identity:"James ${identity:"bond"}"}'
# or this?
'hello ${identity:"James ${identity:'bond'}"}'
# Both are wrong in their own way.

# But this should probably work:
'hello ${identity:"James ${identity:\'bond\'}"}'
```
The above does not require additional double escaping because we are using the other quote style as well.
If we didn't:
```
'hello ${identity:\'James ${identity:\\'bond\\'}'\}'
```

2. Escaping the interpolation itself:
```
\${aa}
\${foo.bar}
\${oc.env:USER}
```
multiple escaping should never be needed here. I think it's required because of your current implementation of the unescaping mechanism.

```
# no escaping:
${foo:${bar:zoo}}
# both
\${foo:\${bar:zoo}}
# inner only:
${foo:\${bar:zoo}}
```

This can be potentially be handled by introducing a token like ESC_INTER_OPEN: `\${`
By the way, this should also work outside of quoted strings because support interpolations in unquoted strings.

PS: None of the examples above are tested so mistakes are likely.

odelalleau: > Escaping is not a thing in bash, They do have a syntax for literal strings though:
> https://unix.stackexchange.com/questions/30903/how-to-escape-quotes-in-shell

Really? In my zsh `echo \\` only displays one `\`. But I haven't dug deeper. Not a big deal.

> Can you create a small example and explain how it leads to double unescaping?
> This will help me play with it.

In the following example, as of the current HEAD in this branch (cd17736):

```python

OmegaConf.clear_resolvers()

cfg = OmegaConf.create({
    "x": r"${identity:'I want two \\\\\\\\ and an escaped \\\\\\${interpolation}'}",
    "interpolation": "do_not_use"
})
print(cfg.x)  # I want two \\ and an escaped \${interpolation}
```
This example prints the desired output. The goal is to cut down the number of `\` (NB: although you could use a few less `\` in the commit I mentioned, I'm assuming that we want to forbid single `\` during an un-escaping phase, so that's the correct number of `\` under this constraint).

What happens is that when we see the quoted string provided as input to the `identity` resolver (i.e., `'I want two \\\\\\\\ and an escaped \\\\\\${interpolation}'`), we do a first un-escaping pass for escaped quotes, which leads (after removing the enclosing quotes) to the string `I want two \\\\ and an escaped \\\${interpolation}`.

This string is then parsed like a regular string describing any other config option, where we do a second un-escaping pass, this time for escaped interpolations. Then it becomes `I want two \\ and an escaped \${interpolation}` which is the desired output.

Removing the double un-escaping should allow (IMO) to get the same output with:
```python
    "x": r"${identity:'I want two \\\\ and an escaped \\\${interpolation}'}",
```

With #621 this gives the desired otuput:
```python
    "x": r"${identity:'I want two \\ and an escaped \\\${interpolation}'}",
```

> 1. Nesting of quoted strings for the purposes of nested interpolations.
>    This is a pretty rare one. and in most cases using a different quote style should be enough.

Yes.

> 'hello ${identity:"James ${identity:'bond'}"}'

Just a note that this one could work. You might remember it was my first approach, so as to reduce the need for escaping. It was based on lexer modes. Not saying we should bring it back.

> 2. Escaping the interpolation itself:
> (...)
> multiple escaping should never be needed here. I think it's required because of your current implementation of the unescaping mechanism.

Based on your examples, I'm not sure we're on the same page here. The "multiple escaping" I have in mind related to escaped interpolations is when you want one (or more) slash before an interpolation: you can't use `C:\${folder}` since it would be interpreted as an escaped interpolation and lead to `C:${folder}`. So instead we need to escape the `\` and write it `C:\\${folder}`, which leads to `C:\<content_of_the_folder_node>`.

> By the way, this should also work outside of quoted strings because support interpolations in unquoted strings.

Escaped interpolations aren't supported in unquoted strings. It's not actually required since you can use a quoted string if you want an escaped interpolation. I don't think this use case is common enough to bother supporting it in unquoted strings.

omry: > > Can you create a small example and explain how it leads to double unescaping?
> > This will help me play with it.
> 
> In the following example, as of the current HEAD in this branch ([cd17736](https://github.com/omry/omegaconf/commit/cd17736ff7493c6f12f05329e8c11f2ab61723ad)):
> 
> ```python
...
> ```

Do we want this to work? 
Can we make it work?
```python
    "x": r"${identity:'I want two \\ and an escaped \${interpolation}'}",
    "y": ${identity:${x}}, # should be the same as x
```
If it doesn't make sense to you, why?
I think it's good to focus on the desired behavior, calling out as many gotchas as we can before proposing changes.

> > 'hello ${identity:"James ${identity:'bond'}"}'
> 
> Just a note that this one could work. You might remember it was my first approach, so as to reduce the need for escaping. It was based on lexer modes. Not saying we should bring it back.

That would be nice to support, but I think we should avoid introducing "new features if possible. We can revisit for a future version.

> > 1. Escaping the interpolation itself:
> >    (...)
> >    multiple escaping should never be needed here. I think it's required because of your current implementation of the unescaping mechanism.
> 
> Based on your examples, I'm not sure we're on the same page here. The "multiple escaping" I have in mind related to escaped interpolations is when you want one (or more) slash before an interpolation: you can't use `C:\${folder}` since it would be interpreted as an escaped interpolation and lead to `C:${folder}`. So instead we need to escape the `\` and write it `C:\\${folder}`, which leads to `C:\<content_of_the_folder_node>`.

I agree with you that double escaping is needed to support this specific 

> > By the way, this should also work outside of quoted strings because support interpolations in unquoted strings.
> 
> Escaped interpolations aren't supported in unquoted strings. It's not actually required since you can use a quoted string if you want an escaped interpolation. I don't think this use case is common enough to bother supporting it in unquoted strings.

That's fine.

odelalleau: > Do we want this to work?
> Can we make it work?
> 
> ```python
>     "x": r"${identity:'I want two \\ and an escaped \${interpolation}'}",
>     "y": ${identity:${x}}, # should be the same as x
> ```

I'm assuming the question is related to your example (and not my example that you quoted).

Personally, I would answer "yes" to both questions, though "work" is ambiguous since you didn't tell what the expected result should be. With #621 printing either `x` or `y` prints:
```
I want two \\ and an escaped ${interpolation}
```
which I believe can be a sensible behavior.

Just note that in my own example, I wanted to also print a `\` in front of `${interpolation}`, so if that's the goal your suggested syntax can't work (since there is a single `\` in front of `${..` it can either be printed or be used to escape the interpolation, but not both).

odelalleau: ## A note on Hydra

I was wondering how un-escaping worked in Hydra's quoted strings. From what I can tell, only `\"` is un-escaped into `"` (for a quoted string enclosed by double quotes). Other `\` characters aren't un-escaped.

This has the advantage of being simple and working almost everywhere... but it relies on the same regex that may not always properly detect the end of a quoted string (I haven't tried to craft a buggy example similar to #617 but I assume it should be possible). I can't think of a way to reliably detect the end of a quoted string without also requiring escaping the `\\` that may come before a quote (i.e., turning `\\\"` into `\"` when inside a quoted string).

omry: > > Do we want this to work?
> > Can we make it work?
> > ```python
> >     "x": r"${identity:'I want two \\ and an escaped \${interpolation}'}",
> >     "y": ${identity:${x}}, # should be the same as x
> > ```
> 
> I'm assuming your questions are related to your example (and not my example that you quoted).

Yes, somehow my very long comment got mangled, not sure how.
I meant my example you quoted.

> Personally, I would answer "yes" to both questions, though "work" is ambiguous since you didn't tell what the expected result should be. With #621 printing either `x` or `y` prints:
> 
> ```
> I want two \\ and an escaped ${interpolation}
> ```
> 
> which I believe can be a sensible behavior.

This is the interpretation I have in mind as well.

> Just note that in my own example, I wanted to also print a `\` in front of `${interpolation}`, so if that's the goal your suggested syntax can't work (since there is a single `\` in front of `${..` it can either be printed or be used to escape the interpolation, but not both).

Yes, if we want to print an escaped interpolation the input should be:
```
\\${interpolation}  # \${interpolation}
```

Another thing, the following example needs to escape the open and close curly braces in order to work:
```python
OmegaConf.register_new_resolver("str.format", str.format, replace=True)
cfg = OmegaConf.create("x: ${str.format:\{\}, 3.14}")
cfg.x # '3.14'
```
This is a bit surprising to me, those are without a $ so they should not be confused with an interpolation.
Am I missing anything?

omry: > ## A note on Hydra
> I was wondering how un-escaping worked in Hydra's quoted strings. From what I can tell, only `\"` is un-escaped into `"` (for a quoted string enclosed by double quotes). Other `\` characters aren't un-escaped.
> 
> This has the advantage of being simple and working almost everywhere... but it relies on the same regex that may not always properly detect the end of a quoted string (I haven't tried to craft a buggy example similar to #617 but I assume it should be possible*). I can't think of a way to reliably detect the end of a quoted string without also requiring escaping the `\\` that may come before a quote (i.e., turning `\\\"` into `\"` when inside a quoted string).
> 
> (*) Edit: actually this may not be an issue in Hydra because currently I don't think you can have anything else outside of quotes when there is a quoted string, so it may not be possible to have ambiguous situations like #617

I suspect once we start to support nested interpolations in the overrides syntax we may have to use a similar grammar to handle the similar similar situations (if we decide that we actually want to concern Hydra with the details of interpolations at all. now that OmegaConf has a proper parser generator, we might be able to simplify the override grammar and make it completely unaware of interpolations. 
We can discuss it once the dust settles on the OmegaConf grammar.

odelalleau: > > Just note that in my own example, I wanted to also print a `\` in front of `${interpolation}`, so if that's the goal your suggested syntax can't work (since there is a single `\` in front of `${..` it can either be printed or be used to escape the interpolation, but not both).
> 
> Yes, if we want to print an escaped interpolation the input should be:
> 
> ```
> \\${interpolation}  # \${interpolation}
> ```

Actually with `\\${interpolation}` I would expect the result to be:
```\<content_of_the_interpolation_node_cast_to_string>```
This is similar to the `C:\\${folder}` example.

To get `\${interpolation}` you would need to use `\\\${interpolation}` (this is why the backslashes need to be escaped before an interpolation).


> 
> Another thing, the following example needs to escape the open and close curly braces in order to work:
> 
> ```python
> OmegaConf.register_new_resolver("str.format", str.format, replace=True)
> cfg = OmegaConf.create("x: ${str.format:\{\}, 3.14}")
> cfg.x # '3.14'
> ```
> 
> This is a bit surprising to me, those are without a $ so they should not be confused with an interpolation.
> Am I missing anything?

This is an unquoted string, backslashes need to be escaped to avoid confusion with a dictionary.

Comment (odelalleau) [reply tests/test_grammar.py:142]: > I suspect once we start to support nested interpolations in the overrides syntax we may have to use a similar grammar to handle the similar similar situations (if we decide that we actually want to concern Hydra with the details of interpolations at all. now that OmegaConf has a proper parser generator, we might be able to simplify the override grammar and make it completely unaware of interpolations.
> We can discuss it once the dust settles on the OmegaConf grammar.

Agreed on finalizing the OmegaConf grammar first (that's why I haven't touched Hydra yet).

I suspect it may be tricky though to make Hydra unaware of interpolations because detecting the end of an interpolation is far from trivial.

Comment (omry) [reply tests/test_grammar.py:142]: Yes, it may be tricky. If we have to we can use similar grammar rules.

Comment (omry) [reply tests/test_grammar.py:142]: > > > Just note that in my own example, I wanted to also print a `\` in front of `${interpolation}`, so if that's the goal your suggested syntax can't work (since there is a single `\` in front of `${..` it can either be printed or be used to escape the interpolation, but not both).
> > 
> > 
> > Yes, if we want to print an escaped interpolation the input should be:
> > ```
> > \\${interpolation}  # \${interpolation}
> > ```
> 
> Actually with `\\${interpolation}` I would expect the result to be:
> `\<content_of_the_interpolation_node_cast_to_string>`
> This is similar to the `C:\\${folder}` example.
> 
> To get `\${interpolation}` you would need to use `\\\${interpolation}` (this is why the backslashes need to be escaped before an interpolation).

What if we made `\${` a token for opening an escaped interpolation?
 
> This is an unquoted string, backslashes need to be escaped to avoid confusion with a dictionary.

Oh, dictionary! how quickly I evict all those details from my limited brain.

Comment (odelalleau) [reply tests/test_grammar.py:142]: > What if we made `\${` a token for opening an escaped interpolation?

Not sure what you mean: `\${` is already a token for escaped interpolations. This is the lexer token:
```
ESC_INTER: '\\${';
```

If you mean `\${a}` resolving to `\${a}` then how would you write a string resolving to `${a}`?


> Oh, dictionary! how quickly I evict all those details from my limited brain.

Haha no worries, honestly between quoted strings, unquoted strings, top-level strings, all having subtle differences, it's easy to lose track of such "details" :p

Comment (omry) [reply tests/test_grammar.py:142]: > Not sure what you mean: `\${` is already a token for escaped interpolations. This is the lexer token:

Is the lexer consuming `\${` as a single token, even inside a quoted string?
I see two possible interpretations to `\\${`:
1. `\` followed by `\${a}`
2. `\\` followed by `${`

```
a: 10
b: '\\${a}'
```
b resolves to
- interpretation 1:  `\${a}`
- interpretation 2:  `\10`

Comment (odelalleau) [reply tests/test_grammar.py:142]: > Is the lexer consuming `\${` as a single token, even inside a quoted string?

Kind of. The lexer actually skips over the content of quoted strings, turning them into a single token `QUOTED_VALUE`.
When the grammar visitor sees a `QUOTED_VALUE`, it removes the enclosing quotes, does an un-escape pass (currently: un-escaping `\"` into `"` and `\\` into `\`, for strings with enclosing double quotes), then re-runs the whole parsing logic over the resulting string => this is when the lexer will see a `\${` as an `ESC_INTER` token (assuming it's at top level).

As example, for a quoted string `"fake \${interpolation}"`:
- Initial lexer pass: `QUOTED_VALUE` (matching `"fake \${interpolation}"`)
- After processing in the visitor: `fake \${interpolation}`
- Second lexer pass: `TOP_STR` (matching `fake `) ESC_INTER (matching `\${`) `TOP_STR` (matching `interpolation}`)

> I see two possible interpretations to `\\${`:
> 
> 1. `\` followed by `\${a}`
> 2. `\\` followed by `${`
> 
> ```
> a: 10
> b: '\\${a}'
> ```
> 
> b resolves to
> 
> * interpretation 1:  `\${a}`
> * interpretation 2:  `\10`

In interpretation 1, how do you write `b` so that it resolves to `\10`?
(in interpretation 2, you can write `b: '\\\${a}` so that it resolves to `\${a}`)

Comment (omry) [reply tests/test_grammar.py:142]: > When the grammar visitor sees a QUOTED_VALUE, it removes the enclosing quotes, does an un-escape pass (currently: un-escaping \" into " and \\ into \, for strings with enclosing double quotes), then re-runs the whole parsing logic over the resulting string => this is when the lexer will see a \${ as an ESC_INTER token (assuming it's at top level).

Are there scenarios in the new code where the second phase is happening more than once on the same characters?
(e.g, what happens in nested quoted strings)?

> In interpretation 1, how do you write b so that it resolves to \10?

No can do.
Do you have a test that ensuring that this is covered?

> (in interpretation 2, you can write b: '\\\${a}' so that it resolves to \${a})

Can you explain the process getting from `\\\${a}` to `\${a}` (in the new code).

Comment (odelalleau) [reply tests/test_grammar.py:142]: > > When the grammar visitor sees a QUOTED_VALUE, it removes the enclosing quotes, does an un-escape pass (currently: un-escaping " into " and \ into , for strings with enclosing double quotes), then re-runs the whole parsing logic over the resulting string => this is when the lexer will see a ${ as an ESC_INTER token (assuming it's at top level).
> 
> Are there scenarios in the new code where the second phase is happening more than once on the same characters?
> (e.g, what happens in nested quoted strings)?

I had a long reply with a detailed example and my laptop crashed on me :/ So here's the short version, let me know if unclear:

1. "The new code" is just a grammar change (it doesn't change how the visitor works)
2. Escaped interpolations `\${` are only un-escaped in top-level strings, so they get un-escaped only once. However, currently the second phase may indeed trigger multiple un-escaping of `\\` into `\` in case of nested quoted strings (this is one thing #621 was meant to avoid).

> 
> > In interpretation 1, how do you write b so that it resolves to \10?
> 
> No can do.
> Do you have a test that ensuring that this is covered?

Yes, this is this test:
```python
("str_top_esc_backslash", r"Esc: \\${str}", r"Esc: \hi"),
```

> > (in interpretation 2, you can write b: '\${a}' so that it resolves to ${a})
> 
> Can you explain the process getting from `\\\${a}` to `\${a}` (in the new code).

It's the same in current master:
1. The lexer turns `\\\${a}` into this sequence of tokens: ESC (matching `\\`), ESC_INTER (matching `\${`), TOP_STR (maching `a}`
2. This is recognized as a `toplevelStr` by the parser
3. The visitor code visiting this node is simply `return self._unescape(ctx.getChildren())` which will concatenate the un-escaped string representation of these 3 tokens, i.e. `\`, `${` and `a}`, yielding the final result `\${a}`

omry: > > > When the grammar visitor sees a QUOTED_VALUE, it removes the enclosing quotes, does an un-escape pass (currently: un-escaping " into " and \ into , for strings with enclosing double quotes), then re-runs the whole parsing logic over the resulting string => this is when the lexer will see a ${ as an ESC_INTER token (assuming it's at top level).
> > 
> > 
> > Are there scenarios in the new code where the second phase is happening more than once on the same characters?
> > (e.g, what happens in nested quoted strings)?
> 
> I had a long reply with a detailed example and my laptop crashed on me :/ So here's the short version, let me know if unclear:
> 
> 1. "The new code" (in this PR) is just a grammar change (it doesn't change how the visitor works)
> 2. Escaped interpolations `\${` are only un-escaped in top-level strings, so they get un-escaped only once. However, currently the second phase may indeed trigger multiple un-escaping of `\\` into `\` in case of nested quoted strings (this is one thing #621 was meant to avoid).

Okay, so once this is in, #621 is supposed to address the repeated un-escaping issue?
 
> > Can you explain the process getting from `\\\${a}` to `\${a}` (in the new code).
> 
> It's the same in current master:
> 
> 1. The lexer turns `\\\${a}` into this sequence of tokens: ESC (matching `\\`), ESC_INTER (matching `\${`), TOP_STR (maching `a}`
> 2. This is recognized as a `toplevelStr` by the parser
> 3. The visitor code visiting this node is simply `return self._unescape(ctx.getChildren())` which will concatenate the un-escaped string representation of these 3 tokens, i.e. `\`, `${` and `a}`, yielding the final result `\${a}`

Gotcha, this answers my question.
Can you explain why we need a toplevelStr?

omegaconf/grammar/OmegaConfGrammarLexer.g4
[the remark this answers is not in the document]
omry: > an even number of backslashes (meaning they are escaped)

What inputs is representing? a string with nothing but slashses?
```
\
\\
\\\
```
?

> a sequence of any character, followed by any non-backslash character, and optionally an even number of backslashes (i.e., also escaped)

It feels like this should support a string that ends with backslashes like in my example above.

tests/test_grammar.py
odelalleau: > Okay, so once this is in, #621 is supposed to address the repeated un-escaping issue?

That was my original plan, though I'm open to other approaches for un-escaping (including not actually fixing it).

> Can you explain why we need a toplevelStr?

This is for string interpolations, so that users can use any character they want outside of the interpolation part, including quotes. You might remember an example we discussed back in the days, that was something like: `"Hi ${name}", he said`. Here, there would be two top-level strings: `"Hi ` and `", he said`.

omegaconf/grammar/OmegaConfGrammarLexer.g4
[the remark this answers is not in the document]
odelalleau: > What inputs is representing? a string with nothing but slashses?
> 
> ```
> \
> \\
> \\\
> ```
> 
> ?

A string with nothing but backslashes is captured by the first option for the content of the string: `('\\\\')*`
The number of backslashes has to be even: `"\"` isn't a valid quoted string since it doesn't have an end quote (the second quote is escaped).

> It feels like this should support a string that ends with backslashes like in my example above.

It does: either there are only backslashes and this is covered above, or there is at least one non-backslash character and this is covered by the second case. But the number of ending backslashes must always be even (for the same reason as above).

tests/test_grammar.py
omry: > That was my original plan, though I'm open to other approaches for un-escaping (including not actually fixing it).

Cool. I think we should try to fix it. I am asking all those questions because I am trying to reach the best solution, not because I think this is not important enough to fix.

> This is for string interpolations, so that users can use any character they want outside of the interpolation part, including quotes. You might remember an example we discussed back in the days, that was something like: "Hi ${name}", he said. Here, there would be two top-level strings: "Hi and ", he said.

Gotcha.

tests/test_grammar.py
```diff
@@ -137,6 +137,23 @@
...
+    ("str_quoted_no_esc_2", '"abc\\\\def"', "abc\\def"),
```
omry, line 143: To make those tests easier to understand, can you use raw strings?
```suggestion
    ("str_quoted_no_esc_2", r'"abc\\def"', r"abc\def"),
```

odelalleau: Done in 3457d0c

(note: one thing a bit annoying with raw strings is they don't "work" for strings that end with a `\` or contain an escaped quote of the same quote type as the enclosing quotes, e.g. `"abc\"def"` => this is why there still remain some non-raw strings)

tests/test_grammar.py
omry: Can you add the missing tests here?
They will make it easier for me to form a mental model of what the behavior is after this change.
some input strings single quoted strings.
```
''
'a' 
'\a'
'a\'
'"'
'\''
'c:\\''
```

Here are some nested quoted strings cases, some may be handled in #621:
```
'${f:"b"}' # nested double quotes
'${f:\'b\'}' # nested escaped single quotes
```
Same as above, but for an escaped interpolation. what is the expected behavior here?
enerally speaking, is parsing inside interpolations the same as parsing outside?
```
'\${f:"b"}'
'\${f:\'b\'}'
```

omry (commented):
Tests looks good.
I am starting to think that maybe we should drive those specific tests from an input text file to eliminate the Python shenanigans. 

I added a comment about possible adding more tests here. (maybe they already covered).

tests/test_grammar.py
odelalleau: > Can you add the missing tests here?

Oh, I now realize that I had probably misunderstood your previous question "This seems like the obvious first test for trailing backslash. do you have it somewhere else?" and as a result my answer was a bit off.

The example you gave:
```python
("str_quoted_trailing_backslash", "'abc\\'", "abc\"),
```
is actually invalid because `"'abc\\'"` is the string with characters `'abc\'` and is an invalid quoted string according to the grammar, since the end quote is escaped (=> it's missing another end quote).
The test for a trailing backslash is the one on the line you highlighted:
```python
("str_quoted_trailing_esc_1", "'abc\\\\'", "abc\\"),
```
where `"'abc\\\\'"` is the string with characters `'abc\\'` which is now a valid quoted string.

> (some examples)

I added all your examples in 7f23c1b
(some are named with "tmp" because they are redundant with existing tests, I still kept them here to make it easier for you to check them out, but I will remove them afterwards)
All of them behave as expected -- or at least how I would expect :)

> Generally speaking, is parsing inside interpolations the same as parsing outside?

Depends on what you mean. Generally not, because outside of interpolations we do nothing except un-escaping (because of escaped interpolations) and otherwise just copy any character that we see (there are no constraints). But when we are inside an interpolation (i.e., parsing resolver arguments), the logic is different (to detect dicts, lists, quoted strings, etc.)

*But* when we enter a quoted string (which would be found inside an interpolation) we start parsing its content as if we were outside of an interpolation (it's the actual implementation: we re-start a parsing from scratch on the content of the quoted string -- after the un-escaping pass for escaped quotes). In that sense, the parsing of a quoted string inside an interpolation is the same as the parsing outside of interpolations.

tests/test_grammar.py
```diff
@@ -137,6 +137,30 @@
...
+    ("str_quoted_esc_quote_double_3", r'"abc\\\\\"def"', r'abc\\"def'),
+    ("str_quoted_esc_quote_double_4", r'"a\"b\"cdef\\\""', r'a"b"cdef\"'),
+    ("str_quoted_esc_quote_double_bad", r'"abc\\"def"', GrammarParseError),
+    ("str_quoted_empty", "''", ""),
+    ("str_quoted_basic", "'a'", "a"),
+    ("str_quoted_tmp_1", r"'\a'", r"\a"),
+    ("str_quoted_tmp_2", r"'a\'", GrammarParseError),
+    ("str_quoted_inside_quote_different", "'\"'", '"'),
+    ("str_quoted_inside_quote_same", r"'\''", "'"),
+    ("str_quoted_extra_quote", r"'c:\\''", GrammarParseError),
```
omry, line 163: This works?
```suggestion
    ("str_quoted_trailing_backslash", r"'c:\\'", "c:\"),
    ("str_quoted_extra_quote", r"'c:\\''", GrammarParseError),
```

Loos like rstrings are working for it:
```
print(r"'c:\\'")
'c:\\'
```
I would expect `\\` to be interpreted as a token equal to `\`, so the content of the quoted string should be `c:\`

tests/test_grammar.py
omry: > is actually invalid because "'abc\\'" is the string with characters 'abc\' and is an invalid quoted string according to the grammar, since the end quote is escaped (=> it's missing another end quote).

I think I meant the characters `'abc\\'`. This is again getting confusing because of Python interpreting the backslashes as well. Is this one of the  cases where r strings are not working?

Thanks for clarifying the parsing logic.

tests/test_grammar.py
odelalleau: That would work (after fixing `"c:\"` which isn't a valid Python string) but there's already a similar test so no need to add it:
```python
("str_quoted_trailing_esc_1", "'abc\\\\'", "abc\\"),
```

r-strings work as long as the "escaped" quote isn't the same type as the enclosing quote.

tests/test_grammar.py
odelalleau: > I think I meant the characters `'abc\\'`. This is again getting confusing because of Python interpreting the backslashes as well. Is this one of the cases where r strings are not working?

Ok so the string with characters `'abc\\'` is the one from the same test:
```python
("str_quoted_trailing_esc_1", "'abc\\\\'", "abc\\"),
````

The reason why I didn't use r-strings is because the last string (`"abc\\"`) can't be converted into an r-string, and I found that using an r-string only for the first one could be confusing (when trying to compare before/after). It would look like this:
```python
("str_quoted_trailing_esc_1", r"'abc\\'", "abc\\"),
````
which I find a bit confusing because at first sight it looks like we have the same number of `\` in both strings.

File: tests/test_grammar.py (hunk 6 above)
Comment (omry) [reply tests/test_grammar.py:163]: That test is similar, but is testing for a double trailing backslash.
Since this is all about backslash and escaping, I think an additional basic test for a single backslash is worth it. 

And now I realize that the double backslash is due to Python interpretation of escaping.
Why not use r-strings here?
```python
In [3]: print(r"'abc\\'")
'abc\\'
```

odelalleau: Yeah it's a bit tricky to get those `\` right :)

> Why not use r-strings here?

I explained my reasoning in https://github.com/omry/omegaconf/pull/630#discussion_r613615877 (I don't mind changing it though if you like the "partial" r-string version better)

odelalleau:
Just rebased on top of master to solve a (minor) merge conflict

tests/test_grammar.py
omry: yeah, let's use r-strings whenever possible, where it's not possible add a comment.
This part of the tests could be better tested by reading raw data from a file instead of encoding the data data in Python but I am not sure it's worth it to do it just for the escaping.

odelalleau: Ok, done in 9011928
I used a couple of tricks to use raw strings everywhere -- I can revert those if you don't like them:

Using `.strip()` for strings ending in with a `\`, e.g.
```python
("str_quoted_trailing_esc_1", r"'abc\\'", r" abc\ ".strip()),
```
(technically the leading whitespace is not needed, but I find it looks better)

Using an f-string to include tabs in r-strings, e.g.
```python
("str_esc_ws_2", fr"\ \{TAB}\{TAB}", f" {TAB}{TAB}"),
```
where `TAB` is a constant defined as
```python
TAB = "\t"  # to be used in raw strings, e.g. `fr"C:\{TAB}foo"`
```

omry:
will take a look soon, this week is the week of PTED21 so we have been focused on it.

tests/test_grammar.py
```diff
@@ -111,17 +113,17 @@
...
-    ("str_esc_ws_2", "\\ \\\t\\\t", " \t\t"),
+    ("str_esc_ws_2", fr"\ \{TAB}\{TAB}", f" {TAB}{TAB}"),
     ("str_esc_comma", r"hello\, world", "hello, world"),
     ("str_esc_colon", r"a\:b", "a:b"),
     ("str_esc_equal", r"a\=b", "a=b"),
     ("str_esc_parentheses", r"\(foo\)", "(foo)"),
     ("str_esc_brackets", r"\[foo\]", "[foo]"),
     ("str_esc_braces", r"\{foo\}", "{foo}"),
-    ("str_esc_backslash", r"\\", "\\"),
+    ("str_esc_backslash", r" \ ".strip(), r" \ ".strip()),
```
omry, line 123: put a comment explaining the strip hack, it's not obvious what it is and why you do it.

tests/test_grammar.py
omry: When reviewing I was scratching my head a bit before figuring out the strip trick.
I think it's fine, but add a comment.
maybe something like:
```
# raw strings does not allow trailing \, adding a space and stripping it.
```

tests/test_grammar.py
```diff
@@ -227,10 +253,14 @@
     # Interpolations in quoted strings.
     ("str_quoted_inter", "'${null}'", "None"),
     ("str_quoted_esc_single_1", r"'ab\'cd\'\'${str}'", "ab'cd''hi"),
-    ("str_quoted_esc_single_2", "'\"\\\\\\\\\\${foo}\\ '", r'"\${foo}\ '),
+    ("str_quoted_esc_single_2", r"""'"\\\\\${foo}\ '""", r'"\${foo}\ '),
```
omry, line 256: Can you explain this one? I find it surprising that we need 5 slashes before the $ and 1 slash after the }.
Is this something that will be addressed by the followup double escaping diff? if so, what will it look like after we eliminate the double unescaping?

tests/test_grammar.py
```diff
@@ -250,7 +280,7 @@
     ("nested_select", "${options.${choice}}", "A"),
     ("nested_select_getitem", "${options[${choice}]}", "A"),
     ("nested_relative", "${${rel_opt}.b}", "B"),
-    ("str_quoted_nested", r"'AB${test:\'CD${test:\\'EF\\'}GH\'}'", "ABCDEFGH"),
+    ("str_quoted_nested", r"'AB${test:\'CD${test:\\\'EF\\\'}GH\'}'", "ABCDEFGH"),
```
omry, line 283: Can you explain this one?
Is this something that will be addressed by the followup double escaping diff?

tests/test_grammar.py
odelalleau: Done in 6015a71

tests/test_grammar.py
odelalleau: Done in 6015a71

tests/test_grammar.py
odelalleau: No, I would expect this one to remain unchanged. Here's what happens:
- Un-escaping of quoted string: `AB${test:\'CD${test:\\\'EF\\\'}GH\'}` ==> `AB${test:'CD${test:\'EF\'}GH'}`
- Un-escaping of top-level strings: no change because the only top-level string is `AB`, the rest is an interpolation and interpolations aren't touched in this step
- We then start processing the interpolation
- Un-escaping of quoted string: `CD${test:\'EF\'}GH` ==> `CD${test:'EF'}GH`
- Un-escaping of top-level strings: no change (the top-level strings are `CD` and `GH`)
- We then process the interpolation which is a normal one (yielding the string `EF`)

The reason why it worked before with *fewer* backslashes in the middle is because a string like `'foo\\'bar'` was accepted as a valid quoted string by the grammar, while this is not the case anymore: now this would be parsed as a quoted string `'foo\\'` followed by `bar'` which is an invalid expression. I don't think we could have kept the previous behavior without facing the subtle bugs of #617.

omry: 1. So the reason we are unescaping in top level strings is just to support escaping interpolations there?
If so, would it simplify things to just say that we only apply the grammar inside interpolations and thus only support escaping of nested interpolations?
2. I am a bit confused by this:
```
input:  r"'AB${test:\'CD${test:\\'EF\\'}GH\'}'"
actual string: 'AB${test:\'CD${test:\\'EF\\'}GH\'}'
```

This is a string outside of interpolation, and it's quoted. why is the result not quoted as well?
I am assuming that this string is interpreted as a value of a config node, so there is no yaml involved here.

3. I don't like how we need to unescape the nested quoted strings.
feels like it adds a lot of complexity to the mental model required to understand the behavior of quoted strings. before diving into this one, I want to understand 2.

odelalleau: > 1. So the reason we are unescaping in top level strings is just to support escaping interpolations there?

Yes.

>    If so, would it simplify things to just say that we only apply the grammar inside interpolations and thus only support escaping of nested interpolations?

I may misunderstand what you're suggesting, but how would you set a config option that evaluates to the string with characters `Please use ${VAR}` if we can't escape interpolations in top-level strings?

> 2. I am a bit confused by this:
> 
> ```
> input:  r"'AB${test:\'CD${test:\\'EF\\'}GH\'}'"
> actual string: 'AB${test:\'CD${test:\\'EF\\'}GH\'}'
> ```
> 
> This is a string outside of interpolation, and it's quoted. why is the result not quoted as well?
> I am assuming that this string is interpreted as a value of a config node, so there is no yaml involved here.

Oh! Careful, there are several tests in the file, and this one belongs to the tests for the `singleElement` parser rule (not `configValue`), so this is parsed as a quoted string like you would use as a resolver argument.


> 3. I don't like how we need to unescape the nested quoted strings.
>    feels like it adds a lot of complexity to the mental model required to understand the behavior of quoted strings. before diving into this one, I want to understand 2.

Let me know how you feel now that hopefully point 2 is clearer.

omry: > I may misunderstand what you're suggesting, but how would you set a config option that evaluates to the string with characters Please use ${VAR} if we can't escape interpolations in top-level strings?

How would you do it in OmegaConf 2.0 (not supported at all as far as I can tell).

In principle, you could use decode (although I think it's not working right now for this)
```
a: ${oc.decode:'Please use \${VAR}'}
```

> Oh! Careful, there are several tests in the file, and this one belongs to the tests for the singleElement parser rule (not configValue), so this is parsed as a quoted string like you would use as a resolver argument.

This is one of the reasons I don't like the current testing approach: it makes it hard to understand what the test data is being used for.

> Let me know how you feel now that hopefully point 2 is clearer.

Okay, so back to nested quoted strings:
```
outer quoted string: 
'AB${test:\'CD${test:\\'EF\\'}GH\'}'
```
At some point we considered using lexer modes to support something like this:
```
'AB${test:'CD${test:'EF'}GH'}'
```
(Correct me if I am wrong and lexer modes will not actually enable this).
This is becoming much more attractive after the back and forth we had on this issue.
Can we pull this off?
With the exception of the time investment, do you see any downsides to suspect additional corner cases?

odelalleau: > > I may misunderstand what you're suggesting, but how would you set a config option that evaluates to the string with characters Please use ${VAR} if we can't escape interpolations in top-level strings?
> 
> How would you do it in OmegaConf 2.0 (not supported at all as far as I can tell).

Indeed, that was a gap in 2.0. Though you could use `${}` as long as it didn't look like an interpolation, for instance `"Please use ${VAR!}"` would work.

> In principle, you could use decode (although I think it's not working right now for this)
> 
> ```
> a: ${oc.decode:'Please use \${VAR}'}
> ```

The example you provide doesn't work because the string that is given as input to `oc.decode` has characters `Please use ${VAR}` so it crashes trying to resolve the interpolation (in `oc.decode`).

I'm actually surprised that it's allowed by the grammar because I didn't realize that you could have interpolations in unquoted strings (`oc.decode` uses the `singleElement` parser rule). So for instance you can write this:
```python
cfg = OmegaConf.create({"x": "foo", "y": "${identity:abc ${x}}"})
assert cfg.y == "abc foo"
```
I'm not sure this is even tested right now, I'll double check later.

This works with `oc.decode`:
```python
cfg = OmegaConf.create({"x": r"""${oc.decode:'"Please use \\\\\${VAR}"'}"""})
assert cfg.x == "Please use ${VAR}"
```

That being said, even if we decide that this is how to do it, it still relies on the escaping of interpolations at the top-level, because the parsing of a quoted string triggers a top-level parsing after the removal of quotes & un-escaping. So I'm not sure what we would gain.

> Okay, so back to nested quoted strings:
> 
> ```
> outer quoted string: 
> 'AB${test:\'CD${test:\\'EF\\'}GH\'}'
> ```
> 
> At some point we considered using lexer modes to support something like this:
> 
> ```
> 'AB${test:'CD${test:'EF'}GH'}'
> ```
> 
> (Correct me if I am wrong and lexer modes will not actually enable this).
> This is becoming much more attractive after the back and forth we had on this issue.
> Can we pull this off?
> With the exception of the time investment, do you see any downsides to suspect additional corner cases?

Almost -- you actually need to alternate quote styles to make it work, otherwise there is no way to differentiate (in the greedy way the lexer works) a closing quote vs. one that starts a new segment. So it should instead be:
```
 'AB${test:"CD${test:'EF'}GH"}'
```
I think it should work fine (it used to), just flagging that it means we would now parse the content of quoted strings when we parse an expression (not really a problem, just to make it clear), instead of triggering this parsing only on evaluation.
The main downside I see is the one you had raised initially: it may be an unusual approach that could surprise people.

omry: > > > I may misunderstand what you're suggesting, but how would you set a config option that evaluates to the string with characters Please use ${VAR} if we can't escape interpolations in top-level strings?
> > 
> > 
> > How would you do it in OmegaConf 2.0 (not supported at all as far as I can tell).
> 
> Indeed, that was a gap in 2.0. Though you could use `${}` as long as it didn't look like an interpolation, for instance `"Please use ${VAR!}"` would work.

This is one of many enhancements in 2.1, but it's not even a requested one so I would be okay with an alternative implementation if it's reasonably clean and simplify out lives.

> > In principle, you could use decode (although I think it's not working right now for this)
> > ```
> > a: ${oc.decode:'Please use \${VAR}'}
> > ```
> 
> The example you provide doesn't work because the string that is given as input to `oc.decode` has characters `Please use ${VAR}` so it crashes trying to resolve the interpolation (in `oc.decode`).

This is a feature of oc.decode:
1. I am not sure why it should resolve the interpolation in the first place.
2. the interpolation is escaped, so it should understand that and not try to resolve it (in principle, I am not talking about the current implementation).

> I'm actually surprised that it's allowed by the grammar because I didn't realize that you could have interpolations in unquoted strings (`oc.decode` uses the `singleElement` parser rule). So for instance you can write this:

I used a quoted string, so I am not sure why you were surprised.

> ```python
> cfg = OmegaConf.create({"x": "foo", "y": "${identity:abc ${x}}"})
> assert cfg.y == "abc foo"
> ```
> 
> I'm not sure this is even tested right now, I'll double check later.

I have done it before, I usually quote just for clarity in scenarios where the string contains spaces.
Let me know if you find why this works :).

> This works with `oc.decode`:
> 
> ```python
> cfg = OmegaConf.create({"x": r"""${oc.decode:'"Please use \\\\\${VAR}"'}"""})
> assert cfg.x == "Please use ${VAR}"
> ```
> 
> That being said, even if we decide that this is how to do it, it still relies on the escaping of interpolations at the top-level, because the parsing of a quoted string triggers a top-level parsing after the removal of quotes & un-escaping. So I'm not sure what we would gain.

Obviously this is not great. but if we can make `a: ${oc.decode:'Please use \${VAR}'}` work I would vote to simplify top-level string and maybe even drop it completely.

> > Okay, so back to nested quoted strings:
> > ```
> > outer quoted string: 
> > 'AB${test:\'CD${test:\\'EF\\'}GH\'}'
> > ```
> > 
> > 
> > At some point we considered using lexer modes to support something like this:
> > ```
> > 'AB${test:'CD${test:'EF'}GH'}'
> > ```
> > 
> > 
> > (Correct me if I am wrong and lexer modes will not actually enable this).
> > This is becoming much more attractive after the back and forth we had on this issue.
> > Can we pull this off?
> > With the exception of the time investment, do you see any downsides to suspect additional corner cases?
> 
> Almost -- you actually need to alternate quote styles to make it work, otherwise there is no way to differentiate (in the greedy way the lexer works) a closing quote vs. one that starts a new segment. So it should instead be:
> 
> ```
>  'AB${test:"CD${test:'EF'}GH"}'
> ```

I am not sure there is no way to differentiate:
Once we are in a quoted string (say single), hitting another single quote closes it.
However, if we hit a double quote we are good, but if we hit an interpolation token (IE, non-greedy), 
we can push a mode that allows us to start a new single quoted string.
when we seen an interpolation end, we pop (By the way, this might make it impossible for regex to handle with the grammar).

> I think it should work fine (it used to), just flagging that it means we would now parse the content of quoted strings when we parse an expression (not really a problem, just to make it clear), instead of triggering this parsing only on evaluation.
> The main downside I see is the one you had raised initially: it may be an unusual approach that could surprise people.


After seeing what where the alternative path leads, I am much more open to a bit of weirdness in exchange for something that is much easier to grok.

odelalleau: > > > In principle, you could use decode (although I think it's not working right now for this)
> > > ```
> > > a: ${oc.decode:'Please use \${VAR}'}
> > > ```
> > 
> > 
> > The example you provide doesn't work because the string that is given as input to `oc.decode` has characters `Please use ${VAR}` so it crashes trying to resolve the interpolation (in `oc.decode`).
> 
> This is a feature of oc.decode:
> 
> 1. I am not sure why it should resolve the interpolation in the first place.
> 2. the interpolation is escaped, so it should understand that and not try to resolve it (in principle, I am not talking about the current implementation).

1. `oc.decode` currently relies on the same parsing (& visitor) code as the one used to parse (& resolve) the `singleElement` rule of the grammar, which resolves interpolations.

2. But `oc.decode` doesn't see the escaped interpolation: it sees the string with characters `Please use ${VAR}` where the interpolation isn't escaped. The reason is that the quoted string `'Please use \${VAR}'` is first resolved into `Please use \${VAR}` *then* passed to `oc.decode`. So what you're trying to achieve would be done in a simpler way with an `identity` resolver that would return this string unchanged (instead of `oc.decode` which parses it according to the grammar rules).


> > I'm actually surprised that it's allowed by the grammar because I didn't realize that you could have interpolations in unquoted strings (`oc.decode` uses the `singleElement` parser rule). So for instance you can write this:
> 
> I used a quoted string, so I am not sure why you were surprised.

Hopefully the above comment makes it clearer: `oc.decode` sees the string with characters `Please use ${VAR}` which is unquoted.

> > ```python
> > cfg = OmegaConf.create({"x": "foo", "y": "${identity:abc ${x}}"})
> > assert cfg.y == "abc foo"
> > ```
> > 
> > 
> > I'm not sure this is even tested right now, I'll double check later.
> 
> I have done it before, I usually quote just for clarity in scenarios where the string contains spaces.
> Let me know if you find why this works :)

Oh, a quick look at the grammar makes it clear why it works. I was just surprised because for some reason I thought it wouldn't. I just want to make sure it's properly tested.

> > This works with `oc.decode`:
> > ```python
> > cfg = OmegaConf.create({"x": r"""${oc.decode:'"Please use \\\\\${VAR}"'}"""})
> > assert cfg.x == "Please use ${VAR}"
> > ```
> > 
> > 
> > That being said, even if we decide that this is how to do it, it still relies on the escaping of interpolations at the top-level, because the parsing of a quoted string triggers a top-level parsing after the removal of quotes & un-escaping. So I'm not sure what we would gain.
> 
> Obviously this is not great. but if we can make `a: ${oc.decode:'Please use \${VAR}'}` work I would vote to simplify top-level string and maybe even drop it completely.

I don't think we can make this one work (for the reason explained above), but if we remove the un-escaping of `\\` from quoted strings (as in #621) then we can probably shave off a few `\` from my example.
I'm not really sure how we would "simplify top-level string and maybe even drop it completely", since the parsing of quoted strings relies on it as well.

> > > At some point we considered using lexer modes to support something like this:
> > > ```
> > > 'AB${test:'CD${test:'EF'}GH'}'
> > > ```
> > > 
> > > 
> > > Can we pull this off?
> > 
> > Almost -- you actually need to alternate quote styles to make it work, otherwise there is no way to differentiate (in the greedy way the lexer works) a closing quote vs. one that starts a new segment. So it should instead be:
> > ```
> >  'AB${test:"CD${test:'EF'}GH"}'
> > ```
> 
> I am not sure there is no way to differentiate:
> Once we are in a quoted string (say single), hitting another single quote closes it.
> However, if we hit a double quote we are good, but if we hit an interpolation token (IE, non-greedy),
> we can push a mode that allows us to start a new single quoted string.
> when we seen an interpolation end, we pop (By the way, this might make it impossible for regex to handle with the grammar).

Oh yes you are right, thanks! I'm not sure where this idea of alternating came from -- for some reason I thought I remembered you had to alternate, but I think I just imagined it.

> After seeing what where the alternative path leads, I am much more open to a bit of weirdness in exchange for something that is much easier to grok.

Alright, I'm happy to do it, but then we also need to decide if we still want to support escaped quotes in quoted strings. I *think* we don't need them anymore because we should always be able to use a different enclosing quote style instead, for instance `"there is a single quote ' in the middle"`. I haven't thought in depth about it though.

omry: > 2. But `oc.decode` doesn't see the escaped interpolation: it sees the string with characters `Please use ${VAR}` where the interpolation isn't escaped. The reason is that the quoted string `'Please use \${VAR}'` is first resolved into `Please use \${VAR}` _then_ passed to `oc.decode`. So what you're trying to achieve would be done in a simpler way with an `identity` resolver that would return this string unchanged (instead of `oc.decode` which parses it according to the grammar rules).

The line of reasoning here seems wrong or missing a step.
But in any case, we shouldn't force decode to do what its' not meant to do.
IF identity works, we can introduce it under some name that makes sense. The use case for not resolving interpolations is very rare and I don't mind if it takes a bit of extra work.

> > Obviously this is not great. but if we can make `a: ${oc.decode:'Please use \${VAR}'}` work I would vote to simplify top-level string and maybe even drop it completely.
> 
> I don't think we can make this one work (for the reason explained above), but if we remove the un-escaping of `\\` from quoted strings (as in #621) then we can probably shave off a few `\` from my example.
> I'm not really sure how we would "simplify top-level string and maybe even drop it completely", since the parsing of quoted strings relies on it as well.

Can you explain what you mean by "since the parsing of quoted strings relies on it as well"?
As far as I understand, quoted strings are only special inside interpolations.
Outside interpolations we can have whatever we want, including quotes.

> > > > At some point we considered using lexer modes to support something like this:
> > > > ```
> > > > 'AB${test:'CD${test:'EF'}GH'}'
> > > > ```
> > > > Can we pull this off?
> > > 
> > > Almost -- you actually need to alternate quote styles to make it work, otherwise there is no way to differentiate (in the greedy way the lexer works) a closing quote vs. one that starts a new segment. So it should instead be:
> > > ```
> > >  'AB${test:"CD${test:'EF'}GH"}'
> > > ```
> > 
> > 
> > I am not sure there is no way to differentiate:
> > Once we are in a quoted string (say single), hitting another single quote closes it.
> > However, if we hit a double quote we are good, but if we hit an interpolation token (IE, non-greedy),
> > we can push a mode that allows us to start a new single quoted string.
> > when we seen an interpolation end, we pop (By the way, this might make it impossible for regex to handle with the grammar).
> 
> Oh yes you are right, thanks! I'm not sure where this idea of alternating came from -- for some reason I thought I remembered you had to alternate, but I think I just imagined it.

Maybe alternating was something we considered before we found out about lexer modes.

> > After seeing what where the alternative path leads, I am much more open to a bit of weirdness in exchange for something that is much easier to grok.
> 
> Alright, I'm happy to do it,

Great, so just to make sure I understand:
In the case of something like:
`${foo: 'outer ${bar: 'inner' }' }`

To be clear, this is the inner interpolation: `${bar: 'inner' }`
It lives in the quotes string `'outer ${bar: 'inner' }'`,
which is a parameter of the resolver `${foo: 'outer ${bar: 'inner' }' }`

What would the tokens be be?
Does it mean that we are promoting quoted string from a token to a parser rule?

> but then we also need to decide if we still want to support escaped quotes in quoted strings. I _think_ we don't need them anymore because we should always be able to use a different enclosing quote style instead, for instance `"there is a single quote ' in the middle"`. I haven't thought in depth about it though.

I think we still need those: how would you create a quoted string that contains both quotes?
```
single: ', double: "
```

omry: In any case, before you start working on it, what do you think about merging what we have now? It will be easier starting from a clean slate.

odelalleau: Just to follow-up on:

> Oh, a quick look at the grammar makes it clear why it works. I was just surprised because for some reason I thought it wouldn't. I just want to make sure it's properly tested.

==> It's indeed properly tested (there is a section "Interpolations in unquoted strings"), so no worries there (except for my failing memory).

> > 1. But `oc.decode` doesn't see the escaped interpolation: it sees the string with characters `Please use ${VAR}` where the interpolation isn't escaped. The reason is that the quoted string `'Please use \${VAR}'` is first resolved into `Please use \${VAR}` _then_ passed to `oc.decode`. So what you're trying to achieve would be done in a simpler way with an `identity` resolver that would return this string unchanged (instead of `oc.decode` which parses it according to the grammar rules).
> 
> The line of reasoning here seems wrong or missing a step.

It may become clearer with a simpler example:
```python
cfg = OmegaConf.create({"x": "foo", "y": "${identity:'x is ${x}'}"})
cfg.y  # 'x is foo'
```
What happens is that in the visitor, when we are to resolve `${identity:'x is ${x}'}` in `visitInterpolationResolver()`, we first obtain the name of the resolver (`identity`) then obtain the list of arguments to the resolver from `visitSequence()`, which calls `visitElement()` on each argument. The quoted value `'x is ${x}'` is a primitive and its resolution triggers a new parsing pass over its content (`x is ${x}`) where the interpolation is resolved, resulting in the string `x is foo`. *Then* this string is passed to the resolver function associated to `identity`, which returns it unchanged.

The main point here is that the quoted string is evaluted *before* it is passed to the resolver. So, now, if you escape the interpolation:
```python
cfg = OmegaConf.create({"x": "foo", "y": r"${identity:'x is \${x}'}"})
cfg.y  # 'x is ${x}'
```
==> the exact same steps are applied, except that the resolution of the quoted value `'x is \${x}'` results in the string `x is ${x}` which is passed as input to the resolver function associated to `identity`.

Hopefully this makes it clearer why `oc.deode` sees the string `Please use ${VAR}` as input in our previous example.

> > > Obviously this is not great. but if we can make `a: ${oc.decode:'Please use \${VAR}'}` work I would vote to simplify top-level string and maybe even drop it completely.
> > 
> > 
> > I don't think we can make this one work (for the reason explained above), but if we remove the un-escaping of `\\` from quoted strings (as in #621) then we can probably shave off a few `\` from my example.
> > I'm not really sure how we would "simplify top-level string and maybe even drop it completely", since the parsing of quoted strings relies on it as well.
> 
> Can you explain what you mean by "since the parsing of quoted strings relies on it as well"?
> As far as I understand, quoted strings are only special inside interpolations.
> Outside interpolations we can have whatever we want, including quotes.

What I mean is that the resolution of a quoted string currently works as follows:
1. Remove the enclosing quotes (let's say these are double quotes `"`)
2. Un-escape `\\` and `\"` (*not* escaped interpolations)
3. Trigger a new parsing + resolution of the resulting string with the `configValue` rule (which relies on top-level strings)

This will change with the changes discussed below though, but maybe we can discuss it in another PR related to these changes.

> Maybe alternating was something we considered before we found out about lexer modes.


> Great, so just to make sure I understand:
> In the case of something like:
> `${foo: 'outer ${bar: 'inner' }' }`
> 
> To be clear, this is the inner interpolation: `${bar: 'inner' }`
> It lives in the quotes string `'outer ${bar: 'inner' }'`,
> which is a parameter of the resolver `${foo: 'outer ${bar: 'inner' }' }`
> 
> What would the tokens be be?
> Does it mean that we are promoting quoted string from a token to a parser rule?

Yes.
I have a backup of my old branch which should make things clear:
Lexer: https://github.com/odelalleau/omegaconf/blob/backup_2020_08_09/omegaconf/grammar/InterpolationLexer.g4
Parser: https://github.com/odelalleau/omegaconf/blob/backup_2020_08_09/omegaconf/grammar/InterpolationParser.g4

> > but then we also need to decide if we still want to support escaped quotes in quoted strings. I _think_ we don't need them anymore because we should always be able to use a different enclosing quote style instead, for instance `"there is a single quote ' in the middle"`. I haven't thought in depth about it though.
> 
> I think we still need those: how would you create a quoted string that contains both quotes?

Yes indeed, I thought of the same example afterwards. Although I'm fine with keeping escaped quotes, just note that since it's a pretty rare situation we could also instead use resolvers (and we could do the same for escaped interpolations). Something like:
```python
"single ' quote, double ${oc.esc.quote:double} quote, escaped ${oc.esc.interpolation:open} interpolation }"
```
which would be resolved into `single ' quote, double " quote, escaped ${ interpolation }`.
(just throwing it out there as an idea, I don't feel strongly either way)

> In any case, before you start working on it, what do you think about merging what we have now? It will be easier starting from a clean slate.

Agreed.

omry: > What I mean is that the resolution of a quoted string currently works as follows:
> 
> 1. Remove the enclosing quotes (let's say these are double quotes `"`)
> 2. Un-escape `\\` and `\"` (_not_ escaped interpolations)
> 3. Trigger a new parsing + resolution of the resulting string with the `configValue` rule (which relies on top-level strings)
> 
> This will change with the changes discussed below though, but maybe we can discuss it in another PR related to these changes.

Yes, it will be easier to discuss in context, and we are considering implementation changes to this anyway.

> > Does it mean that we are promoting quoted string from a token to a parser rule?
> 
> Yes.
> I have a backup of my old branch which should make things clear:
> Lexer: https://github.com/odelalleau/omegaconf/blob/backup_2020_08_09/omegaconf/grammar/InterpolationLexer.g4
> Parser: https://github.com/odelalleau/omegaconf/blob/backup_2020_08_09/omegaconf/grammar/InterpolationParser.g4

Great.

> > > but then we also need to decide if we still want to support escaped quotes in quoted strings. I _think_ we don't need them anymore because we should always be able to use a different enclosing quote style instead, for instance `"there is a single quote ' in the middle"`. I haven't thought in depth about it though.
> > I think we still need those: how would you create a quoted string that contains both quotes?
> Yes indeed, I thought of the same example afterwards. Although I'm fine with keeping escaped quotes, just note that since it's a pretty rare situation we could also instead use resolvers (and we could do the same for escaped interpolations). Something like:
> 
> ```python
> "single ' quote, double ${oc.esc.quote:double} quote, escaped ${oc.esc.interpolation:open} interpolation }"
> ```

I agree that it's a rare case. but the standard approach for it (in most grammars) is to escape and it might be worth supporting directly (people would intuitively try to escape a quote like that).
My gut feeling is that if we switch to the proposed approach, the rare cases where we need to escape will be much easier to understand, so they would no longer be a concern.

> > In any case, before you start working on it, what do you think about merging what we have now? It will be easier starting from a clean slate.
> 
> Agreed.

Will make a final pass soon, not sure I will have time today (Friday is when I have the most meetings).

omry (approved):
You know what, I think this is good to go.
Go ahead.

Issue #426, opened 2020-10-28
Deprecation of `register_resolver()` to introduce new functionalities


odelalleau:
# Migrating custom resolvers from OmegaConf 2.0 to 2.1

## Summary

1. Replace `register_resolver` with `register_new_resolver`
2. Update resolvers to handle non-string arguments (if applicable)
3. Only for resolvers whose output must be cached: ensure `use_cache=True` is passed to `register_new_resolver`

## Detailed migration instructions

Here are the main changes between resolvers registered with `register_new_resolver()` vs the old `register_resolver()`:

* They may take as input data of types supported by the new grammar (e.g. int, float, bool, dict, list) and possibly computed from other interpolations, e.g.: `${func: [3.14, ${oc.env:VAR}], true}`
* Their cache is disabled by default (and can be enabled by passing `use_cache=True` to `register_new_resolver()`)
* They may access the parent and the root config nodes (by declaring `_parent_` and `_root_` to the callable respectively) (#599).

In most cases, migrating an existing resolver from OmegaConf 2.0 to 2.1 means replacing `register_resolver()` with `register_new_resolver()` and ensuring the resolver is correctly handling non-string inputs, if applicable.

As an example, imagine you had the following resolver in OmegaConf 2.0:

```python
def add_ints(*args: str) -> int:
    return sum(int(x) for x in args)  # in 2.0 all arguments are strings and must be manually converted

OmegaConf.register_resolver("add_ints", add_ints)
cfg = OmegaConf.create("sum: ${add_ints:1,2,3}")
assert cfg.sum == 6
```

In this example arguments will now expected to be integers in OmegaConf 2.1:

```python
def add_ints(*args: int) -> int:
    assert all(isinstance(x, int) for x in args)  # if we want to prevent adding non integer values
    return sum(args)

OmegaConf.register_new_resolver("add_ints", add_ints)
cfg = OmegaConf.create("sum: ${add_ints:1,2,3}")
assert cfg.sum == 6
```

Note that here we didn't register the resolver with `use_cache=True`: it is expected that most resolvers should not need to use the cache. If you are depending on the caching behavior by assuming that your resolver is always returning the same value even when the underlying function may return a different value (e.g. `random()`, `time()`), then register your resolver with `use_cache=True`.

The new resolver behavior is documented in detail in [the documentation](https://omegaconf.readthedocs.io/en/latest/custom_resolvers.html).

__Additional remarks:__
[...]

PR #635, opened 2021-03-22
Improvements to the `register_new_resolver()` API

### LINKED ISSUE #370: Improve the API of register_resolver()

odelalleau:
Fixes #370

Note: there is no additional news item because there is a already an item referring to #426, which also includes a pointer to #370 (at the end). Since these changes are pretty small I'm not sure it's worth making them more prominent in the news, so @omry let me know how you feel about that.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -469,6 +469,19 @@ The example below creates a resolver that adds 10 to the given value.
     >>> c.key
     1000
 
+Attempting to register the same resolver twice will raise a ``ValueError`` unless using ``replace=True``:
+
+.. doctest::
+
+    >>> OmegaConf.register_new_resolver("plus_10", lambda x: x + 10.0)
+    Traceback (most recent call last):
+    ...
+    ValueError: resolver 'plus_10' is already registered
+    >>> OmegaConf.register_new_resolver(
+    ...     "plus_10",
+    ...     lambda x: x + 10.0,
+    ...     replace=True,  # OK: replace previous resolver
+    ... )
 
 Custom resolvers support variadic argument lists in the form of a comma separated list of zero or more values.
 Whitespaces are stripped from both ends of each value ("foo,bar" is the same as "foo, bar ").
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -459,7 +459,20 @@ This can be useful for instance to parse environment variables:
 Custom interpolations
 ^^^^^^^^^^^^^^^^^^^^^
 
-You can add additional interpolation types using custom resolvers.
+You can add additional interpolation types by registering custom resolvers with ``OmegaConf.register_new_resolver()``:
+
+.. code-block:: python
+
+    def register_new_resolver(
+        name: str,
+        resolver: Resolver,
+        *,
+        replace: bool = False,
+        use_cache: bool = False,
+    ) -> None
+
+Attempting to register the same resolver twice will raise a ``ValueError`` unless using ``replace=True``.
+
 The example below creates a resolver that adds 10 to the given value.
 
 .. doctest::
@@ -468,20 +481,6 @@ The example below creates a resolver that adds 10 to the given value.
     >>> c = OmegaConf.create({'key': '${plus_10:990}'})
     >>> c.key
     1000
-
-Attempting to register the same resolver twice will raise a ``ValueError`` unless using ``replace=True``:
-
-.. doctest::
-
-    >>> OmegaConf.register_new_resolver("plus_10", lambda x: x + 10.0)
-    Traceback (most recent call last):
-    ...
-    ValueError: resolver 'plus_10' is already registered
-    >>> OmegaConf.register_new_resolver(
-    ...     "plus_10",
-    ...     lambda x: x + 10.0,
-    ...     replace=True,  # OK: replace previous resolver
-    ... )
 
 Custom resolvers support variadic argument lists in the form of a comma separated list of zero or more values.
 Whitespaces are stripped from both ends of each value ("foo,bar" is the same as "foo, bar ").
--- a/omegaconf/omegaconf.py
+++ b/omegaconf/omegaconf.py
@@ -435,16 +435,17 @@ class OmegaConf:
 
         existing = OmegaConf._get_resolver(name)
         if existing is not None:
-            if (
-                # DEPRECATED: remove the very first check in 2.2
-                # (it is only there because some resolvers may be registered with
-                # the legacy function and thus would not be `partial` objects)
-                # Also the typing may be updated accordingly.
-                hasattr(existing, "keywords")
-                and existing.keywords["resolver"] is resolver  # type: ignore
-                and existing.keywords["use_cache"] == use_cache  # type: ignore
-            ):
-                return  # already registered with the same settings => ignore
+            # DEPRECATED: remove this check in 2.2
+            # (it is only there because some resolvers may be registered with
+            # the legacy function and thus would not be `partial` objects)
+            # Also the typing may be updated accordingly.
+            if hasattr(existing, "keywords"):
+                kw = existing.keywords  # type: ignore
+                if (
+                    kw["resolver"] is resolver  # lgtm [py/comparison-using-is]
+                    and kw["use_cache"] == use_cache
+                ):
+                    return  # already registered with the same settings => ignore
             raise ValueError(f"resolver '{name}' is already registered")
 
         try:
--- a/omegaconf/omegaconf.py
+++ b/omegaconf/omegaconf.py
@@ -411,16 +411,19 @@ class OmegaConf:
         name: str,
         resolver: Resolver,
         *,
+        replace: bool = False,
         use_cache: bool = False,
     ) -> None:
         """
         Register a resolver.
 
-        :param name: Name of the resolver. If this name has already been registered,
-            then a `ValueError` is raised.
+        :param name: Name of the resolver.
         :param resolver: Callable whose arguments are provided in the interpolation,
             e.g., with ${foo:x,0,${y.z}} these arguments are respectively "x" (str),
             0 (int) and the value of `y.z`.
+        :param replace: If set to `False` (default), then a `ValueError` is raised if
+            an existing resolver has already been registered with the same name.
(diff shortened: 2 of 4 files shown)


omegaconf/omegaconf.py
```diff
@@ -425,11 +428,25 @@ def register_new_resolver(
...
+            raise TypeError("resolver must be callable")
+        if not name:
+            raise ValueError("cannot use an empty resolver name")
+
+        existing = OmegaConf._get_resolver(name)
+        if existing is not None:
+            # DEPRECATED: remove this check in 2.2
+            # (it is only there because some resolvers may be registered with
+            # the legacy function and thus would not be `partial` objects)
+            # Also the typing may be updated accordingly.
```
omry, line 441: an alternative here is to change the legacy resolver to use a trivial wrapper as well.

omegaconf/omegaconf.py
```diff
@@ -444,34 +461,14 @@ def _should_pass(special: str) -> bool:
...
-                kwargs["_parent_"] = parent
-            if pass_root:
-                kwargs["_root_"] = config
-
-            ret = resolver(*args, **kwargs)
-
-            if use_cache:
-                cache[args_str] = ret
-            return ret
+        resolver_wrapper = partial(
```
omry, line 464: Can you explain what using partial here buys?

omegaconf/omegaconf.py
```diff
@@ -1037,6 +1030,38 @@ def _maybe_wrap(
...
+    config: BaseContainer,
+    parent: Container,
+    args: Tuple[Any, ...],
+    args_str: Tuple[str, ...],
+    name: str,
+    resolver: Resolver,
+    use_cache: bool,
+    pass_parent: bool,
+    pass_root: bool,
+) -> Any:
```
omry, line 1043: The inlined version was significantly simpler because it inherits the context.
it's short enough that I prefer to keep it as an inline method.

What is the motivation for pulling it out of the register_new_resolver?

omegaconf/omegaconf.py
```diff
@@ -411,12 +412,14 @@ def register_new_resolver(
...
-        use_cache: Optional[bool] = False,
+        use_cache: bool = False,
     ) -> None:
         """
         Register a resolver.
 
-        :param name: Name of the resolver.
+        :param name: Name of the resolver. If this name has already been registered,
+            then a `ValueError` is raised (unless `resolver` is the same callable object
+            and `use_cache` has the same value as previously registered).
```
omry, line 422: I think we should simplify here and not add the nice shortcut.
We will probably make caching more complicated in the future, which will also make this handling more complicated (by adding additional keyword to register_resolver).

Le'ts just raise an exception if a resolver with the same name is used even if it's registering the same callable.

omry (changes requested):
I didn't think about use_cache complicating the idea of allowing registering the same resolver twice.

I suggest we drop that idea, see my comment.

omegaconf/omegaconf.py
odelalleau: Do you feel strongly about it? If not I suggest we leave it as is, I'll take care of the clean-up for 2.2.

omegaconf/omegaconf.py
odelalleau: In order to check if the resolver was already registered to the same object, I need to access that object. There doesn't seem to be a "clean" way to achieve this with an inline function (maybe by messing with the closure, but I didn't dig too far as it looked like it would end up relying on CPython implementation details). On the other hand, a `partial` object offers a public API to access its data.

Anyway, since you suggested to drop the idea of re-registering the same function, I will revert this change.

omegaconf/omegaconf.py
odelalleau: Explained in https://github.com/omry/omegaconf/pull/635#discussion_r599854362 => resolving

omegaconf/omegaconf.py
odelalleau: This has been reverted, new version is in 3554985

omegaconf/omegaconf.py
odelalleau: This has been reverted, new version is in 3554985

omegaconf/omegaconf.py
odelalleau: Done by reverting previous changes, new version is in 3554985

tests/interpolation/test_custom_resolvers.py
```diff
@@ -8,22 +8,57 @@
...
     def foo(_: Any) -> int:
         return 10
 
     OmegaConf.register_new_resolver("foo", foo)
-    with raises(AssertionError):
+    with raises(ValueError, match=re.escape("resolver 'foo' is already registered")):
         OmegaConf.register_new_resolver("foo", lambda _: 10)
 
 
+@mark.parametrize("first_use_cache", [True, False])
```
omry, line 20: since the behavior of resiter_new_resolver is much simpler now, I think this test is no longer needed.

tests/interpolation/test_custom_resolvers.py
```diff
@@ -8,22 +8,57 @@
...
 
-def test_register_resolver_twice_error(restore_resolvers: Any) -> None:
+def test_register_resolver_twice_error_new_callable(restore_resolvers: Any) -> None:
     def foo(_: Any) -> int:
         return 10
 
     OmegaConf.register_new_resolver("foo", foo)
-    with raises(AssertionError):
+    with raises(ValueError, match=re.escape("resolver 'foo' is already registered")):
         OmegaConf.register_new_resolver("foo", lambda _: 10)
```
omry, line 17: Since we no longer care if it's the same callable or not, I think we can just have a single test like:
```
test_register_resolver_twice_error
```

omry (changes requested):
looking good. some minor requests.

tests/interpolation/test_custom_resolvers.py
odelalleau: Ok, simplified in e5a490b

tests/interpolation/test_custom_resolvers.py
odelalleau: Also done in e5a490b

omegaconf/omegaconf.py
```diff
@@ -420,16 +421,21 @@ def register_new_resolver(
         :param resolver: Callable whose arguments are provided in the interpolation,
             e.g., with ${foo:x,0,${y.z}} these arguments are respectively "x" (str),
             0 (int) and the value of `y.z`.
+        :param replace: If set to `False` (default), then a `ValueError` is raised if
+            an existing resolver has already been registered with the same name.
+            If set to `True`, then the new resolver replaces the previous one.
```
omry, line 426:
```suggestion
        :param replace: If set to `False` (default), then a `ValueError` is raised if
            an existing resolver has already been registered with the same name.
            If set to `True`, then the new resolver replaces the previous one.
            NOTE: The cache on existing config objects is not affected, use OmegaConf.clear_cache(cfg) to clear it.
```

tests/interpolation/test_custom_resolvers.py
```diff
@@ -38,6 +59,30 @@ def signature_not_inspectable(*args: Any, **kw: Any) -> None:
     assert OmegaConf.create({"x": "${not_inspectable:}"}).x == 123
 
 
+@mark.parametrize(
+    ("use_cache_1", "use_cache_2", "expected"),
+    [
+        (False, False, 2),
+        (False, True, 2),
+        (True, False, 2),
+        (True, True, 1),  # potentially controversial, see #637
```
omry, line 68: Commented. I think this is okay and you can change the comment to indicate that we are getting the value from the cache.
Alternatively you can use a new config object or clear the cache on the existing one.

omry (approved):

omry (changes requested):
Actually - we need to update the docs a bit right?

omegaconf/omegaconf.py
odelalleau: Added in edf97bc

tests/interpolation/test_custom_resolvers.py
odelalleau: Done in cbe3a8e (I prefer to keep the test in case this would change in the future)

odelalleau:
> Actually - we need to update the docs a bit right?

Can't hurt :) Done in c953789

docs/source/usage.rst
```diff
@@ -469,6 +469,19 @@ The example below creates a resolver that adds 10 to the given value.
...
+
+    >>> OmegaConf.register_new_resolver("plus_10", lambda x: x + 10.0)
+    Traceback (most recent call last):
+    ...
+    ValueError: resolver 'plus_10' is already registered
+    >>> OmegaConf.register_new_resolver(
+    ...     "plus_10",
+    ...     lambda x: x + 10.0,
+    ...     replace=True,  # OK: replace previous resolver
+    ... )
```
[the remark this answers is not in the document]
omry: I think we also want to include the signature of the methods we are documenting.
This is what I am doing for resolve:
```rst
OmegaConf.resolve
^^^^^^^^^^^^^^^^^

Resolves all interpolations in the given config object in-place.

.. code-block:: python

    def resolve(cfg: Any) -> Container:
        """
        Resolves all interpolations in the given config object in-place.
        Input config is converted to a DictConfig as needed.
        """

Example:

.. doctest::

    >>> conf = OmegaConf.create({"a": 10, "b": "${a}"})
    >>> OmegaConf.resolve(conf)
    >>> show(container)
    type: dict, value: {'a': 10, 'b': 10}
```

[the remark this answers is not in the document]
odelalleau: What about e879b00?

omry (approved):
Looking good.

PR #639, opened 2021-03-23
Added OmegaConf.resolve()

### LINKED ISSUE #640: OmegaConf.resolve(cfg) : Resolves all interpolations in the given config object in-place
Signature:
```python
    @staticmethod
    def resolve(cfg: Container) -> None:
        """
        Resolves all interpolations in the given config object in-place.
        :param cfg: An OmegaConf container (DictConfig, ListConfig)
                    Raises a ValueError if the input object is not an OmegaConf container.
        """
```

omry:
Introduces OmegaConf.resolve()

Closes #640

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -784,7 +784,7 @@ OmegaConf.resolve
                     Raises a ValueError if the input object is not an OmegaConf container.
         """
 
-Normally interpolations are resolved on access time (lazily). 
+Normally interpolations are resolved lazily, at access time. 
 This function eagerly resolves all interpolations in the given config object in-place.
 Example:
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -773,6 +773,29 @@ as DictConfig, allowing attribute style access on the resulting node.
     >>> show(container["structured_config"])
     type: DictConfig, value: {'port': 80, 'host': 'localhost'}
 
+OmegaConf.resolve
+^^^^^^^^^^^^^^^^^
+
+Normally interpolations are resolved on access time (lazily). 
+This function eagerly resolves all interpolations in the given config object in-place.
+
+.. code-block:: python
+
+    def resolve(cfg: Any) -> Container:
+        """
+        Resolves all interpolations in the given config object in-place.
+        Input config is converted to a DictConfig as needed.
+        """
+
+Example:
+
+.. doctest::
+
+    >>> conf = OmegaConf.create({"a": 10, "b": "${a}"})
+    >>> OmegaConf.resolve(conf)
+    >>> show(conf)
+    type: DictConfig, value: {'a': 10, 'b': 10}
+
 OmegaConf.select
 ^^^^^^^^^^^^^^^^
 OmegaConf.select() allow you to select a config node or value using a dot-notation key.
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -775,25 +775,27 @@ as DictConfig, allowing attribute style access on the resulting node.
 
 OmegaConf.resolve
 ^^^^^^^^^^^^^^^^^
+.. code-block:: python
+
+    def resolve(cfg: Container) -> None:
+        """
+        Resolves all interpolations in the given config object in-place.
+        :param cfg: An OmegaConf container (DictConfig, ListConfig)
+                    Raises a ValueError if the input object is not an OmegaConf container.
+        """
 
 Normally interpolations are resolved on access time (lazily). 
 This function eagerly resolves all interpolations in the given config object in-place.
-
-.. code-block:: python
-
-    def resolve(cfg: Any) -> Container:
-        """
-        Resolves all interpolations in the given config object in-place.
-        Input config is converted to a DictConfig as needed.
-        """
-
 Example:
 
 .. doctest::
 
-    >>> conf = OmegaConf.create({"a": 10, "b": "${a}"})
-    >>> OmegaConf.resolve(conf)
-    >>> show(conf)
+    >>> cfg = OmegaConf.create({"a": 10, "b": "${a}"})
+    >>> show(cfg)
+    type: DictConfig, value: {'a': 10, 'b': '${a}'}
+    >>> assert cfg.a == cfg.b == 10 # lazily resolving interpolation
+    >>> OmegaConf.resolve(cfg)
+    >>> show(cfg)
     type: DictConfig, value: {'a': 10, 'b': 10}
 
 OmegaConf.select
--- a/omegaconf/_impl.py
+++ b/omegaconf/_impl.py
@@ -7,18 +7,18 @@ from omegaconf.errors import InterpolationToMissingValueError
 def _resolve(cfg: Any) -> Any:
     if isinstance(cfg, DictConfig):
         for k in cfg.keys():
-            node = cfg._get_node(k)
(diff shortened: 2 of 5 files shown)


docs/source/usage.rst
```diff
@@ -773,6 +773,31 @@ as DictConfig, allowing attribute style access on the resulting node.
...
+.. code-block:: python
+
+    def resolve(cfg: Container) -> None:
+        """
+        Resolves all interpolations in the given config object in-place.
+        :param cfg: An OmegaConf container (DictConfig, ListConfig)
+                    Raises a ValueError if the input object is not an OmegaConf container.
+        """
+
+Normally interpolations are resolved on access time (lazily).
```
odelalleau, line 787: I would say either "on access" or "at access time", "on access time" sounds a bit weird -- to me at least

omegaconf/_impl.py
```diff
@@ -0,0 +1,24 @@
...
+from omegaconf.errors import InterpolationToMissingValueError
+
+
+def _resolve(cfg: Any) -> Any:
+    if isinstance(cfg, DictConfig):
+        for k in cfg.keys():
+            node = cfg._get_node(k)
+            cfg[k] = _resolve(node)
+
+    if isinstance(cfg, ListConfig):
```
odelalleau, line 12:
```suggestion
    elif isinstance(cfg, ListConfig):
```

omegaconf/_impl.py
```diff
@@ -0,0 +1,24 @@
+from typing import Any
+
+from omegaconf import MISSING, DictConfig, ListConfig, ValueNode
+from omegaconf.errors import InterpolationToMissingValueError
+
+
+def _resolve(cfg: Any) -> Any:
+    if isinstance(cfg, DictConfig):
+        for k in cfg.keys():
```
odelalleau, line 9: What if `cfg._is_inteprolation()`? (same for lists)

omegaconf/_impl.py
```diff
@@ -0,0 +1,24 @@
...
+            cfg[k] = _resolve(node)
+
+    if isinstance(cfg, ListConfig):
+        for i in range(len(cfg)):
+            node = cfg._get_node(i)
+            cfg[i] = _resolve(node)
+
+    elif isinstance(cfg, ValueNode):
+        try:
+            cfg = cfg._dereference_node()
```
odelalleau: I feel like this won't handle well situations where after dereferencing, `cfg` is a container that may contain interpolations inside it. Am I wrong?

tests/test_omegaconf.py
```diff
@@ -469,3 +469,40 @@ def test_is_issubclass() -> None:
...
+    [
+        # dict
+        param({}, {}, id="dict"),
+        param({"a": 10, "b": "${a}"}, {"a": 10, "b": 10}, id="dict"),
+        param({"a": 10, "b": {"a": "${a}"}}, {"a": 10, "b": {"a": 10}}, id="dict"),
+        param({"a": "${b.a}", "b": {"a": 10}}, {"a": 10, "b": {"a": 10}}, id="dict"),
+        param({"a": "???"}, {"a": "???"}, id="dict:missing"),
+        param({"a": "???", "b": "${a}"}, {"a": "???", "b": "???"}, id="dict:missing"),
+        param({"a": 10, "b": "a_${a}"}, {"a": 10, "b": "a_10"}, id="dict:str_inter"),
+        # This seems like a reasonable resolution for a string interpolation pointing to a missing node:
```
odelalleau, line 485: Yeah, I was going to mention it in the code, but catching `InterpolationToMissingValueError` will turn string / nested interpolations into missing values. That's ok with me.

odelalleau (changes requested):
I expect some issues

docs/source/usage.rst
omry: yeah, re-worded before I saw this.

omegaconf/_impl.py
omry: I suspect you were wrong, but in any case take a look at the latest version of the code.
there are tests covering this.

omry:
I am merging this to cut a dev release with it. Feedback is still welcome and if there are change requests I will handle in a followup PR.

omegaconf/_impl.py
```diff
@@ -0,0 +1,44 @@
...
+
+def _resolve(cfg: Node) -> Node:
+    assert isinstance(cfg, Node)
+    try:
+        if cfg._is_interpolation():
+            resolved = cfg._dereference_node()
+            assert resolved is not None
+            cfg._set_value(resolved._value())
+    except InterpolationToMissingValueError:
+        cfg._set_value(MISSING)
```
odelalleau, line 30: MInor (makes it clearer where the exception may be raised):
```python
    if cfg._is_interpolation():
        try:
            resolved = cfg._dereference_node()
        except InterpolationToMissingValueError:
            cfg._set_value(MISSING)
        else:
            assert resolved is not None
            cfg._set_value(resolved._value())
```

omegaconf/_impl.py
```diff
@@ -0,0 +1,44 @@
...
+        for k in cfg.keys():
+            node = cfg._get_node(k)
+            assert isinstance(node, Node)
+            _resolve_container_value(cfg, node, k)
+
+    elif isinstance(cfg, ListConfig):
+        for i in range(len(cfg)):
+            node = cfg._get_node(i)
+            assert isinstance(node, Node)
+            _resolve_container_value(cfg, node, i)
```
odelalleau, line 42: Minor: looks like you could have made the code slightly simpler with:
```python
    if isinstance(cfg, DictConfig):
        for k in cfg.keys():
            _resolve_container_value(cfg, k)

    elif isinstance(cfg, ListConfig):
        for i in range(len(cfg)):
            _resolve_container_value(cfg, i)
```
(and then calling `_get_node()` from within `_resolve_container_value()`)

omegaconf/_impl.py
```diff
@@ -0,0 +1,44 @@
...
+    if node._is_interpolation():
+        try:
+            resolved = node._dereference_node()
+            assert resolved is not None
+            if isinstance(resolved, Container) and isinstance(node, ValueNode):
+                cfg[index] = resolved
+            else:
+                node._set_value(resolved._value())
+        except InterpolationToMissingValueError:
+            node._set_value(MISSING)
```
odelalleau, line 17: This code isn't recursively resolving deeper interpolations. Here's a failure case:
```python
cfg = OmegaConf.create({"x": "${y}", "y": {"z": "${foo}"}, "foo": 0})
OmegaConf.resolve(cfg)
cfg  # <== {'x': {'z': '${foo}'}, 'y': {'z': 0}, 'foo': 0}
```

odelalleau (commented):
Some issues remain

omegaconf/_impl.py
omry: I don't like this syntax very much, feels like it's breaking the flow too much.
trying it to see if it grows on me.

omegaconf/_impl.py
omry: Yes, good call.

omry (commented):
See #642

omegaconf/_impl.py
odelalleau: I agree it's not the prettiest, but after you spend a lot of time investigating a tricky bug caused by a too large `try...except` block, it grows on you pretty fast ;)

PR #641, opened 2021-03-24
Enable the deprecation warning for `register_resolver()`

odelalleau:
Should be a good time to re-enable this warning now that the `register_new_resolver()` API is finalized.

Note that I updated #426, @omry please have a look to check that it's clear enough. Thanks!

### DIFF
--- a/omegaconf/omegaconf.py
+++ b/omegaconf/omegaconf.py
@@ -352,16 +352,15 @@ class OmegaConf:
 
     @staticmethod
     def register_resolver(name: str, resolver: Resolver) -> None:
-        # TODO re-enable warning message before 2.1 release (see also test_resolver_deprecated_behavior)
-        # warnings.warn(
-        #     dedent(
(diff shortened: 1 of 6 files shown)


omry (approved):
looking good.

I edited #426 a bit. It's good in terms of content but I think we should give reorganize the layout:
1. Migration plan is not important for users, it should be last.
2. We should give a tldr that should cover most people at the top for people who are too lazy to read (most of them).

odelalleau:
> looking good.
> 
> I edited #426 a bit. It's good in terms of content but I think we should give reorganize the layout:
> 
> 1. Migration plan is not important for users, it should be last.
> 2. We should give a tldr that should cover most people at the top for people who are too lazy to read (most of them).

I edited #426 accordingly, let me know if ok with you, then I'll merge this PR.

omry:
Looks great, gogo.

PR #642, opened 2021-03-24
Resolve updates

omry:
responding to feedback on #639

### DIFF
--- a/omegaconf/_impl.py
+++ b/omegaconf/_impl.py
@@ -4,8 +4,8 @@ from omegaconf import MISSING, Container, DictConfig, ListConfig, Node, ValueNod
 from omegaconf.errors import InterpolationToMissingValueError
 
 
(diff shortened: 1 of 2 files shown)


odelalleau (approved):
I feel like there ought to be a way to make it a bit simpler, but logic looks good to me.

One small remark is that when calling `_resolve()` from `_resolve_container_value()` we already know that the node we pass isn't an interpolation, so checking it again in `_resolve()` is a waste of time. But we probably don't care much about performance here.

omry:
> I feel like there ought to be a way to make it a bit simpler, but logic looks good to me.
> 
> One small remark is that when calling `_resolve()` from `_resolve_container_value()` we already know that the node we pass isn't an interpolation, so checking it again in `_resolve()` is a waste of time. But we probably don't care much about performance here.

Performance is not a concern here. I agree it's a bit complicated but I care about correctness. It should be straight forward to simplify the logic because we have the tests.

PR #644, opened 2021-03-24
Add new resolvers `oc.dict.keys` and `oc.dict.values`

### LINKED ISSUE #643: Add new resolvers `oc.dict.keys` and `oc.dict.values`
**Description**

Some config options that are stored as a `DictConfig` may sometimes be easier to manipulate as lists, when we care only about the keys or the associated values.

Adding new resolvers `oc.dict.keys` and `oc.dict.values` will make it easy to access either keys or values as a list:

```python
cfg = OmegaConf.create("""
machines:
  node007: 192.168.0.27
  node012: 192.168.0.35
  node075: 192.168.1.89
nodes: ${oc.dict.keys:${machines}}
ips: ${oc.dict.values:${machines}}  
""")

print(f"cfg.nodes: {cfg.nodes}")  # cfg.nodes: ['node007', 'node012', 'node075']
print(f"cfg.ips: {cfg.ips}")      # cfg.ips  : ['192.168.0.27', '192.168.0.35', '192.168.1.89']

```

**Additional context**
See related discussion in https://github.com/omry/omegaconf/pull/578#discussion_r590931279

odelalleau:
Fixes #643

Better check commits separately: the meaningful one is the [second one](https://github.com/omry/omegaconf/pull/644/commits/5fe304f12e39a68a49db0aa2ceafff6c156424da), the other two are refactor commits with no functional changes.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -473,7 +473,8 @@ Some config options that are stored as a ``DictConfig`` may sometimes be easier 
 when we care only about the keys or the associated values.
 
 The resolvers ``oc.dict.keys`` and ``oc.dict.values`` simplify such operations by extracting respectively
-the list of keys and values from ``dict``-like objects like ``DictConfig``:
+the list of keys and values from ``dict``-like objects like ``DictConfig``.
+If a string is given as input, ``OmegaConf.select()`` is used to access the corresponding config node.
 
 .. doctest::
 
@@ -484,8 +485,10 @@ the list of keys and values from ``dict``-like objects like ``DictConfig``:
     ...             "node012": "10.0.0.3",
     ...             "node075": "10.0.1.8",
     ...         },
+    ...         # Explicit interpolation `${machines}` as input.
     ...         "nodes": "${oc.dict.keys:${machines}}",
-    ...         "ips": "${oc.dict.values:${machines}}",
+    ...         # Config node name `machines` as input.
+    ...         "ips": "${oc.dict.values:machines}",
     ...     }
     ... )
     >>> show(cfg.nodes)
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -480,27 +480,23 @@ If a string is given as input, ``OmegaConf.select()`` is used to access the corr
 
     >>> cfg = OmegaConf.create(
     ...     {
-    ...         "machines": {
-    ...             "node007": "10.0.0.7",
-    ...             "node012": "10.0.0.3",
-    ...             "node075": "10.0.1.8",
+    ...         "workers": {
+    ...             "node3": "10.0.0.2",
+    ...             "node7": "10.0.0.9",
     ...         },
     ...         # Obtaining keys with explicit interpolation as input.
-    ...         "nodes": "${oc.dict.keys:${machines}}",
+    ...         "nodes": "${oc.dict.keys:${workers}}",
     ...         # Obtaining values with node name as input.
-    ...         "ips": "${oc.dict.values:machines}",
+    ...         "ips": "${oc.dict.values:workers}",
     ...     }
     ... )
-    >>> nodes = cfg.nodes
-    >>> ips = cfg.ips
-    >>> # The corresponding lists of keys / values are ListConfig nodes.
-    >>> assert isinstance(nodes, ListConfig)
-    >>> assert isinstance(ips, ListConfig)
-    >>> assert nodes == ['node007', 'node012', 'node075']
-    >>> assert ips == ['10.0.0.7', '10.0.0.3', '10.0.1.8']
-    >>> # Values are dynamically updated with the underlying dict.
-    >>> cfg.machines.node012 = "10.0.0.5"
-    >>> assert ips[1] == "10.0.0.5"
+    >>> # Keys are copied from the DictConfig:
+    >>> show(cfg.nodes)
+    type: ListConfig, value: ['node3', 'node7']
+    >>> # Values are dynamically fetched through interpolations:
+    >>> show(cfg.ips)
+    type: ListConfig, value: ['${workers.node3}', '${workers.node7}']
+    >>> assert cfg.ips == ["10.0.0.2", "10.0.0.9"]
 
 
 Custom interpolations
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -485,9 +485,9 @@ If a string is given as input, ``OmegaConf.select()`` is used to access the corr
     ...             "node012": "10.0.0.3",
     ...             "node075": "10.0.1.8",
     ...         },
-    ...         # Explicit interpolation `${machines}` as input.
+    ...         # Obtaining keys with explicit interpolation as input.
     ...         "nodes": "${oc.dict.keys:${machines}}",
-    ...         # Config node name `machines` as input.
+    ...         # Obtaining values with node name as input.
     ...         "ips": "${oc.dict.values:machines}",
     ...     }
     ... )
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -473,8 +473,9 @@ Some config options that are stored as a ``DictConfig`` may sometimes be easier 
 when we care only about the keys or the associated values.
 
 The resolvers ``oc.dict.keys`` and ``oc.dict.values`` simplify such operations by extracting respectively
-the list of keys and values from ``dict``-like objects like ``DictConfig``.
-If a string is given as input, ``OmegaConf.select()`` is used to access the corresponding config node.
+the list of keys and values from existing config nodes.
+They take as input a string that is the path to another config node (using the same syntax
+as interpolations) and return a ``ListConfig`` with its keys / values.
 
 .. doctest::
 
@@ -484,10 +485,8 @@ If a string is given as input, ``OmegaConf.select()`` is used to access the corr
     ...             "node3": "10.0.0.2",
     ...             "node7": "10.0.0.9",
     ...         },
-    ...         # Obtaining keys with explicit interpolation as input.
-    ...         "nodes": "${oc.dict.keys:${workers}}",
-    ...         # Obtaining values with node name as input.
-    ...         "ips": "${oc.dict.values:workers}",
+    ...         "nodes": "${oc.dict.keys: workers}",
+    ...         "ips": "${oc.dict.values: workers}",
     ...     }
     ... )
     >>> # Keys are copied from the DictConfig:
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -472,8 +472,8 @@ Extracting lists of keys / values from a dictionary
 Some config options that are stored as a ``DictConfig`` may sometimes be easier to manipulate as lists,
 when we care only about the keys or the associated values.
 
-The resolvers ``oc.dict.keys`` and ``oc.dict.values`` simplify such operations by extracting respectively
-the list of keys and values from existing config nodes.
+The resolvers ``oc.dict.keys`` and ``oc.dict.values`` simplify such operations by offering an alternative
+view of a dictionary's keys or values as a list.
 They take as input a string that is the path to another config node (using the same syntax
 as interpolations) and return a ``ListConfig`` with its keys / values.
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -1,6 +1,6 @@
 .. testsetup:: *
 
-    from omegaconf import OmegaConf, DictConfig, open_dict, read_write
+    from omegaconf import OmegaConf, DictConfig, ListConfig, open_dict, read_write
     import os
     import sys
     import tempfile
@@ -491,10 +491,16 @@ If a string is given as input, ``OmegaConf.select()`` is used to access the corr
     ...         "ips": "${oc.dict.values:machines}",
     ...     }
     ... )
-    >>> show(cfg.nodes)
-    type: ListConfig, value: ['node007', 'node012', 'node075']
-    >>> show(cfg.ips)
-    type: ListConfig, value: ['10.0.0.7', '10.0.0.3', '10.0.1.8']
+    >>> nodes = cfg.nodes
+    >>> ips = cfg.ips
+    >>> # The corresponding lists of keys / values are ListConfig nodes.
+    >>> assert isinstance(nodes, ListConfig)
+    >>> assert isinstance(ips, ListConfig)
+    >>> assert nodes == ['node007', 'node012', 'node075']
+    >>> assert ips == ['10.0.0.7', '10.0.0.3', '10.0.1.8']
+    >>> # Values are dynamically updated with the underlying dict.
+    >>> cfg.machines.node012 = "10.0.0.5"
+    >>> assert ips[1] == "10.0.0.5"
 
 
 Custom interpolations
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -466,6 +466,34 @@ This can be useful for instance to parse environment variables:
     type: int, value: 3308
 
 
+Extracting lists of keys / values from a dictionary
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Some config options that are stored as a ``DictConfig`` may sometimes be easier to manipulate as lists,
+when we care only about the keys or the associated values.
+
+The resolvers ``oc.dict.keys`` and ``oc.dict.values`` simplify such operations by extracting respectively
+the list of keys and values from ``dict``-like objects like ``DictConfig``:
+
+.. doctest::
+
+    >>> cfg = OmegaConf.create(
+    ...     {
+    ...         "machines": {
+    ...             "node007": "10.0.0.7",
+    ...             "node012": "10.0.0.3",
+    ...             "node075": "10.0.1.8",
+    ...         },
+    ...         "nodes": "${oc.dict.keys:${machines}}",
+    ...         "ips": "${oc.dict.values:${machines}}",
+    ...     }
+    ... )
+    >>> show(cfg.nodes)
+    type: ListConfig, value: ['node007', 'node012', 'node075']
+    >>> show(cfg.ips)
+    type: ListConfig, value: ['10.0.0.7', '10.0.0.3', '10.0.1.8']
+
+
 Custom interpolations
 ^^^^^^^^^^^^^^^^^^^^^
 
--- a/news/643.feature
+++ b/news/643.feature
@@ -1,2 +1,2 @@
-New resolvers `oc.dict.keys` and `oc.dict.values` allow extracting the lists of keys and values of a DictConfig
+New resolvers `oc.dict.keys` and `oc.dict.values` allow extracting the lists of keys and values of config nodes
 
--- a/news/643.feature
+++ b/news/643.feature
@@ -1,2 +1,1 @@
-New resolvers `oc.dict.keys` and `oc.dict.values` allow extracting the lists of keys and values of config nodes
-
+New resolvers `oc.dict.keys` and `oc.dict.values` provide a list view of the keys or values of a DictConfig node.
--- a/news/643.feature
+++ b/news/643.feature
@@ -1,0 +1,2 @@
+New resolvers `oc.dict.keys` and `oc.dict.values` allow extracting the lists of keys and values of a DictConfig
+
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -451,6 +451,7 @@ class Container(Node):
             node that is created to wrap the interpolated value. It is `None` if and only if
             `throw_on_resolution_failure` is `False` and an error occurs during resolution.
         """
+
         try:
             resolved = self.resolve_parse_tree(
                 parse_tree=parse_tree,
@@ -516,6 +517,16 @@ class Container(Node):
             )
         else:
             assert isinstance(resolved, Node)
+            if (
+                parent is not None
+                and resolved._parent is parent
+                and resolved._key() is None
+                and value._key() is not None
+            ):
+                # The interpolation is returning a transient (key-less) node attached
+                # to the current parent. By setting its key to this node's key, we make
+                # it possible to refer to it through an interpolation path.
+                resolved._set_key(value._key())
             return resolved
 
     def _wrap_interpolation_result(
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -517,16 +517,6 @@ class Container(Node):
             )
         else:
             assert isinstance(resolved, Node)
-            if (
-                parent is not None
-                and resolved._parent is parent
-                and resolved._key() is None
-                and value._key() is not None
-            ):
-                # The interpolation is returning a transient (key-less) node attached
-                # to the current parent. By setting its key to this node's key, we make
-                # it possible to refer to it through an interpolation path.
-                resolved._set_key(value._key())
             return resolved
 
     def _wrap_interpolation_result(
--- a/omegaconf/built_in_resolvers.py
+++ b/omegaconf/built_in_resolvers.py
@@ -38,17 +38,25 @@ def dict_keys(
     _root_: BaseContainer,
     _parent_: Container,
 ) -> ListConfig:
-    return _dict_impl(
-        keys_or_values="keys", in_dict=in_dict, _root_=_root_, _parent_=_parent_
+    in_dict = _get_and_validate_dict_input(
+        in_dict, root=_root_, resolver_name="oc.dict.keys"
     )
+    assert isinstance(_parent_, BaseContainer)
+    ret = OmegaConf.create(list(in_dict.keys()), parent=_parent_)
+    assert isinstance(ret, ListConfig)
+    return ret
 
 
 def dict_values(
     in_dict: Union[str, Mapping[Any, Any]], _root_: BaseContainer, _parent_: Container
 ) -> ListConfig:
-    return _dict_impl(
-        keys_or_values="values", in_dict=in_dict, _root_=_root_, _parent_=_parent_
+    in_dict = _get_and_validate_dict_input(
+        in_dict, root=_root_, resolver_name="oc.dict.values"
     )
+    assert isinstance(_parent_, BaseContainer)
+    ret = OmegaConf.create(list(in_dict.values()), parent=_parent_)
+    assert isinstance(ret, ListConfig)
+    return ret
 
 
 def env(key: str, default: Any = _DEFAULT_MARKER_) -> Optional[str]:
@@ -85,24 +93,19 @@ def legacy_env(key: str, default: Optional[str] = None) -> Any:
             raise ValidationError(f"Environment variable '{key}' not found")
 
 
-def _dict_impl(
-    keys_or_values: str,
+def _get_and_validate_dict_input(
     in_dict: Union[str, Mapping[Any, Any]],
-    _root_: BaseContainer,
-    _parent_: Container,
-) -> ListConfig:
+    root: BaseContainer,
+    resolver_name: str,
+) -> Mapping[Any, Any]:
     if isinstance(in_dict, str):
         # Path to an existing key in the config: use `select()`.
-        in_dict = OmegaConf.select(_root_, in_dict, throw_on_missing=True)
+        in_dict = OmegaConf.select(root, in_dict, throw_on_missing=True)
 
     if not isinstance(in_dict, Mapping):
         raise TypeError(
-            f"`oc.dict.{keys_or_values}` cannot be applied to objects of type: "
+            f"`{resolver_name}` cannot be applied to objects of type: "
             f"{type(in_dict).__name__}"
         )
 
-    dict_method = getattr(in_dict, keys_or_values)
-    assert isinstance(_parent_, BaseContainer)
-    ret = OmegaConf.create(list(dict_method()), parent=_parent_)
-    assert isinstance(ret, ListConfig)
-    return ret
+    return in_dict
--- a/omegaconf/built_in_resolvers.py
+++ b/omegaconf/built_in_resolvers.py
@@ -1,0 +1,62 @@
+import os
+import warnings
+from typing import Any, Optional
+
+from ._utils import _DEFAULT_MARKER_, _get_value, decode_primitive
+from .base import Container
+from .errors import ValidationError
+from .grammar_parser import parse
+
+
+# DEPRECATED: remove in 2.2
+def legacy_env(key: str, default: Optional[str] = None) -> Any:
+    warnings.warn(
+        "The `env` resolver is deprecated, see https://github.com/omry/omegaconf/issues/573"
+    )
+
+    try:
+        return decode_primitive(os.environ[key])
+    except KeyError:
+        if default is not None:
+            return decode_primitive(default)
+        else:
+            raise ValidationError(f"Environment variable '{key}' not found")
+
+
+def env(key: str, default: Any = _DEFAULT_MARKER_) -> Optional[str]:
+    """
+    :param key: Environment variable key
+    :param default: Optional default value to use in case the key environment variable is not set.
+                    If default is not a string, it is converted with str(default).
+                    None default is returned as is.
+    :return: The environment variable 'key'. If the environment variable is not set and a default is
+            provided, the default is used. If used, the default is converted to a string with str(default).
+            If the default is None, None is returned (without a string conversion).
+    """
+    try:
+        return os.environ[key]
+    except KeyError:
+        if default is not _DEFAULT_MARKER_:
+            return str(default) if default is not None else None
+        else:
+            raise KeyError(f"Environment variable '{key}' not found")
+
+
+def decode(expr: Optional[str], _parent_: Container) -> Any:
+    """
+    Parse and evaluate `expr` according to the `singleElement` rule of the grammar.
+
+    If `expr` is `None`, then return `None`.
+    """
+    if expr is None:
+        return None
+
+    if not isinstance(expr, str):
+        raise TypeError(
+            f"`oc.decode` can only take strings or None as input, "
+            f"but `{expr}` is of type {type(expr).__name__}"
+        )
+
+    parse_tree = parse(expr, parser_rule="singleElement", lexer_mode="VALUE_MODE")
+    val = _parent_.resolve_parse_tree(parse_tree)
+    return _get_value(val)
--- a/omegaconf/built_in_resolvers.py
+++ b/omegaconf/built_in_resolvers.py
@@ -62,7 +62,7 @@ def dict_values(key: str, _root_: BaseContainer, _parent_: Container) -> ListCon
     assert isinstance(content, dict)
 
     ret = ListConfig([])
-    for k, node in content.items():
+    for k in content:
         ref_node = AnyNode(f"${{{key}.{k}}}")
         ret.append(ref_node)
 
--- a/omegaconf/built_in_resolvers.py
+++ b/omegaconf/built_in_resolvers.py
@@ -2,15 +2,21 @@ import os
 import warnings
 
 # from collections.abc import Mapping, MutableMapping
-from typing import Any, Mapping, Optional, Union
+from typing import Any, List, Mapping, Optional, Union
 
-from ._utils import _DEFAULT_MARKER_, _get_value, decode_primitive
+from ._utils import _DEFAULT_MARKER_, Marker, _get_value, decode_primitive
 from .base import Container
 from .basecontainer import BaseContainer
-from .errors import ValidationError
+from .dictconfig import DictConfig
+from .errors import ConfigKeyError, ValidationError
 from .grammar_parser import parse
 from .listconfig import ListConfig
+from .nodes import AnyNode
 from .omegaconf import OmegaConf
+
+# Special marker use as default value when calling `OmegaConf.select()`. It must be
+# different from `_DEFAULT_MARKER_`, which is used by `OmegaConf.select()`.
+_DEFAULT_SELECT_MARKER_: Any = Marker("_DEFAULT_SELECT_MARKER_")
 
 
 def decode(expr: Optional[str], _parent_: Container) -> Any:
@@ -42,6 +48,7 @@ def dict_keys(
         in_dict, root=_root_, resolver_name="oc.dict.keys"
     )
     assert isinstance(_parent_, BaseContainer)
+
     ret = OmegaConf.create(list(in_dict.keys()), parent=_parent_)
     assert isinstance(ret, ListConfig)
     return ret
@@ -53,6 +60,49 @@ def dict_values(
     in_dict = _get_and_validate_dict_input(
         in_dict, root=_root_, resolver_name="oc.dict.values"
     )
+
+    if isinstance(in_dict, DictConfig):
+        # DictConfig objects are handled in a special way: the goal is to make the
+        # returned ListConfig point to the DictConfig nodes through interpolations.
+
+        dict_key: Optional[str] = None
+        if in_dict._get_root() is _root_:
+            # Try to obtain the full key through which we can access `in_dict`.
+            if in_dict is _root_:
+                dict_key = ""
+            else:
+                dict_key = in_dict._get_full_key(None)
+                if dict_key:
+                    dict_key += "."  # append dot for future concatenation
+                else:
+                    # This can happen e.g. if `in_dict` is a transient node.
+                    dict_key = None
+
+        if dict_key is None:
+            # No path to `in_dict` in the existing config.
+            raise NotImplementedError(
+                "`oc.dict.values` currently only supports input config nodes that "
+                "are accessible through the root config. See "
+                "https://github.com/omry/omegaconf/issues/650 for details."
+            )
+
+        ret = ListConfig([])
+        content = in_dict._content
+        assert isinstance(content, dict)
+
+        for key, node in content.items():
+            ref_node = AnyNode(f"${{{dict_key}{key}}}")
+            ret.append(ref_node)
+
+        # Finalize result by setting proper type and parent.
+        element_type: Any = in_dict._metadata.element_type
+        ret._metadata.element_type = element_type
+        ret._metadata.ref_type = List[element_type]
+        ret._set_parent(_parent_)
+
+        return ret
+
+    # Other dict-like object: simply create a ListConfig from its values.
     assert isinstance(_parent_, BaseContainer)
     ret = OmegaConf.create(list(in_dict.values()), parent=_parent_)
     assert isinstance(ret, ListConfig)
@@ -100,7 +150,12 @@ def _get_and_validate_dict_input(
 ) -> Mapping[Any, Any]:
     if isinstance(in_dict, str):
         # Path to an existing key in the config: use `select()`.
-        in_dict = OmegaConf.select(root, in_dict, throw_on_missing=True)
+        key = in_dict
+        in_dict = OmegaConf.select(
+            root, key, throw_on_missing=True, default=_DEFAULT_SELECT_MARKER_
+        )
+        if in_dict is _DEFAULT_SELECT_MARKER_:
+            raise ConfigKeyError(f"Key not found: '{key}'")
 
     if not isinstance(in_dict, Mapping):
         raise TypeError(
--- a/omegaconf/built_in_resolvers.py
+++ b/omegaconf/built_in_resolvers.py
@@ -58,10 +58,10 @@ def dict_values(key: str, _root_: BaseContainer, _parent_: Container) -> ListCon
         key, parent=_parent_, resolver_name="oc.dict.values"
     )
 
-    ret = ListConfig([])
     content = in_dict._content
     assert isinstance(content, dict)
 
+    ret = ListConfig([])
     for k, node in content.items():
         ref_node = AnyNode(f"${{{key}.{k}}}")
         ret.append(ref_node)
@@ -130,7 +130,6 @@ def _get_and_validate_dict_input(
 
     if in_dict is _DEFAULT_SELECT_MARKER_:
         raise ConfigKeyError(f"Key not found: '{key}'")
-    assert in_dict is not None
 
     if not isinstance(in_dict, DictConfig):
         raise TypeError(
--- a/omegaconf/built_in_resolvers.py
+++ b/omegaconf/built_in_resolvers.py
@@ -1,6 +1,6 @@
 import os
 import warnings
-from typing import Any, Optional
+from typing import Any, Dict, List, Optional
 
 from ._utils import _DEFAULT_MARKER_, _get_value, decode_primitive
 from .base import Container
@@ -26,6 +26,14 @@ def decode(expr: Optional[str], _parent_: Container) -> Any:
     parse_tree = parse(expr, parser_rule="singleElement", lexer_mode="VALUE_MODE")
     val = _parent_.resolve_parse_tree(parse_tree)
     return _get_value(val)
+
+
+def dict_keys(in_dict: Dict[Any, Any]) -> List[Any]:
+    return list(in_dict.keys())
+
+
+def dict_values(in_dict: Dict[Any, Any]) -> List[Any]:
+    return list(in_dict.values())
 
 
 def env(key: str, default: Any = _DEFAULT_MARKER_) -> Optional[str]:
--- a/omegaconf/built_in_resolvers.py
+++ b/omegaconf/built_in_resolvers.py
@@ -8,19 +8,24 @@ from .errors import ValidationError
 from .grammar_parser import parse
 
 
-# DEPRECATED: remove in 2.2
-def legacy_env(key: str, default: Optional[str] = None) -> Any:
-    warnings.warn(
-        "The `env` resolver is deprecated, see https://github.com/omry/omegaconf/issues/573"
-    )
+def decode(expr: Optional[str], _parent_: Container) -> Any:
+    """
+    Parse and evaluate `expr` according to the `singleElement` rule of the grammar.
 
-    try:
-        return decode_primitive(os.environ[key])
-    except KeyError:
-        if default is not None:
-            return decode_primitive(default)
-        else:
-            raise ValidationError(f"Environment variable '{key}' not found")
+    If `expr` is `None`, then return `None`.
+    """
+    if expr is None:
+        return None
+
+    if not isinstance(expr, str):
+        raise TypeError(
+            f"`oc.decode` can only take strings or None as input, "
+            f"but `{expr}` is of type {type(expr).__name__}"
+        )
+
+    parse_tree = parse(expr, parser_rule="singleElement", lexer_mode="VALUE_MODE")
+    val = _parent_.resolve_parse_tree(parse_tree)
+    return _get_value(val)
 
 
 def env(key: str, default: Any = _DEFAULT_MARKER_) -> Optional[str]:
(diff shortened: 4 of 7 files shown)


docs/source/usage.rst
```diff
@@ -458,6 +458,34 @@ This can be useful for instance to parse environment variables:
...
+
+    >>> cfg = OmegaConf.create(
+    ...     {
+    ...         "machines": {
+    ...             "node007": "10.0.0.7",
+    ...             "node012": "10.0.0.3",
+    ...             "node075": "10.0.1.8",
+    ...         },
+    ...         "nodes": "${oc.dict.keys:${machines}}",
+    ...         "ips": "${oc.dict.values:${machines}}",
```
omry, line 480: What do you think about using select syntax instead?

```suggestion
    ...         "nodes": "${oc.dict.keys:machines}",
    ...         "ips": "${oc.dict.values:machines}",
```
Seems significantly cleaner.

omegaconf/built_in_resolvers.py
```diff
@@ -0,0 +1,71 @@
...
+    val = _parent_.resolve_parse_tree(parse_tree)
+    return _get_value(val)
+
+
+def dict_keys(in_dict: Dict[Any, Any]) -> List[Any]:
+    return list(in_dict.keys())
+
+
+def dict_values(in_dict: Dict[Any, Any]) -> List[Any]:
+    return list(in_dict.values())
```
omry, line 36: Handle case where input is of unsupported type.
I think we should support DictConfig and string to be used for selecting a node (which will then be verified as DictConfig). (possibly MutableSequence to support plain dict as well, but can we even get a plain dict here?)

docs/source/usage.rst
odelalleau: Can do, only thing is it makes it a bit less powerful (actually restricting it to existing nodes of the config, while the current implementation could be used on any dict-like object, e.g. coming out of another custom resolver).

omegaconf/built_in_resolvers.py
odelalleau: I was intentionally not checking the type to support any dict-like object. But ok to be stricter. Plain dicts are possible (typically they would come from outputs of other resolvers)

odelalleau:
Something else: I just remembered that I have a WIP change that reverts the automatic conversion of list/dict resolver outputs, relying on explicitly using `oc.create` instead (relevant discussion: https://github.com/omry/omegaconf/pull/606#issuecomment-800589117)

Do you prefer that `oc.dict.keys` output plain dicts / lists (enforcing the usage of `oc.create` to convert them to config containers), or that they output directly DictConfig / ListConfig (the latter is a bit less flexible but probably better for most use cases)

omegaconf/built_in_resolvers.py
omry: You can check `isintance(in_dict, MutableMapping)` to allow for both.

> (typically they would come from outputs of other resolvers)

Yeah,I thought about it but aren't we now converting them to DictConfig automatically?

docs/source/usage.rst
omry: > Something else: I just remembered that I have a WIP change that reverts the automatic conversion of list/dict resolver outputs, relying on explicitly using `oc.create` instead (relevant discussion: [#606 (comment)](https://github.com/omry/omegaconf/pull/606#issuecomment-800589117))
> 
> Do you prefer that `oc.dict.keys` output plain dicts / lists (enforcing the usage of `oc.create` to convert them to config containers), or that they output directly DictConfig / ListConfig (the latter is a bit less flexible but probably better for most use cases, so I'm leaning towards it)

The motivating use case for this is to generate Lists config nodes from dict config nodes, so let's convert to ListConfig automatically.
if someone wants something else they can register their own resolver for now.

omry: > Can do, only thing is it makes it a bit less powerful (actually restricting it to existing nodes of the config, while the current implementation could be used on any dict-like object, e.g. coming out of another custom resolver).
> 
> Unless we support both?

I was thinking both.
See my comment about supported input types.

omry:
Side notes:
1. It becomes more important that we refactor the documentation to give space to detailed resolvers docs. The primary config should contain a list of built in resolvers with a short description of each, and point to another page (not in this PR).
2. There is also `oc.select` to introduce to support interpolation with defaults.

omegaconf/built_in_resolvers.py
odelalleau: > You can check `isintance(in_dict, MutableMapping)` to allow for both.

Yup, was actually going to use `Mapping` (since we're only reading data).

> > (typically they would come from outputs of other resolvers)
> 
> Yeah,I thought about it but aren't we now converting them to DictConfig automatically?

We're not (this is only at the very last step of interpolation resolution, not for intermediate computations, and anyway it's supposed to go away as mentioned [below](https://github.com/omry/omegaconf/pull/644#issuecomment-806180751))

odelalleau:
> 1. It becomes more important that we refactor the documentation to give space to detailed resolvers docs. The primary config should contain a list of built in resolvers with a short description of each, and point to another page (not in this PR).
> 2. There is also `oc.select` to introduce to support interpolation with a default value.

I added an item in #535 for point 1.
Is there already a 2.1 issue to add `oc.select`? (or requiring it?). I don't have it on my radar right now.

omry:
541, I just milestoned it for 2.1. I think it should be straight forward at this point. 
in fact, I already implemented it in a local hack (without testing and docs):

http://paste.ubuntu.com/p/4JRYBMR3Y6/

docs/source/usage.rst
odelalleau: Done in 872eb2d

omegaconf/built_in_resolvers.py
odelalleau: Also done in 872eb2d

omegaconf/built_in_resolvers.py
```diff
@@ -69,3 +84,26 @@ def legacy_env(key: str, default: Optional[str] = None) -> Any:
...
+        raise TypeError(
+            f"`oc.dict.{keys_or_values}` cannot be applied to objects of type: "
+            f"{type(in_dict).__name__}"
+        )
+
+    dict_method = getattr(in_dict, keys_or_values)
+    assert isinstance(_parent_, BaseContainer)
+    ret = OmegaConf.create(list(dict_method()), parent=_parent_)
+    assert isinstance(ret, ListConfig)
+    return ret
```
omry, line 109: IMO would be cleaner to return the dict/DictConfig object and allow the caller to just call the appropriate function.

omegaconf/built_in_resolvers.py
```diff
@@ -28,12 +33,22 @@ def decode(expr: Optional[str], _parent_: Container) -> Any:
...
 
 
-def dict_values(in_dict: Dict[Any, Any]) -> List[Any]:
-    return list(in_dict.values())
+def dict_values(
+    in_dict: Union[str, Mapping[Any, Any]], _root_: BaseContainer, _parent_: Container
+) -> ListConfig:
+    return _dict_impl(
+        keys_or_values="values", in_dict=in_dict, _root_=_root_, _parent_=_parent_
+    )
```
omry, line 51: What is the desired behavior if there is a missing value in the doct selected from?
I think right now this will raise an exception on the call to access the values.

docs/source/usage.rst
```diff
@@ -458,6 +458,37 @@ This can be useful for instance to parse environment variables:
...
+    ...     {
+    ...         "machines": {
+    ...             "node007": "10.0.0.7",
+    ...             "node012": "10.0.0.3",
+    ...             "node075": "10.0.1.8",
+    ...         },
+    ...         # Explicit interpolation `${machines}` as input.
+    ...         "nodes": "${oc.dict.keys:${machines}}",
+    ...         # Config node name `machines` as input.
+    ...         "ips": "${oc.dict.values:machines}",
```
omry, line 483: You are changing two things at once, makes it a bit confusing (using os.dict.values instead of oc.dict.keys and using the alternative selection method). Update the comment to reflect those two changes?

tests/interpolation/built_in_resolvers/test_dict.py
```diff
@@ -0,0 +1,136 @@
...
+def test_dict_keys(restore_resolvers: Any, cfg: Any, key: Any, expected: Any) -> None:
+    OmegaConf.register_new_resolver("sum", lambda x: sum(x))
+
+    cfg = OmegaConf.create(cfg)
+    val = cfg[key]
+    assert val == expected
+    assert type(val) is type(expected)
+
+    if isinstance(val, ListConfig):
+        assert val._parent is cfg
```
omry, line 47: Could be a bit cleaner to split into a basic test_dict_keys function and one that tests it with nesting into sum.
Beyond making the test simpler, the input data can also become pure lists (as the expected output type is always ListConfig).

omry (changes requested):

omegaconf/built_in_resolvers.py
odelalleau: Done in cdc60a8

omegaconf/built_in_resolvers.py
odelalleau: Yeah, I think it's fine, I added tests in 57c3613

We may consider raising an `InterpolationToMissingValueError` instead of a generic `InterpolationResolutionError` wrapping a `MissingMandatoryValue` exception. I can see arguments for either option, so let me know how you feel about this.

docs/source/usage.rst
odelalleau: Clearer in 6898d48?

tests/interpolation/built_in_resolvers/test_dict.py
odelalleau: Done in 018d521

tests/interpolation/built_in_resolvers/test_dict.py
```diff
@@ -134,3 +134,17 @@ def test_dict_values_invalid_type(cfg: Any) -> None:
...
+    [
+        param({"x": "${oc.dict.keys:y}", "y": "???"}, id="keys_select_missing"),
+        param({"x": "${oc.dict.values:y}", "y": "???"}, id="values_select_missing"),
+        param({"x": "${oc.dict.values:y}", "y": {"m": "???"}}, id="missing_value"),
+    ],
+)
+def test_dict_missing(cfg: Any) -> None:
+    cfg = OmegaConf.create(cfg)
+    with raises(InterpolationResolutionError, match="MissingMandatoryValue"):
+        cfg.x
```
omry, line 143: One none obvious behavior here is that those two functions are resolving interpolations eagerly.
OmegaConf interpolations are generally resolved on access.
This can lead to surprising behavior.

```yaml
device: cuda
models_zoo:
  resnet50:
    device: ${device}
    # ...
  alexnet:
    device: ${device}
    # ...

test_models: ${oc.dict.values:models.zoo}
```

```python
cfg = OmegaConf.create(cfg)
tester = ModelsTester(cfg.test_models)
if not torch.cuda.is_available():
  cfg.device = 'cpu'

# At this point, the device in ModelsTester will no longer track cfg.device so the change will not be reflected.
# This is in contrast to the case where `test_models` was a "real" list in the yaml file.
```

I think this is not a huge deal, but it's solvable by iterating on the underlying _content and copying the contained nodes directly).
This would also make the MISSING behavior consistent (and lazy).

tests/interpolation/built_in_resolvers/test_dict.py
```diff
@@ -0,0 +1,143 @@
...
+def test_dict_keys_nested(restore_resolvers: Any) -> None:
+    OmegaConf.register_new_resolver("sum", lambda x: sum(x))
+    cfg = OmegaConf.create({"x": "${sum:${oc.dict.keys:{1: one, 2: two}}}"})
+    assert cfg.x == 3
+
+
+def test_dict_values_nested(restore_resolvers: Any) -> None:
+    OmegaConf.register_new_resolver("sum", lambda x: sum(x))
+    cfg = OmegaConf.create({"x": "${sum:${oc.dict.values:{one: 1, two: 2}}}"})
+    assert cfg.x == 3
```
omry, line 99: Those two can be merged into something like:
```python
@mark.parametrize(
    ("cfg", "expected"),
    [
       param({"x": "${sum:${oc.dict.values:{one: 1, two: 2}}}"}, 3, id="values"),
       param({"x": "${sum:${oc.dict.keys:{1: one, 2: two}}}, 3, id="keys")
    ]
)
def test_nested_oc_dict(restore_resolvers: Any, cfg: Any, expected : Any) -> None:
    OmegaConf.register_new_resolver("sum", lambda x: sum(x))
    cfg = OmegaConf.create(cfg)
    assert cfg.x == expected
```

omry: side question, can you use sum directly  here?
```
OmegaConf.register_new_resolver("sum", sum)
```

odelalleau: Done in f2555ae (no idea why I used a lambda for sum)

tests/interpolation/built_in_resolvers/test_dict.py
odelalleau: Hmm, that's an interesting idea. I have a PoC that works with your example (using `[v for _, v in in_dict.items_ex(resolve=False)]` instead of `list(in_dict.values())` for DictConfigs.

There's a problem though: relative interpolations get broken if we do this. One potential approach would be instead to create nodes in the output ListConfig that are interpolated to the target nodes... but this would start getting a bit hairy => maybe best left for later?

omry: I think changing the parent after copying the nodes to _parent_ should make relative interpolations be relative to the the new node with `${oc.dict.values:...}`.

omry:
```python
ret = []
for _, v in in_dict.items_ex(resolve=False):
  vc = copy.deepcopy(v)
  vc._set_parent(_parent_)
  ret.append(vc)
```

odelalleau: Also, although I agree that it would be nice to support this, note that the current behavior is actually (at least partially) consistent with "interpolations are resolved on access": the reason it doesn't work in your example is because you compute `cfg.test_models` only once, but if you re-access it after changing `cfg.device`, you will get the updated version.
(just making sure that part is clear, doesn't change that I agree "deeper" dynamism would be better)

omry: ah, that won't work. but if you iterate directly on the underlying content as I suggested it will.
items_ex(resolve=False) does not return the underlying nodes.

odelalleau: > I think changing the parent after copying the nodes to `_parent_` should make relative interpolations be relative to the the node with `${oc.dict.values:...}`.

I don’t think that would solve the issue: if we change the parent without changing the interpolation definition (e.g. `${..foo}`) then it may not point to the same node anymore.

omry: This is a matter of definition.
We are creating a virtual node. To me it seems reasonable that relative interpolations be relative to the parent.
This means that if you use it with a new parent it will break relative interpolations that are going up the parents hierarchy.

We can either decide to keep the original parent
```yaml
z:
  a: 10
  b:
    c: ${..a} # 10
x:
  a: 20
  list: ${oc.dict.values:z.b}
```

Is `x.list[0].a` 10 or 20?
if you you print the config, it will hopefully look something like:
```yaml
z:
  a: 10
  b:
    c: ${..a}
x:
  a: 20
  list:
  - c: ${..a}
```

So, should `x.list[0].a` be 10 or 20?

We could provide a resolve flag that would make resolution eager to work around it.

For context:
The purpose of relative interpolation is to make config node relocatable, which is exactly what we are doing here.
a relocatable unit should not have (relative) references to anything outside of it, otherwise it will not relocate properly.

omry: Another advantage of copying the nodes directly is that we are not losing runtime type information. I can expand on this problem if its not clear.

odelalleau: > To me it seems reasonable that relative interpolations be relative to the parent.
> This means that if you use it with a new parent it will break relative interpolations that are going up the parents hierarchy.

I think it would be quite surprising to users. It introduces a subtle coupling between a node set to `${oc.dict.values:foo}` and the node `foo`, where changing the way `foo` is defined (while keeping the same values after resolution) can actually change the values seen by the other node. That sounds pretty bad to me.

> We can either decide to keep the original parent

I was also thinking of this option, I gave it a shot in https://github.com/odelalleau/omegaconf/pull/3/files (WIP code, far from final)
It seems to work in your example. I'm a bit scared of potential side effects of having a `ListConfig` whose child nodes have a different parent, but if you think it's ok we could go for this.

> if you you print the config, it will hopefully look something like:
> 
> ```yaml
> z:
>   a: 10
>   b:
>     c: ${..a}
> x:
>   a: 20
>   list:
>   - c: ${..a}
> ```
> 
> So, should `x.list[0].a` be 10 or 20?

(minor: there should be no `c` in `list`)
I think it should be 10 because we're asking for the first value of `z.b`, which is 10.
Also I'm not sure how you could print that config, since if it's printed with `resolve=False` then `list` should remain an interpolation, and with `resolve=True` then `${..a}` should have been resolved.

> Side note: a workaround to the problem you are describing is to call OmegaConf.resolve() on the node to we are copying from.

To recap, here are the options discussed so far:

1. Keep the current implementation. IMPORTANT: I just realized it actually works in your example (although it breaks with relative interpolations). Somehow yesterday when I tried it I thought it didn't... See below for the working example.
2. Copy nodes (their parent will be the output ListConfig)
3. Copy nodes and set their parent to the current node's parent (not sure exactly what the motivation is, but that's my understanding of https://github.com/omry/omegaconf/pull/644#discussion_r601013752 -- seems weird to me)
4. Copy nodes and set their parent to their original parent (which fixes relative interpolations but could have unintended side effects)
5. `resolve()` the nodes to completely sidestep the interpolation issues (easy and safe, but that's the least dynamic option)
6. Copy nodes and "fix" relative interpolations within them so that they point to the same nodes as before (sounds tricky -- I'd rather not go there)

Here's the working example as of current HEAD (f2555ae):

```python
cfg = OmegaConf.create("""
device: cuda
models_zoo:
  resnet50:
    device: ${device}
    # ...
  alexnet:
    device: ${device}
    # ...

test_models: ${oc.dict.values:models_zoo}
""")

t = cfg.test_models
print(t[1].device)  # cuda
cfg.device = 'cpu'
print(t[1].device)  # cpu
```

omry: # Relative Interpolations

> I think it would be quite surprising to users. It introduces a subtle coupling between a node set to ${oc.dict.values:foo} and the node foo, where changing the way foo is defined (while keeping the same values after resolution) can actually change the values seen by the other node. That sounds pretty bad to me.

Interpolations being evaluated lazily is the surprise. This is an established surprise and I stand behind this design decision.
Such a coupling can only happen if there is a relative interpolation to the parent and in fact can be useful.

If you want your interpolations to be fixed you should use absolute interpolations which are the default.
The only reason to use relative interpolations if you want the resolution to be relative. This means that the location of the node may depends on the location of the node.

```python
c1 = OmegaConf.create({"a": {"b": 10, "c": "${a.b}", "d": "${.b}"}})
print(OmegaConf.to_yaml(c1))
a:
  b: 10
  c: ${a.b}  # fixed
  d: ${.b}    # relative
c2 = OmegaConf.merge(c1, {"zzz": c1})

print(OmegaConf.to_yaml(c2))
a:
  b: 10
  c: ${a.b}
  d: ${.b}
zzz:
  a:
    b: 10
    # changes to a.b are reflected
    c: ${a.b} 
    # changes to a.b are not reflected
    d: ${.b} 
```

This example is using relative interpolation into the subtree. The relocated node is self contained so it makes sense.
In my mind, if you use relative interpolation to the parent, it means it's a part of your logical unit.
Otherwise it's up to you to ensure things still do what you want by changing your config to be friendly to the operations to you performing on it.

> - Keep the current implementation. IMPORTANT: I just realized it actually works in your example (although it breaks with relative interpolations). Somehow yesterday when I tried it I thought it didn't... See below for the working example.

When iterating on the items, ValueNodes are unwrapped but container nodes are returned as is. I think that's why it works.
I do think we should change the parent though, which means we will need to change the implementation to copy the node.

> - Copy nodes and set their parent to the current node's parent (not sure exactly what the motivation is, but that's my understanding of #644 (comment) -- seems weird to me)

We are dynamically creating a config node.
It should be equivalent to a hand coded config with the same values.
The correct parent is the one you are accessing the node from. The resolver is supposed to be transparent and as close as possible to the case where we hand-coded the desired config.

> - resolve() the nodes to completely sidestep the interpolation issues (easy and safe, but that's the least dynamic option)

We could add this as a future enhancement if there is an actual need. Either as a flag to those two functions or as a new oc.resolve built-in resolver (resolve resolver, wdyn!).

> - Copy nodes and "fix" relative interpolations within them so that they point to the same nodes as before (sounds tricky -- I'd rather not go there)

Hell no :).

## flags

Something else we did not discuss at all is the fact that by changing the parent you are also changing the effective flags.
First of all, there are now two possible flags that are effective for a single node, depending on which parent you are accessing it through. this will also have bad interaction with the flags cache.
To me it still makes sense that the parent we are accessing through is the paren't who's flags we should consult. If we copy - we should invalidate the flags cache (if it's being copied as well, I suspect it might not be).


# Repl
Check this out:
https://replit.com/@omry1/OmegaConf-relative-interpolations#main.py

odelalleau: > If you want your interpolations to be fixed you should use absolute interpolations which are the default.
> The only reason to use relative interpolations if you want the resolution to be relative.
> (...)

One main use case (IMO) for relative interpolations is so that you don't have to worry about a subconfig being moved around the global config. So for instance if you write:

```python
cfg = OmegaConf.create("""
model:
    default:
        lr: 1e-3
        activation: linear
        loss: mse
    logistic:
        lr: ${..default.lr}
        activation: sigmoid
        loss: nll
    small_lr:
        lr: 1e-6
        activation: ${..default.activation}
        loss: ${..default.loss}
""")
```
then you can easily rename `model` or put it deeper within the config without having to change its content. That's nice and I think that should be the recommended way of doing things (vs. using absolute interpolations -- I'm not saying it's the best way to define model parameters with default values).

Now let's say you want to get the list of all potential model parameters, by adding a new config entry:
```yaml
model_params: ${oc.dict.values:model}
```

This won't work well because of the relative interpolations. And to me it doesn't make sense to say "it's your fault for using relative interpolations": why would the person writing the `model` parameters have to worry about how other parts of the config may access it through interpolations?

More generally speaking, I strongly feel that if `OmegaConf.resolve(cfg)` works with `cfg` using only absolute interpolations, then it should also give the same result if we convert all interpolations to being relative. I believe this is true currently (correct me if I'm wrong).

(I'll answer to the other points later)

odelalleau: Answering the rest:

> When iterating on the items, ValueNodes are unwrapped but container nodes are returned as is. I think that's why it works.
> I do think we should change the parent though, which means we will need to change the implementation to copy the node.

You're right -- actually the parent is changed, but this isn't great either if we don't copy (since it means the node "loses" its original parent).
 
> > * Copy nodes and set their parent to the current node's parent (not sure exactly what the motivation is, but that's my understanding of #644 (comment) -- seems weird to me)
> 
> We are dynamically creating a config node.
> It should be equivalent to a hand coded config with the same values.
> The correct parent is the one you are accessing the node from. The resolver is supposed to be transparent and as close as possible to the case where we hand-coded the desired config.

I agree in spirit, but with a hand coded config the parent wouldn't be the parent of the node whose interpolation we are resolving. Taking your example, the hand coded config would be:

```yaml
device: cuda
models_zoo:
  resnet50:
    device: ${device}
    # ...
  alexnet:
    device: ${device}
    # ...

# test_models: ${oc.dict.values:models_zoo}
test_models:
  - device: ${device}  # <-- its parent is `test_models`, not the root (= the parent of `test_models`)
  - device: ${device}
```

> Something else we did not discuss at all is the fact that by changing the parent you are also changing the effective flags.

Indeed, something to keep in mind.

> Check this out:
> https://replit.com/@omry1/OmegaConf-relative-interpolations#main.py

That makes sense to me -- anything I should pay specific attention to?


## New idea

Well, not so new since I believe I had mentioned it earlier, but I forgot to put it in my previous list of options: create a `ListConfig` where each element is an interpolated node towards the corresponding node in the config.

Here's a dummy PoC to show what I mean: https://github.com/odelalleau/omegaconf/pull/4/files

And how it works your example:

```python
cfg = OmegaConf.create("""
device: cuda
models_zoo:
  resnet50:
    device: ${device}
    # ...
  alexnet:
    device: ${device}
    # ...

test_models: ${oc.dict.values:models_zoo}
""")
cfg.test_models  # ['${models_zoo.resnet50}', '${models_zoo.alexnet}']
```

This has several benefits:
* Works with relative interpolations
* Avoids the cost of copying the nodes
* Allows modifying the underlying nodes through the list (just like if we have `a: ${b}` we can modify `b` through `a`)

The main drawbacks I can see right now (there are probably more):
* Resolution incurs an extra step to redirect to the correct node
* If we want to expose some kind of type information, using an `AnyNode` isn't a good choice (I guess we could create better typed nodes though, haven't thought too much about it)

Right now I kinda like it. Thoughts?

omry: >> Check this out:
https://replit.com/@omry1/OmegaConf-relative-interpolations#main.py

> That makes sense to me -- anything I should pay specific attention to?

The fact that we can pretty easily share runnable code snippets (based on an available version of OmegaConf, maybe even based on a revision).

I like your new idea. In fact, this is exactly what I suggested that people write to simulate the effect manually: https://github.com/facebookresearch/hydra/issues/1389#issuecomment-776286814

It side steps the interesting problem creating an actual node based on another node dynamically is introducing.
I think we should do it.

odelalleau: > The fact that we can pretty easily share runnable code snippets (based on an available version of OmegaConf, maybe even based on a revision).

Oh ok, cool -- I was wondering if the example itself was highlighting something I may have missed :)

> I think we should do it.

Alright I'll try to flesh it out.

odelalleau: > Alright I'll try to flesh it out.

First attempt in 3394ceb

This commit introduces the PoC of a new node wrapping mechanism, I don't expect it to land "as is", but I wanted to put it out there so we can ground the discussion around it.

The specific problem it addresses in this PR is when `oc.dict.values` must extract the values of a DictConfig that is *not* an existing node in the config. In that case we can't refer to it as an interpolation `${path.to.dictconfig.key}`. The two other solutions I considered are:
- Copying the DictConfig nodes => besides the extra cost, this can break interpolations in the input DictConfig (both absolute and relative)
- Storing the DictConfig nodes directly in the ListConfig but without changing their parent/key. This seems potentially dangerous to me, and doesn't actually seem to solve interpolation problems (I haven't looked too closely why, so maybe I'm wrong on that last point, but I'm still not very comfortable having a ListConfig store nodes that have a different parent)

So I added this new idea of a node being able to wrap another node that doesn't need to be part of the same config. More generally, I feel like this could be potentially useful in scenarios where we wish we had the ability to have multiple parents.
For the purpose of this PR I would actually suggest to create a subclass of `AnyNode` instead of plugging this mechanism directly into `Node` like I did (I didn't do this here to make it quick, so that we can discuss it before the final implementation). It may be moved to `Node` at a later point when/if we have a need for it.

Another remark on this commit: I added some basic typing of the resulting ListConfig. I also considered typing the ListConfig's internal nodes, but I'm not sure it's really useful since in general they should not be used directly.

Let me know what you think!

omry: > The specific problem it addresses in this PR is when oc.dict.values must extract the values of a DictConfig that is not an existing node in the config. In that case we can't refer to it as an interpolation ${path.to.dictconfig.key}. The two other solutions I considered are:

Can you give an example of such a case?

So I added this new idea of a node being able to wrap another node that doesn't need to be part of the same config. More generally, I feel like this could be potentially useful in scenarios where we wish we had the ability to have multiple parents.

> This sounds very close to interpolation to a node (not a value).

I want to understand the reason for it before discussing details. Adding a new low level is something I would like to avoid (it would require defining how it interacts with many other existing concepts, and could require a lot of new tests).

Did you run into issues with the idea of implementing this by returning a list of interpolations?
what are those issues?

odelalleau: > Can you give an example of such a case?

This may happen when the input to `oc.dict.values` is a transient DictConfig that is created by another resolver (which is now more likely to happen, with the upcoming `oc.create` and more generally speaking the new resolver features in 2.1)

Here's a concrete example where the DictConfig comes from a resolver that reads a config from disk:
```python
def load(path, _parent_):
    with open(path) as f:
        data = f.read()
    return OmegaConf.create(data, parent=_parent_)

OmegaConf.register_new_resolver("load", load)

cfg = OmegaConf.create({"coordinates": "${oc.dict.values:${load:test.yaml}}"})

with open("test.yaml", "w") as f:
    f.write("""
x: 1
y: ${.x}
""")

assert cfg.coordinates == [1, 1]
```
We can't create an interpolation to the DictConfig that `load()` creates, because it's not part of the config.

> I want to understand the reason for it before discussing details. Adding a new low level is something I would like to avoid (it would require defining how it interacts with many other existing concepts, and could require a lot of new tests).

Yeah, I get that. Maybe I should just raise a `NotImplementedError` in such situations for now? Then we can revisit later if it's something that may be worth adding in 2.2.
I don't think it's going to be a common use case, and there's actually a workaround for it: create a "dummy" node to store the intermediate resolver result, i.e. something like:
```python
cfg = OmegaConf.create({"coordinates": "${oc.dict.values:ref}", "ref": "${load:test.yaml}"})
```
(won't actually work in this specific example because of an independent bug with `OmegaConf.create()` => I'll open an issue later for that)

omry: In general, I am concerned about the direction of the example.
Hydra 1.1 is enabling very powerful config composition which should be sufficient for all use cases I can think of.
This is trying to create an alternative mechanism that I think more limited and is problematic on many levels.
Loading a config on node access adds a lot of complexity and the semantics are not well defined.
I don't think this is something I want to support (and definitely not for 2.1).

> We can't create an interpolation to the DictConfig that load() creates, because it's not part of the config.

Your example is creating an interpolation inside the loaded config, not to it.
In general, I am not sure I understand the problem with interpolations though (without checking-out your code and running it).
Can you explain what breaks when you interpolate inside the dynamically generated node, or into it?

> Yeah, I get that. Maybe I should just raise a NotImplementedError in such situations for now? Then we can revisit later if it's something that may be worth adding in 2.2.

The actual problem is not clear to me, but if it's a situation you can detect - then yes - good error explaining this is not supported works for me.

odelalleau: > Hydra 1.1 is enabling very powerful config composition which should be sufficient for all use cases I can think of.
> This is trying to create an alternative mechanism that I think more limited and is problematic on many levels.

This is OmegaConf, not Hydra :) You can't tell what people are going to try to do with all the new fancy resolver functionalities...

> > We can't create an interpolation to the DictConfig that load() creates, because it's not part of the config.
> 
> Your example is creating an interpolation inside the loaded config, not to it.
> In general, I am not sure I understand the problem with interpolations though (without checking-out your code and running it).
> Can you explain what breaks when you interpolate inside the dynamically generated node, or into it?

Sorry, that was indeed a bit confusing. Here's what `print(cfg.coordinates)` would print with the workaround that I mentioned afterwards:
```
['${ref.x}', '${ref.y}']
```
This shows that the way `oc.dict.values` works is by creating nodes that interpolate to the desired nodes (here, sub-nodes of `ref`).

However, in my original example (without the `ref` node), I can't create these interpolations since there is no "path" I can define that points to the nodes of the DictConfig that `load()` is creating.

The interpolation `y: ${.x}` is orthogonal to this problem: I only added it as otherwise one may say for instance "just copy the nodes instead", but copying would break this interpolation.

> > Yeah, I get that. Maybe I should just raise a NotImplementedError in such situations for now? Then we can revisit later if it's something that may be worth adding in 2.2.
> 
> The actual problem is not clear to me, but if it's a situation you can detect - then yes - good error explaining this is not supported works for me.

Ok I think I'll do that then (I can detect it, by checking both the parent and the key of the input dictconfig). Probably easier and safer that way.

omry: > This is OmegaConf, not Hydra :) You can't tell what people are going to try to do with all the new fancy resolver functionalities...

I understand that :P.
It does not mean we should encourage people to go in supported directions.
Can you tell me how such config nodes interact with OmegaConf.merge()?

> However, in my original example (without the ref node), I can't create these interpolations since there is no "path" I can define that points to the nodes of the DictConfig that load() is creating.

Okay, it took me trying to actually create the the equivalent config to understand what you mean.
I think we should only support string inputs representing node paths in the enclosing config object.
Trying to support dict/DictConfig input is what gets us into this problem.

omry: The problem with ephemeral generated nodes is that there is no path from the config root to them, so they cannot be accessed via interpolation.
We can only support support config node that are real config nodes in the containing config (same config root as the node of oc.dict.values).

Both of these can work:
```
a:
  b: 10
  c: 20
x: ${oc.dict.values:a}
x: ${oc.dict.values:${a}}
```

odelalleau: > Can you tell me how such config nodes interact with OmegaConf.merge()?

Going to depend on the exact scenario (what are you merging into what?). Not sure it's worth spending too much time on it now -- people merging dynamically generated configs deserve whatever happens to them.

Anyway, I forced-pushed 2e604f9 to override my initial implementation with a new one that doesn't introduce any new low-level mechanics. An error referring to #650 is raised in case someone tries to extract values from a transient DictConfig that's not accessible through a path from the root. I'll fill in the content of #650 once we agree this is a good way to go.

omry: > Going to depend on the exact scenario (what are you merging into what?). Not sure it's worth spending too much time on it now -- people merging dynamically generated configs deserve whatever happens to them.

Dynamically generating config nodes is an alternative to config composition.
Saying people using both deserves whatever happens is not a great approach.
If this feature is not compatible with merging we should be explicit about it and not leave it undefined.
In particular, if you are merging into an interpolating node, the node is expanded and inlined now:
```
In [4]: cfg = OmegaConf.create({"a":  {"b": 10, "c": 20}, "z": "${a}"})

In [5]: cfg
Out[5]: {'a': {'b': 10, 'c': 20}, 'z': '${a}'}

In [6]: OmegaConf.merge(cfg, {"z": {"d": 30}})
Out[6]: {'a': {'b': 10, 'c': 20}, 'z': {'b': 10, 'c': 20, 'd': 30}}
```

odelalleau: > In particular, if you are merging into an interpolating node, the node is expanded and inlined now:

I guess people deserve it to work the same way:
```python
OmegaConf.register_new_resolver(
    "make", lambda _parent_: OmegaConf.create({"b": 10, "c": 20}, parent=_parent_), replace=True)

cfg = OmegaConf.create({"z": "${make:}"})

OmegaConf.merge(cfg, {"z": {"d": 30}})
# {'z': {'b': 10, 'c': 20, 'd': 30}}
```

omry: Great, can you add a test covering this?

odelalleau: Done in e87784c (I also rebased on top of master)

omegaconf/built_in_resolvers.py
```diff
@@ -43,21 +129,39 @@ def env(key: str, default: Optional[str] = _DEFAULT_MARKER_) -> Optional[str]:
...
+    if isinstance(in_dict, str):
+        # Path to an existing key in the config: use `select()`.
+        key = in_dict
+        in_dict = OmegaConf.select(
+            root, key, throw_on_missing=True, default=_DEFAULT_SELECT_MARKER_
+        )
+        if in_dict is _DEFAULT_SELECT_MARKER_:
+            raise ConfigKeyError(f"Key not found: '{key}'")
+
+    if not isinstance(in_dict, Mapping):
```
omry, line 161: If we limit to DictConfig nodes, don't forget to change this.

tests/interpolation/built_in_resolvers/test_dict.py
```diff
@@ -0,0 +1,393 @@
+import re
```
omry, line 1: wow, 400 lines of tests.
I guess breaking the interpolation tests was the right move :)

omry (changes requested):
Went through it except the tests.

Some requests/questions inline.
Also, while playing with it, I found a bug:
```python

cfg = OmegaConf.create(
    {
        "a": {
            "b": {
                "x1": 10,
            },
            "abs_key": "${oc.dict.values:a.b}",
            "abs_dc": "${oc.dict.values:${a.b}}",
            # "relative_key": "${oc.dict.values:.b}",  # this one is breaking
            "relative_dc": "${oc.dict.values:${.b}}",
        },
        "abs_key": "${oc.dict.values:a.b}",
        "abs_dc": "${oc.dict.values:${a.b}}",
        "relative_key": "${oc.dict.values:.a.b}",
        "relative_dc": "${oc.dict.values:${.a.b}}",
    }
)
print("=== unresolved\n\n", OmegaConf.to_yaml(cfg))
print("=== resolved\n\n", OmegaConf.to_yaml(cfg, resolve=True))
```

odelalleau:
> Also, while playing with it, I found a bug:
> (...)

I didn't think of relative interpolations with `select()`. I think there's probably indeed a bug because I'm calling `select(root, ...)` while it should be on the parent if we want relative interpolations to work.

That being said, I couldn't get relative interpolations to work with `select()` at all (outside of this resolver). The following code snippet works without crashing, but (1) I would have liked at least one of these `select()` to actually find the node `.b`, and (2) I would have liked the return value to be 0 when not finding it. So it seems like `select()` doesn't work right now for relative interpolations (unless I'm missing something)

```python

cfg = OmegaConf.create({"a": {"b": 0}, "b": 1})

assert OmegaConf.select(cfg, ".b", default=0) is None
assert OmegaConf.select(cfg.a, ".b", default=0) is None
assert OmegaConf.select(cfg._get_node("a"), ".b", default=0) is None
```

omegaconf/built_in_resolvers.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def dict_values(
+    in_dict: Union[str, Mapping[Any, Any]], _root_: BaseContainer, _parent_: Container
+) -> ListConfig:
+    in_dict = _get_and_validate_dict_input(
+        in_dict, root=_root_, resolver_name="oc.dict.values"
+    )
+
+    if isinstance(in_dict, DictConfig):
```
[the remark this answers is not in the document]
odelalleau: We can. It would prevent it from being used on the output of a resolver that's a dictionary though (instead requiring the somewhat cumbersome and less efficient workaround of creating a temporary node to store the dict as a DictConfig).
I don't feel strongly about it though (it's a niche use case) so let me know if you prefer to support only DictConfigs.

omegaconf/built_in_resolvers.py
```diff
@@ -0,0 +1,167 @@
...
+            # Try to obtain the full key through which we can access `in_dict`.
+            if in_dict is _root_:
+                dict_key = ""
+            else:
+                dict_key = in_dict._get_full_key(None)
+                if dict_key:
+                    dict_key += "."  # append dot for future concatenation
+                else:
+                    # This can happen e.g. if `in_dict` is a transient node.
+                    dict_key = None
```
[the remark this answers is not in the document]
odelalleau: > ```python
> cfg = OmegaConf.create({})
> fake_child = OmegaConf.create({}, parent=cfg)
> ```
> 
> Even though fake_child is in the same root, an interpolation into fake_child is not possible.

But here `fake_child._get_full_key(None)` is equal to `""` so it's properly detected as invalid.
I guess someone could create a fake child with an invalid key, but at this point I think it's fair for some stuff to break.

`OmegaConf.select(_root_, dict_key) is not in_dict` wouldn't actually work for dynamically generated DictConfigs (in addition to add extra overhead). See the test `test_dict_values_dictconfig_resolver_output` for what I mean by that.

omegaconf/base.py
```diff
@@ -516,6 +517,16 @@ def _validate_and_convert_interpolation_result(
...
+            if (
+                parent is not None
+                and resolved._parent is parent
+                and resolved._key() is None
+                and value._key() is not None
+            ):
+                # The interpolation is returning a transient (key-less) node attached
+                # to the current parent. By setting its key to this node's key, we make
+                # it possible to refer to it through an interpolation path.
+                resolved._set_key(value._key())
```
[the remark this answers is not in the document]
odelalleau: It's working for the use case I wanted to support, which is the one in `test_dict_values_dictconfig_resolver_output()`.
We agreed not to support DictConfig objects that couldn't be accessed through interpolations. The specific change you quoted makes it possible to access dynamically generated DictConfig objects through interpolations.

docs/source/usage.rst
```diff
@@ -459,6 +459,43 @@ This can be useful for instance to parse environment variables:
...
+    ...             "node075": "10.0.1.8",
+    ...         },
+    ...         # Obtaining keys with explicit interpolation as input.
+    ...         "nodes": "${oc.dict.keys:${machines}}",
+    ...         # Obtaining values with node name as input.
+    ...         "ips": "${oc.dict.values:machines}",
+    ...     }
+    ... )
+    >>> nodes = cfg.nodes
+    >>> ips = cfg.ips
```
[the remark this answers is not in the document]
odelalleau: How about c3f344b?

docs/source/usage.rst
```diff
@@ -459,6 +459,43 @@ This can be useful for instance to parse environment variables:
...
+    >>> # The corresponding lists of keys / values are ListConfig nodes.
+    >>> assert isinstance(nodes, ListConfig)
+    >>> assert isinstance(ips, ListConfig)
+    >>> assert nodes == ['node007', 'node012', 'node075']
+    >>> assert ips == ['10.0.0.7', '10.0.0.3', '10.0.1.8']
+    >>> # Values are dynamically updated with the underlying dict.
+    >>> cfg.machines.node012 = "10.0.0.5"
+    >>> assert ips[1] == "10.0.0.5"
```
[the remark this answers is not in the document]
odelalleau: I removed it in c3f344b where we show the interpolations

tests/interpolation/built_in_resolvers/test_dict.py
odelalleau: > I guess breaking the interpolation tests was the right move :)

It was! :)

omegaconf/built_in_resolvers.py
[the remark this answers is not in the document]
omry: Okay, let's simplify then.
We can always add support if there is enough demand, but it's much harder to take support away.

docs/source/usage.rst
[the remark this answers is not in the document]
omry: Looking good.

omegaconf/base.py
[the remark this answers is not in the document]
omry: Okay. I will take a closer look at this test to form an opinion (could be a few days).
Hold with any changes for now.

omegaconf/built_in_resolvers.py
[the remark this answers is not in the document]
odelalleau: > Select is just calling _get_node(key) iteratively with fragments of the key, it should be very efficient. (we only need to validate the input node, not all the generated interpolation nodes).

Just flagging that `select()` still has to resolve interpolations along the way (if any).

Maybe if we only support the "select" syntax, i.e. `${oc.dict.values:abc.def}`, then it will be easier since we directly know the correct path.

omry:
I looked at test_dict_values_dictconfig_resolver_output.
Generally speaking, I don't want to push the envelop too far with support for transient config nodes (dynamically generated).

I think oc.dict.values() should support only permanent config nodes in the same config tree.
A way to detect them is to walk up the parent hierarchy and test that each parent contains the key of the node.
This is an alternative approach to using _get_full_key()  + select, which is doing something similar to generate the full key.
Something like:
```python
from omegaconf import OmegaConf, Node, DictConfig, Container

cfg = OmegaConf.create(({"a": {"b": 10}}))


def is_permanent_member_of(node: Node, root: Container) -> bool:
    if node is root:
        return True

    parent = node._get_parent()
    if parent is None:
        return False

    node_in_parent = parent._get_node(
        node._key(),
        throw_on_missing_value=False,
        throw_on_missing_key=False,
        validate_access=False,
    )

    if node_in_parent is not node:
        return False

    # could also be implemented iteratively.
    return is_permanent_member_of(parent, root)


# positive cases
assert is_permanent_member_of(cfg, root=cfg)
assert is_permanent_member_of(cfg.a, root=cfg)
assert is_permanent_member_of(cfg.a._get_node("b"), root=cfg)

# negative cases
assert not is_permanent_member_of(DictConfig(content={}), root=cfg)
assert not is_permanent_member_of(DictConfig(content={}, parent=cfg), root=cfg)
assert not is_permanent_member_of(DictConfig(content={}, parent=cfg.a), root=cfg)
```

As for the select bug, see #656.

odelalleau:
> I think oc.dict.values() should support only permanent config nodes in the same config tree.

Ok, I'll change it in the next few days.

odelalleau:
This is still a WIP but @omry can you please have a look at 27e59f8 to tell me if you agree with the direction?
The main idea is to support *only* the "select" syntax, i.e. `${oc.dict.keys/values: abc.def}` instead of `${oc.dict.keys/values: ${abc.def}}`. This makes the code simpler and there is no confusion regarding the differences between the two syntaxes.

I'll just wait for your green light before updating the tests & documentation.
(current implementation might not be working 100% since I haven't updated the tests yet)

omry:
> This is still a WIP but @omry can you please have a look at [27e59f8](https://github.com/omry/omegaconf/commit/27e59f87cbd1f7db9a6d440e6117a308fe663061) to tell me if you agree with the direction?
> The main idea is to support _only_ the "select" syntax, i.e. `${oc.dict.keys/values: abc.def}` instead of `${oc.dict.keys/values: ${abc.def}}`. This makes the code simpler and there is no confusion regarding the differences between the two syntaxes.
> 
> I'll just wait for your green light before updating the tests & documentation.

I think that's a good idea.
It cuts right through the Gordian knot we got ourselves into by trying to support so many different kind of inputs.

odelalleau:
> > This is still a WIP but @omry can you please have a look at [27e59f8](https://github.com/omry/omegaconf/commit/27e59f87cbd1f7db9a6d440e6117a308fe663061) to tell me if you agree with the direction?
> > The main idea is to support _only_ the "select" syntax, i.e. `${oc.dict.keys/values: abc.def}` instead of `${oc.dict.keys/values: ${abc.def}}`. This makes the code simpler and there is no confusion regarding the differences between the two syntaxes.
> > I'll just wait for your green light before updating the tests & documentation.
> 
> I think that's a good idea.
> It cuts right through the Gordian knot we got ourselves into by trying to support so many different kinds of inputs.

Alright done in 5a1f49c
(NB: I kept the test accessing the config root in spite of other discussions towards forbidding it, because right now it is working, but it'll be ok to break it later if we don't want to support it anymore)

omegaconf/built_in_resolvers.py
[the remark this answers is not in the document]
odelalleau: Should be solved now with 27e59f8 (and companion commit 5a1f49c)

omegaconf/base.py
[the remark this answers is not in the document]
odelalleau: Should also be resolved by 27e59f8

omegaconf/built_in_resolvers.py
odelalleau: That was done in 27e59f8

docs/source/usage.rst
```diff
@@ -466,6 +466,38 @@ This can be useful for instance to parse environment variables:
     type: int, value: 3308
 
 
+Extracting lists of keys / values from a dictionary
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Some config options that are stored as a ``DictConfig`` may sometimes be easier to manipulate as lists,
+when we care only about the keys or the associated values.
+
+The resolvers ``oc.dict.keys`` and ``oc.dict.values`` simplify such operations by extracting respectively
```
omry, line 475: Extracting sounds like it's creating a new list.
Maybe something along the lines of "by offering an alternative view of a dictionary's keys or values as a list"

news/643.feature
```diff
@@ -0,0 +1,2 @@
+New resolvers `oc.dict.keys` and `oc.dict.values` allow extracting the lists of keys and values of config nodes
+
```
omry, line 2:
```suggestion
New resolvers `oc.dict.keys` and `oc.dict.values` provides a list view of the keys or values or a DictConfig node.
```

omegaconf/built_in_resolvers.py
```diff
@@ -0,0 +1,140 @@
+import os
+import warnings
+from typing import Any, List, Optional
```
omry, line 3: In retrospect, it would have been better to make the refactoring a standalone pull request.
With so much moved code, it makes it hard to review.

omegaconf/built_in_resolvers.py
```diff
@@ -0,0 +1,140 @@
...
+        ref_node = AnyNode(f"${{{key}.{k}}}")
+        ret.append(ref_node)
+
+    # Finalize result by setting proper type and parent.
+    element_type: Any = in_dict._metadata.element_type
+    ret._metadata.element_type = element_type
+    ret._metadata.ref_type = List[element_type]
+    ret._set_parent(_parent_)
+
+    return ret
```
omry, line 75: Great, so much simpler.

omry (changes requested):
Looking good. a few nits about docs and news fragment.

docs/source/usage.rst
odelalleau: > Extracting sounds like it's creating a new list.

It *is* creating a new list.

> Maybe something along the lines of "by offering an alternative view of a dictionary's keys or values as a list"

Done in 70e61fc

news/643.feature
odelalleau: Done in 8a65897 (up to minor typo fixes)

omegaconf/built_in_resolvers.py
odelalleau: Yeah, it was initially meant to be a very simple PR easily reviewed by individual commits... I should have cleaned it up when it started getting more complex.
To make it easier I rebased to have the two refactoring commits as the first two commits, and I create a dummy PR on my fork where you can see the diff without these refactoring commits: https://github.com/odelalleau/omegaconf/pull/8/files

omegaconf/built_in_resolvers.py
odelalleau: Looking at it again, there was one more simplification to do: ccba9b8

omry (approved):
shipit!

PR #647, opened 2021-03-26
Support for getitem notation

### LINKED ISSUE #179: [Feature Request] Array Element Overrides from dotlist via getitem syntax
Currently, OmegaConf supports overriding as `training_parameters.lr_steps.1=500` for overriding a specific element of an array. This is not intuitive and we should support proper getitem syntax for array element overrides. 

`training_parameters.lr_steps[1]=500` should be supported.

odelalleau:
Fixes #179

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -317,8 +317,9 @@ OmegaConf supports variable interpolation. Interpolations are evaluated lazily o
 
 Config node interpolation
 ^^^^^^^^^^^^^^^^^^^^^^^^^
-The interpolated variable can be the dot-path to another node in the configuration, and in that case
+The interpolated variable can be the path to another node in the configuration, and in that case
 the value will be the value of that node.
+This path may use either dot-notation (``foo.1``), brackets (``[foo][1]``) or a mix of both (``foo[1]``, ``[foo].1``).
 
 Interpolations are absolute by default. Relative interpolation are prefixed by one or more dots:
 The first dot denotes the level of the node itself and additional dots are going up the parent hierarchy.
@@ -361,7 +362,7 @@ Interpolations may be nested, enabling more advanced behavior like dynamically s
     ...            "B": "plan B",
     ...        },
     ...        "selected_plan": "A",
-    ...        "plan": "${plans.${selected_plan}}",
+    ...        "plan": "${plans[${selected_plan}]}",
     ...    }
     ... )
     >>> cfg.plan # default plan
@@ -814,7 +815,7 @@ Example:
 
 OmegaConf.select
 ^^^^^^^^^^^^^^^^
-OmegaConf.select() allow you to select a config node or value, using either a dot-notation or brackets to denote sub-keys.
+OmegaConf.select() allows you to select a config node or value, using either a dot-notation or brackets to denote sub-keys.
 
 .. doctest::
 
@@ -851,7 +852,7 @@ OmegaConf.select() allow you to select a config node or value, using either a do
 
 OmegaConf.update
 ^^^^^^^^^^^^^^^^
-OmegaConf.update() allow you to update values in your config using either a dot-notation or brackets to denote sub-keys.
+OmegaConf.update() allows you to update values in your config using either a dot-notation or brackets to denote sub-keys.
 
 The merge flag controls the behavior if the input is a dict or a list. If it's true, those are merged instead of
 being assigned.
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -814,7 +814,7 @@ Example:
 
 OmegaConf.select
 ^^^^^^^^^^^^^^^^
-OmegaConf.select() allow you to select a config node or value using a dot-notation key.
+OmegaConf.select() allow you to select a config node or value, using either a dot-notation or brackets to denote sub-keys.
 
 .. doctest::
 
@@ -835,7 +835,8 @@ OmegaConf.select() allow you to select a config node or value using a dot-notati
     >>> assert OmegaConf.select(cfg, "foo.bar") == {
     ...     "zonk" : 10, 
     ... }
-    >>> assert OmegaConf.select(cfg, "foo.bar.zonk") == 10
+    >>> assert OmegaConf.select(cfg, "foo.bar.zonk") == 10    # dots
+    >>> assert OmegaConf.select(cfg, "foo[bar][zonk]") == 10  # brackets
     >>> assert OmegaConf.select(cfg, "no_such_key", default=99) == 99
     >>> assert OmegaConf.select(cfg, "foo.missing") is None
     >>> assert OmegaConf.select(cfg, "foo.missing", default=99) == 99
@@ -850,7 +851,7 @@ OmegaConf.select() allow you to select a config node or value using a dot-notati
 
 OmegaConf.update
 ^^^^^^^^^^^^^^^^
-OmegaConf.update() allow you to update values in your config using a dot-notation key.
+OmegaConf.update() allow you to update values in your config using either a dot-notation or brackets to denote sub-keys.
 
 The merge flag controls the behavior if the input is a dict or a list. If it's true, those are merged instead of
 being assigned.
@@ -858,11 +859,14 @@ being assigned.
 .. doctest::
 
     >>> cfg = OmegaConf.create({"foo" : {"bar": 10}})
-    >>> OmegaConf.update(cfg, "foo.bar", 20, merge=True) # merge has no effect because the value is a primitive
+    >>> # Merge flag has no effect because the value is a primitive
+    >>> OmegaConf.update(cfg, "foo.bar", 20, merge=True)
     >>> assert cfg.foo.bar == 20
-    >>> OmegaConf.update(cfg, "foo.bar", {"zonk" : 30}, merge=False) # set   
+    >>> # Set dictionary value (using dot notation)
+    >>> OmegaConf.update(cfg, "foo.bar", {"zonk" : 30}, merge=False)
     >>> assert cfg.foo.bar == {"zonk" : 30}
-    >>> OmegaConf.update(cfg, "foo.bar", {"oompa" : 40}, merge=True) # merge
+    >>> # Merge dictionary value (using bracket notation)
+    >>> OmegaConf.update(cfg, "foo[bar]", {"oompa" : 40}, merge=True)
     >>> assert cfg.foo.bar == {"zonk" : 30, "oompa" : 40}
 
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -860,18 +860,16 @@ being assigned.
 .. doctest::
 
     >>> cfg = OmegaConf.create({"foo" : {"bar": 10}})
-    >>> # Set dictionary value (dot notation).
-    >>> OmegaConf.update(cfg, "foo.bar", {"zonk" : 30}, merge=False)
-    >>> assert cfg.foo.bar == {"zonk" : 30}
-    >>> # Set dictionary value (bracket notation).
-    >>> OmegaConf.update(cfg, "foo[bar]", {"zonk" : 30}, merge=False)
-    >>> assert cfg.foo.bar == {"zonk" : 30}
-    >>> # Merge dictionary value.
-    >>> OmegaConf.update(cfg, "foo.bar", {"oompa" : 40}, merge=True)
-    >>> assert cfg.foo.bar == {"zonk" : 30, "oompa" : 40}
     >>> # Merge flag has no effect because the value is a primitive
     >>> OmegaConf.update(cfg, "foo.bar", 20, merge=True)
     >>> assert cfg.foo.bar == 20
+    >>> # Set dictionary value (using dot notation)
+    >>> OmegaConf.update(cfg, "foo.bar", {"zonk" : 30}, merge=False)
+    >>> assert cfg.foo.bar == {"zonk" : 30}
+    >>> # Merge dictionary value (using bracket notation)
+    >>> OmegaConf.update(cfg, "foo[bar]", {"oompa" : 40}, merge=True)
+    >>> assert cfg.foo.bar == {"zonk" : 30, "oompa" : 40}
+
 
 
 OmegaConf.masked_copy
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -860,16 +860,18 @@ being assigned.
 .. doctest::
 
     >>> cfg = OmegaConf.create({"foo" : {"bar": 10}})
+    >>> # Set dictionary value (dot notation).
+    >>> OmegaConf.update(cfg, "foo.bar", {"zonk" : 30}, merge=False)
+    >>> assert cfg.foo.bar == {"zonk" : 30}
+    >>> # Set dictionary value (bracket notation).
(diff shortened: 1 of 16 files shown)


omegaconf/grammar/OmegaConfGrammarParser.g4
```diff
@@ -41,7 +41,13 @@ sequence: (element (COMMA element?)*) | (COMMA element?)+;
 // Interpolations.
 
 interpolation: interpolationNode | interpolationResolver;
-interpolationNode: INTER_OPEN DOT* (configKey (DOT configKey)*)? INTER_CLOSE;
+interpolationNode:
+      INTER_OPEN DOT*                                            // dots for relative interpolations
+      (
+        configKey                                                // first config key
+        (DOT configKey | BRACKET_OPEN configKey BRACKET_CLOSE)*  // .foo, [foo], .foo[bar], [foo].bar[baz]
```
omry, line 48: I think the first can also be bracket index:
```python
lc = OmegaConf.create([
  {"a": 10}, 
  {"b": 20}
])

assert OmegaConf.select(lc, "[0].a") == 10
```

omegaconf/grammar_parser.py
```diff
@@ -19,7 +19,8 @@
 # Build regex pattern to efficiently identify typical interpolations.
 # See test `test_match_simple_interpolation_pattern` for examples.
 _config_key = r"[$\w]+"  # foo, $0, $bar, $foo_$bar123$
-_node_path = f"(\\.)*({_config_key}(\\.{_config_key})*)?"  # foo, .foo.$bar
+_node_access = f"\\.{_config_key}|\\[{_config_key}\\]"  # .foo, [foo], [$bar]
+_node_path = f"(\\.)*({_config_key}({_node_access})*)?"  # foo, .foo.$bar, foo[bar]
```
omry, line 23: Can you add another long string to the benchmark that contains this style of key to see performance is good?
e.g:
```
("$a[1].a[1].a[1].a[1].a[1].a[1].a[1].a[1].a[1].a[1].a[1]}", ValueKind.INTERPOLATION),
```

omegaconf/grammar_visitor.py
```diff
@@ -136,16 +136,26 @@ def visitInterpolation(
...
         for child in ctx.getChildren():
             if isinstance(child, TerminalNode):
-                if child.symbol.type == OmegaConfGrammarLexer.DOT:
-                    inter_key_tokens.append(".")  # preserve dots
+                if child.symbol.type in (
+                    OmegaConfGrammarLexer.DOT,
+                    OmegaConfGrammarLexer.BRACKET_OPEN,
+                ):
+                    # Always use dot notation for simplicity.
+                    inter_key_tokens.append(".")
```
omry, line 155: I am assuming that this is why you did not have to change `_resolve_key_and_root()`.
One thing I am suspecting that this will cause some unexpected error messages where the input contains brackets and the error message is in pure dot notation.

omegaconf/grammar/OmegaConfGrammarParser.g4
odelalleau: So, I thought about this, but I think it may be confusing because `x[0]` is equivalent to `x.0`, but `[0]` wouldn't be equivalent to `.0`
At the same time, I was wondering if we should make `[0]` equivalent to `.0`, and I thought this may not be very intuitive precisely because someone may instead expect the behavior you describe here. So I thought "let's just not support it :)"
But it's a small point so let me know what you prefer.

omegaconf/grammar_visitor.py
odelalleau: > I am assuming that this is why you did not have to change `_resolve_key_and_root()`.

No, I didn't change `_resolve_key_and_root()` because as far as I can tell it only relies on the `.` notation for the "relative" part of the interpolation (i.e., the initial dots). And I didn't touch that part.

> One thing I am suspecting that this will cause some unexpected error messages where the input contains brackets and the error message is in pure dot notation.

Indeed! Do you believe it matters? If yes I can preserve the bracket notation instead.

omegaconf/grammar_parser.py
odelalleau: Done in c64d446 => looks good to me (first two rows):
```
ame (time in us)                                                                                                     Min                    Max                Mean              StdDev              Median                IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_get_value_kind[${a[1].a[1].a[1].a[1].a[1].a[1].a[1].a[1].a[1].a[1].a[1]}-ValueKind.INTERPOLATION-True]        6.0050 (1.87)        122.2720 (1.70)       6.7207 (1.87)       2.2021 (1.60)       6.5400 (1.87)      0.1320 (2.49)     467;5170      148.7930 (0.54)      52629           1
test_get_value_kind[${a[1].a[1].a[1].a[1].a[1].a[1].a[1].a[1].a[1].a[1].a[1]}-ValueKind.INTERPOLATION-False]       3.2710 (1.02)        111.6820 (1.55)       3.6911 (1.03)       1.5701 (1.14)       3.5760 (1.02)      0.0920 (1.74)    519;13440      270.9206 (0.98)     103897           1
test_get_value_kind[${a:b,c,d}-ValueKind.INTERPOLATION-False]                                                      3.2090 (1.0)         116.0530 (1.62)       3.6442 (1.01)       2.1572 (1.57)       3.5210 (1.00)      0.0960 (1.81)     609;5267      274.4071 (0.99)     104385           1
test_get_value_kind[${long_string1xxx}_${long_string2xxx:${key}}-ValueKind.INTERPOLATION-False]                    3.2320 (1.01)         71.8260 (1.0)        3.5997 (1.0)        1.4992 (1.09)       3.5240 (1.01)      0.1190 (2.25)     508;1239      277.7977 (1.0)      100868           1
test_get_value_kind[simple-ValueKind.VALUE-False]                                                                  3.2540 (1.01)         96.0400 (1.34)       3.6567 (1.02)       1.6797 (1.22)       3.5450 (1.01)      0.0620 (1.17)     540;6027      273.4733 (0.98)      97638           1
test_get_value_kind[simple-ValueKind.VALUE-True]                                                                   3.2690 (1.02)         88.2840 (1.23)       3.6320 (1.01)       1.7461 (1.27)       3.5060 (1.0)       0.1080 (2.04)      118;427      275.3303 (0.99)      18425           1
test_get_value_kind[${a:${b}}-ValueKind.INTERPOLATION-False]                                                       3.3010 (1.03)         87.7480 (1.22)       3.6397 (1.01)       1.4013 (1.02)       3.5710 (1.02)      0.0840 (1.58)     369;6208      274.7492 (0.99)      98252           1
test_get_value_kind[${${b}}-ValueKind.INTERPOLATION-False]                                                         3.3310 (1.04)        254.8840 (3.55)       3.6705 (1.02)       1.8336 (1.33)       3.5620 (1.02)      0.0630 (1.19)     472;6240      272.4389 (0.98)     100091           1
test_get_value_kind[${a}-ValueKind.INTERPOLATION-False]                                                            3.3660 (1.05)         74.4900 (1.04)       3.6511 (1.01)       1.3773 (1.0)        3.5700 (1.02)      0.0530 (1.0)      130;1378      273.8892 (0.99)      35797           1
test_get_value_kind[${a}-ValueKind.INTERPOLATION-True]                                                             4.3350 (1.35)         90.1880 (1.26)       4.7217 (1.31)       1.6762 (1.22)       4.5880 (1.31)      0.0860 (1.62)     345;1835      211.7900 (0.76)      47697           1
test_get_value_kind[${a:b,c,d}-ValueKind.INTERPOLATION-True]                                                       4.7510 (1.48)         99.3830 (1.38)       5.3331 (1.48)       1.9930 (1.45)       5.1620 (1.47)      0.1160 (2.19)     580;4874      187.5073 (0.67)      54958           1
test_get_value_kind[${${b}}-ValueKind.INTERPOLATION-True]                                                        249.8020 (77.84)       435.5180 (6.06)     260.7636 (72.44)     26.3446 (19.13)    252.9805 (72.16)     3.1620 (59.66)       14;17        3.8349 (0.01)        166           1
test_get_value_kind[${a:${b}}-ValueKind.INTERPOLATION-True]                                                      333.7430 (104.00)   10,315.0600 (143.61)   387.6692 (107.69)   546.8473 (397.06)   338.8680 (96.65)    35.7640 (674.79)       2;16        2.5795 (0.01)        334           1
test_get_value_kind[${long_string1xxx}_${long_string2xxx:${key}}-ValueKind.INTERPOLATION-True]                   530.4720 (165.31)    1,142.6960 (15.91)    583.2386 (162.02)    67.1833 (48.78)    562.0560 (160.31)   14.7300 (277.92)      23;77        1.7146 (0.01)        270           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

omegaconf/grammar_visitor.py
omry: Can you show an a short session that results in such an error?
Will help get a feeling of how bad it looks.

omegaconf/grammar/OmegaConfGrammarParser.g4
omry: This is in part why get_full_key is such a complicated function (by the way, it uses index notation for lists, but looks like not very consistently (that's a small bug there):
```python
In [24]: cfg = OmegaConf.create("[{'a':10}, 2]")

In [25]: cfg
Out[25]: [{'a': 10}, 2]

In [26]: cfg[0]._get_full_key("")
Out[26]: 0

In [27]: cfg[0]._get_full_key("a")
Out[27]: '[0].a'
```
making [0] equivalent to .0 will make it relative by default: its meaning will also differ from 0, I think that's not good.

I think we should make those equivalent:
```
a == [a]
0 == [0]
foo.bar == [foo].bar == foo[bar]
a.b.c == [a].b[c] == a.b[c]
```
```
# start of key:
[k] == k
# middle of key:
[k] == .k
```

While this may look weird, I think people would not be surprised by the behavior.
The case of a top level list is a bit unusual, but the case with relative interpolations may be more common:
```yaml
list:
  - a:
      b: 10
  - ${0.a}
  - ${[0].a}
```

omegaconf/grammar_visitor.py
odelalleau:
```python
cfg = OmegaConf.create({
    "x": "${foo[a][b]}",
})
cfg.x
```

```
InterpolationKeyError                     Traceback (most recent call last)
(...)
InterpolationKeyError: Interpolation key 'foo.a.b' not found
    full_key: x
    object_type=dict
```

omegaconf/grammar/OmegaConfGrammarParser.g4
odelalleau: Ok, done in 8131646

Note that your example isn't actually a relative interpolation, here's the correct version (as of the above commit):
```yaml
list:
  - a:
      b: 10
  - ${.0.a}
  - ${.[0].a}
```

omegaconf/grammar_visitor.py
odelalleau: I changed it in 8131646 to show the correct path (it was important to keep the brackets to support https://github.com/omry/omegaconf/pull/647#discussion_r602629863)

odelalleau:
There seems to be a failure of test_jupyter_notebook-{3.6,3.7} probably caused by the update of an external dependency (it also fails on master)

omry:
Should be good now (https://github.com/python/importlib_metadata/issues/293).

omegaconf/_utils.py
```diff
@@ -789,3 +803,44 @@ def is_generic_dict(type_: Any) -> bool:
...
 def is_container_annotation(type_: Any) -> bool:
     return is_list_annotation(type_) or is_dict_annotation(type_)
+
+
+def split_key(key: str) -> List[str]:
+    """
+    Split a full key path into its individual components.
+
+    This is similar to `key.split(".")` but also works with the getitem syntax:
+        "a.b"       -> ["a", b]
```
omry, line 813:
```suggestion
        "a.b"       -> ["a", "b"]
```

news/179.feature
```diff
@@ -0,0 +1 @@
+Brackets may be used to access sub-keys (instead of the dot notation), e.g. foo.bar -> foo[bar], or foo.0.1 -> foo[0][1]
```
omry, line 1:
```suggestion
Bracketed key may be used as an alternative form to dot notation, e.g. foo.1 is equivalent to foo[1], [foo].1 and [foo][1].
The affects OmegaConf.{update, select} and interpolation usage.
```

docs/source/usage.rst
```diff
@@ -814,7 +814,7 @@ Example:
 
 OmegaConf.select
 ^^^^^^^^^^^^^^^^
-OmegaConf.select() allow you to select a config node or value using a dot-notation key.
+OmegaConf.select() allow you to select a config node or value, using either a dot-notation or brackets to denote sub-keys.
```
omry, line 817: Can you mention it in the `Config node interpolation` as well?

omry (changes requested):
Looking good, a few nits/request to update the docs.

news/179.feature
odelalleau: Done in 9db0ca9 (with two minor tweaks)

docs/source/usage.rst
odelalleau: Done in d7b7b00

docs/source/usage.rst
```diff
@@ -850,19 +852,22 @@ OmegaConf.select() allow you to select a config node or value using a dot-notati
...
+    >>> # Merge flag has no effect because the value is a primitive
+    >>> OmegaConf.update(cfg, "foo.bar", 20, merge=True)
     >>> assert cfg.foo.bar == 20
-    >>> OmegaConf.update(cfg, "foo.bar", {"zonk" : 30}, merge=False) # set   
+    >>> # Set dictionary value (using dot notation)
+    >>> OmegaConf.update(cfg, "foo.bar", {"zonk" : 30}, merge=False)
     >>> assert cfg.foo.bar == {"zonk" : 30}
-    >>> OmegaConf.update(cfg, "foo.bar", {"oompa" : 40}, merge=True) # merge
+    >>> # Merge dictionary value (using bracket notation)
+    >>> OmegaConf.update(cfg, "foo[bar]", {"oompa" : 40}, merge=True)
```
omry, line 870: This is intertwining the merge and the key syntax.
Maybe something like:
```
    >>> # Merge flag has no effect because the value is a primitive
    >>> # Examples with different keys (no merge flag)
    >>> ...
    >>> # Examples with merge flag (one key style)
    >>> ...
```

omry (changes requested):

docs/source/usage.rst
odelalleau: What about e54a860?
I moved the first example in last position as it seemed easier to understand after seeing what merge=True usually does.

omry: Looking at this, I prefer your previous version.
Let's go with that.

omry (approved):

docs/source/usage.rst
odelalleau: Ok, so I'm reverting e54a86026fac57483dafa998e5c7a69fa3463909 then merging once all green

PR #649, opened 2021-03-28
`OmegaConf.create()` now properly sets the parent for string inputs

### LINKED ISSUE #648: `OmegaConf.create(some_string, parent=parent)` doesn't actually set the parent
**Describe the bug**

The following tests all fail in current master because the parent isn't actually set:
```python
@mark.parametrize(
    "data",
    [
        param("", id="empty"),
        param("hello", id="name_only"),
        param("a: b", id="dictconfig"),
        param("- a", id="listconfig"),
    ]
)
def test_create_from_str_check_parent(data: str) -> None:
    parent = OmegaConf.create({})
    cfg = OmegaConf.create(data, parent=parent)
    assert cfg._get_parent() is parent  # <== FAIL!
```

odelalleau:
Fixes #648

### DIFF
--- a/news/648.bugfix
+++ b/news/648.bugfix
@@ -1,0 +1,1 @@
+OmegaConf.create() was ignoring the `parent` argument when taking a string as input
--- a/news/648.bugfix
+++ b/news/648.bugfix
@@ -1,1 +1,1 @@
-OmegaConf.create() was ignoring the `parent` argument when taking a string as input
+Fix OmegaConf.create() to set the provided `parent` when creating a config from a YAML string.
--- a/omegaconf/omegaconf.py
+++ b/omegaconf/omegaconf.py
@@ -825,12 +825,12 @@ class OmegaConf:
             if isinstance(obj, str):
                 obj = yaml.load(obj, Loader=get_yaml_loader())
                 if obj is None:
-                    return OmegaConf.create({}, flags=flags)
+                    return OmegaConf.create({}, parent=parent, flags=flags)
                 elif isinstance(obj, str):
-                    return OmegaConf.create({obj: None}, flags=flags)
+                    return OmegaConf.create({obj: None}, parent=parent, flags=flags)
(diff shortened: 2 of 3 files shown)


news/648.bugfix
```diff
@@ -0,0 +1 @@
+OmegaConf.create() was ignoring the `parent` argument when taking a string as input
```
omry, line 1:
```suggestion
Fix OmegaConf.create() to set the provided `parent` when creating a config from a dict, list and a YAML string.
```

omry (approved):
Looks good. fix the news fragment.

news/648.bugfix
odelalleau: Actually the bug is only for YAML strings => check new text in 59b7bdd

omegaconf/omegaconf.py
```diff
@@ -825,12 +825,12 @@ def _create_impl(  # noqa F811
...
                 if obj is None:
-                    return OmegaConf.create({}, flags=flags)
+                    return OmegaConf.create({}, parent=parent, flags=flags)
                 elif isinstance(obj, str):
-                    return OmegaConf.create({obj: None}, flags=flags)
+                    return OmegaConf.create({obj: None}, parent=parent, flags=flags)
                 else:
                     assert isinstance(obj, (list, dict))
-                    return OmegaConf.create(obj, flags=flags)
+                    return OmegaConf.create(obj, parent=parent, flags=flags)
```
omry, line 833: You fixed three code paths:
None, yaml string, and dict/list.

odelalleau: They're all from Yaml strings:
* None comes from empty string
* The second case is for a string like "hello"
* Dict/list are the "normal" cases

See the tests I added: they're all based on strings.

omry (approved):
Ah, I missed the `isinstance(obj, str)` at line 845.

PR #652, opened 2021-03-29
Cosmetic: remove some whitespaces before colons

odelalleau:
There were a bunch of them, thought it'd be good to do a pass.

This uncovered an invalid `# type : ignore` comment which I removed.


omry (approved):
Thanks.

PR #653, opened 2021-03-30
in get_structured_config_data: check for dict subclass data

### LINKED ISSUE #584: Dict subclass data not preserved during assignment
**Describe the bug**
When a `Dict` subclass is assigned to an omegaconf Node, the data from the dict is thrown away.

**To Reproduce**
```python
$ cat tmp.py
from typing import Dict

@dataclass
class Str2Str(Dict[str, str]):
    x: int = 0

a = Str2Str()
assert a.x == 0
a["y"] = 1
assert a["y"] == 1
cfg_a = OmegaConf.create(a)
assert cfg_a.x == 0
assert "y" in cfg_a
assert cfg_a["y"] == 1
```
```bash
$ python tmp.py
  File "/Users/lappy486/omegaconf/tmp.py", line 18, in <module>
    assert "y" in cfg_a
AssertionError
```

**Expected behavior**
We expect the above assertions to succeed.

**Additional context**
- OmegaConf version: 2.1.0dev20
- Python version: 3.9.1
- Operating system : MacOS

Jasha10:
Closes #584

TODO:
- [x] add a test case to `test_errors.py` for the new `try: ... except ValidationError: ...` code path.
- [x] add bugfix news fragment

### DIFF
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -237,6 +237,23 @@ def get_attr_data(obj: Any, allow_objects: Optional[bool] = None) -> Dict[str, A
         except ValidationError as ex:
             format_and_raise(node=None, key=name, value=value, cause=ex, msg=str(ex))
         d[name]._set_parent(None)
+    if is_dict_subclass(obj_type) and not isinstance(obj, type):
+        key_type, element_type = get_dict_key_value_types(obj_type)
+        for name, value in obj.items():
+            is_optional, type_ = _resolve_optional(element_type)
+            type_ = _resolve_forward(type_, element_type.__module__)
+            try:
+                d[name] = _maybe_wrap(
+                    ref_type=type_,
+                    is_optional=is_optional,
+                    key=name,
+                    value=value,
+                    parent=dummy_parent,
+                )
+            except ValidationError as ex:
+                format_and_raise(
+                    node=None, key=name, value=value, cause=ex, msg=str(ex)
+                )
     return d
 
 
@@ -248,7 +265,8 @@ def get_dataclass_data(
     flags = {"allow_objects": allow_objects} if allow_objects is not None else {}
     dummy_parent = OmegaConf.create({}, flags=flags)
     d = {}
-    resolved_hints = get_type_hints(get_type_of(obj))
+    obj_type = get_type_of(obj)
+    resolved_hints = get_type_hints(obj_type)
     for field in dataclasses.fields(obj):
         name = field.name
         is_optional, type_ = _resolve_optional(resolved_hints[field.name])
@@ -280,6 +298,23 @@ def get_dataclass_data(
         except ValidationError as ex:
             format_and_raise(node=None, key=name, value=value, cause=ex, msg=str(ex))
         d[name]._set_parent(None)
+    if is_dict_subclass(obj_type) and not isinstance(obj, type):
+        key_type, element_type = get_dict_key_value_types(obj_type)
+        for name, value in obj.items():
+            is_optional, type_ = _resolve_optional(element_type)
+            type_ = _resolve_forward(type_, element_type.__module__)
+            try:
+                d[name] = _maybe_wrap(
+                    ref_type=type_,
+                    is_optional=is_optional,
+                    key=name,
+                    value=value,
+                    parent=dummy_parent,
+                )
+            except ValidationError as ex:
+                format_and_raise(
+                    node=None, key=name, value=value, cause=ex, msg=str(ex)
+                )
     return d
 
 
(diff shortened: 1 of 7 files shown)


tests/structured_conf/test_structured_config.py
```diff
@@ -904,6 +904,12 @@ def test_str2str(self, module: Any) -> None:
         with raises(KeyValidationError):
             cfg[Color.RED]
 
+    def test_dict_subclass_data_preserved_upon_node_creation(self, module: Any) -> None:
+        src = module.DictSubclass.Str2Str()
+        src["baz"] = "qux"
+        cfg = OmegaConf.structured(src)
+        assert cfg.baz == "qux"
```
Jasha10, line 911: If run against the master branch, the above test raises `ConfigAttributeError: Missing key baz`.
This PR addresses the error by inspecting for dict subclass data within the `get_attr_data` and `get_dataclass_data` functions.

omegaconf/_utils.py
```diff
@@ -237,6 +237,23 @@ def get_attr_data(obj: Any, allow_objects: Optional[bool] = None) -> Dict[str, A
...
+                    ref_type=type_,
+                    is_optional=is_optional,
+                    key=name,
+                    value=value,
+                    parent=dummy_parent,
+                )
+            except ValidationError as ex:
+                format_and_raise(
+                    node=None, key=name, value=value, cause=ex, msg=str(ex)
+                )
```
omry, line 256: Looks identical to get_dataclass_data, can you refactor to use a common function for this block?

omry: 1. What is the purpose of `not isinstance(obj, type)` in the if?
2. It feels like DictConf shoudl correctly extract the fields from it in the primary loop inside _maybe_wrap(), did you look to see why it's not happening? (I didn't run debug this so I may be wrong).

omry (changes requested):

omegaconf/_utils.py
Jasha10: > Looks identical to get_dataclass_data, can you refactor to use a common function for this block?

Ok, will do.

Jasha10: >     1. What is the purpose of `not isinstance(obj, type)` in the if?

This way, we only get class instances, not classes.

For example `isinstance(MyDataclass(), type)` is false but `isinstance(MyDataclass, type)` is true.

Calling `instance.items()` makes sense, but calling `MyDataclass.items()` does not work.

Jasha10: >     2. It feels like DictConf shoudl correctly extract the fields from it in the primary loop inside _maybe_wrap(), did you look to see why it's not happening? (I didn't run debug this so I may be wrong).

It's because `obj` does not get passed to `_maybe_wrap`; only the fields/values are getting passed to `_maybe_wrap`, one at a time.

Jasha10: > Looks identical to get_dataclass_data, can you refactor to use a common function for this block?

Refactored in https://github.com/omry/omegaconf/pull/653/commits/d30c7d70e92df32ee0b3e61327f38e7d136f7f9c.

omegaconf/_utils.py
```diff
@@ -201,6 +201,37 @@ def _resolve_forward(type_: Type[Any], module: str) -> Type[Any]:
         return type_
 
 
+def extract_dict_subclass_data(obj: Any, parent: Any) -> Optional[Dict[str, Any]]:
+    """Check if obj is an instance of a subclass of Dict. If so, extract the Dict keys/values."""
```
omry, line 205: How does this interact with class fields?
```python
@dataclass
class DictSubclassWithField(Dict[str, Any]):
  age: int = 10

cfg = OmegaConf.create(DictSubclassWithField)
cfg.age = 11
cfg.name = "Commander Keen"

cfg2 = OmegaConf.create(cfg)
assert cfg2 == cfg
```

Jasha10: > How does this interact with class fields?

I don't think there is a change in behavior for class fields.

The only difference should be that on this branch we have
```python
src = DictSubclassWithField(age=11)
src["name"] = "Commander Keen"
print(OmegaConf.create(src))  # prints {'age': 11, 'name': 'Commander Keen'}
```
and on the master branch we have
```python
src = DictSubclassWithField(age=11)
src["name"] = "Commander Keen"
print(OmegaConf.create(src))  # prints {'age': 11}
```

omry (approved):
lgtm

omry:
1. make sure your todo is addressed (in the bug description).
2. Consider adding a bugfix news fragment (if this is a bug in 2.0).

omry (changes requested):
looks like there is a new conflict.

Jasha10:
Should be ready to go :)

omry (approved):

PR #659, opened 2021-04-02
oc.env: converts none-string inputs to string instead of raising an exception

omry:
This changes the behavior of oc.env to convert the input default value to string if it's used (except for None).

Followup: update #573.


Jasha10 (approved):

odelalleau (commented):
Looks good!

Issue #367, opened 2020-09-09
Intruction of merge flag to OmegaConf.update()


omry:
OmegaConf.update() behavior when the input value was a list or a dictionary was to assign the value instead of to merge it:
```python
cfg = OmegaConf.create({"a": {"b": 10}})
OmegaConf.update(cfg, "a", {"c": 20}})
# cfg is now {"a": {"c": 20}}
```

The intended behavior for update lines up better with a merge:
```python
OmegaConf.update(cfg, "a", {"c": 20}})
# cfg is now {"a": {"b": 10, "c": 20}}
```
To support both use cases, a new `merge` flag is introduced to update:
```python
def update(
    cfg: Container, key: str, value: Any = None, merge: Optional[bool] = None
) -> None:
```

- None: issue a warning, default to False to maintain compatibility.
- True : Merge config
- False: Assign config

The default value for merge will become True in a future version.
To avoid unexpected behavior changes, please call update with `merge=[True|False]`.

PR #665, opened 2021-04-07
Add force_add to OmegaConf.update(), effectively using open_dict for all nodes along the path

### LINKED ISSUE #664: Allow OmegaConf.update to ignore struct flag and structured configs and add anyway
Motivation:
In https://github.com/facebookresearch/hydra/issues/1515, an attempt to use `+foo.bar=10` (`+` indicates forced addition in Hydra) is failing if foo is Structured Config.

The current implementation in Hydra is using open_dict on the config root, but Structured Configs are considered closed by default which means the flag override is not effective.
To address this, I will add a force_add flag to `OmegaConf.update()`, that will override the struct node directly on the manipulated node as it traverses the provided path.

omry:
Closes #664.

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -149,3 +149,20 @@ def test_get_value_kind(
 
 def test_is_missing_literal(benchmark: Any) -> None:
     assert benchmark(_is_missing_literal, "???")
+
+
+@mark.parametrize("force_add", [False, True])
+def test_update_force_add(
+    large_dict_config: Any, force_add: bool, benchmark: Any
+) -> None:
+    if force_add:
+        OmegaConf.set_struct(large_dict_config, True)
+
+    benchmark(
+        OmegaConf.update,
+        large_dict_config,
+        "a.a.a.a.a.a.a.a.a.a.a",
+        10,
+        merge=True,
+        force_add=force_add,
+    )
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -158,6 +158,14 @@ def test_update_force_add(
     if force_add:
         OmegaConf.set_struct(large_dict_config, True)
 
+    def recursive_is_struct(node: Any) -> None:
+        if OmegaConf.is_config(node):
+            OmegaConf.is_struct(node)
+            for val in node.values():
+                recursive_is_struct(val)
+
+    recursive_is_struct(large_dict_config)
+
     benchmark(
         OmegaConf.update,
         large_dict_config,
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -863,7 +863,7 @@ OmegaConf.update() allows you to update values in your config using either a dot
 
 The merge flag controls the behavior if the input is a dict or a list. If it's true, those are merged instead of
 being assigned.
-The force_add flag will add the key even if the config or a node along the way a Structured Config or is in struct mode.
+The force_add flag ensures a the path is created even if it will result in insertion of new values into struct nodes.
 
 .. doctest::
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -863,7 +863,7 @@ OmegaConf.update() allows you to update values in your config using either a dot
 
 The merge flag controls the behavior if the input is a dict or a list. If it's true, those are merged instead of
 being assigned.
-The force_add flag ensures a the path is created even if it will result in insertion of new values into struct nodes.
+The force_add flag ensures that the path is created even if it will result in insertion of new values into struct nodes.
 
 .. doctest::
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -863,6 +863,7 @@ OmegaConf.update() allows you to update values in your config using either a dot
 
 The merge flag controls the behavior if the input is a dict or a list. If it's true, those are merged instead of
 being assigned.
+The force_add flag will add the key even if the config or a node along the way a Structured Config or is in struct mode.
 
 .. doctest::
 
@@ -876,6 +877,10 @@ being assigned.
     >>> # Merge dictionary value (using bracket notation)
     >>> OmegaConf.update(cfg, "foo[bar]", {"oompa" : 40}, merge=True)
     >>> assert cfg.foo.bar == {"zonk" : 30, "oompa" : 40}
+    >>> # force_add ignores nodes in struct mode and updates anyway.
+    >>> OmegaConf.set_struct(cfg, True)
+    >>> OmegaConf.update(cfg, "a.b.c.d", 10, merge=True, force_add=True)
+    >>> assert cfg.a.b.c.d == 10
 
 
 
--- a/news/664.feature
+++ b/news/664.feature
@@ -1,1 +1,1 @@
-force_add flag added to OmegaConf.update(), ensuring a the path is created even if it will result in insertion of new values into struct nodes.
+force_add flag added to OmegaConf.update(), ensuring that the path is created even if it will result in insertion of new values into struct nodes.
--- a/news/664.feature
+++ b/news/664.feature
@@ -1,1 +1,1 @@
-OmegaConf.update() accept a new flag force_add (default False), which ensures the config path will be created even if a node along the way is in strut mode.
+force_add flag added to OmegaConf.update(), ensuring a the path is created even if it will result in insertion of new values into struct nodes.
--- a/news/664.feature
+++ b/news/664.feature
@@ -1,0 +1,1 @@
+OmegaConf.update() accept a new flag force_add (default False), which ensures the config path will be created even if a node along the way is in strut mode.
--- a/omegaconf/omegaconf.py
+++ b/omegaconf/omegaconf.py
@@ -727,7 +727,12 @@ class OmegaConf:
 
     @staticmethod
     def update(
-        cfg: Container, key: str, value: Any = None, merge: Optional[bool] = None
+        cfg: Container,
+        key: str,
+        value: Any = None,
+        *,
+        merge: Optional[bool] = None,
+        force_add: bool = False,
     ) -> None:
         """
         Updates a dot separated key sequence to a value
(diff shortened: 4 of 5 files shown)


benchmark/benchmark.py
```diff
@@ -149,3 +149,20 @@ def test_get_value_kind(
 
 def test_is_missing_literal(benchmark: Any) -> None:
     assert benchmark(_is_missing_literal, "???")
+
+
+@mark.parametrize("force_add", [False, True])
+def test_update_force_add(
+    large_dict_config: Any, force_add: bool, benchmark: Any
+) -> None:
```
omry, line 157: Added a benchmark to verify the excessive flag overrides are not causing performance regression.
The concern is that the recursive flags cache invalidation when a flag is set will slow things down.
It does not seem to cause a significant slowdown for a large dict_config here though.

docs/source/usage.rst
```diff
@@ -863,6 +863,7 @@ OmegaConf.update() allows you to update values in your config using either a dot
 
 The merge flag controls the behavior if the input is a dict or a list. If it's true, those are merged instead of
 being assigned.
+The force_add flag will add the key even if the config or a node along the way a Structured Config or is in struct mode.
```
omry, line 866: Happy to take alternative phrasings for this one.

docs/source/usage.rst
```diff
@@ -863,6 +863,7 @@ OmegaConf.update() allows you to update values in your config using either a dot
 
 The merge flag controls the behavior if the input is a dict or a list. If it's true, those are merged instead of
 being assigned.
+The force_add flag ensures a the path is created even if it will result in insertion of new values into struct nodes.
```
odelalleau, line 866: Typo
```suggestion
The force_add flag ensures that the path is created even if it will result in insertion of new values into struct nodes.
```

news/664.feature
```diff
@@ -0,0 +1 @@
+force_add flag added to OmegaConf.update(), ensuring a the path is created even if it will result in insertion of new values into struct nodes.
```
odelalleau, line 1:
```suggestion
force_add flag added to OmegaConf.update(), ensuring that the path is created even if it will result in insertion of new values into struct nodes.
```

omegaconf/omegaconf.py
```diff
@@ -727,7 +727,12 @@ def select(
 
     @staticmethod
     def update(
-        cfg: Container, key: str, value: Any = None, merge: Optional[bool] = None
+        cfg: Container,
+        key: str,
+        value: Any = None,
+        *,
+        merge: Optional[bool] = None,
```
odelalleau, line 734: Probably not a big deal, but just flagging that making `merge` keyword-only is an undocumented breaking change

tests/test_update.py
```diff
@@ -195,3 +196,25 @@ def test_merge_deprecation() -> None:
...
+        param({}, "a.b", 10, {"a": {"b": 10}}, id="add_value"),
+        param({}, "a", {"b": 10}, {"a": {"b": 10}}, id="add_dict"),
+        param({}, "a.b", {"c": 10}, {"a": {"b": {"c": 10}}}, id="add_dict"),
+        param({}, "a", [1, 2], {"a": [1, 2]}, id="add_list"),
+        param({}, "a.b", [1, 2], {"a": {"b": [1, 2]}}, id="add_list"),
+    ],
+)
+def test_update_force_add(cfg: Any, key: str, value: Any, expected: Any) -> None:
+    cfg = _ensure_container(cfg)
+    OmegaConf.set_struct(cfg, True)
```
odelalleau, line 214: Can you add a test where the struct node isn't the root, but rather a node in the middle of the update path?

odelalleau (changes requested):
I only had time for a quick look, I'll need to re-review it more carefully, but already sharing a few comments

omegaconf/omegaconf.py
omry: yeah, I am okay with this one.
I am planning to change the default of merge to be True in the future and all the examples as well as the [deprecation warning link](https://github.com/omry/omegaconf/issues/367) shows it used with an a keyword.

tests/test_update.py
omry: Added something similar with a Structured Config.

omry (commented):
Updated, thanks.

omegaconf/omegaconf.py
```diff
@@ -739,7 +744,8 @@ def update(
         :param merge: If value is a dict or a list, True for merge, False for set.
             True to merge
             False to set
-            None (default) : deprecation warning and default to False
+            None (default): deprecation warning and default to False
+        :param force_add: insert the entire path regardless of Struct flag or Structured Config nodes.
```
odelalleau, line 748: It may be surprising for `force_add` not to allow adding to readonly configs, is that intended?

(if it's changed, it may be easier to rename it into `force` so that it can force both adding new values and overriding existing ones, as otherwise you'd need to make the difference between both situations for readonly configs, which seems tricky)

odelalleau (commented):
Overall looks good, just a few comments & suggestions

tests/test_update.py
```diff
@@ -195,3 +196,32 @@ def test_merge_deprecation() -> None:
...
+    "cfg,key,value,expected",
+    [
+        param({}, "a", 10, {"a": 10}, id="add_value"),
+        param({}, "a.b", 10, {"a": {"b": 10}}, id="add_value"),
+        param({}, "a", {"b": 10}, {"a": {"b": 10}}, id="add_dict"),
+        param({}, "a.b", {"c": 10}, {"a": {"b": {"c": 10}}}, id="add_dict"),
+        param({}, "a", [1, 2], {"a": [1, 2]}, id="add_list"),
+        param({}, "a.b", [1, 2], {"a": {"b": [1, 2]}}, id="add_list"),
+        param(
+            {"user": User(name="Bond", age=7)},
```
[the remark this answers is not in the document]
omry: This was how I initially attempted to implement it.
the downside is that flags are recursive, and in this case you want the flag to apply only to the structured config node itself and not to any nested dictionaries.

In general, Structured Configs works without the struct flag.
The struct flag can be used to open a Structured Config to changes if it's to False on an SC node. Usually this is done temporarily.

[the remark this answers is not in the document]
odelalleau: Yeah I thought about the recursive issue after posting this message. I guess it'd be possible for a structured config to explicitly set `struct=False` onto its unstructured nested dicts & lists... Another idea could be to use a different flag specifically for structured configs... but anyway, like I said, probably not worth changing at this point. As long as the user-facing behavior is sensible, it's not a big deal.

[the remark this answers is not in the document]
omry: Recursively setting a flag on child nodes will have unintended consequences and in general very tricky in the presence of a dynamic object model.
as an example, if each child has an explicit struct flag set to False, it will be hard to set a specific node to struct mode (now we will also need to iterate on all the children to set the flag to True on them).

The problem with Structured Configs is that the the expected behavior is not recursive, not that they  are sharing the struct flag. They are in fact not using the struct flag, EXCEPT to temporarily remove the structness from an SC node.

If we ever need a second non-recursive flag we can consider adding a generic mechanism for it and port the behavior of the Structured Configs to use it.

omegaconf/omegaconf.py
omry: The [reported problem](https://github.com/facebookresearch/hydra/issues/1515) is specifically about adding a field to a Structured Config.
The original implementation in Hydra was using `open_dict on the root of the config and this was not sufficient because of Structured Configs are not inheriting the struct flag from their parent.

read-only is used as a mechanism to protect against changes to nodes [here](https://hydra.cc/docs/patterns/write_protect_config_node).
In contrast to the reported problem, using `with read_write(cfg):` will work fine because it's behavior normal and in fact frozen support for dataclasses is implemented by setting the recursive read-only flag on the DictConfig node.

tests/test_update.py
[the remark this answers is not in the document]
odelalleau: Right -- so maybe what we need is a mechanism to set a flag in a non-recursive way 🤔 (i.e., children wouldn't inherit from it)

omegaconf/omegaconf.py
omry: That future is actually now. Added #667.

omegaconf/omegaconf.py
odelalleau: I see. I feel like it may be possible to have a version of `open_dict()` (or, more generally, `flag_override()`) that behaves as if *all* children had their flag set to the desired value (which would avoid the issue of intermediate nodes causing trouble because of their local flags). It'd be a bit tricky to get it to work efficiently though (you don't want to actually propagate the desired flag to all children -- only those whose flags are being accessed within the context)...

omegaconf/omegaconf.py
```diff
@@ -757,12 +763,14 @@ def update(
...
-                root[key_] = {}
-            root = root[key_]
+            struct_override = False if force_add else root._get_node_flag("struct")
+            with flag_override(root, "struct", struct_override):
+                k = split[i]
+                # if next_root is a primitive (string, int etc) replace it with an empty map
+                next_root, key_ = _select_one(root, k, throw_on_missing=False)
+                if not isinstance(next_root, Container):
+                    root[key_] = {}
+                root = root[key_]
```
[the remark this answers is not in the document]
omry: The code here will not apply cleanly, (see syntax error in 775).
In fact we do have a `does_not_raise` context manager similar to `nullcontext` in the tests. we can promote it to utils.
Feel free to followup with a PR optimizing both places.

tests/test_update.py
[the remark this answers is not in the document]
omry: Something to consider for another PR :)

omegaconf/omegaconf.py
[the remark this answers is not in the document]
odelalleau: Ah, oops, looks like my mouse select skills need improving. I just added the missing bracket.

benchmark/benchmark.py
omry: Yeah, I didn't notice the gating in _invalidate_flags_cache.
Good point about the benchmark not doing much.

I fixed the benchmark to populate the flags cache on all nodes. We don't have a baseline though.
Overall I am not too concerned about relative performance here though, at 400us per set on the big config this is plenty fast enough.

PR #668, opened 2021-04-07
Refactoring: `_maybe_dereference_node`

Jasha10:
On `master`, the `Node._dereference_node` method returns `Optional[Node]`. This leads to some `assert node is not None` statements in the code (for the benefit of `mypy`).

This PR does two things:
  - introduce a new method, `Node._maybe_dereference_node`, which returns `Optional[Node]`
  - change the type signature of `_dereference_node` to retrurn `Node` (not optional).

This change lets us delete some of the `node is not None` assertions.

This PR is a followup to https://github.com/omry/omegaconf/pull/502#discussion_r608711688.

### DIFF
--- a/omegaconf/_impl.py
+++ b/omegaconf/_impl.py
@@ -13,7 +13,6 @@ def _resolve_container_value(cfg: Container, key: Any) -> None:
         except InterpolationToMissingValueError:
             node._set_value(MISSING)
         else:
-            assert resolved is not None
             if isinstance(resolved, Container):
                 _resolve(resolved)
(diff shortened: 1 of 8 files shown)


odelalleau (approved):
Looking good!

omry (commented):
Looking for _dereference_node under the omegaconf module, I think there are more cases where you can eliminate the assert.

Jasha10:
> Looking for _dereference_node under the omegaconf module, I think there are more cases where you can eliminate the assert.

I can't find any more...

omry (approved):
In that case shipit!

PR #669, opened 2021-04-07
Optimization in update()

odelalleau:
This speeds up benchmark/benchmark.py::test_update_force_add by ~50%.

This is the optimization from https://github.com/omry/omegaconf/pull/665#discussion_r608598996

I might do the other one later (no time right now)


omry:
likestamp.

PR #675, opened 2021-04-08
Detect interpolation cycles

### LINKED ISSUE #662: Fail early on interpolation to self or to ancestor node
Using the interpolation `${}` should be illegal because it necessarily causes cycles in the config object.

For example, the following config is pathological: `{"x": "${}"}`.

Based on a comment originally posted by @omry in https://github.com/omry/omegaconf/issues/661#issuecomment-814238029

omry:
This is an initial take at  #662.
Pretty sure there are additional conditions it will not handle correctly.
Feel free to check-out and try to break it.

It detects two conditions:
1. interpoalton into parent (which works but can get other logic into trouble). 
2. cyclic interpolation (e.g. a -> b -> a)

By the way, this is very nice in the debugger:
image

Fixes #662


omry:
Identified an annoying false positive with the simple algorithm I got there.
not sure how to address it (added a test for it).
any ideas?

odelalleau:
> Identified an annoying false positive with the simple algorithm I got there.
> not sure how to address it (added a test for it).
> any ideas?

I haven't looked at the code closely yet but this passes all tests:
```python
        def node_interpolation_callback(
            inter_key: str, memo: Set[int]
        ) -> Optional["Node"]:
            return self._resolve_node_interpolation(
                inter_key=inter_key, memo=copy.copy(memo)  # <== copy the memo here
            )
```

omegaconf/base.py
```diff
@@ -542,10 +552,20 @@ def _wrap_interpolation_result(
...
+        while parent is not None:
+            if parent is target:
+                self._format_and_raise(
+                    key=node._key(),
+                    value=None,
+                    cause=InterpolationResolutionError(
+                        "Interpolation to parent node detected"
+                    ),
+                )
+            parent = parent._get_parent()
```
odelalleau, line 566: There's a potential for an infinite loop here (in case of accidental cycles in the parent graph) => can you add a safety check? (yay, another memo!)

omegaconf/base.py
```diff
@@ -598,6 +620,22 @@ def _maybe_resolve_interpolation(
         key: Any,
         value: "Node",
         throw_on_resolution_failure: bool,
+    ) -> Any:
+        return self._maybe_resolve_interpolation_impl(
+            parent=parent,
+            key=key,
+            value=value,
+            throw_on_resolution_failure=throw_on_resolution_failure,
+            memo=set(),
```
odelalleau, line 629: Shouldn't `memo` contain `id(self)` here?

omegaconf/base.py
```diff
@@ -598,6 +620,22 @@ def _maybe_resolve_interpolation(
...
+        )
+
+    def _maybe_resolve_interpolation_impl(
```
odelalleau, line 632: What's the motivation for splitting `_maybe_resolve_interpolation()`?

omegaconf/base.py
```diff
@@ -626,8 +666,10 @@ def resolve_parse_tree(
         """
         from .nodes import StringNode
 
-        def node_interpolation_callback(inter_key: str) -> Optional["Node"]:
-            return self._resolve_node_interpolation(inter_key=inter_key)
+        def node_interpolation_callback(
+            inter_key: str, memo: Set[int]
```
odelalleau, line 670: Probably that `quoted_string_callback` and `resolver_interpolation_callback` should also use `memo` (for the latter, most resolvers won't use it, but we could at least support it in relevant builtin resolvers like `oc.decode`)

omegaconf/basecontainer.py
```diff
@@ -563,7 +563,8 @@ def assign(value_key: Any, val: ValueNode) -> None:
             elif input_config:
                 assign(key, value)
             else:
-                self.__dict__["_content"][key] = wrap(key, value)
+                value_node = wrap(key, value)
+                self.__dict__["_content"][key] = value_node
```
odelalleau, line 567: noop change?

odelalleau (changes requested):
Some comments below.

Overall it's the kind of logic that's a bit hard to follow and I'm not convinced that the added complexity is worth it. But why not.

Regarding the false positive that I "solved" with a `copy.copy()`, I suspect that it highlights a more generic issue: we want to remember the nodes we've been through in a chain of function calls, but when a function returns, we don't care anymore which nodes it has visited. So I think the same problem exists in other places right now. I haven't really thought of the ideal approach to solve this though.

omegaconf/base.py
```diff
@@ -395,11 +400,17 @@ def _select_impl(
         )
         if value is None:
             return root, last_key, None
-        value = root._maybe_resolve_interpolation(
+        vid = id(value)
+        if vid in memo:
+            raise InterpolationResolutionError("Recursive interpolation detected")
+        memo.add(vid)
```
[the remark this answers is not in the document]
Jasha10: By "transient node", do you mean nodes that are created and then discarded relatively quickly? I don't think that such a node would be garbage collected if the set still holds a reference to the node.

omry:
Okay, I got a fix for the false positive that I think is clean and will probably solve other problems as well:
we are preventing a stack overflow, right?
we should use a fucking stack :).
will look at the rest of the feedback tomorrow or possibly next week.
You are still welcome to check-out find examples that are either false positives or false negatives.

odelalleau:
> Okay, I got a fix for the false positive that I think is clean and will probably solve other problems as well:

I agree, I came to a similar conclusion after giving it more thought (though your solution is better than what I had in mind -- I was going to suggest to copy `memo` in `_select_impl()`, but doing `add/remove` is more efficient)
I think my other comments still hold.

omegaconf/base.py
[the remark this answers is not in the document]
odelalleau: > I don't think that such a node would be garbage collected if the set still holds a reference to the node.

That's the thing: a set doesn't hold a reference to a node, while a dict does. This is the reason why I'm suggesting to use a dict instead.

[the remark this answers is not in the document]
odelalleau: However, looking at it again, in the new version it becomes clearer what is the scope of an entry in `memo`, and it's actually not possible for the associated node to get garbage collected while its id is still in `memo`.
So this concern is addressed => resolving.

omegaconf/base.py
odelalleau: Just to expand a bit on resolvers, what I have in mind is adding a new `_memo_` resolver argument (using the same mechanism as `_parent_` and `_root_`). It could be used by `oc.decode`, `oc.dict.{keys,values}` and `oc.select` for instance.

odelalleau (commented):
Should there be a news item regarding forbidding interpolation to a parent node? (it seems like a breaking change)

By the way, thinking of potential reasons why someone may want to interpolate to the root, there's a situation that isn't so uncommon where there is a subset of the config that represents a specific object's parameters, and you may want for instance to use Hydra's instantiation mechanism to create the object with these parameters.

But then, you realize that your object also needs access to other parts of the config, and the most convenient way to do so could be to add a pointer to the root. Something like:
```yaml
datasets:
  imagenet:
    n_samples: 123456
logging:
  level: debug
model:
  n_hidden: 30
  cost: mse
  global_cfg: ${}
```
By instantiating the model with `cfg.model`, it would still have access to the global config through `global_cfg`.

The current PR would make this a bit more cumbersome to achieve. Just wanted to point it out.

Jasha10:
> By instantiating the model with `cfg.model`, it would still have access to the global config through `global_cfg`.

I think this behavior could still theoretically be supported if we scrap the call to `_validate_not_dereferencing_to_parent`, and just keep the `if vid in memo: raise InterpolationResolutionError` to prevent an infinite loop situation.

Jasha10:
> By instantiating the model with `cfg.model`, it would still have access to the global config through `global_cfg`.

This reminds me of an idea I had for handling cyclic or diamond-shaped configs in `OmegaConf.to_container` and `OmegaConf.to_object`.

Currently, the following results in a RecursionError:
```python
cfg = OmegaConf.create({"a": {"b": "${a}"}})
OmegaConf.to_container(cfg, resolve=True)
```

However, it is possible to create a (perfectly valid) python dict that looks like what the output of OmegaConf.to_container "should be":
```python
a = {"b": "tmp"}
a["b"] = a
```

I think that such a behavior could be supported by `OmegaConf.{to_container,to_object}` if we use a dict to keep track of which nodes have already been converted to container.

omry (commented):
I didn't see your copy suggestion before implementing my solution.
what made me realize that we need a stack was you mentioning that we want to track function calls.
Initially I used an actual stack (deque) but decided to keep the set because we also check for containment and overall a set is more efficient for all the operations we need.

omry:
> Should there be a news item regarding forbidding interpolation to a parent node? (it seems like a breaking change)

Yes.

> By the way, thinking of potential reasons why someone may want to interpolate to the root, there's a situation that isn't so uncommon where there is a subset of the config that represents a specific object's parameters, and you may want for instance to use Hydra's instantiation mechanism to create the object with these parameters.
> 
> But then, you realize that your object also needs access to other parts of the config, and the most convenient way to do so could be to add a pointer to the root. Something like:
> 
> ```yaml
> datasets:
>   imagenet:
>     n_samples: 123456
> logging:
>   level: debug
> model:
>   n_hidden: 30
>   cost: mse
>   global_cfg: ${}
> ```
> 
> By instantiating the model with `cfg.model`, it would still have access to the global config through `global_cfg`.

If you must do it you can pass hand it the global config directly.
zooming out a bit, needing access to the global config is an indication or poorly designed code.

> The current PR would make this a bit more cumbersome to achieve. Just wanted to point it out.

Good. :)
It also makes it harder for people to crash their entire program when they try to print a resolved config.

odelalleau:
> > By instantiating the model with `cfg.model`, it would still have access to the global config through `global_cfg`.
> 
> I think this behavior could still theoretically be supported if we scrap the call to `_validate_not_dereferencing_to_parent`, and just keep the `if vid in memo: raise InterpolationResolutionError` to prevent an infinite loop situation.

Yeah I was actually wondering if `_validate_not_dereferencing_to_parent()` was still needed, but the answer is yes in the motivating example from #662. I haven't looked closely at it but it's probably because we "exit" the interpolation loop and yet still recurse further into other functions that call again the interpolation, etc. (that's the problem with carrying `memo` around: it's hard to make sure it is propagated everywhere you'd want it to)

> I think that such a behavior could be supported by `OmegaConf.{to_container,to_object}` if we use a dict to keep track of which nodes have already been converted to container. (Keys of the dict would be `id(node)`, values in the dict would be the corresponding python container)

Probably -- though this PR is meant to explicitly forbid it so probably not worth going that direction ;)

omegaconf/base.py
odelalleau: Hmm yeah it may be tricky to trigger it without something else blowing up. But since you wrote the code why not keep it?
(you can always write a test that directly calls `_validate_not_dereferencing_to_parent()`)

Main reason why I'm pointing it out is that it's an infinite loop and not a recursion error (and stuck programs suck).

omegaconf/base.py
odelalleau: I think it's not a big deal because the consequence is that you may wait an extra step before triggering the exception. I see no reason not to have it though, do you?
If we wanted to test it we'd need to check which key triggers the cycle detection.

omegaconf/base.py
odelalleau: Your new quoted strings tests aren't actually testing quoted strings, they're all top level strings. You can use `oc.decode` to test quoted strings, e.g.:
```python
cfg = OmegaConf.create({"a": "${oc.decode:'x ${a}'}"})
cfg.a
```

On resolvers: I think it could remain documented and have a docstring that says "internal use only" (or we could create a separate private register function). But I'm fine with keeping it for later.

omry: Quoted strings: Added a proper test and fixed.

Custom resolvers: 
Feel free to introduce it when we have a custom resolver that can actually benefit from it.
Standard usage of custom resolvers does not have this issue because they are not the ones doing derefencing the interpolations.

omegaconf/base.py
odelalleau: Agreed, 10 mins is plenty :)

omry:
> > Should there be a news item regarding forbidding interpolation to a parent node? (it seems like a breaking change)
> 
> Yes.
> 
> > By the way, thinking of potential reasons why someone may want to interpolate to the root, there's a situation that isn't so uncommon where there is a subset of the config that represents a specific object's parameters, and you may want for instance to use Hydra's instantiation mechanism to create the object with these parameters.
> > But then, you realize that your object also needs access to other parts of the config, and the most convenient way to do so could be to add a pointer to the root. Something like:
> > ```yaml
> > datasets:
> >   imagenet:
> >     n_samples: 123456
> > logging:
> >   level: debug
> > model:
> >   n_hidden: 30
> >   cost: mse
> >   global_cfg: ${}
> > ```

By the way: This is doable with a custom resolver like:
```
OmegaConf.register_new_resolver("root", lambda _root_: _root_)
cfg = OmegaConf.create({"a": {"b": "${root:}"}})
print(cfg)
print(cfg.a.b)
print(cfg.a.b.a.b)
OmegaConf.resolve(cfg)  # boom
```
Outputs:
```
{'a': {'b': '${root:}'}}
{'a': {'b': '${root:}'}}
{'a': {'b': '${root:}'}}
...
omegaconf.errors.InterpolationResolutionError: RecursionError raised while resolving interpolation: maximum recursion depth exceeded in comparison
```

tests/interpolation/test_interpolation.py
```diff
@@ -444,3 +444,53 @@ def test_type_validation_error_no_throw() -> None:
...
+        param({"a": "${a}"}, "a", IRE, id="self_interpolation"),
+        param({"a": "${b}", "b": "${a}"}, "a", IRE, id="ping-pong"),
+        param({"a": {"a": "${b}"}, "b": "${a.a}"}, "a.a", IRE, id="ping-pong"),
+        param({"a": {"a": "${.a}"}}, "a.a", IRE, id="self-relative"),
+        # quoted strings
+        param({"a": "${oc.decode:'${a}'}"}, "a", IRE, id="quoted"),
+        # custom resolver
+        param({"a": "${oc.identity:${a}}"}, "a", IRE, id="custom-resolver"),
+        param(
+            {"a": "${oc.identity:${oc.identity:${a}}}"}, "a", IRE, id="custom-resolver"
```
odelalleau, line 480: Should actually be `identity` and not `oc.identity`. Technically the test will work with any resolver name since the exception will be raised before checking the resolver, but if for any reason the recursion exception isn't raised then it would be an issue (since the unknown resolver name would also trigger an `IRE` and thus the test would still pass).

omegaconf/base.py
odelalleau: > Custom resolvers:
> Feel free to introduce it when we have a custom resolver that can actually benefit from it.
> Standard usage of custom resolvers does not have this issue because they are not the ones doing derefencing the interpolations.

To be honest I probably won't bother, but right now `oc.decode` could already benefit from it since it's able to resolve interpolations.

omegaconf/grammar/OmegaConfGrammarParser.g4
```diff
@@ -41,13 +41,14 @@ sequence: (element (COMMA element?)*) | (COMMA element?)+;
...
 interpolation: interpolationNode | interpolationResolver;
+
 interpolationNode:
       INTER_OPEN
       DOT*                                                       // relative interpolation?
       (
         (configKey | BRACKET_OPEN configKey BRACKET_CLOSE)       // foo, [foo]
         (DOT configKey | BRACKET_OPEN configKey BRACKET_CLOSE)*  // .foo, [foo], .foo[bar], [foo].bar[baz]
-      )?
+      )
```
odelalleau, line 51: Probably the parentheses can be removed?
Also can you update grammar_visitor:
`assert ctx.getChildCount() >= 2` should become `assert ctx.getChildCount() >= 3` and the comment above that assert should be updated with the new definition

odelalleau: + the regex version and associated tests probably need a look too

odelalleau (changes requested):

tests/interpolation/test_interpolation.py
omry: True and good catch.
However, another IRE will not have the message "Recursive interpolation detected", so I am not sure what is going on here.

omry: oh, actually I know what is going on here: the recursive interpolations are resolved before calling oc.identity and discovering there is no such thing.

omegaconf/base.py
omry: Can you show oc.decode usage that is failing to detect an interpolation cycle with this PR?

omegaconf/grammar/OmegaConfGrammarParser.g4
omry: Yeah. I took a look at the regex pattern and it wasn't obvious to me how to fix it (this is no longer a simple revert because of the addition of brackets to the node path syntax). can you take a look? (I moved the relevant passing cases to the negative test).

tests/interpolation/test_interpolation.py
```diff
@@ -444,3 +441,45 @@ def test_type_validation_error_no_throw() -> None:
...
+    ("cfg", "key", "expected"),
+    [
+        param({"a": "${a}"}, "a", IRE, id="self_interpolation"),
+        param({"a": "${b}", "b": "${a}"}, "a", IRE, id="ping-pong"),
+        param({"a": {"a": "${b}"}, "b": "${a.a}"}, "a.a", IRE, id="ping-pong"),
+        param({"a": {"a": "${.a}"}}, "a.a", IRE, id="self-relative"),
+        # quoted strings
+        param({"a": "${oc.decode:'${a}'}"}, "a", IRE, id="quoted"),
+        # custom resolver
+        param({"a": "${oc.identity:${a}}"}, "a", IRE, id="custom-resolver"),
```
odelalleau, line 469:
```suggestion
        param({"a": "${identity:${a}}"}, "a", IRE, id="custom-resolver"),
```

odelalleau (changes requested):

omegaconf/base.py
odelalleau:
```python
cfg = OmegaConf.create({"a": "${oc.decode:${oc.env:VAR}}", "b": r"${oc.decode:'\${b}'}"})
os.environ["VAR"] = "${a}"
cfg.a  # or cfg.b
```

The case with `cfg.a` shows how it could happen in practice. The case with `cfg.b` is what I would put in a unit test.

tests/interpolation/test_interpolation.py
odelalleau: > oh, actually I know what is going on here: the recursive interpolations are resolved before calling oc.identity and discovering there is no such thing.

Yes exactly, my concern was what would happen if for any reason there would be no error anymore due to the cycle: then the test would have still passed by failing on the resolver name.
However, I hadn't noticed that the test was actually matching the text of the exception, so it wouldn't actually have been a real issue.

omegaconf/grammar/OmegaConfGrammarParser.g4
odelalleau: I pushed it directly, hope that's ok:
* cb902788d25f7ac5ed1dc4d2d7bbe818800e7083 adds a consistency check between the regex and the full grammar
* 9fe900e9eb7b75d73f51ef5be368d117359b91e6 is the actual fix to the regex

omegaconf/base.py
```diff
@@ -346,7 +347,11 @@ def _resolve_key_and_root(self, key: str) -> Tuple["Container", str]:
             return root, key
 
     def _select_impl(
-        self, key: str, throw_on_missing: bool, throw_on_resolution_failure: bool
+        self,
+        key: str,
+        throw_on_missing: bool,
+        throw_on_resolution_failure: bool,
+        memo: Set[int],
```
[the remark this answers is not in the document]
odelalleau: `b` should have been `a.a.a` in your example I think.

Here's a minimal test case:
```python
cfg = OmegaConf.create({"a": "${b}", "b": "${a.a}"})
cfg.a
```

omegaconf/base.py
odelalleau: Going to resolve this one since I don't think it matters much to delay the detection by one step

omegaconf/grammar/OmegaConfGrammarParser.g4
omry: yeah, it's fine - thanks.

omry:
Okay, I think I got all the issues. let me know if I missed anything.

grouped the commits until today into two:
1. adding support for the two detections.
2. removing the grammar support for {}, {..} etc.
On top of those two, I have two commits - one adding a the test case you provided and the second fixing it. 
I will combine the last two with 1 before I land this.

tests/test_grammar.py
```diff
@@ -363,7 +364,8 @@ def test_single_element_no_interpolation(
         visitor = grammar_visitor.GrammarVisitor(
             node_interpolation_callback=None,  # type: ignore
             resolver_interpolation_callback=None,  # type: ignore
-            quoted_string_callback=lambda s: s,
+            quoted_string_callback=lambda s, memo: s,
+            memo=set(),
```
odelalleau, line 368: Probably that `memo` could be `None` here

odelalleau (approved):
All good! Just a minor comment that doesn't really need to be addressed.

tests/test_grammar.py
omry: Yup.
I only looked for set() in under the omegaconf module so I missed that one.

omry:
Okay,  merging once green.
1. rebased and removed newly added test for the behavior of oc.dict.values with ${}.
2. Updates news fragment to be about Interpolation cycles in  general and not the interpolation to a parent specifically.

odelalleau:
> 2. Updates news fragment to be about Interpolation cycles in  general and not the interpolation to a parent specifically.

Not a big deal, but the breaking change is the interpolation to a parent (that doesn't necessarily trigger a cycle). Cycles were already crashing before (just in a different way -- that nobody should have been relying on).

omry:
It felt worth mentioning that we this is now officially detected.

PR #676, opened 2021-04-11
refactor built-in resolvers

omry:
Reorganizing built-in resolvers.

I am also thinking about an alternative layout with a single file per module:
```
oc/__init__.py # oc module
oc/dict.py    # oc.dict module 
```


odelalleau (approved):
Sounds good! (not sure we care about defining `__all__` but why not)

omry:
`__all__` there is to allow `from resolvers.oc import env` (if you just import in init and not use it's triggering a lint warning).
the alternative layout should probably eliminate the need for it.

Jasha10 (approved):
Looks good to me.

omry:
Switched to new layout, I think it's significantly cleaner.

PR #677, opened 2021-04-12
Add new resolver `oc.create`

### LINKED ISSUE #645: Add new resolver `oc.create`
**Motivation**

A new resolver `oc.create` would make it easier to dynamically generate config objects from lists and dictionaries.

For instance imagine you have the following JSON file `model.json` with the following content:
```json
{"name": "resnet", "n_units_per_layer": [100, 1000, 500]}
```

You want to load this JSON within an existing config. This could be achieved with:

```python
import json
from omegaconf import DictConfig, OmegaConf

# Define a custom resolver to load JSON files.

def load_json(path):
    with open(path) as f:
        return json.load(f)

OmegaConf.register_new_resolver("load_json", load_json)

# Combine it with `oc.create` to generate a DictConfig with the content of this file.

cfg = OmegaConf.create({"model": "${oc.create:${load_json:'model.json'}}"})
assert isinstance(cfg.model, DictConfig)
assert cfg.model.n_units_per_layer[1]  == 1000
```

If `oc.create` wasn't used, then `cfg.model` would be a plain dictionary that wouldn't support all DictConfig features (like accessing a field with the syntax `cfg.model.n_units_per_layer`)

**Alternatives**

It is possible for custom resolvers to explicitly call `OmegaConf.create(..., parent=_parent_)` to always generate config nodes. The objective of introducing `oc.create` is (i) to make this more straightforward and (ii) to add more flexibility, so that config nodes are only generated when explicitly requested (there are situations where keeping plain dicts / lists may be preferred).

**Additional context**
* In current master (d93905e) when a resolver outputs a dict/list, it is automatically converted into a DictConfig / ListConfig. However this may not always be a good thing and will be reverted when introducing `oc.create` (see discussion: https://github.com/omry/omegaconf/pull/606#issuecomment-800589117)

odelalleau:
Probably best reviewed commit by commit:
* abc60ae753b902dd945c6867dbf1487e53de3c35 makes it so that dictionaries and lists output by resolvers aren't converted into DictConfig and ListConfig automatically anymore (something we discussed a while back). It breaks a few things which is why it's not an independent PR by itself (tests marked as broken in this commit are fixed in the next one). It also removes the readonly flag from transient nodes generated when resolving interpolations (its main use was for containers and having it on other types of node was more a side effect than an important feature)
* f5fb0cc5044c8b196aca633409c2730102f7157c adds the `oc.create` resolver which makes it possible again to easily create DictConfigs and ListConfigs from dicts/lists
* 71b615e3343b1e525ad4e10255c23be519d76444 removes the readonly flag from containers created with `oc.create`, because I realized it would otherwise prevent people from merging configs easily. I made it a separate commit because there may be other alternatives (ex: making merge work with readonly configs?)

Fixes #645


tests/interpolation/test_custom_resolvers.py
```diff
@@ -345,44 +346,28 @@ def test_clear_cache(restore_resolvers: Any) -> None:
...
-    OmegaConf.register_new_resolver("dict", lambda: cfg)
+@mark.parametrize("readonly", [True, False, None])
+def test_resolver_output_dict(restore_resolvers: Any, readonly: bool) -> None:
+    some_dict = {"a": 0, "b": "${y}"}
+    OmegaConf.register_new_resolver("dict", lambda: some_dict)
     c = OmegaConf.create({"x": "${dict:}", "y": -1})
-    assert isinstance(c.x, DictConfig)
-    assert c.x == expected
-    assert dereference_node(c, "x")._get_flag("readonly")
+    c._set_flag("readonly", readonly)
```
omry, line 354: OmegaConf.set_readonly(c, readonly)

tests/interpolation/test_custom_resolvers.py
```diff
@@ -345,44 +346,28 @@ def test_clear_cache(restore_resolvers: Any) -> None:
...
-        ({"a": 0, "b": "${x.a}"}, {"a": 0, "b": 0}),
-        ({"a": 0, "b": "${.a}"}, {"a": 0, "b": 0}),
-        ({"a": "${..y}"}, {"a": -1}),
-    ],
-)
-def test_resolver_output_dict_to_dictconfig(
-    restore_resolvers: Any, cfg: Dict[str, Any], expected: Dict[str, Any]
-) -> None:
-    OmegaConf.register_new_resolver("dict", lambda: cfg)
+@mark.parametrize("readonly", [True, False, None])
```
omry, line 349: I think we can go with True and False only. I know that in other places we are also testing None but it's probably redundant in most of them.

tests/interpolation/test_custom_resolvers.py
```diff
@@ -345,44 +346,28 @@ def test_clear_cache(restore_resolvers: Any) -> None:
...
+def test_resolver_output_dict(restore_resolvers: Any, readonly: bool) -> None:
+    some_dict = {"a": 0, "b": "${y}"}
+    OmegaConf.register_new_resolver("dict", lambda: some_dict)
     c = OmegaConf.create({"x": "${dict:}", "y": -1})
-    assert isinstance(c.x, DictConfig)
-    assert c.x == expected
-    assert dereference_node(c, "x")._get_flag("readonly")
+    c._set_flag("readonly", readonly)
+    assert c.x == some_dict
```
omry, line 355:
```
assert isinstance(c.x, dict)
```

DictConfig will will also return True:
```
In [2]: OmegaConf.create({"a": 0, "b": "${y}"}) == {"a": 0, "b": "${y}"}
Out[2]: True
```

tests/interpolation/test_custom_resolvers.py
```diff
@@ -345,44 +346,28 @@ def test_clear_cache(restore_resolvers: Any) -> None:
...
+@mark.parametrize("readonly", [True, False, None])
+def test_resolver_output_list(restore_resolvers: Any, readonly: bool) -> None:
+    some_list = ["a", 0, "${y}"]
+    OmegaConf.register_new_resolver("list", lambda: some_list)
     c = OmegaConf.create({"x": "${list:}", "y": -1})
-    assert isinstance(c.x, ListConfig)
-    assert c.x == expected
-    assert dereference_node(c, "x")._get_flag("readonly")
+    c._set_flag("readonly", readonly)
+    assert c.x == some_list
```
omry, line 367: assert isinstance(c.x, list)

tests/interpolation/test_interpolation.py
```diff
@@ -402,39 +369,53 @@ def test_interpolation_type_validated_error(
...
+            ["a", "b"],
+            ListConfig,
+            id="list_type_mismatch",
+        ),
+        param(
+            MissingDict(dict=SI("${identity:{0: b, 1: d}}")),
+            "dict",
+            {0: "b", 1: "d"},
+            DictConfig,
+            id="dict_key_type_mismatch",
```
omry: confused by those. why are they not expecting list and dict objects?

omry (commented):
review for first commit in stack.

tests/interpolation/built_in_resolvers/test_dict.py
```diff
@@ -234,33 +234,18 @@ def test_dict_values_dictconfig_resolver_output(
...
-            lambda: {"a": 0, "b": 1},
-            [0, 1],
-            ["${y.a}", "${y.b}"],
-            id="plain_dict",
-        ),
-    ],
-)
-def test_dict_values_transient_interpolation(
+@mark.parametrize("dict_func", ["oc.dict.values", "oc.dict.keys"])
+def test_extract_from_dict_resolver_output(
```
omry, line 238: I don't understand what this function is testing from its name.
Is this new test related to the changes in this PR?

docs/source/usage.rst
```diff
@@ -432,14 +432,19 @@ The following example falls back to default passwords when ``DB_PASSWORD`` is no
 Decoding strings with interpolations
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Strings may be converted using ``oc.decode``:
+With ``oc.decode``, strings can be converted into their corresponding data types using the OmegaConf grammar.
+This grammar recognizes typical data types like ``bool``, ``int``, ``float``, ``dict`` and ``list``,
+e.g. ``"true"``, ``"1"``, ``"1e-3"``, ``"{a: b}"``, ``"[a, b, c]"``.
+It will also resolve interpolations like ``"${foo}"``, returning the corresponding value of the node.
```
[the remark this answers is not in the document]
odelalleau: Yeah, I'm thinking of the string `"${foo}"` being given as input to `oc.decode`. It's best seen in the code example below when we set
```python
os.environ["DB_TIMEOUT"] = "${.port}"
```

However, it's tricky to explain succinctly here. Maybe I just shouldn't mention it?

docs/source/usage.rst
```diff
@@ -432,14 +432,19 @@ The following example falls back to default passwords when ``DB_PASSWORD`` is no
 Decoding strings with interpolations
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Strings may be converted using ``oc.decode``:
+With ``oc.decode``, strings can be converted into their corresponding data types using the OmegaConf grammar.
+This grammar recognizes typical data types like ``bool``, ``int``, ``float``, ``dict`` and ``list``,
+e.g. ``"true"``, ``"1"``, ``"1e-3"``, ``"{a: b}"``, ``"[a, b, c]"``.
```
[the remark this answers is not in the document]
odelalleau: True -- I'll take a note to refer to it once we have this doc.

docs/source/usage.rst
```diff
@@ -432,14 +432,19 @@ The following example falls back to default passwords when ``DB_PASSWORD`` is no
...
-- Dictionaries and lists (e.g., ``"{a: b}"``, ``"[a, b, c]"``)  are returned as transient config nodes (DictConfig and ListConfig)
-- Interpolations (e.g., ``"${foo}"``) are automatically resolved
-- ``None`` is the only valid non-string input to ``oc.decode`` (returning ``None`` in that case)
+Note that:
 
-This can be useful for instance to parse environment variables:
+- When providing as input to ``oc.decode`` a string that is meant to be decoded into another string, in general
+  the input string should be quoted (since only a subset of characters are allowed by the grammar in unquoted
+  strings). For instance, a proper string interpolation could be: ``"'Hi! My name is: ${name}'"`` (with extra quotes).
+- ``None`` (written as ``null`` in the grammar) is the only valid non-string input to ``oc.decode`` (returning ``None`` in that case)
```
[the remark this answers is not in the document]
odelalleau: Right, I can see how this can be confusing. I'm talking about the string being provided as input to `oc.decode`. When you write:
```python
"c": "${oc.deprecated:a, 'message: ${msg}'}",
```
The string provided to `oc.deprecated` isn't `"message: ${msg}"`, it is `"message: E.T. call home"`.

In general strings provided as input to `oc.decode` are much more likely to come from resolvers than to be written explicitly in the config (because if you can write it explicitly then there is no need for `oc.decode`). That's why my example also uses `oc.env`.

I'll need to think a bit more about it to try and figure out how to make it understandable in the doc. Suggestions welcome ;)

tests/interpolation/test_interpolation.py
odelalleau: I took a shortcut here as I was planning to keep the DictConfig / ListConfig but only after introducing `oc.create`, so in this commit the test is partially broken (I commented out the type check) and it's fixed in the next commit. Sorry it's a bit confusing.

docs/source/usage.rst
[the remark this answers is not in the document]
omry: yeah, I think we can omit it.

docs/source/usage.rst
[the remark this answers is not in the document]
omry: I suggest we don't mention it.

tests/interpolation/built_in_resolvers/test_dict.py
odelalleau: The name means "test what happens when you try to extract [keys / values] from a dictionary which is the output of a resolver".

The relation with this PR is that before this commit, in this test `y` would be a DictConfig and thus you would be able to do `${oc.dict.keys:y}` or `${oc.dict.values:y}`.
But now `y` is a plain Python dictionary which isn't supported by `oc.dict.{keys,values}`.

This is essentially the replacement for the previous test `test_dict_values_transient_interpolation[plain_dict]` (which was poorly named because the interpolation isn't transient anymore, it was legacy from an earlier version)

The other previous test `test_dict_values_transient_interpolation[dictconfig_with_parent]` is gone because it seemed too redundant with `test_dict_values_dictconfig_resolver_output[basic]` to be worth keeping it.

docs/source/usage.rst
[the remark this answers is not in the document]
odelalleau: Removed in 01ddb95

odelalleau:
I rebased on top of master -- the first commit that you had already reviewed hasn't changed much (I just had to apply the doc changes to the new location)

docs/source/usage.rst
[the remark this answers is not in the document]
odelalleau: Also removed in 01ddb95

omegaconf/base.py
```diff
@@ -552,31 +552,36 @@ def _wrap_interpolation_result(
...
-                ref_type=value._metadata.ref_type,
-            )
-        except (KeyValidationError, ValidationError) as e:
-            if throw_on_resolution_failure:
-                self._format_and_raise(
-                    key=key,
+
+        if is_primitive_type(type(resolved)):
+            # Primitive types get wrapped using the same logic as when setting the
+            # value of a node (i.e., through `_node_wrap()`).
```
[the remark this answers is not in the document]
odelalleau: Done in b3cc6ab

omegaconf/base.py
```diff
@@ -552,31 +552,36 @@ def _wrap_interpolation_result(
...
-            return None
-        # Since we created a new node on the fly, future changes to this node are
-        # likely to be lost. We thus set the "readonly" flag to `True` to reduce
-        # the risk of accidental modifications.
-        wrapped._set_flag("readonly", True)
+            except ValidationError:  # pragma: no cover
+                # This is not supposed to happen because primitive types that must
+                # be wrapped should have already been validated inside
+                # `_validate_and_convert_interpolation_result()`.
+                assert False
```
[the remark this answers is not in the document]
odelalleau: Done in 348e2ae

omegaconf/base.py
```diff
@@ -552,31 +552,36 @@ def _wrap_interpolation_result(
...
+        else:
+            # Other objects get wrapped into an `AnyNode` with `allow_objects` set
+            # to True.
+            wrapped = AnyNode(value=None, key=key, parent=parent)
+            wrapped._set_flag("allow_objects", True)
+            with flag_override(wrapped, "readonly", False):
+                wrapped._set_value(resolved)
```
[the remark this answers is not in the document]
odelalleau: Done in 7e514e4 (generic ability for flags to be set in the constructor) and c372a58 (the corresponding simplification of this code block)

tests/interpolation/test_custom_resolvers.py
odelalleau: Done in ac26495

tests/interpolation/test_custom_resolvers.py
odelalleau: Done in 55f8041

tests/interpolation/test_custom_resolvers.py
odelalleau: Done in a5978e9

tests/interpolation/test_custom_resolvers.py
odelalleau: Also done in a5978e9

docs/source/custom_resolvers.rst
```diff
@@ -213,6 +200,29 @@ The following example falls back to default passwords when ``DB_PASSWORD`` is no
...
+    >>> cfg = OmegaConf.create(
+    ...     {
+    ...         "model": "${oc.create:${oc.decode:${oc.env:MODEL}}}",
+    ...     }
+    ... )
+    >>> os.environ["MODEL"] = "{name: my_model, layer_size: [100, 200]}"
+    >>> show(cfg.model.layer_size)
+    type: ListConfig, value: [100, 200]
+
+
```
omry, line 225: This seems a bit much (using 3 resolvers to explain 1).
Maybe something like this?
```python
cfg = OmegaConf.create(
    {
        "plain_dict": "${identity:{a:10}}",
        "dict_config": "${oc.create:{a:10}}",
        "dict_config_from_env": "${oc.create:${oc.env:YAML_ENV}}",
    }
)

os.environ["YAML_ENV"] = "A: 10\nb: 20\nC: ${.A}" 
show(cfg.plain_dict)
show(cfg.dict_config)
show(cfg.dict_config_from_env)
# interpolations works because this is a DictConfig
print(cfg.dict_config_from_env.C)
```

docs/source/custom_resolvers.rst
```diff
@@ -245,14 +255,18 @@ It takes two parameters:
...
+e.g. ``"true"``, ``"1"``, ``"1e-3"``, ``"{a: b}"``, ``"[a, b, c]"``.
+
+Note that:
 
-- Primitive values (e.g., ``"true"``, ``"1"``, ``"1e-3"``) are automatically converted to their corresponding type (bool, int, float)
-- Dictionaries and lists (e.g., ``"{a: b}"``, ``"[a, b, c]"``)  are returned as transient config nodes (DictConfig and ListConfig)
-- Interpolations (e.g., ``"${foo}"``) are automatically resolved
-- ``None`` is the only valid non-string input to ``oc.decode`` (returning ``None`` in that case)
+- When providing as input to ``oc.decode`` a string that is meant to be decoded into another string, in general
+  the input string should be quoted (since only a subset of characters are allowed by the grammar in unquoted
```
omry, line 265: This is a bit cumbersome, maybe:
```suggestion
- Input strings provided to `oc.decode` should be quoted because only a subset of the characters is allowed in unquoted strings
```

(make sure it formats properly).

omry: I also think it would be good to show decoding of a literal input string as well (not the result of another resolver).

docs/source/usage.rst
```diff
@@ -22,6 +22,8 @@ Just pip install::
 
 OmegaConf requires Python 3.6 and newer.
 
+.. _creating:
+
```
omry, line 26: rst stinks.

tests/interpolation/built_in_resolvers/test_dict.py
omry, line 238:
```suggestion
def test_extract_from_plain_dict_unsupported(
```
Also, I think we have decided that the only supported input the string keys.
dunno if it makes sense to test something so intricate now that the behavior is so simple. I vote to just remove this test as well.

tests/interpolation/test_custom_resolvers.py
```diff
@@ -345,44 +346,30 @@ def test_clear_cache(restore_resolvers: Any) -> None:
...
     c = OmegaConf.create({"x": "${list:}", "y": -1})
-    assert isinstance(c.x, ListConfig)
-    assert c.x == expected
-    assert dereference_node(c, "x")._get_flag("readonly")
+    OmegaConf.set_readonly(c, readonly)
+    assert isinstance(c.x, list)
+    assert c.x == some_list
+    x_node = dereference_node(c, "x")
+    assert isinstance(x_node, AnyNode)
+    assert x_node._get_flag("allow_objects")
```
omry, line 372: I think those two tests can be combined into one test using identity to pass through dict and list from the parameterization.

omry (changes requested):

docs/source/custom_resolvers.rst
odelalleau: Done in 8107676 (with a few small changes)

docs/source/custom_resolvers.rst
odelalleau: > This is a bit cumbersome, maybe:
> (...)

Done in 8ea1549 (with minor changes)

> I also think it would be good to show decoding of a literal input string as well (not the result of another resolver).

I'm a bit reluctant to do so because:
* It would break a bit the "flow" of the current example
* I can't think of a situation where one would want to use `oc.decode` on a literal input string

Do you have a specific suggestion in mind?

tests/interpolation/built_in_resolvers/test_dict.py
odelalleau: I'm not particularly attached to this test => removed in 9050018
(note: it was still using a string key as input, it's just that the corresponding key was a plain dict -- I think using a resolver is the only way to get a plain dict in a config)

docs/source/custom_resolvers.rst
omry: No, just a suggestion. we can leave it like that.

tests/interpolation/test_custom_resolvers.py
odelalleau: Yes, good point, done in e4503d3 (I didn't use `identity` though because it would be a bit more cumbersome and harder to read IMO, it's easier to register a resolver that outputs the dict/list we want)

omry (approved):
lgtm

docs/source/usage.rst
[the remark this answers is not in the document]
odelalleau: Just a note that this was addressed in 243712c496e48b6639fce5e5b768f5ae2f41dc1a

PR #679, opened 2021-04-15
oc.deprecated

### LINKED ISSUE #681: Add oc.deprecated to enable deprecating config nodes

omry:
1. Refactor docs, moving resolvers to a dedicated page.
2. Adding support for oc.deprecated.

Most of changed lines here are from the docs refactor.
Try to go over it anyway and let me know if you have any feedback.

Closes #681.

### DIFF
--- a/docs/source/custom_resolvers.rst
+++ b/docs/source/custom_resolvers.rst
@@ -2,6 +2,9 @@
 
     from omegaconf import OmegaConf, DictConfig
     import os
+    import pytest
+    os.environ['USER'] = 'omry'
+
     def show(x):
       print(f"type: {type(x).__name__}, value: {repr(x)}")
 
@@ -131,14 +134,14 @@ the inputs themselves:
 
 Custom interpolations can also receive the following special parameters:
 
-- ``_parent_`` : the parent node of an interpolation.
+- ``_parent_``: the parent node of an interpolation.
 - ``_root_``: The config root.
 
 This can be achieved by adding the special parameters to the resolver signature.
-Note that special parameters must be defined as named keywords (after the `*`):
+Note that special parameters must be defined as named keywords (after the `*`).
 
-In this example, we use ``_parent_`` to implement a sum function that defaults to 0 if the node does not exist.
-(In contrast to the sum we defined earlier where accessing an invalid key, e.g. ``"a_plus_z": ${sum:${a}, ${z}}`` will result in an error).
+In the example below, we use ``_parent_`` to implement a sum function that defaults to 0 if the node does not exist.
+(In contrast to the sum we defined earlier where accessing an invalid key, e.g. ``"a_plus_z": ${sum:${a}, ${z}}`` would result in an error).
 
 .. doctest::
 
@@ -189,6 +192,53 @@ In such a case, the default value is converted to a string using ``str(default)`
 
 The following example falls back to default passwords when ``DB_PASSWORD`` is not defined:
 
+.. doctest::
+
+    >>> cfg = OmegaConf.create(
+    ...     {
+    ...         "database": {
+    ...             "password1": "${oc.env:DB_PASSWORD,password}",
+    ...             "password2": "${oc.env:DB_PASSWORD,12345}",
+    ...             "password3": "${oc.env:DB_PASSWORD,null}",
+    ...         },
+    ...     }
+    ... )
+    >>> # default is already a string
+    >>> show(cfg.database.password1)
+    type: str, value: 'password'
+    >>> # default is converted to a string automatically
+    >>> show(cfg.database.password2)
+    type: str, value: '12345'
+    >>> # unless it's None
+    >>> show(cfg.database.password3)
+    type: NoneType, value: None
+
+.. _oc.deprecated:
+
+oc.deprecated
+^^^^^^^^^^^^^
+``oc.deprecated`` enables you to deprecate a config node.
+It takes two parameters:
+
+- ``key``: An interpolation key representing the new key you are migrating to. This parameter is required.
+- ``message``: A message to use as the warning when the config node is being accessed. The default message is
+  ``'$OLD_KEY' is deprecated. Change your code and config to use '$NEW_KEY'``.
+
+.. doctest::
+
+    >>> conf = OmegaConf.create({
+    ...   "rusty_key": "${oc.deprecated:shiny_key}", 
+    ...   "custom_msg": "${oc.deprecated:shiny_key, 'Use $NEW_KEY'}",
+    ...   "shiny_key": 10
+    ... })
+    >>> # Accessing rusty_key will issue a deprecation warning 
+    >>> # and return the new value automatically
+    >>> warning = "'rusty_key' is deprecated. Change your" \
+    ...           " code and config to use 'shiny_key'"
+    >>> with pytest.warns(UserWarning, match=warning):
+    ...   assert conf.rusty_key == 10
+    >>> with pytest.warns(UserWarning, match="Use shiny_key"):
+    ...   assert conf.custom_msg == 10
 
 .. _oc.decode:
 
--- a/docs/source/custom_resolvers.rst
+++ b/docs/source/custom_resolvers.rst
@@ -1,0 +1,262 @@
+.. testsetup:: *
+
+    from omegaconf import OmegaConf, DictConfig
+    import os
+    def show(x):
+      print(f"type: {type(x).__name__}, value: {repr(x)}")
+
+.. _custom_resolvers:
+
+Custom resolvers
+----------------
+
+You can add additional interpolation types by registering custom resolvers with ``OmegaConf.register_new_resolver()``:
+
+.. code-block:: python
+
+    def register_new_resolver(
+        name: str,
+        resolver: Resolver,
+        *,
+        replace: bool = False,
+        use_cache: bool = False,
+    ) -> None
+
+Attempting to register the same resolver twice will raise a ``ValueError`` unless using ``replace=True``.
+
+The example below creates a resolver that adds 10 to the given value.
+
+.. doctest::
+
+    >>> OmegaConf.register_new_resolver("plus_10", lambda x: x + 10)
+    >>> c = OmegaConf.create({'key': '${plus_10:990}'})
+    >>> c.key
+    1000
+
+Custom resolvers support variadic argument lists in the form of a comma separated list of zero or more values.
+Whitespaces are stripped from both ends of each value ("foo,bar" is the same as "foo, bar ").
+You can use literal commas and spaces anywhere by escaping (:code:`\,` and :code:`\ `), or
+simply use quotes to bypass character limitations in strings.
+
+.. doctest::
+
+    >>> OmegaConf.register_new_resolver("concat", lambda x, y: x+y)
+    >>> c = OmegaConf.create({
+    ...     'key1': '${concat:Hello,World}',
+    ...     'key_trimmed': '${concat:Hello , World}',
+    ...     'escape_whitespace': '${concat:Hello,\ World}',
+    ...     'quoted': '${concat:"Hello,", " World"}',
+    ... })
+    >>> c.key1
+    'HelloWorld'
+    >>> c.key_trimmed
+    'HelloWorld'
+    >>> c.escape_whitespace
+    'Hello World'
+    >>> c.quoted
+    'Hello, World'
+
+
+Custom resolvers can return lists or dictionaries, that are automatically converted into DictConfig and ListConfig:
+
+.. doctest::
+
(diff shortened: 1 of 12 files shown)


omegaconf/resolvers/oc/__init__.py
```diff
@@ -46,8 +49,42 @@ def decode(expr: Optional[str], _parent_: Container) -> Any:
     return _get_value(val)
 
 
+def deprecated(
+    key: str,
+    message: str = "'$KEY' is deprecated. Change your code and config to use '$NEW_KEY'",
```
odelalleau, line 54: Out of curiosity (no need to change it) is there a reason to prefer a string Template vs. regular string formatting, i.e.:
```python
    message: str = "'{old_key}' is deprecated. Change your code and config to use '{new_key}'",
```

By the way I have a slight preference to use `$OLD_KEY` instead of `$KEY` because here `key` is the new key (but one naturally tends to think that `key` is mapped to `$KEY` because they have the same name)

tests/interpolation/built_in_resolvers/test_oc_deprecated.py
```diff
@@ -0,0 +1,107 @@
...
+            1,
+            "'b' is deprecated. Change your code and config to use 'a'",
+            id="list_value",
+        ),
+    ],
+)
+def test_deprecated(
+    cfg: Any, key: str, expected_value: Any, expected_warning: str
+) -> None:
+    cfg = _ensure_container(cfg)
```
odelalleau, line 72: Just for my own understanding, why not `OmegaConf.create()`?

docs/source/custom_resolvers.rst
```diff
@@ -0,0 +1,292 @@
...
+.. _oc.deprecated:
+
+oc.deprecated
+^^^^^^^^^^^^^
+``oc.deprecated`` enables you to deprecate a config node.
+It takes two parameter:
+- key: An interpolation key representing the new key you are migrating to. This parameter is required.
+- message: A message to use as the warning when the config node is being accessed. The default message is
+``'$KEY' is deprecated. Change your code and config to use '$NEW_KEY'``. Note that $KEY and $NEW_KEY 
+do not use interpolation syntax.
```
odelalleau, line 205:
```suggestion
It takes two parameters:

- ``key``: An interpolation key representing the new key you are migrating to. This parameter is required.
- ``message``: An optional message to use as the warning when the config node is being accessed. The default message is
  ``'$KEY' is deprecated. Change your code and config to use '$NEW_KEY'``. Note that ``$KEY`` and ``$NEW_KEY``
  do not use interpolation syntax.
```

odelalleau (changes requested):
Bunch of small things

docs/source/custom_resolvers.rst
```diff
@@ -0,0 +1,292 @@
...
+    >>> assert c.cached_2 == c.cached_1 == 122
+    >>> # same string literal "${uncached}" => same value
+    >>> assert c.cached_3 == c.cached_3 == 1192
+
+
+Custom interpolations can also receive the following special parameters:
+
+- ``_parent_`` : the parent node of an interpolation.
+- ``_root_``: The config root.
+
```
Jasha10, line 139: Is the new `_node_` special parameter going to become a public API? This would be the place to mention it :)

Jasha10: Also, if it will be public, we should add a test or two to `tests/interpolation/test_custom_resolvers.py`.

docs/source/custom_resolvers.rst
```diff
@@ -0,0 +1,292 @@
...
+    ...   "rusty_key" : "${oc.deprecated:shiny_key}", 
+    ...   "rusty_key_custom_msg" : "${oc.deprecated:shiny_key, 'Why you no use $NEW_KEY?'}",
+    ...   "shiny_key": 10
+    ... })
+    >>> # Accessing rusty_key will issue a deprecation warning 
+    >>> # and return the new value automatically
+    >>> warning = "'rusty_key' is deprecated. Change your" \
+    ...           " code and config to use 'shiny_key'"
+    >>> with pytest.warns(UserWarning, match=warning):
+    ...   assert conf.rusty_key == 10
```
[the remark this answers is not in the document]
omry: It would have been. unfortunately the warning is showing up when building the and is not captured.
This is not only making it harder to read, but also make the doc building depends on pytest, which could cause issues when building the docs.

docs/source/custom_resolvers.rst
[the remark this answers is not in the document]
omry: done in 1f32b11

omry: assuming this is only about using \`\` around key and message, done.

docs/source/custom_resolvers.rst
```diff
@@ -190,6 +193,33 @@ In such a case, the default value is converted to a string using ``str(default)`
...
+- key: An interpolation key representing the new key you are migrating to. This parameter is required.
+- message: A message to use as the warning when the config node is being accessed. The default message is
+``'$KEY' is deprecated. Change your code and config to use '$NEW_KEY'``. Note that $KEY and $NEW_KEY 
+do not use interpolation syntax.
+
+.. doctest::
+
+    >>> conf = OmegaConf.create({
+    ...   "rusty_key" : "${oc.deprecated:shiny_key}", 
+    ...   "rusty_key_custom_msg" : "${oc.deprecated:shiny_key, 'Why you no use $NEW_KEY?'}",
```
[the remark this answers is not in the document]
omry: https://imgflip.com/memegenerator/5214848/Why-you-no

Fixed both in f4aa8f2.

docs/source/usage.rst
```diff
@@ -383,272 +383,28 @@ Interpolated nodes can be any node in the config, not just leaf nodes:
...
-    ...             "a_plus_z": "${sum2:a,z}",
-    ...         },
-    ...     }
-    ... )
-    >>> cfg.node.a_plus_b
-    3
-    >>> cfg.node.a_plus_z
-    1
+* :ref:`oc.decode`: Parsing an input string using interpolation grammar
+* :ref:`oc.env`: Accessing environment variables.
```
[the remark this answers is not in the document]
omry: They are ordered alphabetically within each namespace.

docs/source/usage.rst
```diff
@@ -383,272 +383,28 @@ Interpolated nodes can be any node in the config, not just leaf nodes:
...
-    >>> c.escape_whitespace
-    'Hello World'
-    >>> c.quoted
-    'Hello, World'
-
-
-Custom resolvers can return lists or dictionaries, that are automatically converted into DictConfig and ListConfig:
+Resolvers
+^^^^^^^^^
+You can add additional interpolation types by registering resolvers using ``OmegaConf.register_new_resolver()``.
```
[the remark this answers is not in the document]
omry: fixed.

omegaconf/base.py
```diff
@@ -633,7 +633,17 @@ def _evaluate_custom_resolver(
         resolver = OmegaConf._get_resolver(inter_type)
         if resolver is not None:
             root_node = self._get_root()
-            return resolver(root_node, self, inter_args, inter_args_str)
+            node = None
+            if key is not None:
+                node = self._get_node(key, validate_access=True)
```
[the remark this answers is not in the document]
omry: Feel free to follow up with a PR improving it.

omegaconf/resolvers/oc/__init__.py
omry: 1. no strong reason. I used string Template when formatting exception messages and this was similar.
In terms of user facing syntax, $NEW_KEY feels cleaner to me than {new_key}. it's also more explicit and does not establish a protocol via local variable names in some function. It also does not potentially leak the value of other local variables).
2. changed to OLD_KEY.

omegaconf/resolvers/oc/__init__.py
```diff
@@ -46,8 +49,42 @@ def decode(expr: Optional[str], _parent_: Container) -> Any:
...
+    if not isinstance(message, str):
+        raise ValueError(
+            f"oc.deprecated: interpolation message type is not a string ({type(message).__name__})"
+        )
+    assert _node_ is not None
+    full_key = _node_._get_full_key(key=None)
+    target_node = select_node(_parent_, key, absolute_key=True)
+    if target_node is None:
+        raise ConfigKeyError(
+            f"In oc.deprecate resolver at '{full_key}': Key not found: '{key}'"
```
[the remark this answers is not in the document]
omry: maybe as a followup. I prefer having the custom message here indicating where this is coming from anyway.

tests/interpolation/built_in_resolvers/test_oc_deprecated.py
omry: In this location they are identical.
generally, OmegaConf.create will always re-wrap an object. _ensure_container will not touch DictConfig or ListConfig.

docs/source/usage.rst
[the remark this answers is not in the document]
odelalleau: It doesn't look fixed in current version but it's definitely not a big deal so feel free to just resolve :)

docs/source/usage.rst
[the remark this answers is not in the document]
odelalleau: Ah, I see, the "namespaced" resolvers would be at the bottom then. Makes sense.

odelalleau (changes requested):

omegaconf/resolvers/oc/__init__.py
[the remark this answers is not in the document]
odelalleau: Ok, probably that when an exception is raised in a resolver we should always provide the resolver name (this doesn't seem to be the case right now). Going to resolve this for now.

docs/source/custom_resolvers.rst
```diff
@@ -0,0 +1,311 @@
...
+    >>> # unless it's None
+    >>> show(cfg.database.password3)
+    type: NoneType, value: None
+
+.. _oc.deprecated:
+
+oc.deprecated
+^^^^^^^^^^^^^
+``oc.deprecated`` enables you to deprecate a config node.
+It takes two parameter:
```
odelalleau, line 221:
```suggestion
It takes two parameters:

```

docs/source/custom_resolvers.rst
```diff
@@ -0,0 +1,311 @@
...
+- ``key``: An interpolation key representing the new key you are migrating to. This parameter is required.
+- ``message``: A message to use as the warning when the config node is being accessed. The default message is
```
odelalleau, line 223:
```suggestion
- ``message``: An optional message to use as the warning when the config node is being accessed. The default message is
```

odelalleau (changes requested):
The missing changes I had suggested

docs/source/custom_resolvers.rst
odelalleau: It was more than that => I created another diff

docs/source/usage.rst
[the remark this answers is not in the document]
odelalleau: Oh never mind, it's because it's in two places in the doc... resolving.

docs/source/custom_resolvers.rst
```diff
@@ -0,0 +1,292 @@
...
+Custom interpolations can also receive the following special parameters:
+
+- ``_parent_`` : the parent node of an interpolation.
+- ``_root_``: The config root.
+
+This can be achieved by adding the special parameters to the resolver signature.
+Note that special parameters must be defined as named keywords (after the `*`):
+
+In this example, we use ``_parent_`` to implement a sum function that defaults to 0 if the node does not exist.
+(In contrast to the sum we defined earlier where accessing an invalid key, e.g. ``"a_plus_z": ${sum:${a}, ${z}}`` will result in an error).
```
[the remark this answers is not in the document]
omry: > Ok, maybe we should actually check it and raise an exception if we want to really enforce it.

I don't care about enforcing it, this is just a gotcha that people will hit if they try to use this with a parameter that has a default value.
I am not even sure this is detectable from Python.
Feel free to send a followup PR to improve this though.

docs/source/custom_resolvers.rst
omry: It's not optional but since it has a default value you don't need to worry about it.

omry:
I think I got everything.

docs/source/custom_resolvers.rst
odelalleau: It's optional in the sense that you don't need to provide it. But anyway, not a big deal => resolving.

odelalleau (approved):
All good! There was a doc error due to my "fix" (which fixed something but broke another) => I pushed the fix directly, should be good to merge once all green

Issue #674, opened 2021-04-08
`OmegaConf.select()` with default value doesn't work with keys "", ".", "..", etc.


odelalleau:
**Describe the bug**

The following code runs without error, but the lines marked with `# !!!` should have failed:
```python
cfg = OmegaConf.create({"x": {}})

assert OmegaConf.select(cfg, "") is cfg
assert OmegaConf.select(cfg, "", default=0) == 0  # !!!

assert OmegaConf.select(cfg.x, ".") is cfg.x
assert OmegaConf.select(cfg.x, ".", default=0) == 0  # !!!

assert OmegaConf.select(cfg.x, "..") is cfg
assert OmegaConf.select(cfg.x, "..", default=0) == 0  # !!!
```

**Expected behavior**

The lines marked with `# !!!` should have instead returned the same object as the line just above.

**Additional context**

I haven't looked closely at it but I suspect this is caused by the `_last_key not in _root` check triggering the return of the default value.

ppwwyyxx:
I encountered a similar issue:
```python
cfg = OmegaConf.create({"x": {}})

assert OmegaConf.select(cfg, "y", default=0) == 0  # works
assert OmegaConf.select(cfg, "y.z", default=0) == 0   # fails?
```

Not sure if the failure is expected or not. From a user's perspective it seems natural to allow the above usage.

omry:
Smells like the same issue. What version is this against?

ppwwyyxx:
I'm on `2.1.0dev22` from pypi

PR #683, opened 2021-04-16
678 select

### LINKED ISSUE #678: OmegaConf.select raises an exception when selecting an invalid key in some cases
#444 fixed an assertion error but the fix is inconsistent:
Generally, the contact of OmegaConf.select is that it returns None if the field is not found.
This fix is causing inconsistent behavior:
```python
In [14]: cfg
Out[14]: {'a': 10, 'b': None}

In [15]: print(OmegaConf.select(cfg, "foo.bar"))
None

In [16]: print(OmegaConf.select(cfg, "b.bar"))
...
ConfigKeyError: Error trying to access b.bar: node `b` is not a container and thus cannot contain `bar`
    full_key: b.bar
    object_type=dict
```

I can see the value of raising an exception instead of returning None, but:
1. it should be optional to maintain backward compatibility (default to return None).
2. It should be consistent (always or never).

Adding support for it is potentially beyond the scope of fixing this bug.

omry:
Followup to discussion 678. easier to review individual commits.
1. eliminating redundant tests.
2. Change exception raised when trying to drill into a non-container with select from ConfigKeyError to ConfigTypeError.
3. Change OmegaConf.select to return None if user attempts to drill into a non existing node (into a ValueNode or into None).

Closes #678

### DIFF
--- a/news/678.bugfix
+++ b/news/678.bugfix
@@ -1,0 +1,1 @@
+OmegaConf.select now returns None when attempting to select a child of a value or None node
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -611,10 +611,6 @@ class Container(Node):
             raise InterpolationToMissingValueError(
                 f"MissingMandatoryValue while resolving interpolation: {exc}"
             ).with_traceback(sys.exc_info()[2])
-        except ConfigKeyError as exc:
-            raise InterpolationKeyError(
-                f"ConfigKeyError while resolving interpolation: {exc}"
-            ).with_traceback(sys.exc_info()[2])
 
         if parent is None or value is None:
             raise InterpolationKeyError(f"Interpolation key '{inter_key}' not found")
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -19,6 +19,7 @@ from ._utils import (
 )
 from .errors import (
     ConfigKeyError,
+    ConfigTypeError,
     InterpolationKeyError,
     InterpolationResolutionError,
     InterpolationToMissingValueError,
@@ -399,7 +400,7 @@ class Container(Node):
             if ret is not None and not isinstance(ret, Container):
                 parent_key = ".".join(split[0 : i + 1])
                 child_key = split[i + 1]
-                raise ConfigKeyError(
+                raise ConfigTypeError(
                     f"Error trying to access {key}: node `{parent_key}` "
                     f"is not a container and thus cannot contain `{child_key}`"
                 )
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -280,9 +280,9 @@ params = [
         Expected(
             create=lambda: OmegaConf.create({"foo": "${int.missing}", "int": 0}),
             op=lambda cfg: getattr(cfg, "foo"),
-            exception_type=InterpolationKeyError,
+            exception_type=InterpolationResolutionError,
             msg=(
-                "ConfigKeyError while resolving interpolation: Error trying to access int.missing: "
+                "ConfigTypeError raised while resolving interpolation: Error trying to access int.missing: "
                 "node `int` is not a container and thus cannot contain `missing`"
             ),
             key="foo",
--- a/omegaconf/_impl.py
+++ b/omegaconf/_impl.py
@@ -1,7 +1,7 @@
 from typing import Any
 
 from omegaconf import MISSING, Container, DictConfig, ListConfig, Node, ValueNode
-from omegaconf.errors import ConfigKeyError, InterpolationToMissingValueError
+from omegaconf.errors import ConfigTypeError, InterpolationToMissingValueError
 
 from ._utils import _DEFAULT_MARKER_, _get_value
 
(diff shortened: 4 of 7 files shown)


tests/test_errors.py
```diff
@@ -280,9 +280,9 @@ def finalize(self, cfg: Any) -> None:
         Expected(
             create=lambda: OmegaConf.create({"foo": "${int.missing}", "int": 0}),
             op=lambda cfg: getattr(cfg, "foo"),
-            exception_type=InterpolationKeyError,
+            exception_type=InterpolationResolutionError,
             msg=(
-                "ConfigKeyError while resolving interpolation: Error trying to access int.missing: "
+                "ConfigTypeError raised while resolving interpolation: Error trying to access int.missing: "
```
odelalleau, line 285: In base.py there's this try/except block that was triggering the previous exception => the second except should be removed:
```python
        try:
            parent, last_key, value = root_node._select_impl(
                inter_key,
                throw_on_missing=True,
                throw_on_resolution_failure=True,
                memo=memo,
            )
        except MissingMandatoryValue as exc:
            raise InterpolationToMissingValueError(
                f"MissingMandatoryValue while resolving interpolation: {exc}"
            ).with_traceback(sys.exc_info()[2])
        except ConfigKeyError as exc:
            raise InterpolationKeyError(
                f"ConfigKeyError while resolving interpolation: {exc}"
            ).with_traceback(sys.exc_info()[2])
```

omegaconf/base.py
```diff
@@ -399,7 +400,7 @@ def _select_impl(
             if ret is not None and not isinstance(ret, Container):
                 parent_key = ".".join(split[0 : i + 1])
                 child_key = split[i + 1]
-                raise ConfigKeyError(
+                raise ConfigTypeError(
```
odelalleau, line 403: Not going to push for it, but just a suggestion to restrict this new feature to `AnyNodes` that are `None` (because it still doesn't make much sense to me to let people try to select into other nodes):
* Check here if `isinstance(ret, AnyNode) and ret._is_none()` and if it is, return something that will make `select()` return `None` (or the default value). It may be worth looking into #674 at the same time since it will probably require some changes regarding when to return the default value, which might also mean changing what to return here.
* Otherwise, raise the exception (and don't catch it in the caller)

odelalleau (changes requested):
Besides the comment below:
* News item?
* Maybe also consider fixing the case with a None container in this PR: `{"x": "${y.z}", "y": DictConfig(None)}` (ok too to leave it for another PR if you prefer)

omegaconf/base.py
omry: The contact of OmegaConf.select is that it returns None if the selection cannot be performed.
It makes it harder to use if it sometimes raises an exception due to subtle reasons.

Hydra is using OmegaConf.select to determine if a config node exists in a path. raising an exception there makes the API significantly more cumbersome to use.
We can potentially introduce a specific API to do that in the future (but certainly not for 2.1), but in general I think using select for it is very reasonable.

tests/test_errors.py
omry: Yes, this is why the code coverage failed.

omry:
> Besides the comment below:
> 
> * News item?
> * Maybe also consider fixing the case with a None container in this PR: `{"x": "${y.z}", "y": DictConfig(None)}` (ok too to leave it for another PR if you prefer)

1. Added news item.
2. Do you mean:
```
OmegaConf.create({"x": "${y.z}", "y": DictConfig(None)}).x
```
Raising the wrong exception?
this seems to be an existing (low priority) bug that is not related to this one.

File a separate issue?

omegaconf/base.py
odelalleau: Got it

odelalleau:
> File a separate issue?

New issue: #685

news/678.bugfix
```diff
@@ -0,0 +1 @@
+OmegaConf.select now returns None is attempting to select a child of a value node
```
odelalleau, line 1:
```suggestion
OmegaConf.select now returns None when attempting to select a child of a value node
```

odelalleau (changes requested):
Good to merge after fixing this small typo!

omry:
> > File a separate issue?
> 
> New issue: #685

Actually not the same issue I was thinking about.
Run this:
```
OmegaConf.create({"x": "${y.z}", "y": DictConfig(None)}).x
```

no OmegaConf.select involved.

odelalleau:
> > > File a separate issue?
> > 
> > 
> > New issue: #685
> 
> Actually not the same issue I was thinking about.
> Run this:
> 
> ```
> OmegaConf.create({"x": "${y.z}", "y": DictConfig(None)}).x
> ```
> 
> no OmegaConf.select involved.

Hopefully both can be fixed at the same time :)

omry:
Looks like the fix accidentally fixed bug:
```python
OmegaConf.create({"x": "${y.z}", "y": DictConfig(None)}).x
...
InterpolationKeyError: Interpolation key 'y.z' not found
    full_key: x
    object_type=dict

```

odelalleau:
> Looks like the fix accidentally fixed the other bug:

Nice -- that's what I was hoping for since both were raising the same exception.

PR #684, opened 2021-04-16
Optimization to avoid re-fetching the node on each resolver call

odelalleau:
Note that the node is now guaranteed to always be available.

This is my suggestion from https://github.com/omry/omegaconf/pull/679#discussion_r614137195

### DIFF
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -480,7 +480,7 @@ class Container(Node):
 
         try:
             resolved = self.resolve_parse_tree(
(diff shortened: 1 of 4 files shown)


omry (commented):
Looks good.
Can you explain how you were able to make the node non-optional? was the case for None node only coming from the grammar test?

odelalleau:
> Can you explain how you were able to make the node non-optional? was the case for None node only coming from the grammar test?

Yes actually the only place where `node` was `None` in the previous code was in these grammar tests, due to not providing the key.

omry (approved):
shipit.

PR #687, opened 2021-04-17
541 oc.select

### LINKED ISSUE #541: Allow "default" in interpolation when target config node does not exist
Currently, when the key in interpolation doesn't exist, e.g. in:
```python
dic = DictConfig({
        "c": "${a}"
    })
print(dic.c)
```
it throws exception about "str interpolation key 'a' not found".
I hope there is a way to specify default values.

the `env:` resolver already supports defaults: https://omegaconf.readthedocs.io/en/latest/usage.html#environment-variable-interpolation
and the syntax seems nice. 
The above can be written as:
```python
dic = DictConfig({
        "c": "${a,123}"
})
```

omry:
Closes #541


docs/source/custom_resolvers.rst
```diff
@@ -298,6 +298,45 @@ This resolver can be useful for instance to parse environment variables:
     type: int, value: 3308
 
 
+.. _oc.select:
+
+oc.select
+^^^^^^^^^
+``oc.select`` enables selection similar to that performed with node interpolation, but is a bit more flexible.
+Using oc.select, you can provide a default value to use in case the primary interpolation key is not found.
```
odelalleau, line 306:
```suggestion
Using ``oc.select``, you can provide a default value to use in case the primary interpolation key is not found.
```

docs/source/custom_resolvers.rst
```diff
@@ -298,6 +298,45 @@ This resolver can be useful for instance to parse environment variables:
...
+The following example uses /tmp as the default value for the node output:
```
odelalleau, line 307:
```suggestion
The following example uses "/tmp" as the default value for the node output:
```

docs/source/custom_resolvers.rst
```diff
@@ -298,6 +298,45 @@ This resolver can be useful for instance to parse environment variables:
...
+    ... })
+    >>> print(cfg.a)
+    Saving output to /tmp
+    >>> cfg.output = "/etc/config"
+    >>> print(cfg.a)
+    Saving output to /etc/config
+
+``oc.select`` can also be used to select keys that are otherwise illegal interpolation keys.
+The following example has a key with a colon, such key looks like a custom resolver and therefore
+cannot be accessed using a regular interpolation.
```
odelalleau, line 322:
```suggestion
cannot be accessed using a regular interpolation:
```

docs/source/custom_resolvers.rst
```diff
@@ -298,6 +298,45 @@ This resolver can be useful for instance to parse environment variables:
...
+    ...    # yes, there is a : in this key
+    ...    "a:b": 10,
+    ...    "a": "${a:b}",
+    ...    "b": "${oc.select:'a:b'}",
+    ... })
+    >>> print(cfg.a)
+    Traceback (most recent call last):
+    ...
+    UnsupportedInterpolationType: Unsupported interpolation type a
+    >>> print(cfg.b)
```
odelalleau, line 336: It's a bit confusing to have `a:b`, `a` and `b` mixed together:
```suggestion
    ...    "bad": "${a:b}",
    ...    "good": "${oc.select:'a:b'}",
    ... })
    >>> print(cfg.bad)
    Traceback (most recent call last):
    ...
    UnsupportedInterpolationType: Unsupported interpolation type a
    >>> print(cfg.good)
```

docs/source/usage.rst
```diff
@@ -407,6 +407,7 @@ OmegaConf comes with a set of built-in custom resolvers:
 * :ref:`oc.decode`: Parsing an input string using interpolation grammar
 * :ref:`oc.deprecated`: Deprecate a key in your config
 * :ref:`oc.env`: Accessing environment variables
+* :ref:`oc.select`: Select an interpolation key, similar to interpolation but more flexible
```
odelalleau, line 410: Just for consistency with (most of) the other items in the list:
```suggestion
* :ref:`oc.select`: Selecting an interpolation key, similar to interpolation but more flexible
```

news/541.feature
```diff
@@ -0,0 +1 @@
+Add oc.select, enabling node selection with with a default value to use if the node cannot be selected
```
odelalleau, line 1:
```suggestion
Add oc.select, enabling node selection with a default value to use if the node cannot be selected
```

omegaconf/resolvers/oc/__init__.py
```diff
@@ -92,10 +94,25 @@ def deprecated(
...
+    default: Any = _DEFAULT_SELECT_MARKER_,
+    *,
+    _parent_: Container,
+) -> Any:
+    from omegaconf._impl import select_value
+
+    if default is _DEFAULT_SELECT_MARKER_:
+        return select_value(cfg=_parent_, key=key, absolute_key=True)
+    else:
+        return select_value(cfg=_parent_, key=key, absolute_key=True, default=default)
```
odelalleau, line 108: Why use a different marker?
```suggestion
def select(
    key: str,
    default: Any = _DEFAULT_MARKER_,
    *,
    _parent_: Container,
) -> Any:
    from omegaconf._impl import select_value

    return select_value(cfg=_parent_, key=key, absolute_key=True, default=default)
```

docs/source/custom_resolvers.rst
```diff
@@ -298,6 +298,45 @@ This resolver can be useful for instance to parse environment variables:
     type: int, value: 3308
 
 
+.. _oc.select:
+
+oc.select
+^^^^^^^^^
```
odelalleau, line 304: General doc comment:
There is another situation where `oc.select` can be useful which I believe is worth documenting: selecting from a missing value (which either yields `None` or the provided default -- by the way it looks like only the second case is tested).
It's an important difference with interpolations that don't allow interpolating to a missing node.

tests/interpolation/built_in_resolvers/test_oc_select.py
```diff
@@ -0,0 +1,93 @@
+from omegaconf import OmegaConf
```
odelalleau, line 1: (not related to this file)

Shouldn't `test_dict.py` be renamed into `test_oc_dict.py` for consistency?

tests/interpolation/built_in_resolvers/test_oc_select.py
```diff
@@ -0,0 +1,93 @@
...
+def test_oc_select_using_default() -> None:
+    cfg = OmegaConf.create(
+        {
+            "a0": "${oc.select:zz, 10}",
+            "a1": "${oc.select:.zz, 11}",
+            "a2": "${oc.select:zz, null}",
+            "a3": "${oc.select:zz, ${value}}",
+            "a4": "${oc.select:missing, 10}",
+            "value": 20,
+            "missing": "???",
```
odelalleau, line 86: As mentioned in the doc comment, I think the missing case should also be tested without a default value.

Can you also add a test when selecting a node whose value is `None` to make sure it doesn't use the default value? (it's working fine, but would be good to have a test for it). What I mean is:
```python
cfg = OmegaConf.create({"x": None, "y": "${oc.select:x, 3}"})
assert cfg.y is None
```

tests/interpolation/built_in_resolvers/test_oc_select.py
```diff
@@ -0,0 +1,93 @@
...
+            },
+            "k": 10,
+        },
+    )
+
+    n = cfg.nested
+    assert n.a0 == n.a1 == n.a2 == 10
+
+
+def test_oc_select_using_default() -> None:
```
odelalleau, line 77: Minor: convert to parameterized test?

odelalleau (changes requested):
Behavior looks good, just a few small things

omegaconf/resolvers/oc/__init__.py
```diff
@@ -92,10 +94,25 @@ def deprecated(
     return target_node
 
 
+def select(
```
odelalleau, line 97: Oh -- something else I want to point out (though I don't feel strongly about it) is that it's possible to select a parent node / create interpolation cycles with `oc.select`. We might want to prevent them similar to interpolations.

Jasha10 (changes requested):
Looks good :)
Just one grammar nit-pick beyond Olivier's suggestions:

omegaconf/resolvers/oc/__init__.py
omry: good point. Initially I was passing it into select_value.

tests/interpolation/built_in_resolvers/test_oc_select.py
omry: yes.

tests/interpolation/built_in_resolvers/test_oc_select.py
omry: added.

tests/interpolation/built_in_resolvers/test_oc_select.py
omry: sure.

omegaconf/resolvers/oc/__init__.py
odelalleau: I think it can be simplified further (see my suggestion: no more `if`)

omry: yeah, for some reason I thought using the same marker object select_value is using would cause issues.

PR #693, opened 2021-04-23
to_object no enum_to_str

Jasha10:
What do we think about removing the `enum_to_str` keyword argument from `OmegaConf.to_object`?
Here is my argument for removal:
- make the public API simpler,
- users can still call `OmegaConf.to_container(..., enum_to_str=True)` if they need it, and
- make sure that the `DictConfig._to_object` method is consistent with the `OmegaConf.to_object` method. Currently, `DictConfig._to_object` calls `BaseContainer._to_content` with `enum_to_str=False`, which will result in an inconsistency if the client calls `OmegaConf.to_object(cfg, enum_to_str=True)`, i.e. we could get some `enum_to_str=True` and some `enum_to_str=False` in the same deserialized object.

### DIFF
--- a/omegaconf/omegaconf.py
+++ b/omegaconf/omegaconf.py
@@ -555,11 +555,7 @@ class OmegaConf:
         )
 
     @staticmethod
-    def to_object(
-        cfg: Any,
-        *,
-        enum_to_str: bool = False,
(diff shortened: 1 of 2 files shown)


omry:
I support it.
any doc changes are needed?

Jasha10:
Nope, there is no mention of it in the docs.

omry (approved):

PR #696, opened 2021-04-27
Prevent assigning None to non-optional element_type

### LINKED ISSUE #579: When validating the assignment to a list, `None` is not properly validated
**Describe the bug**
The following code should crash (because `None` should be rejected) but currently does not:

```python
@dataclass
class Foo:
    bar: List[int] = "???"

cfg.bar = [1, None]
```

**Expected behavior**
There should be an exception because it is `List[int]` and not `List[Optional[int]]`.

**Additional context**
- [X] OmegaConf version: current master (cb5e556)

Jasha10:
This PR is to close #579.

### DIFF
--- a/news/579.bugfix
+++ b/news/579.bugfix
@@ -1,1 +1,1 @@
-Improve type validation of assignment to structured configs.
+Prevent assignment of List[NoneType] or Dict[..., NoneType] to a ListConfig/DictConfig whose values are typed as non-optional.
--- a/news/579.bugfix
+++ b/news/579.bugfix
@@ -1,1 +1,1 @@
-Prevent assignment of List[NoneType] or Dict[..., NoneType] to a ListConfig/DictConfig whose values are typed as non-optional.
+Prevent appending or inserting None in a ListConfig/DictConfig with non-optional element_type.
--- a/news/579.bugfix
+++ b/news/579.bugfix
@@ -1,1 +1,1 @@
-Prevent appending or inserting None in a ListConfig/DictConfig that has non-optional element_type.
+Prevent appending or inserting None in a ListConfig/DictConfig that has non-optional element/value type.
--- a/news/579.bugfix
+++ b/news/579.bugfix
@@ -1,1 +1,1 @@
-Prevent appending or inserting None in a ListConfig/DictConfig with non-optional element_type.
+Prevent appending or inserting None in a ListConfig/DictConfig that has non-optional element_type.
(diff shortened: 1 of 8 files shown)


news/579.bugfix
```diff
@@ -0,0 +1 @@
+Improve type validation of assignment to structured configs.
```
omry, line 1: This is too broad. be more specific.

omry (changes requested):
nice.

news/579.bugfix
Jasha10: Done

news/579.bugfix
```diff
@@ -0,0 +1 @@
+Prevent assignment of List[NoneType] or Dict[..., NoneType] to a ListConfig/DictConfig whose values are typed as non-optional.
```
omry, line 1: Doesn't it prevent assignment of None values to a List Config with a non optional element type?

Jasha10: I think you're right.

Jasha10: Hmm... It turns out that `list_cfg.append(None)` was not as expected. I just added another failing test...

Jasha10: Ok, should be cleared up now. The news fragment is improved in https://github.com/omry/omegaconf/pull/696/commits/00322883311ded5ca76ed87c4f46ef8c6ca5ca34.

> Hmm... It turns out that `list_cfg.append(None)` was not as expected.

Nevermind about this, I was confused because I didn't realize that `cfg = ListConfig([], element_type=int)` has `is_optional(cfg)==True` by default.

omry (commented):
Can you add an explicit test that verifies that assingment of a None value to a ListConfig and DictConfig with non optional value type?

Jasha10:
Added some explicit tests in https://github.com/omry/omegaconf/pull/696/commits/45c67d5855ef50e5d7ecf07408b2d8f2d3a36bd8.

omry (approved):
Looking good.
See inline comments for followup on OmegaConf.is_optional.

tests/test_omegaconf.py
```diff
@@ -280,6 +280,11 @@ def test_is_optional(fac: Any, is_optional: bool) -> None:
     assert OmegaConf.is_optional(cfg, "node") == is_optional
 
 
+def test_is_optional_non_node() -> None:
+    assert OmegaConf.is_optional("not_a_node")
+    assert OmegaConf.is_optional("???")
+
```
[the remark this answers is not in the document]
Jasha10: > Can you file a followup issue for it for 2.1?

Will do.

PR #700, opened 2021-04-28
Deprecate OmegaConf.is_optional

### LINKED ISSUE #698: Deprecate public OmegaConf.is_optional() API
For deprecation of the public OmegaConf.is_optional() method in favor of a private (internal) method.

_Based on comment posted by @omry in https://github.com/omry/omegaconf/pull/696#discussion_r622415886_

If you have a use case for OmegaConf.is_optional() please let us know what it is.

Jasha10:
Closes #698

### DIFF
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -197,10 +197,8 @@ def _is_optional(obj: Any, key: Optional[Union[int, str]] = None) -> bool:
     if key is not None:
         assert isinstance(obj, Container)
         obj = obj._get_node(key)
-    if isinstance(obj, Node):
-        return obj._is_optional()
-    else:
-        return True
+    assert isinstance(obj, Node)  # raises if key could not be found
+    return obj._is_optional()
 
 
 def _resolve_forward(type_: Type[Any], module: str) -> Type[Any]:
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -179,6 +179,7 @@ def _is_union(type_: Any) -> bool:
 
 
 def _resolve_optional(type_: Any) -> Tuple[bool, Any]:
+    """Check whether `type_` is equivalent to `typing.Optional[T]` for some T."""
     if getattr(type_, "__origin__", None) is Union:
         args = type_.__args__
         if len(args) == 2 and args[1] == type(None):  # noqa E721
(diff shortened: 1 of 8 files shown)


omegaconf/omegaconf.py
```diff
@@ -585,15 +586,14 @@ def is_missing(cfg: Any, key: DictKeyType) -> bool:
...
-            obj = obj._get_node(key)
-        if isinstance(obj, Node):
-            return obj._is_optional()
-        else:
-            return True
+        warnings.warn(
+            "`OmegaConf.is_optional()` is deprecated, see https://github.com/omry/omegaconf/issues/698",
+            stacklevel=2,
+        )
+        return _is_optional(obj, key)
```
omry, line 596: I am not sure there is a good reason behind defaulting to True.
If there isn't any, I think you can maintain the old deprecated implementation here for now and make a more strict internal _is_optional.

Jasha10: I think that the defaulting to True is for the case where `_is_optional(obj, key)` is called and `key` cannot be found.

Currently, no code in the `omegaconf/` directory is using this "defaulting to True" branch, as all the calls use the one-argument signature `_is_optional(obj)` (though some code in the `tests/` directory does use the two-argument signature).

Instead of returning `True` in the case where `key` cannot be found, what do you think about raising `ConfigKeyError`?

omegaconf/_utils.py
```diff
@@ -189,6 +190,19 @@ def _resolve_optional(type_: Any) -> Tuple[bool, Any]:
...
+    """Check `obj` metadata to see if the given node is optional."""
+    from .base import Container, Node
+
+    if key is not None:
+        assert isinstance(obj, Container)
+        obj = obj._get_node(key)
+    if isinstance(obj, Node):
+        return obj._is_optional()
+    else:
+        return True
```
omry, line 203: now that we have a separate implementation for the internal one, is there a need to support objects that are not nodes?

Jasha10: There is not.
But the `else` block still catches the case where `key` cannot be found.
What do you think of raising `ConfigKeyError` if `key` is not found, instead of returning True? (see my https://github.com/omry/omegaconf/pull/700#discussion_r623369771).

omegaconf/omegaconf.py
omry: Since it's an internal function and the branch is not used now, I suggest using assertion to ensure that it's not tripping anyone.

omegaconf/_utils.py
omry: missed your other question. answered there.
resolving.

omegaconf/omegaconf.py
Jasha10: > Make a pass on the usage to convince yourself it's not current;y possible for a user to hit this branch though.

Yup, it's not possible for a user to hit.

omry (approved):
lgtm

PR #703, opened 2021-04-29
add __delattr__ to DictConfig

### LINKED ISSUE #682: support __delattr__
```python

a = OmegaConf.create()
a.b = {}
a.b["c"] = 1
del a.b["c"]   # works

a.b.c = 1
del a.b.c  # fails
```

It seems counter-intuitive that this type of deletion is not supported.

jieru-hu:
Closes #682

### DIFF
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -378,8 +378,6 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
         :param key:
         :return:
         """
-        if key == "__name__":
-            raise AttributeError()
         self._delkey(key)
 
     def __delitem__(self, key: DictKeyType) -> None:
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -378,13 +378,22 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
         :param key:
         :return:
         """
-        self._delkey(key, attr=True)
+        if self._get_flag("readonly"):
+            self._format_and_raise(
+                key=key,
+                value=None,
+                cause=ReadonlyConfigError(
+                    "DictConfig in read-only mode does not support deletion"
+                ),
+            )
+        try:
+            del self.__dict__["_content"][key]
+        except KeyError:
+            msg = "Attribute not found: '$KEY'"
+            self._format_and_raise(key=key, value=None, cause=ConfigAttributeError(msg))
 
     def __delitem__(self, key: DictKeyType) -> None:
         key = self._validate_and_normalize_key(key)
-        self._delkey(key)
-
-    def _delkey(self, key: DictKeyType, attr: bool = False) -> None:
         if self._get_flag("readonly"):
             self._format_and_raise(
                 key=key,
@@ -414,8 +423,7 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
             del self.__dict__["_content"][key]
         except KeyError:
             msg = "Key not found: '$KEY'"
-            error = ConfigAttributeError if attr else ConfigKeyError
-            self._format_and_raise(key=key, value=None, cause=error(msg))
+            self._format_and_raise(key=key, value=None, cause=ConfigKeyError(msg))
 
     def get(self, key: DictKeyType, default_value: Any = None) -> Any:
         """Return the value for `key` if `key` is in the dictionary, else
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -372,8 +372,21 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
         except Exception as e:
             self._format_and_raise(key=key, value=None, cause=e)
 
+    def __delattr__(self, key: str) -> None:
+        """
+        Allow deleting dictionary values as attributes
+        :param key:
+        :return:
+        """
+        if key == "__name__":
+            raise AttributeError()
+        self._delkey(key)
+
     def __delitem__(self, key: DictKeyType) -> None:
         key = self._validate_and_normalize_key(key)
+        self._delkey(key)
+
+    def _delkey(self, key: DictKeyType) -> None:
         if self._get_flag("readonly"):
             self._format_and_raise(
                 key=key,
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -378,13 +378,13 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
         :param key:
         :return:
         """
-        self._delkey(key)
+        self._delkey(key, attr=True)
 
     def __delitem__(self, key: DictKeyType) -> None:
         key = self._validate_and_normalize_key(key)
         self._delkey(key)
 
-    def _delkey(self, key: DictKeyType) -> None:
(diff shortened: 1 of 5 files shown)


Jasha10 (approved):
Looks good, nice work!

omegaconf/dictconfig.py
```diff
@@ -372,8 +372,21 @@ def __getitem__(self, key: DictKeyType) -> Any:
...
 
+    def __delattr__(self, key: str) -> None:
+        """
+        Allow deleting dictionary values as attributes
+        :param key:
+        :return:
+        """
+        if key == "__name__":
+            raise AttributeError()
+        self._delkey(key)
```
omry, line 383: Deleting from an object raises an AttributeError.
When deleting an attribute from a DictConfig backed by a Structured Config I think we should raise ConfigAttributeError.

jieru-hu:
>what do you need special treatment for "__name__" ?


Yea, I'm not super sure about this one - I added this case to be  consistent with the `__getattr__` [behavior](https://github.com/omry/omegaconf/blob/master/omegaconf/dictconfig.py#L341-L348) which seems to make sense. I can  remove it if this is not needed.

omry:
> > what do you need special treatment for "**name**" ?
> 
> Yea, I'm not super sure about this one - I added this case to be consistent with the `__getattr__` [behavior](https://github.com/omry/omegaconf/blob/master/omegaconf/dictconfig.py#L341-L348) which seems to make sense. I can remove it if this is not needed.

1. Yes, remove it. I don't want to add logic without a very clear reason.
2. Can you respond on the actual comment so it's easier to respond in context?

omry (changes requested):
See previous comments.

omegaconf/dictconfig.py
omry: Did I miss your response to this?
I am not seeing a reply, and I am not seeing this addressed either.

Jasha10: FYI @jieru-hu I think that in this case the right way to check if a DictConfig is backed by a Structured Config would be to evaluate `is_structured_config(self._metadata.object_type)`.

jieru-hu: Sorry I missed this one! I've updated the PR to address this comment.

Jasha10: Oh, please disregard my previous comment, what you've got now looks good (raising ConfigAttributeError for both structured and non-structured)

Jasha10 (approved):

tests/test_readonly.py
```diff
@@ -51,12 +51,6 @@
             raises(ReadonlyConfigError, match="a"),
             id="dict_delitem",
         ),
-        param(
-            {"a": 10},
-            lambda c: c.__delattr__("a"),
-            raises(ReadonlyConfigError, match="a"),
-            id="dict_delattr",
-        ),
```
omry: This one actually makes sense, why did you remove it?

omry (commented):
Since Python in general allow deleting of fields from objects with delattr, I think we should allow it for DictConfig in struct mode and for DictConfig backed by a Structured Config. (if they are not readonly).

tests/test_errors.py
```diff
@@ -199,28 +199,6 @@ def finalize(self, cfg: Any) -> None:
...
-        Expected(
-            create=lambda: OmegaConf.create({"foo": "bar"}),
-            op=lambda cfg: cfg.__delattr__("not_found"),
-            key="not_found",
-            exception_type=ConfigKeyError,
-            msg="Key not found: 'not_found'",
-        ),
-        id="dict:delattr_invalid",
-    ),
-    param(
```
Jasha10: Since `__delattr__` is no longer just an alias for `__delitem__`, does it make sense to put some of these tests back in?

jieru-hu: Yes - I think I've added back the tests that still apply. Now we allow deleting attributes from struct and structured,  some of the old tests (in the first commit) in `test_errors` does not apply to `__delattr__` any more since they are expecting the call to fail.

jieru-hu: The new behavior is not covered in `test_delattr` :)

tests/test_basic_ops_list.py
```diff
@@ -738,6 +739,24 @@ def test_getattr() -> None:
     assert getattr(cfg, "2") == src[2]
 
 
+@mark.parametrize(
+    "getconfig, struct",
+    [
+        (lambda: OmegaConf.create({"name": "alice", "age": 1}), False),
+        (lambda: OmegaConf.create({"name": "alice", "age": 1}), True),
+        (lambda: OmegaConf.structured(User(name="alice", age=1)), False),
```
omry, line 747: Why not:
```suggestion
        (OmegaConf.create({"name": "alice", "age": 1}), False),
        (OmegaConf.create({"name": "alice", "age": 1}), True),
        (OmegaConf.structured(User(name="alice", age=1)), False),
```

tests/test_basic_ops_list.py
```diff
@@ -738,6 +739,24 @@ def test_getattr() -> None:
     assert getattr(cfg, "2") == src[2]
 
 
+@mark.parametrize(
```
omry, line 742: `tests/test_basic_ops_list.py` seems like the wrong file for these tests.
move to test_basic_ops_dict ?

tests/test_basic_ops_list.py
jieru-hu: fixed!

tests/test_basic_ops_list.py
jieru-hu: thanks, yes, accidentally added to the wrong file :)

tests/test_basic_ops_dict.py
```diff
@@ -75,6 +75,23 @@ def test_getattr_dict() -> None:
...
+    "cfg, struct",
+    [
+        (OmegaConf.create({"name": "alice", "age": 1}), False),
+        (OmegaConf.create({"name": "alice", "age": 1}), True),
+        (OmegaConf.structured(User(name="alice", age=1)), False),
+    ],
+)
+def test_delattr(cfg: DictConfig, struct: bool) -> None:
+    if struct:
+        OmegaConf.set_struct(cfg, True)
```
omry, line 88:
```suggestion
    OmegaConf.set_struct(cfg, struct)
```

tests/test_basic_ops_dict.py
```diff
@@ -75,6 +75,23 @@ def test_getattr_dict() -> None:
...
 
 
+@mark.parametrize(
+    "cfg, struct",
+    [
+        (OmegaConf.create({"name": "alice", "age": 1}), False),
+        (OmegaConf.create({"name": "alice", "age": 1}), True),
+        (OmegaConf.structured(User(name="alice", age=1)), False),
+    ],
+)
```
omry, line 85: Try to parameterize like this. This will give all 4 combinations.
```python
@mark.parametrize("struct" : [True, False])
@mark.parametrize("cfg" : [
        param(OmegaConf.create({"name": "alice", "age": 1}), id="dict"),
        param(OmegaConf.structured(User(name="alice", age=1)), id="structured_config"),
])
```

jieru-hu: nice!

omry (approved):
shipit!

PR #709, opened 2021-05-07
Fix crash with "interpolation-like" strings from interpolations

### LINKED ISSUE #666: Escaped interpolations trigger a GrammarParseError if they don't follow the expected interpolation grammar
**Describe the bug**
The following code triggers a GrammarParseError:

```python
cfg = OmegaConf.create({"x": "\\${not an interpolation}"})
cfg.x
```

**Expected behavior**

There should be no grammar syntax constraint after a `\${` since it is not a real interpolation.

**Discussion**

What happens is that when we dereference `cfg.x`, the resulting string (`"${not an interpolation}"`) is wrapped within a node (`AnyNode` because here `x` is untyped), whose constructor calls `ValueNode._set_value()`, that checks if the the value is an interpolation... and this check raises an exception because it's a badly formed interpolation.

The bandaid fix is to set `strict_interpolation_validation=False` at https://github.com/omry/omegaconf/blob/a03f681713a4c551c9f25ed8172d958b16fc92d0/omegaconf/nodes.py#L37
It would break the tests that specifically test for this check, but overall things would still work fine.

This bandaid fix, however, wouldn't change the fact that `cfg._get_node("x")._dereference_node()._is_interpolation()` is True, i.e., when we dereference `cfg.x` we obtain a node that believes it is an interpolation (!). Right now this isn't really a problem because we don't do anything with this node except getting its value, but it may bite us later.

Right now I don't have a solution that I really like in mind, so posting this issue for discussion.
Here are options I've thought about so far:
* The bandaid fix mentioned above (currently my preferred option for 2.1)
* Creating a subclass of `StringNode` specifically when an interpolation result is a string with `${`. This subclass would override what it takes to avoid a crash. One issue with this is that if `cfg.x` is actually typed, we'd like to return another type of node (ex: `IntegerNode`), so there should be an exception instead (it seems doable, just a bit clumsy)
* Adding a flag to `StringNode` that says "I am not an interpolation" even if the value looks like one. This would be similar to the previous idea, but without adding an extra class.
* Add a more general flag to all nodes, `is_interpolation={True,False,None}`, where `None` would be the current behavior (check if the value is a string looking like an interpolation), `False` would make it possible to "declare" the value as not being an interpolation (even if it looks like one), and `True` would declare it as being an interpolation (regardless of the value). Besides fixing this issue, this could enable more things like (i) caching the interpolation status in that flag for faster checks, or (ii) enabling new kinds of interpolations that aren't tied to the string content (e.g. to point to config nodes that aren't accessible by a path from the root config, as I once did in a draft PR). I kinda like this option but I expect it to be non trivial and uncover issues I haven't thought about yet, so I'd rather not go there in 2.1

Thoughts?

odelalleau:
Fixes #666

This is a first stab at it in order to ground discussions. I did a bit more than the "bandaid" fix I had suggested in #666, to get a cleaner behavior (at the expense of passing around a new "can_be_interpolation" flag).

In this first version I only included this flag in `AnyNode` and `StringNode` so as to somewhat limit potential side effects, though it's a bit ugly because it leaks into `ValueNode._set_value_impl()`. I wouldn't mind making it an attribute of `ValueNode`, which could serve as a first step towards caching the interpolation status in the future, but I preferred to wait until we'd discuss it.

### DIFF
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -568,6 +568,9 @@ class Container(Node):
                 value=resolved,
                 key=key,
                 ref_type=value._metadata.ref_type,
+                # Since `resolved` was obtained by resolving an interpolation, it cannot
+                # be itself an interpolation even if may look like one (ex: "${foo}").
+                can_be_interpolation=False,
             )
         else:
             # Other objects get wrapped into an `AnyNode` with `allow_objects` set
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -15,7 +15,6 @@ from ._utils import (
     _is_missing_value,
     format_and_raise,
     get_value_kind,
-    is_primitive_type,
     split_key,
 )
 from .errors import (
@@ -504,7 +503,7 @@ class Container(Node):
         resolved: Any,
         throw_on_resolution_failure: bool,
     ) -> Optional["Node"]:
-        from .nodes import AnyNode, ValueNode
+        from .nodes import AnyNode, InterpolationResultNode, ValueNode
 
         # If the output is not a Node already (e.g., because it is the output of a
         # custom resolver), then we will need to wrap it within a Node.
@@ -533,63 +532,10 @@ class Container(Node):
                 resolved = conv_value
 
         if must_wrap:
-            return self._wrap_interpolation_result(
-                parent=parent,
-                value=value,
-                key=key,
-                resolved=resolved,
-                throw_on_resolution_failure=throw_on_resolution_failure,
-            )
+            return InterpolationResultNode(value=resolved, key=key, parent=parent)
         else:
             assert isinstance(resolved, Node)
             return resolved
-
-    def _wrap_interpolation_result(
-        self,
-        parent: Optional["Container"],
-        value: Node,
-        key: Any,
-        resolved: Any,
-        throw_on_resolution_failure: bool,
-    ) -> Optional["Node"]:
-        from .basecontainer import BaseContainer
-        from .nodes import AnyNode, StringInterpolationResultNode
-        from .omegaconf import _node_wrap
-
-        assert parent is None or isinstance(parent, BaseContainer)
-
-        if is_primitive_type(type(resolved)):
-            # Primitive types get wrapped using `_node_wrap()`, ensuring value is
-            # validated and potentially converted.
-            # There is one exception when the following two conditions are met:
-            #   - `resolved` is a string, and
-            #   - the target type is either unspecified / Any / str
-            # In that case, we wrap `resolved` inside a `StringInterpolationResultNode`
-            # so as to ensure that it cannot be considered itself as an interpolation,
-            # e.g., when `resolved` would be a string like "${foo}".
-            target_type = value._metadata.ref_type
-            if isinstance(resolved, str) and target_type in [None, Any, str]:
-                wrapped: Node = StringInterpolationResultNode(
-                    value=resolved, key=key, parent=parent
-                )
-            else:
-                wrapped = _node_wrap(
-                    type_=target_type,
-                    parent=parent,
-                    is_optional=value._metadata.optional,
-                    value=resolved,
-                    key=key,
-                    ref_type=target_type,
-                )
-        else:
-            # Other objects get wrapped into an `AnyNode` with `allow_objects` set
-            # to True.
-            wrapped = AnyNode(
-                value=resolved, key=key, parent=None, flags={"allow_objects": True}
-            )
-            wrapped._set_parent(parent)
-
-        return wrapped
 
     def _validate_not_dereferencing_to_parent(self, node: Node, target: Node) -> None:
         parent: Optional[Node] = node
--- a/omegaconf/base.py
+++ b/omegaconf/base.py
@@ -553,7 +553,7 @@ class Container(Node):
         throw_on_resolution_failure: bool,
     ) -> Optional["Node"]:
         from .basecontainer import BaseContainer
-        from .nodes import AnyNode
+        from .nodes import AnyNode, StringInterpolationResultNode
         from .omegaconf import _node_wrap
 
         assert parent is None or isinstance(parent, BaseContainer)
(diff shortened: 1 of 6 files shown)


omegaconf/nodes.py
```diff
@@ -33,15 +33,19 @@ def _set_value(self, value: Any, flags: Optional[Dict[str, bool]] = None) -> Non
...
             value, strict_interpolation_validation=True
-        ) in (
+        ) not in (
             ValueKind.INTERPOLATION,
             ValueKind.MANDATORY_MISSING,
-        ):
-            self._val = value
-        else:
-            self._val = self.validate_and_convert(value)
+        )
```
omry, line 48: This is pretty hard to parse. can you reorganize it to be easier to understand?

omegaconf/nodes.py
```diff
@@ -390,3 +394,50 @@ def __deepcopy__(self, memo: Dict[int, Any]) -> "EnumNode":
...
+        key: Any = None,
+        parent: Optional[Container] = None,
+        flags: Optional[Dict[str, bool]] = None,
+    ):
+        # In general we should not try to write into interpolation results.
+        if flags is None:
+            flags = {"readonly": True}
+        elif "readonly" not in flags:
+            flags = flags.copy()
+            flags["readonly"] = True
```
omry, line 416: can you set readonly after you call `super().__init__`?

omegaconf/nodes.py
omry: Instead of introducing should_validate(), can the new class override _set_value?

tests/interpolation/test_interpolation.py
```diff
@@ -463,3 +465,35 @@ def test_circular_interpolation(cfg: Any, key: str, expected: Any) -> None:
...
+
+    else:
+        # Structured config.
+
+        @dataclass
+        class Config:
+            x: node_type = value  # type: ignore
+            y: str = "{foo"
+
+        cfg = OmegaConf.structured(Config)
```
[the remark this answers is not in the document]
omry: I think this can work:
```
OmegaConf.create({"a": AnyNode("b")})
OmegaConf.create({"a": Stringode("b")})
```

omegaconf/base.py
```diff
@@ -553,22 +553,34 @@ def _wrap_interpolation_result(
...
+                )
+            else:
+                wrapped = _node_wrap(
+                    type_=target_type,
+                    parent=parent,
+                    is_optional=value._metadata.optional,
+                    value=resolved,
+                    key=key,
+                    ref_type=target_type,
+                )
```
omry, line 583: iirc, we use the resulting node for validation by calling validate_and_convert on it.

An idea is to have a TempNode that is delegating validate_and_convert to the function on the destination node.
This would allow us to potentially simplify here by always returning the same type from this function.
I am not 100% sure it's worth it though.

tests/interpolation/test_interpolation.py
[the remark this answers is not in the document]
odelalleau: Ah, thanks! Done in ff75b04

omry (approved):
Overall looking good.
I would have not bothered with the intricate test for the readonly flag on the StringInterpolation node but it doesn't hurt.

omegaconf/base.py
omry: This is even simpler than what I had in mind. If you can make it work like this I think it's better.

1. Your current logic is raising a validation error for container nodes, which I am not sure is correct.
2. At some point we added the validation logic for interpolation somewhere else, this should also be cleaned up.

odelalleau: Ok I'll give it a shot a bit later today. Regarding your points:
1. I'll need to look into it more closely. Before this PR there was no proper validation for containers (see last point in this section: https://omegaconf.readthedocs.io/en/latest/structured_config.html#interpolations)
2. I'm not sure what you are referring to, can you be more specific? (before this PR the validation logic was from within `_node_wrap()`)

omry: 1. Not validating is not the same as raising a validation error :)
2. If it was all localized to this function then I may be just misremembering or remembering an intermediate state.

odelalleau: So, you were right regarding (2), the validation actually happened just before the call to `_node_wrap()`. That call was thus just to return a node of the correct type, and since we don't care about it anymore, this whole block can go away.

I didn't try to address the validation of containers (1) since I believe it's better left for a different PR.

Result is in b88350d

omegaconf/nodes.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, parent: Optional[Container], value: Any, metadata: Metadata):
 
         super().__init__(parent=parent, metadata=metadata)
         with read_write(self):
-            self._set_value(value)
+            self._set_value(value)  # lgtm [py/init-calls-subclass]
```
omry, line 27: lgtm is still complaining about it. maybe some syntax issue with the suppression comment?

tests/interpolation/test_interpolation.py
```diff
@@ -257,7 +261,7 @@ def test_none_value_in_quoted_string(restore_resolvers: Any) -> None:
             User(name="Bond", age=SI("${cast:int,'7'}")),
             "age",
             7,
-            IntegerNode,
+            InterpolationResultNode,
```
omry, line 264: can you add 
```python
assert type(val) is type(expected_value)
```
to ensure the type of the value is the right one and this is not some `__eq__` magic?

tests/test_nodes.py
```diff
@@ -499,14 +500,31 @@ def test_deepcopy(obj: Any) -> None:
...
     assert (node != value) != expected
     assert (value == node) == expected
     assert (value != node) != expected
-    assert (node.__hash__() == value.__hash__()) == expected
+
+    try:
+        assert (node.__hash__() == value.__hash__()) == expected
+    except TypeError as exc:
+        # Skip this check if unhashable.
+        assert "unhashable type" in str(exc)
```
omry, line 527: If the problem is that InterpolationResultNode is not hashable, be more specific:
```suggestion
    if not isinstance(node, InterpolationResultNode):
        assert (node.__hash__() == value.__hash__()) == expected
```

Alternatively you can fix InterpolationResultNode to return hash of the value.

omegaconf/nodes.py
odelalleau: I'm a bit confused by lgtm, it seems like the warning will be there until merged, since they in the warning link they say: "If you make changes to the suppression of alerts in a pull request, please note that theses changes will only be effective once that pull request is merged"
(this sounds dumb, but I guess there must be a reason)

tests/interpolation/test_interpolation.py
odelalleau: Done in f764f5c (actually had to update the DictConfig/ListConfig tests that were relying on `__eq__` magic)

tests/test_nodes.py
odelalleau: No, the problem is when wrapping a dict/list which aren't hashable. I changed it in c4004bf to make it less generic but more explicit.

omry (approved):
shipit

omry:
If `1 for `__init__` method calls overridden method` is still an issue we can fix it post rc1.

PR #710, opened 2021-05-07
Avoid warnings in tests

odelalleau:


omry:
Did we lose coverage?

Jasha10:
> Did we lose coverage?

By default, pytest does not fail if there are warnings. It just prints the warnings to stdout after the tests are finished.

omry:
You can change the noxfile in omegaconf to fail on warnings like it does on Hydra.

Jasha10 (approved):

odelalleau:
I'm merging so that #711 can go in

PR #711, opened 2021-05-07
Pytest to fail on warnings

Jasha10:
I've added a '-Werror' flag to pytest's ini options.

I expect that the CI tests will fail (#710 is fixing some warnings that are currently issued).


Jasha10:
I have confirmed (on my local computer) that nox tests pass when this branch is rebased against #710.

omry (approved):
rebase and merge once green.

PR #712, opened 2021-05-12
throw error if subclassing Dict[!=str, ...]

### LINKED ISSUE #657: Error on subclassing Dict[Any, ...]
Based on https://github.com/omry/omegaconf/pull/502#discussion_r605832350, it is proposed to deprecate support for structured configs that subclass `Dict[Any, ...]`, with only subclasses of `Dict[str, ...]` being supported going forward.

For example, creating a structured config node based on the following dataclass would issue a deprecation warning,

```python
@dataclass
class MyConf(Dict[int, Any]):
    ...
```

while a structured config node based on the following would be fully supported:

```python
@dataclass
class MyConf(Dict[str, Any]):
    ...
```

Jasha10:
Closes #657 

TODO:
- [ ] add test case to test_errors.py

### DIFF
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -240,6 +240,12 @@ def extract_dict_subclass_data(obj: Any, parent: Any) -> Optional[Dict[str, Any]
     subclasses_dict = is_dict_subclass(obj_type)
 
     if subclasses_dict:
+        key_type, _element_type = get_dict_key_value_types(obj_type)
+        if key_type != str:
+            raise ValueError(
+                f"Dict subclass '{obj_type.__name__}' must have key_type==str."
+                + f" Type '{key_type.__name__}' is not supported."
+            )
         warnings.warn(
             f"Class `{obj_type.__name__}` subclasses `Dict`."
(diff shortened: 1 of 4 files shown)


omegaconf/_utils.py
```diff
@@ -240,6 +240,12 @@ def extract_dict_subclass_data(obj: Any, parent: Any) -> Optional[Dict[str, Any]
     subclasses_dict = is_dict_subclass(obj_type)
 
     if subclasses_dict:
+        key_type, _element_type = get_dict_key_value_types(obj_type)
+        if key_type != str:
+            raise ValueError(
+                f"Dict subclass '{obj_type.__name__}' must have key_type==str."
+                + f" Type '{key_type.__name__}' is not supported."
+            )
```
omry, line 248: People might already be extending Any and using it for str.
Let's think about this one after rc1.
I think the deprecation warning may be sufficient.

Jasha10: Sounds good.

omry:
I think it's not worth it. we are already issuing a deprecation warning when a Structured Config extends Dict regardless of the key type.

PR #716, opened 2021-05-14
Ensure the grammar cache is thread-safe

### LINKED ISSUE #715: Thread-safety concern related to interpolation parsing
**Describe the bug**

It occurred to me that we may want to ensure that OmegaConf is thread-safe (at least on simultaneous reads). But the [cache mechanism](https://github.com/omry/omegaconf/blob/2186b1da6134129c9f4ca89e4b479be445a2690d/omegaconf/grammar_parser.py#L121) I introduced in #445 to speed-up the parsing of interpolations is not thread-safe.

**Expected behavior**

Parsing interpolations in parallel from multiple threads should not fail because of the cache mentioned above.

odelalleau:
Fixes #715

Do we need a news item for bugfixes in-between RC releases?

Note: the diff is easier to read when [ignoring whitespaces](https://github.com/omry/omegaconf/pull/716/files?w=1)

### DIFF
--- a/omegaconf/grammar_parser.py
+++ b/omegaconf/grammar_parser.py
@@ -106,8 +106,8 @@ def parse(
         lexer.removeErrorListeners()
         lexer.addErrorListener(error_listener)
         lexer.mode(l_mode)
-        tokens = CommonTokenStream(lexer)
-        parser = OmegaConfGrammarParser(tokens)
+        token_stream = CommonTokenStream(lexer)
+        parser = OmegaConfGrammarParser(token_stream)
         parser.removeErrorListeners()
         parser.addErrorListener(error_listener)
 
@@ -116,13 +116,13 @@ def parse(
         # from antlr4 import PredictionMode
         # parser._interp.predictionMode = PredictionMode.SLL
 
-        _grammar_cache.data = lexer, tokens, parser
+        _grammar_cache.data = lexer, token_stream, parser
 
     else:
-        lexer, tokens, parser = cached
+        lexer, token_stream, parser = cached
         lexer.inputStream = istream
         lexer.mode(l_mode)
-        tokens.setTokenSource(lexer)
+        token_stream.setTokenSource(lexer)
         parser.reset()
 
     try:
--- a/omegaconf/grammar_parser.py
+++ b/omegaconf/grammar_parser.py
@@ -1,6 +1,6 @@
 import re
 import threading
-from typing import Any, Tuple
+from typing import Any
 
 from antlr4 import CommonTokenStream, InputStream, ParserRuleContext
 from antlr4.error.ErrorListener import ErrorListener
@@ -15,10 +15,8 @@ from .grammar_visitor import (  # type: ignore
 )
 
 # Used to cache grammar objects to avoid re-creating them on each call to `parse()`.
-_grammar_cache = None
-
-# Lock to ensure we do not use cached objects in parallel threads.
-_cache_lock = threading.Lock()
+# We use a per-thread cache to make it thread-safe.
+_grammar_cache = threading.local()
 
 # Build regex pattern to efficiently identify typical interpolations.
 # See test `test_match_simple_interpolation_pattern` for examples.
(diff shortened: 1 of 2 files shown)


omegaconf/grammar_parser.py
```diff
@@ -16,6 +17,9 @@
 # Used to cache grammar objects to avoid re-creating them on each call to `parse()`.
 _grammar_cache = None
 
+# Lock to ensure we do not use cached objects in parallel threads.
+_cache_lock = threading.Lock()
+
```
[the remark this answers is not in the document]
odelalleau: Interesting, thanks, I didn't know about this feature. I implemented a version based on it in cb57d01
(probably easier to check the new full diff vs. master)

omegaconf/grammar_parser.py
```diff
@@ -14,7 +15,8 @@
 )
 
 # Used to cache grammar objects to avoid re-creating them on each call to `parse()`.
-_grammar_cache = None
+# We use a per-thread cache to make it thread-safe.
+_grammar_cache = threading.local()
```
omry, line 19: Since someone else might also be calling threading.local() to get the exact object, let's call it `thread_local` and the data object inside `grammar_cache`.

tests/test_grammar.py
```diff
@@ -767,3 +769,43 @@ def create() -> DictConfig:
         # Other invalid characters should be detected at creation time.
         with raises(GrammarParseError):
             create()
+
+
+def test_grammar_cache_is_thread_safe() -> None:
```
omry, line 774: how long does this test take to run?

omegaconf/grammar_parser.py
omry: Also, Maybe we shouldn't store the thread-local-storage object in the global scope as it might get serialized in some scenarios.

It would probably be better to not store the global at all and always access the grammar cache directly from the TLS in the call-site.

omegaconf/grammar_parser.py
```diff
@@ -115,10 +116,10 @@ def parse(
         # from antlr4 import PredictionMode
         # parser._interp.predictionMode = PredictionMode.SLL
 
-        _grammar_cache = lexer, tokens, parser
+        _grammar_cache.data = lexer, tokens, parser
```
omry, line 119: 1. Can you rename tokens to tokenizer or token_stream?
2. I find it a bit odd that we are caching the input stream (via the lexer).
Can you explain the caching strategy?

omegaconf/grammar_parser.py
odelalleau: Wait -- `threading.local()` creates a new storage on each call as far as I can tell:
```python
In [67]: x = threading.local()

In [68]: y = threading.local()

In [69]: x is y
Out[69]: False

In [70]: x.a = 1

In [71]: y.a
AttributeError                            Traceback (most recent call last)
<ipython-input-71-3c3a974596a6> in <module>
----> 1 y.a

AttributeError: '_thread._local' object has no attribute 'a'
```

tests/test_grammar.py
odelalleau: 0.15s on my laptop when I run only this test (including pytest's overhead)

It's actually slower if I remove the `sleep()` calls because of Python's terrible multi-threading capabilities.

omegaconf/grammar_parser.py
omry: Gotcha. I think that's a bit unconventional but maybe I am wrong. It's been a while since I last used TLS and that was not in Python.

omegaconf/grammar_parser.py
odelalleau: 1. Done in a5c66fc
2. At high-level the strategy is "let's try to cache the lexer and parser so that we don't need to re-create them every time we parse a new string". Caching the input stream is just a side effect of caching the lexer, we don't really need it since we override it when we re-use the cached lexer. I guess we could set `lexer.inputStream = None` after parsing to avoid keeping a reference to the input string -- it'd be very surprising for it to ever matter, but I can easily add it if you think it's better.

omry: This is more about the semantics than about an actual runtime concern.
A comment explaining that we are replacing the input stream when we reuse the cache would be sufficient.

odelalleau: Got it -- done in 2b4aba8

omry:
> Do we need a news item for bugfixes in-between RC releases?

Generally yes. but in this particular one I am not sure it's worth a news item because it does not change anything for users.

omry (approved):

odelalleau:
> Generally yes. but in this particular one I am not sure it's worth a news item because it does not change anything for users.

Ok (FYI when writing the test I took the opportunity to check what would happen if the cached objects were shared between mutliple threads, and this caused it to hang, so I think it's good that we fixed it)

PR #719, opened 2021-05-19
OmegaConf.missing_keys to get a set of missing keys

Ohad31415:


Jasha10:
Closes #720

Ohad31415:
@omry @Jasha10 I have a gap at the docs session, I've added a segment under docs/source/usage.rst for sphinx and seems like I'm missing something. 
Would like to get a hand here 😀

Jasha10:
From the doctest output:
```
File "usage.rst", line 791, in default
Failed example:
    OmegaConf.missing_keys({
        "foo": {"bar": "???"},
        "missing": "???",
         "list": ["a", None, "???"]
    })
Expected:
    {'list.2', 'foo.bar', 'missing'}
Got:
    {'missing', 'foo.bar', 'list.2'}
```
The elements of a set are not guaranteed to be printed in any particular order...
Here is something that should make the tests pass, though it is sacrifices clarity for the reader:
```python
>>> sorted(OmegaConf.missing_keys({
...     "foo": {"bar": "???"},
...     "missing": "???",
...     "list": ["a", None, "???"]
... }))
['foo.bar', 'list.2', 'missing']
```

tests/test_omegaconf.py
```diff
@@ -518,3 +518,128 @@ def test_resolve(cfg: Any, expected: Any) -> None:
...
+                        "foo": "???",
+                        "bar": None
+                    }),
+                    "???",
+                    ["???", 3, False]
+                ],
+                "x": "y",
+                "y": "???"
+            },
+            {"list.1.foo", "list.2", "list.3.0", "y"}
```
omry: for lists, this would be a better syntax:
```suggestion
            {"list[1].foo", "list[2]", "list[3][0]", "y"}
```

Ohad31415: @omry What about a plain list like [1, ???, 2] - do you think it's better like `{[1]}`, or  `{1}`?

Ohad31415:
> From the doctest output:
> 
> ```
> File "usage.rst", line 791, in default
> Failed example:
>     OmegaConf.missing_keys({
>         "foo": {"bar": "???"},
>         "missing": "???",
>          "list": ["a", None, "???"]
>     })
> Expected:
>     {'list.2', 'foo.bar', 'missing'}
> Got:
>     {'missing', 'foo.bar', 'list.2'}
> ```
> 
> The elements of a set are not guaranteed to be printed in any particular order...
> Here is something that should make the tests pass, though it is sacrifices clarity for the reader:
> 
> ```python
> >>> sorted(OmegaConf.missing_keys({
> ...     "foo": {"bar": "???"},
> ...     "missing": "???",
> ...     "list": ["a", None, "???"]
> ... }))
> ['foo.bar', 'list.2', 'missing']
> ```

Yeah it does sacrifice readability, I wonder if there is a way around it to make it verify the output rather than the output's representation.

omry:
> Yeah it does sacrifice readability, I wonder if there is a way around it to make it verify the output rather than the output's representation.

```python
assert ret == {'list.2', 'foo.bar', 'missing'}
```

tests/test_omegaconf.py
omry: Yes, I think it's okay to always use brackets for lists.
In fact, you can use `node._get_full_key(key)` on nodes to get the proper representation.
try it as an alternative to constructing the key yourself.

Ohad31415: Great I dropped constructing the prefix on my own, using _get_full_key instead.

omegaconf/omegaconf.py
```diff
@@ -785,6 +787,38 @@ def resolve(cfg: Container) -> None:
...
+        :param cfg: An `OmegaConf.Container`,
+                    or a convertible object via `OmegaConf.create` (dict, list, ...).
+        :return: set of strings of the missing keys.
+        :raises ValueError: On input not representing a config.
+        """
+        if not isinstance(cfg, Container):
+            try:
+                cfg = OmegaConf.create(cfg)
+            except ValidationError:
+                raise ValueError(f"Could not create a config out of {cfg}")
```
Jasha10, line 803: I think that it would be appropriate to use the `_ensure_container` function here.
```suggestion
        cfg = _ensure_container(cfg)
```

omry: Yes. once you do, you can also change your tests to take plain dict/list instead of a Container to simplfy.

omry (commented):
Please a news fragment file. This is similar to how it's done for Hydra.
Context [here](https://hydra.cc/docs/next/development/documentation#news-entries).

omegaconf/omegaconf.py
Ohad31415: @omry The tests contain inputs of all types: ListConfig, DictConfig, plain list, dict. I think it doesn't hurt having it so, but if you think it's better with only plain lists/dicts I'll change it 😃

Ohad31415:
> Please a news fragment file. This is similar to how it's done for Hydra.
> Context [here](https://hydra.cc/docs/next/development/documentation#news-entries).

I've added `news/720.feature`. Let me know if something is wrong/missing.

news/720.feature
```diff
@@ -0,0 +1 @@
+New `OmegaConf.missing_keys` method that enables to get a config's set of missing keys represented in a dotlist style.
```
omry, line 1:
```suggestion
New `OmegaConf.missing_keys` method that returns the missing keys in a config object
```

omegaconf/omegaconf.py
omry: Let's test everything with plain dict and list, but keep a single test for DictConfig and a single test for ListConfig.

tests/test_omegaconf.py
Jasha10: The `_get_full_key` method gets the key relative to the root node.
```python
>>> cfg = OmegaConf.create({"a": {"b": {"c": "???"}}})
>>> subcfg = cfg.a
>>> print(subcfg)
{'b': {'c': '???'}}
>>> OmegaConf.missing_keys(subcfg)
{'a.b.c'}
```
I would have expected instead that the missing keys are relative to the config that is passed to `OmegaConf.missing_keys`:
```python
>>> OmegaConf.missing_keys(subcfg)
{'b.c'}  # This is what I would expect
```

Jasha10: If you both agree with me about the expected behavior, there are a few routes to implementation.
I think the easiest would be to temporarily set the `_parent` and the `_metadata.key` for the `cfg`, so that `_get_full_key` will stop moving up the tree at that point.
```python
    @staticmethod
    def missing_keys(cfg: Any) -> Set[str]:

        ...  # same as before

        old_parent = cfg.__dict__["_parent"]
        old_key = cfg._metadata.key
        try:
            cfg.__dict__["_parent"] = None
            cfg._metadata.key = None
            gather(cfg)
            return missings
        finally:
            cfg.__dict__["_parent"] = old_parent
            cfg._metadata.key = old_key
```

omry: Good point about the config root.
Yes, I think it would ideally behave as you are describing.
A solution can either be what you are proposing or allowing `_get_full_key` to take an optional root (if not provided it can use the real root, otherwise it can use the provided root).

When it hits the root (instead of when there is no parent as is the case now) it can stop building the key.

This is more complicated, but also cleaner because it's thread safe. Currently OmegaConf is not thread safe so it can work, but in general it's better form to not mutate an object when you are inspecting it, not even temporarily.

Thoughts?

Jasha10: Yes, passing a `root` to `_get_full_key` is appealing, and the functionality might be useful elsewhere in the library. Definitely harder to implement though :)

> Currently OmegaConf is not thread safe so it can work, but in general it's better form to not mutate an object when you are inspecting it, not even temporarily.

Ok, interesting. I think it will be thread-safe (though less efficient) if we use `deepcopy`:
```
    @staticmethod
    def missing_keys(cfg: Any) -> Set[str]:

        ...  # same as before

        cfg = deepcopy(cfg)
        cfg.__dict__["_parent"] = None
        cfg._metadata.key = None
        gather(cfg)
        return missings
```


> 3. Even if they do call it on a config node, it's not strictly wrong to produce the full keys.

Fair enough.

omry: In general, the use case for this function is to determine the missing keys in the config for the purposes of producing an error. This will almost certainly be performed on the entire config object and not on a subnode.

Jasha10: Yup, makes sense. Ok, resolving this as low priority.

omegaconf/omegaconf.py
Ohad31415: 👌 I've simplified it adding two tests at the end one with DictConfig and one with ListConfig.

docs/source/usage.rst
```diff
@@ -782,6 +782,24 @@ and ``OmegaConf.is_list(cfg)`` is equivalent to ``isinstance(cfg, ListConfig)``.
     >>> assert not OmegaConf.is_dict(l)
 
 
+OmegaConf.missing_keys
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Gives a set of missing keys represented in a dotlist style.
```
Jasha10, line 787:
```suggestion
``OmegaConf.missing_keys(cfg)`` returns a set of missing keys present in the input ``cfg``.
Each missing key is represented as a ``str``, using a dotlist style.
```

docs/source/usage.rst
```diff
@@ -782,6 +782,24 @@ and ``OmegaConf.is_list(cfg)`` is equivalent to ``isinstance(cfg, ListConfig)``.
...
+This utility function can be used at the end of creating configuration, after merging sources and so on,
+to check for remaining mandatory fields and prompt a proper error message.
```
Jasha10, line 789:
```suggestion
This utility function can be used after creating a config object, after merging sources and so on,
to check for missing mandatory fields and aid in creating a proper error message.
```

docs/source/usage.rst
```diff
@@ -782,6 +782,24 @@ and ``OmegaConf.is_list(cfg)`` is equivalent to ``isinstance(cfg, ListConfig)``.
...
+
+    >>> missings = OmegaConf.missing_keys({
+    ...     "foo": {"bar": "???"},
+    ...     "missing": "???",
+    ...      "list": ["a", None, "???"]
+    ... })
+    >>> assert missings == {'list[2]', 'foo.bar', 'missing'}
+
+A plain dict, list or any valid input for `OmegaConf.create` is also acceptable - the function would raise a `ValueError` exception on wrong input.
+
```
Jasha10, line 801:
```suggestion
```

Since we are using `_ensure_container` now instead of raising the `ValidationError` explicitly, I don't think we should document the exact circumstances under which an error is raised (`_ensure_container` is an internal API).

Jasha10 (changes requested):
A few docs updates:

docs/source/usage.rst
Ohad31415: The docstring of the function says that it raises ValueError on input not representing a config - maybe that is better way to put it?

Jasha10: That sounds good to me.

Ohad31415: 👌 I've changed the phrasing

Jasha10 (approved):
Looks good to me! Thanks @Ohad31415.

Jasha10:
I think we can now revisit this PR since work on OmegaConf 2.2 has started.

Jasha10:
Merged.
Thanks again @Ohad31415!

PR #730, opened 2021-05-21
OmegaConf.to_container(..., throw_on_missing) keyword arg

### LINKED ISSUE #501: Feature: throw_on_missing keyword argument for OmegaConf.to_container

### LINKED ISSUE #727: Calling `OmegaConf.to_container` on interpolation-to-missing errors
The behavior of the `OmegaConf.to_container` method has changed with respect to interpolation-to-missing nodes.

With OmegaConf v2.0.6, no exception is raised on encountering interpolation-to-missing:
```python
>>> cfg = OmegaConf.create({"missing": "???", "x": "${missing}"})
>>> OmegaConf.to_container(cfg, resolve=True)
{'missing': '???', 'x': '???'}
```

Currently on the `master` branch, this no longer works: an `InterpolationToMissingValueError` is raised.
This behavior change resulted from #545.

Jasha10:
This PR introduces a `throw_on_missing keyword` arg to the `OmegaConf.to_container` signature.
- `MandatoryMissingValue` exceptions are raised or suppressed depending on whether `throw_on_missing` is set (closes #501).
- ~`InterpolationToMissingValueError` exceptions are raised or suppressed depending on whether `throw_on_missing` is set (closes #727).~ As discussed in #727, we will keep the current behavior for raising `InterpolationResolutionError`: they are raised if and only if `OmegaConf.to_container` is called with `resolve=True`.

The `OmegaConf.to_object` method uses `throw_on_missing==True` when calling into `OmegaConf.to_container`.

This PR replaces #503, which is outdated.

TODO:
- [x] Update the docs (I'll write the docs after the discussion in #727 is resolved)
- [x] news fragment


odelalleau:
Just a note that it's going to take me a little while to get through this one (due to other priorities).

odelalleau (changes requested):
Looks mostly good to me -- the main point to discuss is whether or not we want to allow interpolations to missing values when `throw_on_missing` is True

omegaconf/dictconfig.py
```diff
@@ -720,14 +723,12 @@ def _to_object(self) -> Any:
...
+            except InterpolationToMissingValueError as e:
+                self._format_and_raise(key=k, value=None, cause=e)
             if isinstance(node, Container):
-                v = BaseContainer._to_content(
-                    node,
-                    resolve=True,
-                    enum_to_str=False,
-                    structured_config_mode=SCMode.INSTANTIATE,
-                )
+                v = OmegaConf.to_object(node)
```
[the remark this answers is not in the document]
Jasha10: Yes, good catch. Done in https://github.com/omry/omegaconf/pull/730/commits/2ca05382472664d1fd853300d6d4f4ec8ff3d352.

omegaconf/omegaconf.py
```diff
@@ -526,6 +526,7 @@ def to_container(
         cfg: Any,
         *,
         resolve: bool = False,
+        throw_on_missing: bool = False,
```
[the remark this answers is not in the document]
Jasha10: Updated in https://github.com/omry/omegaconf/pull/730/commits/40278d87b2b359e368725e36cef0075ed3b4f4f6.

tests/test_to_container.py
```diff
@@ -462,3 +472,98 @@ def test_enum_to_str_for_list(
...
+                id="to_container",
+            ),
+            param(OmegaConf.to_object, id="to_object"),
+        ],
+    )
+    @mark.parametrize(
+        "cfg",
+        [
+            # to_container: throw_on_missing
+            param(DictConfig("???"), id="dict,missing"),
```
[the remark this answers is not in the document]
Jasha10: Updated in https://github.com/omry/omegaconf/pull/730/commits/28bcd62c77b1ce1ce44f96d52bd318d0a727a41f. Thanks!

omegaconf/basecontainer.py
```diff
@@ -194,7 +223,16 @@ def convert(val: Node) -> Any:
             assert isinstance(inter, str)
             return inter
         elif conf._is_missing():
-            return MISSING
+            if throw_on_missing:
+                conf._format_and_raise(
+                    key=None,
+                    value=None,
+                    cause=MissingMandatoryValue(
+                        f"Encountered a missing {type(conf).__name__} with `throw_on_missing==True`"
```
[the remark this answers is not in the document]
Jasha10: Yes, they were different. I made the change in https://github.com/omry/omegaconf/pull/730/commits/52d5c89f7cdd31575128e6943f15d8db7942727c.

omegaconf/omegaconf.py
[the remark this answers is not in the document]
odelalleau: Thanks -- not resolving this comment for now to be sure we don't forget to remove the TODO later.

tests/test_to_container.py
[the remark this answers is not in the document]
odelalleau: I don't really want to make a big deal of this, so I'm resolving this comment, but FYI even when there are more than two items, I think usually `:` is used instead of commas (as you did in 834a0bcff56f84c83e57bb6ec2d6a6bb04b7bfda). One example is:
https://github.com/omry/omegaconf/blob/a2e25ddc7aa34bf904ce6f111accc08f50557e58/tests/test_merge.py#L260

omegaconf/basecontainer.py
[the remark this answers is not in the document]
odelalleau: Actually, looking at it more closely, I realize now that this exception can only be raised for the config `_to_container()` is  called on. If there is a missing container (e.g., `DictConfig("???")`) inside the the config, this will be caught before we reach this point.

As a result, I'm not sure anymore that my comment was relevant, sorry. I actually wonder if we really want to support this use case, because in general we want something like this to be true: 
```
OmegaConf.create(OmegaConf.to_container(cfg, throw_on_missing=False)) == cfg
```
But this is not true if `cfg` is `DictConfig("???")`.

As a result, it might be better to always make it an error. But it'd be good to have @omry's opinion on it.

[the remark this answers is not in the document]
omry: Let's discuss this on #727. See my last comment there.

As I said there, to_container is losing type safety. Missing containers are used in conjunction with Structured Configs (when you have a missing SC field):
```python
@dataclass
class Config:
  user: User = MISSING # DictConfig("???") with a backing User type.
```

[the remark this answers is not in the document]
odelalleau: > Missing containers are used in conjunction with Structured Configs (when you have a missing SC field):
> 
> ```python
> @dataclass
> class Config:
>   user: User = MISSING # DictConfig("???") with a backing User type.
> ```

Yes but in your example this situation still won't occur because:
- `to_container(cfg, throw_on_missing=True)` would raise an exception when we detect that `cfg.user` is missing (this check is done before we get a chance to call `to_container(cfg._get_node("user"))`)
- `to_container(cfg.user, throw_on_missing=True)` would raise an exception before `to_container()` is called (trying to access `cfg.user` will fail).

As far as I can tell this error can only be triggered if a user ends up manipulating directly a `DictConfig("???")`, but I'm not sure it's something that is supposed to happen in practice. Which is why I was thinking that maybe we shouldn't even try to support it.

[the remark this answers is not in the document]
omry: > As far as I can tell this error can only be triggered if a user ends up manipulating directly a DictConfig("???"), but I'm not sure it's something that is supposed to happen in practice. Which is why I was thinking that maybe we shouldn't even try to support it.

I agree. and this is not a part of the public API.
In fact, DictConfig itself is not meant to be instantiated directly at all. The documented way to create configs is OmegaConf.create().

tests/test_to_container.py
```diff
@@ -230,22 +227,6 @@ def test_string_interpolation_with_readonly_parent() -> None:
...
-    [
-        param(DictConfig(content="${bar}"), "${bar}", id="DictConfig"),
-        param(
-            OmegaConf.create({"foo": DictConfig(content="${bar}")}),
-            {"foo": "${bar}"},
-            id="nested_DictConfig",
-        ),
-    ],
-)
-def test_to_container_missing_inter_no_resolve(src: Any, expected: Any) -> None:
```
odelalleau, line 234: Why is this test being removed?

tests/test_to_container.py
```diff
@@ -487,83 +485,60 @@ class TestThrowOnMissing:
...
-    def test_interpolation_to_missing_throws(self, op: Any, src: Any) -> None:
-        """
-        Interpolation to missing: Test that InterpolationToMissingValueError
-        is raised when resolve==True and throw_on_missing==True.
-        """
+    def test_resolve_raises(self, op: Any, src: Any) -> None:
         cfg = OmegaConf.create(src)
-        with raises(InterpolationToMissingValueError):
-            op(cfg.subcfg)
+        with raises(InterpolationResolutionError):
```
odelalleau, line 535: Can you add the specific expected exception type as a test parameter, so that interpolations to missing test explicitly for `InterpolationToMissingValueError`?

tests/test_to_container.py
```diff
@@ -474,6 +455,23 @@ def test_enum_to_str_for_list(
...
 
 
+@mark.parametrize(
+    "cfg",
+    [
+        # to_container: throw_on_missing
+        # param(DictConfig("???"), id="dict:missing"),
+        param(DictConfig({"a": "???"}), id="dict:missing_value"),
+        param(DictConfig({"a": {"b": "???"}}), id="dict:nested"),
+        # param(ListConfig("???"), id="list:missing"),
```
odelalleau, line 465: Are you testing the two commented test cases elsewhere?

odelalleau (changes requested):

omegaconf/basecontainer.py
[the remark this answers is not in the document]
odelalleau: So, actually, looking at it more closely, this situation (a `DictConfig("???")` being processed by `to_container()`) can occur with structured configs, because the `to_container()` implementation takes a different code path (compared to non-structured configs).
As a result, we should indeed support it => resolving this comment.

omegaconf/omegaconf.py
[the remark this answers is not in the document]
odelalleau: Just a reminder to remove the TODO

omegaconf/basecontainer.py
```diff
@@ -186,6 +188,33 @@ def convert(val: Node) -> Any:
...
+                conf._format_and_raise(key=key, value=None, cause=e)
+            assert isinstance(node, Node)
+            if resolve:
+                try:
+                    node = node._dereference_node()
+                except InterpolationToMissingValueError as e:
+                    if throw_on_missing:
+                        conf._format_and_raise(key=key, value=None, cause=e)
+                    else:
+                        return "???"
```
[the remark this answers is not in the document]
odelalleau: Has been addressed => resolving

omegaconf/basecontainer.py
```diff
@@ -186,6 +188,30 @@ def convert(val: Node) -> Any:
 
             return value
 
+        def get_node_value(key: Union[DictKeyType, int]) -> Any:
+            try:
+                node = conf._get_node(key, throw_on_missing_value=throw_on_missing)
+            except MissingMandatoryValue as e:
+                conf._format_and_raise(key=key, value=None, cause=e)
```
omry, line 195: Generally speaking, it's better to format exceptions at high level functions, otherwise you risk formatting them multiple times.

(e.g: in OmegaConf.to_container() or other high level functions calling _to_content()).

omegaconf/omegaconf.py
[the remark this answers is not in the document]
omry: Yup, I think we have a decision.

tests/__init__.py
```diff
@@ -99,6 +99,17 @@ class FoobarParams:
...
+class NestedInterpolationToMissing:
+    name: str = MISSING
+
+    @dataclass
+    class BazParams:
+        baz: str = "${..name}"
+
+    subcfg: BazParams = BazParams()
+
+
```
omry, line 112: You can test this without a full fledged dataclass.
See test_dict_values_with_missing_value.

tests/test_to_container.py
omry: In general, do not commit commented out code. This looks like an oversight. Either remove the comments and explain why they are removed or enable the tests.

Jasha10: Sorry about that. Those tests are failing because `"???" == DictConfig("???")` evaluates to `False`.
Same deal for `ListConfig`.

Here are the options I see for this:
1. submit a PR to make sure `"???" == DictConfig("???")` is `True`, and mark these cases as `xfail` for now,
2. delete these test cases, or
3. change the testing strategy to work around the problem with `DictConfig.__eq__`.

odelalleau: Option 1 seems best to me because currently this equality check works with `ValueNode`s, so it seems to me that it should also be valid for containers.

tests/test_to_container.py
Jasha10: My thinking was that this is redundant with the new `TestResolveBadInterpolation.test_no_resolve` method below.
However, it looks like I missed one of the test cases (the case of `DictConfig(content="${bar}")`). I'll make sure to add it back in, either by restoring this test or by adding to the `TestResolveBadInterpolation` class below.

tests/test_to_container.py
Jasha10: Ok, sounds good to me :)

omegaconf/basecontainer.py
Jasha10: Hmm... If I do formatting in `OmegaConf.to_container()`, then I will not be able to set the `$KEY` properly (since the for loop iterating through keys happens in this lower-level `_to_content` function).

omegaconf/omegaconf.py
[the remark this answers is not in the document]
Jasha10: Done in https://github.com/omry/omegaconf/pull/730/commits/df47f0e0a6e4b1b55128f3a0de2a0a3f7e1a4744

tests/test_to_container.py
Jasha10: Ok the tests are uncommented (in https://github.com/omry/omegaconf/pull/730/commits/0df8e1a97ae7033b3679434f3544d7f4e1eeca09).

omegaconf/basecontainer.py
omry: Is this related to this change? `raise MissingMandatoryValue("Missing mandatory value: $KEY")` ?
Why not just use the key directly when raising the exception?

omry:
1. Update the PR to indicate that issue (s) it's closing.
2. I think we should be able to get the docs and news fragment already.

omegaconf/basecontainer.py
Jasha10: > Is this related to this change?

Yes, I think it's related -- In the past, a `MissingMandatoryValue` would not have been raised.

> Why not just use the key directly when raising the exception?

Something like this?
```python
raise MissingMandatoryValue(f"Missing mandatory value: {key}")
```

My point was: if we call `format_and_raise` in `OmegaConf.to_container` instead of here in `BaseContainer._to_content`, then the `format_and_raise` function will not have the correct `key` passed in to it (since the high-level function does not know what `key` caused the exception).

tests/test_to_container.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/730/commits/6f696f3b56d2cc5783eb2a490a103fbd00b8ec0b.

tests/test_to_container.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/730/commits/05221aa1235b298ae6bd5232af9f60bf27bd2ac7.

omry:
Unsubscribing for now, please request review when ready.

tests/__init__.py
Jasha10: That works for the `to_container` tests, but I think a dataclass is necessary for testing `OmegaConf.to_object`/`DictConfig._to_object`.

omegaconf/basecontainer.py
omry: > Something like this?
> 
> ```python
> raise MissingMandatoryValue(f"Missing mandatory value: {key}")
> ```

Yes.
Depending on exactly what is going on here, you may need to pass the key from the outside or use the key from the offending node (node._key()).

tests/__init__.py
omry: It is. Resolving.

tests/test_errors.py
```diff
@@ -1288,6 +1290,94 @@ def finalize(self, cfg: Any) -> None:
...
+            create=lambda: ListConfig(["???"]),
+            op=lambda cfg: OmegaConf.to_container(cfg, throw_on_missing=True),
+            exception_type=MissingMandatoryValue,
+            msg="Missing mandatory value: 0",
+            key=0,
+            full_key="[0]",
+            child_node=lambda cfg: cfg._get_node(0),
+        ),
+        id="to_container:throw_on_missing,list_item",
+    ),
```
omry, line 1380: This is adding almost 200 lines to an already huge test file.
We should refactor this file somehow at some point.

tests/test_errors.py
```diff
@@ -1288,6 +1290,94 @@ def finalize(self, cfg: Any) -> None:
         ),
         id="to_object:structured-missing-field",
     ),
+    param(
+        Expected(
+            create=lambda: OmegaConf.structured(NestedInterpolationToMissing).subcfg,
```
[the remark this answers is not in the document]
omry: Gotcha.
What if you change the NestedInterpolationToMissing to:
```python
**@dataclass
class NestedInterpolationToMissing:
    @dataclass
    class BazParams:
        baz: str = "${..name}"

    subcfg: BazParams = BazParams()

    name: str = MISSING
```

tests/test_to_container.py
odelalleau: Hmm it seems like you duplicated the list of test parameters, why? (when using a class it's ok to have a function ignore one of the test parameters, if it's not relevant -- at least it seems better to me than duplicating parameters)

tests/test_to_container.py
odelalleau: I may be missing something, but I don't see the test case with just `DictConfig(content="${bar}"))` in this commit (?)

docs/source/usage.rst
```diff
@@ -621,7 +646,7 @@ Note that here, ``container["structured_config"]`` is actually an instance of
 ``MyConfig``.
 
 The call ``OmegaConf.to_object(conf)`` is equivalent to
-``OmegaConf.to_container(conf, resolve=True,
+``OmegaConf.to_container(conf, resolve=True, throw_on_missing=True
```
odelalleau, line 649:
```suggestion
``OmegaConf.to_container(conf, resolve=True, throw_on_missing=True,
```

odelalleau (changes requested):
One minor typo in the doc

tests/test_to_container.py
Jasha10: > ...at least it seems better to me than duplicating parameters

I guess I got carried away trying to make sure that each function only accepts params that it uses. I'll deduplicate :)

tests/test_to_container.py
Jasha10: Nope, you're absolutely right. Thanks for keeping an eye on the details...

docs/source/usage.rst
Jasha10: Thanks.

tests/test_errors.py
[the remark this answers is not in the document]
omry: awesome.

tests/test_to_container.py
Jasha10: Deduplicated in https://github.com/omry/omegaconf/pull/730/commits/e11bb050e8f89dd63ff1178bad7b9249e6280c2f.

tests/test_to_container.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/730/commits/a059a8fc07b733ff3c52b413b0772009ed8c772a.

omegaconf/basecontainer.py
```diff
@@ -189,16 +191,52 @@ def convert(val: Node) -> Any:
...
             assert isinstance(inter, str)
             return inter
-        elif conf._is_missing():
-            return MISSING
-        elif isinstance(conf, DictConfig):
+
+        if resolve:
+            _conf = conf._dereference_node()
+            assert isinstance(_conf, Container)
+            conf = _conf
```
odelalleau, line 237: I'm curious if this might also be fixing a bug with *valid* top-level interpolations? (it seems that the previous code wasn't calling `_dereference_node()`)

If this is correct, then tests should be added for Dict/ListConfig interpolations.

Side note: `typing.cast()` would make this piece of code cleaner -- but we would lose the assert safety. So far we haven't been using `typing.cast()` so not suggesting you should change it, just an FYI.

omegaconf/basecontainer.py
```diff
@@ -189,16 +191,52 @@ def convert(val: Node) -> Any:
...
+        elif conf._is_missing():
+            if throw_on_missing:
+                conf._format_and_raise(
+                    key=None,
+                    value=None,
+                    cause=MissingMandatoryValue("Missing mandatory value"),
+                )
+            else:
+                return MISSING
         elif conf._is_interpolation() and not resolve:
```
odelalleau, line 229: Minor optimization:
```suggestion
        elif not resolve and conf._is_interpolation():
```

tests/test_to_container.py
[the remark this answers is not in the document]
Jasha10: Interesting... I hadn't noticed the use of multiple colons before.

I was building from what I'd seen in `test_errors.py`, where there are lots of mixed colon+comma id's (e.g. `id="structured:merge,invalid_field_type"`).
It feels almost like colons are a sort of "nesting" of test-case categories, whereas commas feel like taking an intersection of categories.

omegaconf/basecontainer.py
Jasha10: > I'm curious if this might also be fixing a bug with _valid_ top-level interpolations? (it seems that the previous code wasn't calling `_dereference_node()`)

The previous code was calling `_dereference_node()` inside the `get_node_value` function defined near the top of the `BaseContainer._to_content` method, and inside the `DictConfig._to_object` method (for the case of a `DictConfig` nested inside a structured config, when `SCMode.INSTANTIATE` is being used).

This reminds me of our [previous discussion](https://github.com/omry/omegaconf/pull/730#discussion_r641741533) about about `DictConfig("???")`.
You remarked that such a config could validly appear nested inside a structured config.
The same applies for `DictConfig("${foo}")`: it can appear nested inside a structured config.

That being said, I think that our documented APIs do not currently allow for users to handle `DictConfig("???")` or `DictConfig("${foo}")` directly (as attempting to access them within a structured config would result in `MissingMandatoryValue` in the first case, and in a resolved value or an `InterpolationResolutionError` in the second case). In other words, I think that neither `DictConfig("???")` nor `DictConfig("${foo}")` will ever be passed directly to `OmegaConf.to_container` unless a non-standard/internal APIs is being used.

Currently at the top level of the `BaseContainer._to_content` method there is an `elif conf._is_missing()` branch and an `if resolve` branch (which you highlighted above); these branches are precisely for this situation, where `DictConfig("???")` or `DictConfig("${foo}")` is being passed directly to `OmegaConf.to_container`.

So I think the question is, should we support this even though it's not part of the public API? The alternative would be to fail noisily (with e.g. a `ValueError`) if `DictConfig("???")` or `DictConfig("${foo}")` is passed directly to `OmegaConf.to_container`.

> Side note: `typing.cast()` would make this piece of code cleaner -- but we would lose the assert safety. So far we haven't been using `typing.cast()` so not suggesting you should change it, just an FYI.

Yup, `typing.cast()` came to my mind as I was writing this code :)

tests/test_to_container.py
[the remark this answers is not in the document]
odelalleau: Makes sense :)

omegaconf/basecontainer.py
odelalleau: Just to make my point clear: I think that if we test the case `DictConfig(content="${bar}")` where the `bar` node doesn't exist (current test ID `dictconfig_interp_key_error`), I think we should also test the case where `bar` exists. And I suspect it didn't work before because the code was iterating on the DictConfig keys with `for key in conf.keys():` and `keys()` is empty if `conf` is an interpolation.

An orthogonal question is whether we actually care about this use case -- I'm not 100% sure, but in doubt I'd rather have test cases covering the current behavior, so that we notice if it changes in the future.

Jasha10: > Just to make my point clear: I think that if we test the case `DictConfig(content="${bar}")` where the `bar` node doesn't exist (current test ID `dictconfig_interp_key_error`), I think we should also test the case where `bar` exists.

You're right, I'll add some tests.

> And I suspect it didn't work before because the code was iterating on the DictConfig keys with `for key in conf.keys():` and `keys()` is empty if `conf` is an interpolation (but maybe I'm missing something -- I didn't try it).

It actually did work even before commit https://github.com/omry/omegaconf/pull/730/commits/a059a8fc07b733ff3c52b413b0772009ed8c772a was added.
The following works after `git checkout a059a8f~1`, checking out the previous commit:
```
>>> cfg = OmegaConf.create({"a": {"b": 123}, "c": DictConfig("${a}")})
>>> print(OmegaConf.to_container(cfg, resolve=True))
{'a': {'b': 123}, 'c': {'b': 123}}
```
It works because calling `to_container(cfg, resolve=True)` will dereference the interpolation (in `get_node_value`, with `key=="c"`) before calling `_to_content` on that node.

> An orthogonal question is whether we actually care about this use case -- I'm not 100% sure, but in doubt I'd rather have test cases covering the current behavior, so that we notice if it changes in the future.

Sounds good to me.

Jasha10: > You're right, I'll add some tests.

I noticed just now that in the `test_to_container` function near the top of `test_to_container.py`, we have the following test cases:
```
        param(
            {"foo": DictConfig(content="${bar}"), "bar": 10},
            {"foo": "${bar}", "bar": 10},
            {"foo": 10, "bar": 10},
            id="dict_inter_dictconfig",
        ),
...
        param(
            {"foo": ListConfig(content="${bar}"), "bar": 10},
            {"foo": "${bar}", "bar": 10},
            {"foo": 10, "bar": 10},
            id="dict_inter_listconfig",
        ),
```
I think that this covers the use-case in question, though it's a bit awkward because the integer `10` is not of the correct type (it should be a dict or list).

omegaconf/basecontainer.py
Jasha10: I tried it in [this commit](https://github.com/Jasha10/omegaconf/commit/ae306919e96fb16109ded231479e705abbd96ef7) on a separate branch (which is based off of this PR branch).
The tests pass, but we are losing some information in the error messages. For example, here is the error before the commit:
```python
>>> OmegaConf.to_object(OmegaConf.create({"a": {"b": "???"}}))
...
omegaconf.errors.MissingMandatoryValue: Missing mandatory value: b
    full_key: a.b
    object_type=dict
```
and here is the error after the commit:
```python
>>> OmegaConf.to_object(OmegaConf.create({"a": {"b": "???"}}))
...
omegaconf.errors.MissingMandatoryValue: Missing mandatory value: b
    full_key:
    object_type=dict
```
There is no longer information about the `full_key` that caused the error.

> Generally speaking, it's better to format exceptions at high level functions, otherwise you risk formatting them multiple times.

Is it really that bad to format multiple times?

omegaconf/basecontainer.py
odelalleau: This is what I was referring to (I just verified that it was indeed not working previously, and your code fixes it):
```
cfg = OmegaConf.create({"foo": DictConfig(content="${bar}"), "bar": {"a": 0}})._get_node("foo")
assert OmegaConf.to_object(cfg) == {"a": 0}
```
==> I think there should be a test for it (both for Dict & ListConfig)

omegaconf/dictconfig.py
```diff
@@ -469,7 +469,7 @@ def _get_node(
             if throw_on_missing_key:
                 raise ConfigKeyError(f"Missing key {key}")
         elif throw_on_missing_value and value._is_missing():
-            raise MissingMandatoryValue("Missing mandatory value")
+            raise MissingMandatoryValue("Missing mandatory value: $KEY")
```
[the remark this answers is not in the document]
Jasha10: See my [comment](https://github.com/omry/omegaconf/pull/730/files#r648662687) above.

omegaconf/basecontainer.py
Jasha10: Oh, got it :)
Thanks for doing the digging.
I've added the tests in https://github.com/omry/omegaconf/pull/730/commits/f98b1f00e7ae0124395cd3e078d4d03ac5fea2b1.

tests/test_to_container.py
```diff
@@ -187,14 +208,25 @@ def test_scmode(
...
 def test_to_container(src: Any, expected: Any, expected_with_resolve: Any) -> None:
     if expected is None:
         expected = src
     if expected_with_resolve is None:
         expected_with_resolve = expected
-    cfg = OmegaConf.create(src)
+    if isinstance(src, Container):
+        cfg = src
+    else:
+        cfg = OmegaConf.create(src)
```
odelalleau, line 229: Would `cfg = _ensure_container(src)` work here?

Jasha10: Yup, that works. Updated in https://github.com/omry/omegaconf/pull/730/commits/e67739babdb19837e7b7c877698f58e6bc4d25fb.

odelalleau (approved):
Good on my side, thanks @Jasha10!

omry (approved):
lgtm, thanks.

PR #735, opened 2021-06-01
typing: DictConfig.keys to return AbstractSet

Jasha10:
This PR does two things:
- change return type of `DictConfig.keys` from `Any` to `AbstractSet[DictKeyType]`, and
- ensure that `DictConfig.keys` always returns an `AbstractSet`. Previously, an empty `list()` was returned in the case where the the config was an interpolation/missing/none.

### DIFF
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -206,7 +206,7 @@ class BaseContainer(Container, ABC):
             ):
                 return conf._to_object()
 
-            retdict: Dict[str, Any] = {}
+            retdict: Dict[DictKeyType, Any] = {}
             for key in conf.keys():
                 node = conf._get_node(key)
                 assert isinstance(node, Node)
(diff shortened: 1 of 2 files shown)


omry:
>  Previously, an empty list() was returned in the case where the the config was an interpolation/missing/none.

What in the PR fixes that?

Jasha10:
> > Previously, an empty list() was returned in the case where the the config was an interpolation/missing/none.
> 
> What in the PR fixes that?

On line 509 of `dictconfig.py`, I've replaced `return list()` with `return set()`.

omry:
Ok.
What is the motivation here?
Is keys() in general expected to return an AbstractSet?

Jasha10:
Yes, exactly -- keys is supposed to return an AbstractSet.
I learned about this because mypy complained when I tried to replace `def keys(self) -> Any:` with `def keys(self) -> Iterator[DictKeyType]:`

omry (approved):
roger.

PR #737, opened 2021-06-02
BaseContainer: error if `parent` has wrong type

Jasha10:
Previously the following examples either gave an `AssertionError` or returned a config object with bad metadata:
```python
OmegaConf.create( {}, {"a", "b"})  # bad metadata
OmegaConf.create({"zz": 10}, {"a", "b"})  # used to raise AssertionError
OmegaConf.create(Dog, {"a", "b"})  # used to raise AssertionError
```
Now all three examples raise `ConfigTypeError`.

### DIFF
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -28,6 +28,7 @@ from ._utils import (
 from .base import Container, ContainerMetadata, DictKeyType, Node, SCMode
 from .errors import (
     ConfigCycleDetectedException,
+    ConfigTypeError,
     MissingMandatoryValue,
     ReadonlyConfigError,
     ValidationError,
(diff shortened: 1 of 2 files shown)


omegaconf/basecontainer.py
```diff
@@ -42,6 +43,8 @@ class BaseContainer(Container, ABC):
     _resolvers: Dict[str, Any] = {}
 
     def __init__(self, parent: Optional["Container"], metadata: ContainerMetadata):
+        if not (parent is None or isinstance(parent, Container)):
+            raise ConfigTypeError("Node parent should be a Container")
```
omry, line 47:
```suggestion
            raise ConfigTypeError("Parent node type is not omegaconf.Container")
```

omry (approved):
Nit about error message.

omry (approved):

PR #738, opened 2021-06-02
ListConfig append deepcopies input config nodes

### LINKED ISSUE #601: ListConfig.append() does not deep copy nodes
**Describe the bug**

If `cfg` is a `ListConfig`, `cfg.append(node)` does not deep copy the node (contrary to `cfg[idx] = node`).
This can cause several issues, including the following that raises a max recursion error:

```python
c = OmegaConf.create({"x": []})
c.x.append(c)
c._get_flag("test")  # (RecursionError)
```

**To Reproduce**
See above

**Expected behavior**
Similar behavior to what happens with assigment:

```python
c = OmegaConf.create({"x": [None]})
c.x[0] = c
c._get_flag("test")  # None (no error)
```

**Additional context**
See related discussion in https://github.com/omry/omegaconf/pull/599

omry:
Closes #601

Not worried about perf:
This is something we are already doing for DictConfig, and ListConfig is not used nearly as much.

### DIFF
--- a/news/601.bugfix
+++ b/news/601.bugfix
@@ -1,0 +1,1 @@
+ListConfig append now copies input config nodes
--- a/omegaconf/listconfig.py
+++ b/omegaconf/listconfig.py
@@ -252,6 +252,8 @@ class ListConfig(BaseContainer, MutableSequence[Any]):
             index = len(self)
             self._validate_set(key=index, value=item)
 
+            if isinstance(item, Node):
+                item = copy.deepcopy(item)
             node = _maybe_wrap(
                 ref_type=self.__dict__["_metadata"].element_type,
                 key=index,
--- a/omegaconf/listconfig.py
+++ b/omegaconf/listconfig.py
@@ -253,7 +253,10 @@ class ListConfig(BaseContainer, MutableSequence[Any]):
             self._validate_set(key=index, value=item)
 
             if isinstance(item, Node):
-                item = copy.deepcopy(item)
+                do_deepcopy = not self._get_flag("no_deepcopy_set_nodes")
+                if do_deepcopy:
+                    item = copy.deepcopy(item)
+
             node = _maybe_wrap(
                 ref_type=self.__dict__["_metadata"].element_type,
                 key=index,
--- a/tests/test_basic_ops_dict.py
+++ b/tests/test_basic_ops_dict.py
@@ -18,6 +18,7 @@ from omegaconf import (
     _utils,
     flag_override,
     open_dict,
+    AnyNode,
 )
 from omegaconf._utils import _ensure_container
 from omegaconf.basecontainer import BaseContainer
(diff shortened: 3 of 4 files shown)


Jasha10:
The `BaseContainer._set_item_impl` method uses a `"no_deepcopy_set_nodes"` flag to decide whether deepcopy shold be done.
I think that `ListConfig.append` should probably do the same...

omry:
> The `BaseContainer._set_item_impl` method uses a `"no_deepcopy_set_nodes"` flag to decide whether deepcopy shold be done.
> I think that `ListConfig.append` should probably do the same...

This is an optimization, so not super critical because ListConfigs are not common.
I added it anyway. One thing I did not add was a logic that copies anyway if the assigned node is from the same config tree. I think this is handing some edge case but I would rather see it first before I enable the same logic in append.

Jasha10:
> One thing I did not add was a logic that copies anyway if the assigned node is from the same config tree. I think this is handing some edge case but I would rather see it first before I enable the same logic in append.

I commented out the edge-case logic to see what test would fail. The following case from `test_merge.py` would fail:
```python
input1 = {"n": {"a": 10}, "i": "${n}"}
input2 = {"i": {"b": 20}}
expected = {"n": {"a": 10}, "i": {"a": 10, "b": 20}}
assert expected == OmegaConf.merge(input1, input2)
assert expected == OmegaConf.unsafe_merge(input1, input2)  # fails: unsafe_merge returns {'n': {'a': 10, 'b': 20}, 'i': {'a': 10, 'b': 20}} 
```
It seems that this edge case only happens when merging into an interpolation; the deepcopy prevents the merge from mutating other parts of the tree.
`ListConfig.append` does not mutate other parts of the tree, so I think we're safe.

Jasha10 (approved):

news/601.bugfix
```diff
@@ -0,0 +1 @@
+ListConfig append now copies input config nodes
```
odelalleau, line 1: Easier to read (at least to me):
```suggestion
ListConfig.append now copies input config nodes
```

odelalleau (commented):
I guess I'm a few minutes late on the review, sorry. Feel free to ignore my minor comment below.

At high level I'm fine with it but I wonder if `append()` wouldn't be better implemented as "append `None` to the underlying content list, then call `set_item()` on the last index". There's more stuff happening in `_set_item_impl()` that I don't fully grasp, which may -- or may not -- matter as well for `append()`.

omry:
> I guess I'm a few minutes late on the review, sorry. Feel free to ignore my minor comment below.
> 
> At high level I'm fine with it but I wonder if `append()` wouldn't be better implemented as "append `None` to the underlying content list, then call `set_item()` on the last index". There's more stuff happening in `_set_item_impl()` that I don't fully grasp, which may -- or may not -- matter as well for `append()`.

Good call.
It did cross my mind briefly but I didn't try.
Got it to work in #740.

PR #740, opened 2021-06-03
Updated ListConfig.append to use _set_item_impl

omry:


omegaconf/listconfig.py
```diff
@@ -247,25 +247,11 @@ def __setitem__(self, index: Union[int, slice], value: Any) -> None:
...
-                value=item,
-                is_optional=_is_optional(item),
-                parent=self,
-            )
-            self.__dict__["_content"].append(node)
+            index = len(self.__dict__["_content"])
+            self.__dict__["_content"].append(None)
+            self._set_item_impl(index, item)
         except Exception as e:
+            del self.__dict__["_content"][index]
```
odelalleau, line 254: Looks good to me, I would just move the first couple of lines outside of the `try` block, and maybe assign `self.__dict__["_content"]` for readability. Something like (this time I tested, it works :) ):
```python
    def append(self, item: Any) -> None:
        content = self.__dict__["_content"]
        index = len(content)
        content.append(None)
        try:
            self._set_item_impl(index, item)
        except Exception as e:
            del content[index]
            self._format_and_raise(key=index, value=item, cause=e)
            assert False
```

omry: done.

odelalleau (approved):
Go!

Jasha10 (approved):
Looks good :)

PR #742, opened 2021-06-03
Make missing {Dict,List}Config compare equal to "???"

### LINKED ISSUE #741: DictConfig("???") == "???" evaluates to False
**Describe the bug**
Currently we have the following:
```python
>>> DictConfig("???") == "???"
False
>>> ListConfig("???") == "???"
False
```
It is proposed to change `DictConfig.__eq__` and `ListConfig.__eq__` so that the above are true. This would be consistent with the behavior for `ValueNode`s:
```python
>>> StringNode("???") == "???"
True
>>> FloatNode("???") == "???"
True
```

**Motivation**

This change would make it easier to write certain unit tests, e.g. `node == expected` in the case where `expected="???"`. This came up in the comments for #730.
(Of course, in testing, there are other options, such as using `node._is_missing()` instead of `node == "???"`)

Anyway, I think that it is good for an API to be consistent, even if it's an internal API.

This change should not affect typical OmegaConf users. A typical user will encounter a `MissingMandatoryValue` exception before being able to compare a missing node with a `"???"` string.

**Additional context**
- OmegaConf version: 2.1.rc1

Jasha10:
Closes #741.

I've added an additional branch to each of `DictConfig.__eq__` and `ListConfig.__eq__`:
```
if self._is_missing:
    return _is_missing_literal(other)
```

This allows for `assert DictConfig("???") == "???"` and `assert ListConfig("???") == "???"`.

### DIFF
--- a/tests/test_config_eq.py
+++ b/tests/test_config_eq.py
@@ -109,6 +109,8 @@ def test_eq(i1: Any, i2: Any) -> None:
 def test_missing_container_string_eq(cfg: Any, other: Any) -> None:
     assert cfg == other
     assert other == cfg
+    assert not (cfg != other)
+    assert not (other != cfg)
 
 
 @mark.parametrize(
--- a/tests/test_config_eq.py
+++ b/tests/test_config_eq.py
@@ -100,6 +100,18 @@ def test_eq(i1: Any, i2: Any) -> None:
 
 
 @mark.parametrize(
+    "cfg,other",
+    [
+        param(DictConfig("???"), "???", id="missing_dictconfig"),
+        param(ListConfig("???"), "???", id="missing_listconfig"),
+    ],
+)
+def test_missing_container_string_eq(cfg: Any, other: Any) -> None:
+    assert cfg == other
+    assert other == cfg
+
+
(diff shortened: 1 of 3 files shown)


tests/test_config_eq.py
```diff
@@ -99,6 +99,18 @@ def eq(a: Any, b: Any) -> None:
...
+@mark.parametrize(
+    "cfg,other",
+    [
+        param(DictConfig("???"), "???", id="missing_dictconfig"),
+        param(ListConfig("???"), "???", id="missing_listconfig"),
+    ],
+)
+def test_missing_container_string_eq(cfg: Any, other: Any) -> None:
+    assert cfg == other
+    assert other == cfg
```
omry, line 111: for completeness, add neq tests.
```suggestion
def test_missing_container_string_eq(cfg: Any, other: Any) -> None:
    assert cfg == other
    assert other == cfg
    assert not (cfg != other)
    assert not (other != cfg)
```

omry (approved):
lgtm.
@odelalleau , wdyt?

odelalleau (approved):

PR #744, opened 2021-06-07
IntegerNode/FloatNode ValidationError message: include value type

### LINKED ISSUE #743: FloatNode/IntegerNode ValidationError message: include value type
When creating a `FloatNode` from a value that has unsupported type, the raised `ValidationError` might be confusing if the value "looks like a float".
For example, `numpy.float32` is not currently supported as a value type for creation of `FloatNode`s (though there is an open issue, #725, to change this), and the error message on creation does not clearly indicate the problem:
```python
>>> import numpy as np
>>> DictConfig({"a": np.float32(.9)}, element_type=float)
...
omegaconf.errors.ValidationError: Value '0.9' could not be converted to Float
    full_key: a
    object_type=dict
```

Based on this [comment](https://github.com/omry/omegaconf/issues/725#issuecomment-855154981_), it is proposed to change the error
```
ValidationError: Value '0.9' could not be converted to Float
```
to something like:
```
ValidationError: Value '0.9' of type 'np.float32` could not be converted to Float
```

The same applies to `IntegerNode`.

Jasha10:
Closes #743 

The `ValidationError` raised by `IntegerNode` and `FloatNode` now contains the type of the value that triggered the error.

### DIFF
--- a/news/743.api_change
+++ b/news/743.api_change
@@ -1,1 +1,1 @@
-In the error message when assigning a mis-typed value to a float-typed or int-typed field, include the type of the mis-typed value.
+Improved error message when assigning an invalid value to int or float config nodes
--- a/news/743.api_change
+++ b/news/743.api_change
@@ -1,0 +1,1 @@
+In the error message when assigning a mis-typed value to a float-typed or int-typed field, include the type of the mis-typed value.
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -95,6 +95,7 @@ class ConcretePlugin(Plugin):
     @dataclass
     class FoobarParams:
         foo: int = 10
+        bar: float = 123.456
 
     params: FoobarParams = FoobarParams()
 
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -100,11 +100,6 @@ class ConcretePlugin(Plugin):
 
 
 @dataclass
-class TypedFields:
-    bar: float = 123.456
-
-
-@dataclass
 class StructuredWithMissing:
     num: int = MISSING
     opt_num: Optional[int] = MISSING
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -95,9 +95,13 @@ class ConcretePlugin(Plugin):
     @dataclass
     class FoobarParams:
         foo: int = 10
-        bar: float = 123.456
 
     params: FoobarParams = FoobarParams()
+
+
+@dataclass
+class TypedFields:
+    bar: float = 123.456
 
 
 @dataclass
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -821,7 +821,7 @@ def format_and_raise(
     _raise(ex, cause)
 
 
-def type_str(t: Any) -> str:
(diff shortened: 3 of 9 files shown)


tests/__init__.py
```diff
@@ -99,6 +99,11 @@ class FoobarParams:
     params: FoobarParams = FoobarParams()
 
 
+@dataclass
+class TypedFields:
+    bar: float = 123.456
+
+
```
omry, line 106: As an alternative to creating a type here, you can use:
```
DictConfig({"bar": FloatNode(123.456)})
```

omry (commented):
looks good, one nit.

tests/__init__.py
Jasha10: Ok, I made the change in https://github.com/omry/omegaconf/pull/744/commits/f1b441d0a9cf3fc4d148200a1f379601bf1b76cc.

Jasha10:
Currently we have this error message for numpy floats:
```python
>>> cfg.bar = np.float32(0.9)
...
omegaconf.errors.ValidationError: Value '0.9' of type 'float32' could not be converted to Float
```

I've been experimenting with a diff to `_utils.format_and_raise` that includes the module name in the error message:
```python
>>> cfg.bar = np.float32(0.9)
...
omegaconf.errors.ValidationError: Value '0.9' of type 'numpy.float32' could not be converted to Float
```

The diff looks like this:
```python
diff --git a/omegaconf/_utils.py b/omegaconf/_utils.py
index d3dafe9..72bcc07 100644
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -768,13 +768,18 @@ def format_and_raise(
         ref_type = get_ref_type(node)
         ref_type_str = type_str(ref_type)

+    value_type = type(value)
+    if hasattr(value_type, "__module__") and value_type.__module__ != "builtins":
+        value_type_str = value_type.__module__ + "." + value_type.__name__
+    else:
+        value_type_str = value_type.__name__
     msg = string.Template(msg).safe_substitute(
         REF_TYPE=ref_type_str,
         OBJECT_TYPE=object_type_str,
         KEY=key,
         FULL_KEY=full_key,
         VALUE=value,
-        VALUE_TYPE=f"{type(value).__name__}",
+        VALUE_TYPE=value_type_str,
         KEY_TYPE=f"{type(key).__name__}",
     )
```
@omry, what do you think?

omry:
Since 2.1.0 is already out, please add a news fragment. otherwise looks good.

omry:
> @omry, what do you think?

lgtm, extract it to a function in _utils.
`_fully_qualified_type_name`?

Alternatively, and probably better - is to add support for fully-qualified-name to `_utils.type_str`.
It can handle compound types and annotations as well.

Either way works.

news/743.api_change
```diff
@@ -0,0 +1 @@
+In the error message when assigning a mis-typed value to a float-typed or int-typed field, include the type of the mis-typed value.
```
omry, line 1:
```suggestion
Improved error message when assigning an invalid value to int or float config nodes
```

omry (approved):

PR #749, opened 2021-06-11
Improve Optional[type] validation

### LINKED ISSUE #460: Dict[str, Optional[str]] fails
```python
from dataclasses import *
from typing import *
@dataclass
class Foo:
    a : Dict[str, Optional[str]]

OmegaConf.structured(Foo)
```

Raises an assertion failure

### LINKED ISSUE #579: When validating the assignment to a list, `None` is not properly validated

Jasha10:
This PR is to improve validation for `Optional[...]` typing.
The main contribution is to allow `element_type` to be optional, e.g.
```python
DictConfig({}, element_type=Optional[int])
```
is now legal.

Summary of changes:
- In the `wrap` function, check whether the parent's `_metadata.element_type` is optional. This means that, for the new child value,  `is_optional==True` if `element_type==Any` and `is_optional==False` if e.g. `element_type==int`. This closes #579.
- Modify `_utils.valid_value_annotation_type` so that `Optional[...]` values are considered valid. Check `valid_value_annotation_type(element_type)` in `ContainerMetadata.__post_init__`. This closes #460.
- In `ContainerMetadata.__post_init__`: replace an assertion with a `raise ValidationError`; remove redundant `raise ValidationError` statements from `ListConfig.__init__` and `DictConfig.__init__`.
- Use `_utils._resolve_optional` as needed to adapt code for optional `element_type`.
- Updates to `OmegaConf.merge` such that optional element_type is handled properly.


tests/test_merge.py
```diff
@@ -277,7 +277,7 @@
         ),
         param(
             (
-                DictConfig({"user007": None}, element_type=User),
+                DictConfig({"user007": "???"}, element_type=User),
```
odelalleau, line 280: The test is called "merging into None" => maybe it's the `element_type` that should be `Optional[User]` rather than replacing `None` with missing?

tests/test_optional.py
```diff
@@ -0,0 +1,30 @@
...
+    ],
+)
+def test_assign_none(cls: Any, key: str, assignment: Any, error: bool) -> None:
+    cfg = OmegaConf.structured(cls)
+    if error:
+        with raises(ValidationError):
+            cfg.__setattr__(key, assignment)
+    else:
+        cfg.__setattr__(key, assignment)
+        assert cfg.__getattr__(key) == assignment
```
odelalleau, line 30: A bit more readable IMO:
```suggestion
    if error:
        with raises(ValidationError):
            cfg[key] = assignment
    else:
        cfg[key] = assignment
        assert cfg[key] == assignment
```

tests/test_optional.py
```diff
@@ -0,0 +1,30 @@
+from typing import Any
```
odelalleau, line 1: I'll let @omry comment on this, but I'm not sure it's worth adding a new test file just for this (could be part of `test_basic_ops_{list,dict}`)
(don't change it just based on my opinion though)

tests/test_optional.py
```diff
@@ -0,0 +1,30 @@
...
+
+from omegaconf import OmegaConf
+from omegaconf.errors import ValidationError
+from tests import SubscriptedDict, SubscriptedList
+
+
+@mark.parametrize(
+    ("cls,key,assignment,error"),
+    [
+        param(SubscriptedList, "list", [None], True, id="list_elt"),
```
odelalleau, line 13: Can you add a test when assigning a *partially* valid list, e.g., `[0, 1, None]`?

odelalleau (changes requested):

tests/test_merge.py
Jasha10: Good point.

tests/test_optional.py
Jasha10: My motivation was to be able to write `Optional[...]` tests in one place instead of having to split them up according to `Dict[...,Optional]` / `List[Optional]` / `Optional[User]` / ...
I'm hoping to make more improvements to optional type handling in the near future.

tests/test_optional.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/749/commits/c97b3f16245315b685496348ad5a4160b55c1d5b.

tests/test_merge.py
Jasha10: This relates to https://github.com/omry/omegaconf/issues/460, which I'll be working on in the near future :)

tests/test_optional.py
Jasha10: (Maybe something like `test_typing.py` would be more future-proof...)

tests/test_merge.py
odelalleau: If this doesn't work yet, maybe you can hack around it by generating the DictConfig for this test in a separate function, that would set the `element_type` only after the content has been set?
(keeping a note to change it when fixing #460)

Jasha10: After playing with it for a bit I found that the test does not pass even if `element_type` is set to `Optional[User]` manually. This is because `merge` complains about about an optional element type.

Anyway, I've reverted the test to it's original state in https://github.com/omry/omegaconf/pull/749/commits/007a879cca593bf5c6351ef33e0f0f15187fd86f (expecting a CI failure).

Jasha10: Since #579 and #460 have similar scope (both dealing with optional/non-optional `element_type` in containers), I decided to tackle them both in this PR.

tests/test_typing.py
```diff
@@ -0,0 +1,158 @@
...
+            User("Bond", 7),
+            False,
+            id="Optional[User]<-User_succeeds",
+        ),
+        param(
+            ListConfig([User], element_type=User),
+            0,
+            None,
+            True,
+            id="Optional[User]<-str_fails",
```
odelalleau, line 54: Two comments:
1. This kind of `id` formatting clashes with what's been used so far -- I don't really mind myself but just highlighting it for @omry 
2. In this one, should `str` be `None`?

tests/test_typing.py
```diff
@@ -0,0 +1,158 @@
...
+            None,
+            ["a"],
+            int,
+            True,
+            id="dict_opt_int",
+        ),
+        param(DictConfig({"a": 123}), None, ["a"], Any, True, id="dict_any"),
+        param(
+            ListConfig([], element_type=int),
+            lambda cfg: cfg.insert(0, 123),
```
odelalleau, line 107: I think you could test the same thing without introducing an extra `op` with:
```python
ListConfig([], element_type=int) + [123]
```

omegaconf/basecontainer.py
```diff
@@ -328,13 +328,12 @@ def expand(node: Container) -> None:
...
-                dest_node is None
-                and is_structured_config(dest._metadata.element_type)
-                and not missing_src_value
-            ):
+            is_optional, et = _resolve_optional(dest._metadata.element_type)
+            if dest_node is None and is_structured_config(et) and not missing_src_value:
                 # merging into a new node. Use element_type as a base
-                dest[key] = DictConfig(content=dest._metadata.element_type, parent=dest)
+                dest[key] = DictConfig(
+                    et, parent=dest, ref_type=et, is_optional=is_optional
```
odelalleau, line 335: More of a meta question for @omry: should a `DictConfig` created from a structured config type automatically set its `ref_type` to that type? (here, that would mean not needing to specify `ref_type=et`)

tests/test_typing.py
```diff
@@ -0,0 +1,158 @@
...
+        ),
+    ],
+)
+def test_ref_type(
+    src: Any, op: Any, keys: Any, ref_type: Any, is_optional: bool
+) -> None:
+    cfg = _ensure_container(src)
+    if callable(op):
+        op(cfg)
+    for k in keys:
```
odelalleau, line 155: Minor: I would suggest to accept a single key as well, i.e.:
```python
if not isinstance(keys, (list, tuple)):
    keys = [keys]
```
=> makes test cases simpler and avoids confusing errors when forgetting to put the string into a list

odelalleau (changes requested):
Overall looks great, just a few small things (NB: there's also a pickling error with 3.6 to look into)

tests/test_optional.py
odelalleau: That's fine by me => will just let @omry resolve this point.

Jasha10:
> Overall looks great, just a few small things (NB: there's also a pickling error with 3.6 to look into)

Apparently Python 3.6 pickling doesn't play well with Optional:
```python
$ python
Python 3.6.13
>>> from typing import Optional
>>> import pickle
>>> pickle.dumps(Optional[int])
_pickle.PicklingError: Can't pickle typing.Union[int, NoneType]: it's not the same object as typing.Union
```

The CI failure was triggered because I added fields to the `tests.SubscriptedList` and `tests.SubscriptedDict` dataclasses, which are used in `test_serialization.py` to test pickling.

tests/test_typing.py
Jasha10: Oh, nice! Done in https://github.com/omry/omegaconf/pull/749/commits/da7457330e40d5d166dae84e89bbb3d90f8b2cec.

tests/test_typing.py
Jasha10: Right, thanks. I'd been meaning to change those (especially after our previous discussion on `id` conventions).

Updated in https://github.com/omry/omegaconf/pull/749/commits/a661d91000f87a157cf2355203b95443a462375e:
`"Optional[User]<-User_succeeds"` -> `"set_optional_user"`
`"Optional[User]<-str_fails"` -> `"illegal_set_user_to_none"`

omegaconf/base.py
```diff
@@ -73,7 +75,10 @@ def __post_init__(self) -> None:
             self.ref_type = Any
         assert self.key_type is Any or isinstance(self.key_type, type)
         if self.element_type is not None:
-            assert self.element_type is Any or isinstance(self.element_type, type)
+            if not valid_value_annotation_type(self.element_type):
+                raise ValidationError(
```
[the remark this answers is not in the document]
Jasha10: I would not be opposed to using `TypeError`.
That being said, there are a few preexisting tests that require a `ValidationError` when an unsupported `element_type` is supplied; using `TypeError` would be an API change.

(Previously the `ValidationError` was raised in `DictConfig.__init__` and `ListConfig.__init__`, now it is being raised in `BaseContainer.__post_init__`)

tests/test_typing.py
Jasha10: I was trying to test `ListConfig.insert` specifically, since the `insert` method required an update to handle `element_type` properly.
In the case of `ListConfig([], element_type=int) + [123]`, The `ListConfig.append` method is used instead of `ListConfig.insert`.

This being said, I'm becoming suspicious about the differences between `ListConfig.insert` and `ListConfig.append`. Why should `insert` need to access `self._metadata.element_type` when `append` does not? Perhaps something is missing from the logic in `append`, or maybe the logic in `insert` could be simplified.

omegaconf/base.py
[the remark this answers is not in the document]
odelalleau: Oh ok, better leave it alone then.

tests/test_typing.py
odelalleau: Ah, I think I had read the code too quickly and misunderstood where the related change was taking place, never mind.

I think `append()` should check the `element_type` through `_set_item_impl()` (in `wrap()`). But I also think that `insert()` could be simplified in a way similar to `append()`, by first inserting `None` into `_content` then setting its value.

Jasha10: I'm able to get the tests to pass after removing the `_maybe_wrap()` call from `insert()`, and calling `_set_item_impl()` directly like in `append()`.

Jasha10:
> Apparently Python 3.6 pickling doesn't play well with Optional:

Resolved in https://github.com/omry/omegaconf/pull/749/commits/e2b2800149ed618240c8947b4a062d1d15da0013 using `pytest.skipif`.

omegaconf/listconfig.py
```diff
@@ -103,8 +98,10 @@ def _validate_set(self, key: Any, value: Any) -> None:
...
                     )
 
-        target_type = self._metadata.element_type
+        is_optional, target_type = _resolve_optional(self._metadata.element_type)
         value_type = OmegaConf.get_type(value)
+        if is_optional and value_type is None:
+            return
         if is_structured_config(target_type):
             if (
                 target_type is not None
```
Jasha10, line 107: I wasn't sure how to test this particular diff. The tests still pass even if this diff is not included in the PR.

odelalleau: Looking at it, this seems to be because the validation error is raised anyway when trying to set the value, even if `_validate_set()` didn't catch it.

You could add a test that explicitly tests `_validate_set()` in a situation like below (it raises an exception if and only if this diff is included):
```python
@dataclass
class A:
    x: int = 0

cfg = ListConfig([], element_type=Optional[A])
cfg._validate_set(0, 3)
```

tests/test_typing.py
odelalleau: > I'm able to get the tests to pass after removing the `_maybe_wrap()` call from `insert()`, and calling `_set_item_impl()` directly like in `append()`.

Cool, maybe you can leave it for another PR to avoid adding even more stuff into that one.

odelalleau: Regarding the addition of `op` that is used in only one test case, you can use a trick like this:
```python
(lambda cfg: cfg.insert(0, 123) or cfg)(ListConfig([], element_type=int))
```
Though if you go this route it should definitely be commented to explain what's going on here... not sure if it's really a better option, but I thought I'd mention it anyway :)

tests/test_typing.py
```diff
@@ -0,0 +1,182 @@
+from typing import Any, Dict, List, Optional
```
omry, line 1: This can probably go to test_base_config.py

tests/test_typing.py
omry: It looks like you made the 15-20 test cases + the test code itself more complicated for the benefit of a single test case.
In such cases, I prefer to create a dedicated test that handles the special case.
I would also rename test_ref_type to test_something, e.g. test_assignment_optional_behavior and test_list_insert_optional_behavior. (also, what about testing of ListConfig.append?)

tests/test_errors.py
```diff
@@ -690,6 +691,16 @@ def finalize(self, cfg: Any) -> None:
...
+    param(
+        Expected(
+            create=lambda: None,
+            op=lambda cfg: OmegaConf.structured(ErrorDictUnsupportedValue),
+            exception_type=ValidationError,
+            msg="Unsupported value type: 'tests.structured_conf.data.dataclasses.RegularClass'",
+            key="dict",
+        ),
+        id="structured:create_with_unsupported_element_type",
+    ),
```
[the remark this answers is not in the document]
Jasha10: Ok, made a change in https://github.com/omry/omegaconf/pull/749/commits/941554f064b9ab3d15c538c99ec2437121493e8b.

tests/test_optional.py
Jasha10: Resolved in https://github.com/omry/omegaconf/pull/749/commits/7f5892e45897c65ca2b8fcda876e83b792dffd67: I've put the new tests in `test_base_config.py`.

tests/test_typing.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/749/commits/7f5892e45897c65ca2b8fcda876e83b792dffd67.

tests/test_typing.py
Jasha10: > It looks like you made the 15-20 test cases + the test code itself more complicated for the benefit of a single test case.
> In such cases, I prefer to create a dedicated test that handles the special case.

Got it. Done in https://github.com/omry/omegaconf/pull/749/commits/c488af67e6aa3ac2f8ccb0b81b3812f2039f4887.

> I would also rename test_ref_type to test_something, e.g. test_assignment_optional_behavior and test_list_insert_optional_behavior.

Renamed in https://github.com/omry/omegaconf/pull/749/commits/aeda10ccb86cf0a45c03ab0627963ff215d43409.

> (also, what about testing of ListConfig.append?)

I've added some tests for append in https://github.com/omry/omegaconf/pull/749/commits/ceb5dd1922f0582b315a7d067e265eec90cc7d80.
Originally I tested `insert` only because this PR does not touch `append`'s code.

After some discussion (see the above), I became suspicious about why `append`'s code is so different from `insert`'s -- it seems that the two have diverged. I'm planning a followup PR to reconcile this.

omegaconf/basecontainer.py
Jasha10: > In the context of merge, I think the ref_type should be preserved from the target node.

Agreed. This PR is consistent with preserving ref_type from the target node.
> If there is no target node (new node), I think the ref type should be Any (default, irrc).

Usually, but I think that if parent's `element_type` is set
then (for a new node) we should have `parent_element_type == child_ref_type`.
For example:
```python
from omegaconf import OmegaConf, DictConfig, ValidationError
from tests import User
from typing import Any
from pytest import raises

cfg = OmegaConf.merge(DictConfig({}, element_type=User), {"a": User("Bond")})

# On this branch:
assert cfg.a._metadata.ref_type == User

# On master branch:
assert cfg.a._metadata.ref_type == Any
```
This change is consistent with how assignment works: `child_ref_type` is
inherited from `parent_element_type`.
```python
cfg2 = DictConfig({}, element_type=User)
cfg2.a = User("Bond")
# On this branch AND master branch:
assert cfg2.a._metadata.ref_type == User
```

omegaconf/listconfig.py
Jasha10: That works, thanks! I added some tests in https://github.com/omry/omegaconf/pull/749/commits/2bc4023eaac93690fbc22f9a1134def4c30ee06b, and additionally (with the help of TDD) I found another edge case where ValidationError should be raised (`value_type==None` and `element_type` is not optional).

odelalleau: Sounds good, resolving.

odelalleau (approved):
Good for me once ongoing discussions with @omry are resolved, thanks!

tests/test_merge.py
```diff
@@ -379,6 +381,332 @@ def test_merge(
...
+            True,
+            id="missing_optional_user_missing",
+        ),
+    ],
+)
+def test_optional_element_type_merge(
+    inputs: Any, expected: Any, ref_type: Any, is_optional: bool
+) -> None:
+    configs = [_ensure_container(c) for c in inputs]
+    if isinstance(expected, AbstractContextManager):
```
odelalleau, line 696: Other tests typically use `RaisesContext` in this situation, any reason not to do the same?

tests/test_merge.py
```diff
@@ -379,6 +381,332 @@ def test_merge(
...
+            ),
+            {"foo": MISSING},
+            str,
+            True,
+            id="missing_optional_str_missing",
+        ),
+        param(
+            (
+                DictConfig(content={"foo": User("Bond")}, element_type=User),
+                {"foo": User("Bond")},
```
odelalleau, line 534: Can you make it merge a different user name? (ex: `User("007")`)
(same for `optional_user` below)

tests/test_merge.py
```diff
@@ -379,6 +381,332 @@ def test_merge(
...
+            None,
+            None,
+            id="user_none",
+        ),
+        param(
+            (
+                DictConfig(content={"foo": User("Bond")}, element_type=User),
+                {"foo": MISSING},
+            ),
+            {"foo": MISSING},
```
odelalleau, line 556: Shouldn't this one be `{"foo": User("Bond")}`?

tests/test_merge.py
```diff
@@ -379,6 +381,332 @@ def test_merge(
...
+        param(
+            (DictConfig(content={}, element_type=User), {"foo": User("Bond")}),
+            {"foo": User("Bond")},
+            User,
+            False,
+            id="new_user",
+        ),
+        param(
+            (DictConfig(content={}, element_type=User), {"foo": None}),
+            raises(ValidationError, match="child 'foo' is not Optional"),
```
odelalleau, line 600: Minor comment: may be worth looking at the full error message to make sure it makes sense (the excerpt being tested could be confusing)

odelalleau (changes requested):
A few comments on the latest commit

omry:
Python 3.6 eol is coming in a couple of months: https://endoflife.date/python

It would be good to issue a better error than `Can't pickle typing.Union[int, NoneType]: it's not the same object as typing.Union` on Python 3.6 when we run into this problem.
Furthermore, we can consider dropping formal Python 3.6 support for OmegaConf 2.2.

tests/test_merge.py
Jasha10: Good idea. Done in https://github.com/omry/omegaconf/pull/749/commits/e81dd715b3b7223e05fe44b38f18aabc18dbe6b9.

tests/test_merge.py
Jasha10: In https://github.com/omry/omegaconf/pull/749/commits/0d8196cea552b6f409be7ca20942d200147fbcf1 I've updated the tests to match the full error message (excluding the one with id="new_str_none", which is currently not passing).

tests/test_merge.py
Jasha10: Good catch! Fixed in https://github.com/omry/omegaconf/pull/749/commits/9a2293956d9a6663d305e11b57ce077e867d21de, and the test is passing now :)

Jasha10:
> It would be good to issue a better error than `Can't pickle typing.Union[int, NoneType]: it's not the same object as typing.Union` on Python 3.6 when we run into this problem.

Currently I can't reproduce that pickling error on the `master` branch (`element_type==Optional[int]` causes an `AssertionError` before the DictConfig instance can be full initialized).

I think catching the pickling error would require adding logic to `BaseContainer.__getstate__` to handle the case where `python_version<=3.7` and a `Union` type appears in `self._metadata`.
I'll follow up on that once the problem can be reproduced on `master`.

tests/test_merge.py
Jasha10: Oh, I hadn't realized that other tests in this repo are using `RaisesContext`.
Updated in https://github.com/omry/omegaconf/pull/749/commits/1ab08a6b51dd6f33296c76c589e991dfde32cc0c.

tests/test_merge.py
odelalleau: Thanks -- though personally I tend to prefer checking snippets to make tests more concise (leaving full checks to test_errors.py) -- but anyway the new version looks good to me.

This actually confirms that this error message ("child 'foo' is not Optional") is a bit confusing (IMO). I can see where it may be coming from (we first add a `foo` node of type `User` then try to set it to `None`), but a clearer error could be something like: "Can't merge None values into a DictConfig whose element type is not Optional".

Definitely not something to change in this PR (and may not be worth changing it at all) -- I just thought I'd mention it because it caught my eye :)

odelalleau:
> Edit: or should I add that logic to this PR?

Future PR sounds better to me, can you just make sure there is an xfail test for Python 3.6 in this one? (if not already the case)

odelalleau:
> Merge branch 'master' into closes579

Meta-discussion: I personally prefer to rebase on top of master and force-push (doing it only if needed) => this allows to easily see all changes (commits) done on top of master, without having obscure merge commits that may hide how some conflicts were handled.

=> This is only me though and I wonder if there's a good argument in favor of merging master like you did

odelalleau (approved):
Should be good for me once everyone else's happy (including CI)

Jasha10:
> > Merge branch 'master' into closes579
> 
> Meta-discussion: I personally prefer to rebase on top of master and force-push (doing it only if needed) => this allows to easily see all changes (commits) done on top of master, without having obscure merge commits that may hide how some conflicts were handled.

Good point about hiding how merge conflicts were handled.

> => This is only me though and I wonder if there's a good argument in favor of merging master like you did

My main motivation was to prevent earlier commits from becoming orphaned (so that review comments linking to specific commits would not be out-of-date).
I was also worried that the review comments themselves would become orphaned (i.e. that they would no longer show up in GitHub's "Files changed" tab). This seems not to be the case, however: based on my experiment in a [test repo](https://github.com/Jasha10/test_repo), GitHub is smart enough to re-apply the comments in the correct place on the rebased branch. I'll be rebasing in the future :)

omry:
> Meta-discussion: I personally prefer to rebase on top of master and force-push (doing it only if needed) => this allows to easily see all changes (commits) done on top of master, without having obscure merge commits that may hide how some conflicts were handled.

Not 100% sure we are talking about the same thing, but I prefer not to force push to master because it breaks anyone that has a checkout (or is keeping a git ref id stored).
You can reorganize things on your own branch (squash, or splitting into multiple commits) and force push and finally do a regular push when merging the PR.

odelalleau:
> GitHub is smart enough to re-apply the comments in the correct place on the rebased branch. I'll be rebasing in the future :)

Yeah it's been working fine for me so far, I haven't noticed any major issue. The main problem I've found is that comments referencing specific commits would now refer to the old version of these commits. It's often not a real problem because the diff is usually the same after rebase, but it's something to be aware of (especially if you have many unresolved discussions pointing to various commits).

> Not 100% sure we are talking about the same thing, but I prefer not to force push to master because it breaks anyone that has a checkout (or is keeping a git ref id stored). You can reorganize things on your own branch (squash, or splitting into multiple commits) and force push and finally do a regular push when merging the PR.

We're not talking about the same thing :) What I was talking about is when you want to bring in some changes from `master` into your PR branch (typically to solve a conflict, or to fix a broken test). You can either merge master into your branch, or rebase your branch on top of master (and I was saying that I personally favor the second approach).

Jasha10:
Per offline discussion with Omry: I'm planning to add serialization tests (serializing e.g. containers with optional element type) and will try to modify `BaseContainer.__getstate__` as to address the issue with pickle+optional+python36 that came up earlier in this PR.

Jasha10:
Rebased against `master` to incorporate changes from #800 (fixing the CI).

tests/test_serialization.py
```diff
@@ -153,24 +158,101 @@ def test_load_empty_file(tmpdir: str) -> None:
...
+        param(
+            SubscriptedDict,
+            "dict_enum",
+            int,
+            Color,
+            False,
+            Dict[Color, int],
+            id="dict_enum",
+        ),
+        param(SubscriptedList, "list", int, Any, False, List[int], id="list_int"),
```
Jasha10, line 214: The above diff is purely cosmetic; pytest id's have been added, and nothing else has changed.

Jasha10:
@odelalleau would you be willing to take a quick look at the recent commit https://github.com/omry/omegaconf/pull/749/commits/b593f4fd749b003286edb5e886f2c40c6a19bcc4? Outside of that diff, there have been no significant changes since your  previous review (besides a `git rebase`).

odelalleau:
> @odelalleau would you be willing to take a quick look at the recent commit [b593f4f](https://github.com/omry/omegaconf/commit/b593f4fd749b003286edb5e886f2c40c6a19bcc4)? Outside of that commit, there have been no significant changes since your previous review (besides a `git rebase`).

Yeah, sorry I said nothing but I had a look and it seems good to me. Since you're asking though, I remember @omry's been a bit reluctant in the past to add new exception subclasses unless there's a clear need, but that's ok as far as I'm concerned :)

Jasha10:
Ok, thanks! I'll mention it to Omry before landing this.

Jasha10:
After exchanging emails with Omry, I'm changing back to `OmegaConfBaseException` instead of introducing the new `ConfigSerializationError` type, as currently the use case for `ConfigSerializationError` is very rare (it only applies to python3.6 + pickle + optional element type).

PR #752, opened 2021-06-16
Fix regression:  merging list into optional tuple

### LINKED ISSUE #751: Regression in handling of Tuple with merge in 2.1
There is a regression in handling `Tuple[int, ...]` in 2.1.

The following works on 2.0 and fails on 2.1:
```python
from omegaconf import *
from typing import *

@dataclass
class Foo:
    x : Optional[Tuple[int, ...]] = None

cfg = OmegaConf.merge(Foo, {"x": [2,1]})

assert cfg == {"x": [2,1]}
```

Jasha10:
Closes #751.

Previously (in OmegaConf 2.0) it was possible to merge a list into a ListConfig(None) that had `ref_type==Tuple[int, ...]` and `is_optional=True`.
In 2.1, this was no longer possible.

This PR patches the regression.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,7 +1,7 @@
 import re
 from dataclasses import dataclass, field
 from enum import Enum
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import attr
 from pytest import warns
@@ -237,6 +237,11 @@ class InterpolationDict:
 @dataclass
 class Str2Int(Dict[str, int]):
     pass
+
+
+@dataclass
+class OptTuple:
+    x: Optional[Tuple[int, ...]] = None
 
 
 def warns_dict_subclass_deprecated(dict_subclass: Any) -> Any:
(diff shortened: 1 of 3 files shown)


tests/__init__.py
```diff
@@ -239,6 +239,11 @@ class Str2Int(Dict[str, int]):
     pass
 
 
+@dataclass
+class OptTuple:
+    x: Optional[Tuple[int, ...]] = None
```
omry, line 244: What is the element_type if the resulting ListConfig?

omry (approved):
See question, I am hoping the answer is `int` :).

cc @jieru-hu, this should address the issue you saw.

tests/__init__.py
Jasha10: The `element_type` is `int` :)

Jasha10: FYI in the following case we also have `element_type==int`:
```python
@dataclass
class OptTuple:
    x: Optional[Tuple[int, str]]
```
Something to watch out for...

PR #754, opened 2021-06-17
Improve error message for InterpolationValidationError

### LINKED ISSUE #753: InterpolationValidationError: error message improvements
Running the below file produces an InterpolationValidationError, which is expected.
```python
from typing import Optional

@dataclass
class Foo:
    x: Optional[int] = None
    y: int = "${.x}"

print(cfg.y)
```
```bash
$ python3 tmp.py
  ...
omegaconf.errors.InterpolationValidationError: Non optional field cannot be assigned None
    full_key: y
    object_type=Foo
```

Two things could be improved in this error message:
- the offending interpolation key ${.x} is relevant and could be included in the message, and
- the message text "cannot be assigned" is misleading, since no assignment is occurring.

Jasha10:
This PR improves error messages raised when interpolating to `None` from a node whose `ref_type` is not optional.
- The interpolation key is now included in the error message.
- The error message now includes the `ref_type` of the target field.
- The error message no longer uses the word "assign", which was misleading because (semantically) assignment is not taking place when an interpolation is dereferenced.

If this PR is merged into `master`, I'll plan to cherry pick the updates and open a similar PR against the OmegaConf `2.1_branch`.

Closes #753.

### DIFF
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -319,6 +319,7 @@ params = [
             create=lambda: OmegaConf.structured(StructuredInterpolationValidationError),
             op=lambda cfg: getattr(cfg, "y"),
             exception_type=InterpolationValidationError,
+            object_type=StructuredInterpolationValidationError,
             msg=(
                 "While dereferencing interpolation '${.x}': "
                 "Incompatible value 'None' for field of type 'int'"
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -27,6 +27,7 @@ from omegaconf.errors import (
     InterpolationKeyError,
     InterpolationResolutionError,
     InterpolationToMissingValueError,
+    InterpolationValidationError,
     KeyValidationError,
     MissingMandatoryValue,
     OmegaConfBaseException,
@@ -42,6 +43,7 @@ from tests import (
     Plugin,
     Str2Int,
     StructuredInterpolationKeyError,
+    StructuredInterpolationValidationError,
     StructuredWithBadDict,
     StructuredWithBadList,
     StructuredWithMissing,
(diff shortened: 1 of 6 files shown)


tests/test_errors.py
```diff
@@ -312,6 +314,20 @@ def finalize(self, cfg: Any) -> None:
...
         id="dict,accessing_missing_nested_interpolation",
     ),
+    param(
+        Expected(
+            create=lambda: OmegaConf.structured(StructuredInterpolationValidationError),
+            op=lambda cfg: getattr(cfg, "y"),
+            exception_type=InterpolationValidationError,
+            msg=(
+                "While dereferencing interpolation '${.x}': "
+                "Incompatible value 'None' for field of type 'int'"
```
omry, line 324: Does the error indicate the type containing the problem? (StructuredInterpolationValidationError)?

I would expect `object_type == StructuredInterpolationValidationError`.

tests/test_nodes.py
```diff
@@ -600,8 +600,12 @@ def test_dereference_missing() -> None:
 )
 def test_validate_and_convert_none(make_func: Any) -> None:
     node = make_func("???", is_optional=False)
+    ref_type_str = node._metadata.ref_type.__name__
```
omry, line 603: use the same function: type_str.

tests/test_errors.py
Jasha10: Yes. Here is the full error message:
```
omegaconf.errors.InterpolationValidationError: While dereferencing interpolation '${.x}': Incompatible value 'None' for field of type 'int'
    full_key: y
    object_type=StructuredInterpolationValidationError
```
To be explicit, I added the `object_type=StructuredInterpolationValidationError` requirement to the `test_errors.py` test-case in https://github.com/omry/omegaconf/pull/754/commits/094a3ee37faadfca0faace720fae1ef13bbaceaa.

tests/test_nodes.py
Jasha10: Done in https://github.com/omry/omegaconf/pull/754/commits/9058f9df3d2996963b35eddc133b7a229e2e27d6.

omry (approved):
Thanks!

PR #764, opened 2021-06-30
Update regular expression so tests will pass on pypy

Jasha10:
This PR should fix the pypy error message that is causing https://github.com/conda-forge/omegaconf-feedstock/pull/19 build to fail.


omry:
Thanks.
Can you try to also add one version of pypy (say pypy 3.8) to the circleci file the project is using?
Not sure how hard this would be.

omry (approved):

Jasha10:
> Not sure how hard this would be.

I'll give it a try.

omry:
Let's limit to Linux only.

PR #765, opened 2021-06-30
circleci: add pypy3.7

Jasha10:


Jasha10:
@omry I am stuck.
In the pypy job, I'm getting `/bin/bash: sudo: command not found`.
Any idea what to do?

omry:
not sure why the pypy image is missing sudo.
what os is it? what user is it running under?

Jasha10:
Looks like Debian, running under `root` user.
```bash
# cat etc/os-release
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
...
# whoami
root
```

Jasha10:
The other docker images we're using (`circleci/python:3.x`) are also running Debian, though they are running under user `circleci`, not `root`.

omry:
you can apt-get install sudo.
I suspect the other images are actually Ubuntu.

Jasha10:
It seems that [mypy does not run on pypy](https://github.com/python/mypy/blob/3bdef9fe6d401f3d2e0cacf4964bd315550c3394/docs/source/faq.rst#does-it-run-on-pypy), hence the build failures.

One idea for a workaround: Have another version of `requirements/dev.txt` (e.g. `dev_pypy.txt`) which skips `mypy`. Then `noxfile.py` can pick which requirements file to install from.

omry:
Skipping mypy is fine for pypy.

Try to get noxfile to work correctly locally first on an a conda env with pypy instead of python.
You can use a different nox session for it that will do the right thing for pypy and only run the tests.

```
nox -s test_omegaconf_pypy
```

Or make the default session smart enough to handle it.

.circleci/config.yml
```diff
@@ -6,36 +6,68 @@ commands:
...
+              sudo apt-get update
+              sudo apt-get install -y openjdk-11-jre
+              sudo pip install nox
+            fi
       - run:
           name: "Testing OmegaConf"
           command: |
             export NOX_PYTHON_VERSIONS=<< parameters.py_version >>
-            nox --add-timestamp
+            nox --add-timestamp << parameters.nox_args >>
```
Jasha10, line 30: In a pypy conda env, running `nox -s test_omegaconf_pypy` succeeds, but running `NOX_PYTHON_VERSIONS=pypy3.7 nox` fails, since some other sessions (e.g. docs, lint, etc) will not be able to run against the pypy executable. This is why I added `<< parameters.nox_args >>` to the circleci config file.

noxfile.py
```diff
@@ -22,6 +22,14 @@ def omegaconf(session):
     session.run("pytest")
 
 
+@nox.session(python="pypy")
+def test_omegaconf_pypy(session):
+    deps(
+        session, editable_installl=False, requirements="requirements/dev_pypy.txt"
+    )  # ensure we test the regular install
+    session.run("pytest", "-Wdefault")
```
Jasha10, line 30: The `-Wdefault` flag here allows warnings in the pypy session (whereas in other sessions, warnings would cause pytest to fail).
In the pypy session, a warning is being caused by the `import yaml` statement:
`ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__`

omry:
I am starting to think it's not worth the effort at this point.
What do you think about abandoning it?

.circleci/config.yml
omry: Adding a new session specifically for pypy (something like this) could work:
```python
@nox.session(python=["pypy-3.8"])
def omegaconf_pypy(session):
    deps(session)
    session.run("pytest")
```

omry (commented):
After looking at the changes, I feel like this is making CI more complicated for a pretty marginal gain.

I think we can keep pypy as a best effort platform for now (in both OmegaConf and Hydra).

.circleci/config.yml
Jasha10: That's basically what I ended up doing with the `def test_omegaconf_pypy(session): ...` below.

The `parameters.nox_args` here is just to make sure that other sessions besides `test_omegaconf_pypy` don't get run.
Without the `parameters.nox_args`, nox tries to run e.g. the `benchmark` and `coverage` sessions using the `python` executable it finds in `$PATH` (which is aliased to pypy).

Jasha10:
> I think we can keep pypy as a best effort platform for now (in both OmegaConf and Hydra).

Ok, closing now. The tests are passing now (as of the last commit); we can always come back to this PR later if needed.

PR #776, opened 2021-08-13
Update MANIFEST.in to fix packaging issue.

Jasha10:

PR #777, opened 2021-08-16
fix packaging issue on master branch

Jasha10:
Cherry picked diff #776 from `2.1_branch` to `master`

### DIFF
--- a/news/772.misc
+++ b/news/772.misc
@@ -1,1 +1,0 @@
-Fix a packaging issue (missing sdist dependency)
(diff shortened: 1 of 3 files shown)


news/772.misc
```diff
@@ -0,0 +1 @@
+Fix a packaging issue (missing sdist dependency)
```
Jasha10, line 1: It makes sense to have this same fragment on both `2.1_branch` and `master`, right?

omry: No, I think we should only report it once.
Remove it from the cherry pick.

Jasha10: Done in https://github.com/omry/omegaconf/pull/777/commits/c68fd10cec667fd0222b6808b80c6bc7e475aaf1.

PR #790, opened 2021-09-29
Refactor titles of the structured configs documentation

odelalleau:
The objective is to make it easier to understand that all content in the
`structured_config.rst` file is related to structured configs, when
looking at the left navigation pane. More precisely, the main changes
are that:

- The "Lists" and "Dicts" sections are now moved inside the "Structured
  configs" main section (getting rid of the intermediate "Containers"
  section to reduce nesting)
- The "Misc" section is also moved inside the "Structured Configs" main
  section (and is also renamed to give a better idea of what's inside
  it, now that its detailed content isn't shown anymore in the left
  navigation pane)
- Finally the "Merging with other configs" section is also moved inside
  the "Structured Configs" main section so as to make it clear that it
  is related to structured configs.

### DIFF
--- a/docs/source/structured_config.rst
+++ b/docs/source/structured_config.rst
@@ -1,4 +1,4 @@
-.. _structured_config:
+.. _structured_configs:
 
 .. testsetup:: *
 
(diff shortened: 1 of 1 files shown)


Jasha10 (approved):
lgtm!

PR #792, opened 2021-09-30
Add documentation for the OmegaConf grammar

odelalleau:
Some "design choices" I made, happy to discuss them here:
* Not dumping the whole set of parser rules as is done in [Hydra's documentation](https://hydra.cc/docs/advanced/override_grammar/basic): this is because it would be pretty much unreadable with the current documentation formatting (in particular due to the limited width). Instead, I chose to focus on a few relevant excerpts, and link to the file if someone wants to see the whole thing
* Making a few simplifications (not mentioning all details), in order to (hopefully) keep it digestible

I'm requesting reviews for diverse people to get feedback both from those familiar vs unfamiliar with it.

For ease of reviewing I suggest you build it locally:
```
cd docs && make html && open build/html/index.html
```

### DIFF
--- a/docs/source/grammar.rst
+++ b/docs/source/grammar.rst
@@ -89,7 +89,8 @@ with OmegaConf 2.0, but this has been deprecated (see* `#572 <https://github.com
 Element types
 ^^^^^^^^^^^^^
 
-There are four main types of elements:
+As seen in the ``sequence`` rule above, each resolver argument is parsed by an ``element`` rule,
+which currently supports four main types of arguments:
 
     .. code-block:: antlr
 
--- a/docs/source/grammar.rst
+++ b/docs/source/grammar.rst
@@ -170,14 +170,20 @@ For instance:
     >>> print(c.path)  # does *not* interpolate into the `dir` node
     ${dir}
 
-Since the ``\`` character has a special meaning in front of ``${``,
-it must itself be escaped with ``\\`` if you actually intend to follow it with an interpolation:
+If you actually want to follow a ``\`` with a resolved interpolation, this backslash
+needs to be escaped into ``\\`` to differentiate it from an escaped interpolation:
 
 .. doctest::
 
     >>> c = OmegaConf.create({"path": r"C:\\${dir}", "dir": "tmp"})
     >>> print(c.path)  # *does* interpolate into the `dir` node
     C:\tmp
+
+Note that we use Python raw strings here to make examples
+more readable -- otherwise all ``\`` characters would need be duplicated due to how Python handles
+escaping in regular string literals. For instance, the above example would become ``"C:\\\\${dir}"``
+without using a raw string.
+
 
 Finally, since the ``\`` character has no special meaning unless followed by ``${``,
 it does *not* need to be escaped anywhere else:
--- a/docs/source/grammar.rst
+++ b/docs/source/grammar.rst
@@ -1,0 +1,240 @@
+.. testsetup:: *
+
+    from omegaconf import OmegaConf
+
+The OmegaConf grammar
+---------------------
+
+OmegaConf uses an `ANTLR <https://www.antlr.org/>`_-based grammar to parse string expressions,
+where the `lexer rules <https://github.com/omry/omegaconf/blob/master/omegaconf/grammar/OmegaConfGrammarLexer.g4>`_
+rules define the tokens used by the `parser rules <https://github.com/omry/omegaconf/blob/master/omegaconf/grammar/OmegaConfGrammarParser.g4>`_.
+Currently this grammar's main usage is in the parsing of :ref:`interpolations<interpolation>`, detailed below.
+
+
+String interpolations
+^^^^^^^^^^^^^^^^^^^^^
+
+A string interpolation is a string concatenating an interpolation with additional characters
+(that may or may not be other interpolations).
+These are all examples of string interpolations:
+
+    - ``https://${host}:${port}``
+    - ``Hello ${name}``
+    - ``${a}${oc.env:B}${c}``
+
+Such strings are matched by the following parser rule, which is the main rule used
+to parse any string containing an interpolation:
+
+    .. code-block:: antlr
+
+        text: (interpolation |
+               ANY_STR | ESC | ESC_INTER | TOP_ESC | QUOTED_ESC)+;
+
+In this rule, all tokens except ``interpolation`` represent regular string fragments
+(with a special handling of escaped characters, see :ref:`escaping-in-sring-interpolations` below).
+
+
+Interpolation types
+^^^^^^^^^^^^^^^^^^^
+
+An ``interpolation`` as found in the rule above can either be a :ref:`config-node-interpolation`
+(e.g., ``${host}``) or a call to a :ref:`resolver<resolvers>` (e.g., ``${oc.env:B}``).
+This is reflected in the following parser rules:
+
+    .. code-block:: antlr
+
+        interpolation: interpolationNode | interpolationResolver;
+
+        interpolationNode:
+            INTER_OPEN  // ${
+            DOT* 
+            (configKey | BRACKET_OPEN configKey BRACKET_CLOSE)
+            (DOT configKey | BRACKET_OPEN configKey BRACKET_CLOSE)*
+            INTER_CLOSE;  // }
+
+        interpolationResolver:
+            INTER_OPEN  // ${
+            resolverName COLON sequence?
+            BRACE_CLOSE;  // }
+
+The following are all valid examples of config node interpolations according to the ``interpolationNode`` rule
+(note in particular that it supports both dot and bracket notations to access child nodes):
+
+    - ``${host}``
+    - ``${.sibling}``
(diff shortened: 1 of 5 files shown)


Jasha10 (commented):
Very nice!

docs/source/grammar.rst
```diff
@@ -0,0 +1,240 @@
...
+            BRACE_CLOSE;  // }
+
+The following are all valid examples of config node interpolations according to the ``interpolationNode`` rule
+(note in particular that it supports both dot and bracket notations to access child nodes):
+
+    - ``${host}``
+    - ``${.sibling}``
+    - ``${..uncle.cousin}``
+    - ``${some_list[3]}``
+    - ``${some_deep_dict[key1][subkey2].subsubkey3}``
```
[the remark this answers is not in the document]
odelalleau: I can see how this may look a bit confusing. The underlying reason is that the leading dots notation for relative interpolations is "special" (those dots aren't the same as the dots used to separate a parent node from its child).

For instance `${.}` denotes the parent node (this isn't allowed anymore because it can lead to cycles in the config, but it used to work at some point). So if you think as a leading dot as meaning "parent", `${.foo}` means `${parent.foo}` and `${.[foo]}` would mean `${parent[foo]}`.

I'll let @omry chime in if he thinks we should change the current behavior. Personally I would keep it, because I like that relative interpolations are consistently using leading dots (if `${[key]}` became a relative interpolation we would lose this consistency).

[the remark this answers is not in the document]
Jasha10: This explanation makes sense to me.
Also, I think that if we make `${[key]}` invalid then there will be no easy way to access the root config with brackets notation.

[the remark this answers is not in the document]
Jasha10: This explanation makes sense to me. The "parental" dots at the beginning of the interpolation have a different meaning from the "getattr" dots that appear between keys.

docs/source/grammar.rst
```diff
@@ -0,0 +1,240 @@
...
+    '\thi u\t\t'
+
+Escaping in unquoted strings can lead to hard-to-read expressions, and it is recommended
+to switch to quoted strings instead of relying heavily on the above escape sequences.
+
+Escaping in quoted strings
+++++++++++++++++++++++++++
+
+As can be seen from the definition of the ``quotedValue`` parser rule above, a quoted string
+is essentially a string interpolation surrounded by quotes.
```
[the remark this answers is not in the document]
odelalleau: Good point, this is a consequence of some over-simplifications I made. I tried to make it clearer in bb77fc1 => let me know if that's better.

I changed "string interpolation" into "interpolation string" because "string interpolation" actually means something very specific: the concatenation of at least one interpolation with either some regular string or another interpolation (which always results in a string result, e.g., `https://${host}:${port}`). But the ``text`` parser rule is also used to parse single interpolations like ``${foo}`` (that aren't string interpolations).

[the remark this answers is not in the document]
Jasha10: Looks good!

Jasha10 (approved):
Sorry I guess I forgot to give a formal approval on this one.

docs/source/grammar.rst
```diff
@@ -0,0 +1,242 @@
...
+For instance:
+
+.. doctest::
+
+    >>> c = OmegaConf.create({"path": r"\${dir}", "dir": "tmp"})
+    >>> print(c.path)  # does *not* interpolate into the `dir` node
+    ${dir}
+
+Since the ``\`` character has a special meaning in front of ``${``,
+it must itself be escaped with ``\\`` if you actually intend to follow it with an interpolation:
```
omry, line 174: Not sure I understand what you mean here.
In the above example you use an rstring instead of escaping it.
escaping the `\` is a Python requirement in string literals. Maybe provide an second example without an rstring calling out that this is because `\` has to be escaped in Python strings?

docs/source/grammar.rst
```diff
@@ -0,0 +1,242 @@
...
+Escaping in unquoted strings
+++++++++++++++++++++++++++++
+
+Unquoted strings can be found in a number of contexts, including dictionary keys/values,
+list elements, etc. As a result, the  escape sequences are used for some
+special characters
+(``\\``, ``\[``, ``\]``, ``\{``, ``\}``, ``\(``, ``\)``, ``\:``, ``\=``, ``\,``),
+for instance:
+
+    - ``C\:\\$\{dir\}`` resolves to the string ``"C:\${dir}"``
```
omry, line 203: Why do you have to escape the colon?

omry (approved):
Overall looks good. See some inline comments/questions.

docs/source/grammar.rst
odelalleau: Does e71ac8d make it clearer?

docs/source/grammar.rst
odelalleau: Ah, good question, I forgot that escaping the colon is only required in dictionary keys for obvious reasons (but escaping is optional in other unquoted strings, i.e., both `\:` and `:` lead to the same result).

I'm tempted to not go into this detail for the documentation, because it's not a big deal if people escape a colon that doesn't require it, and in most situations I expect users to just use `:` without second thoughts and it will usually work.

Let me know though if you believe this should be fully documented here.

Jasha10:
I believe that the CI should pass if changes from PR #800 are incorporated into this branch.

odelalleau:
> I believe that the CI should pass if changes from PR #800 are incorporated into this branch.

Thanks, I'll rebase before merging to get it green.

pixelb (approved):
Very nice. I had questions at the start of the change, which were all answered by the end.

I added a comment about always requiring to escape a backslash to provide forward compat,
but as for documenting the current logic, this is excellent.

thanks!

docs/source/grammar.rst
```diff
@@ -0,0 +1,248 @@
...
+    C:\tmp
+
+Note that we use Python raw strings here to make examples
+more readable -- otherwise all ``\`` characters would need be duplicated due to how Python handles
+escaping in regular string literals. For instance, the above example would become ``"C:\\\\${dir}"``
+without using a raw string.
+
+
+Finally, since the ``\`` character has no special meaning unless followed by ``${``,
+it does *not* need to be escaped anywhere else:
```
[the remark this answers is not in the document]
odelalleau: Yeah, that's definitely a valid point. It's a trade-off and Omry and I had some discussions about this in the past (too many discussions related to escaping to dig it up though)

There are two main reasons for the current implementation:
1. Any additional escaping of `\` can quickly make things quite unreadable when combined with Python string escaping (outside of raw strings), shell escaping, or nested interpolations. Limiting the amount of escaping to the strict minimum helps keep things "sane".
2. This keeps interpolation strings close (in behavior) to static strings, making it easier to introduce interpolations into existing strings. For instance if you had `cfg = OmegaConf.create({"msg", "Please type 'start' to get started"})` you can change this message to be dynamic with `cfg = OmegaConf.create({"msg", "Please type '${start_command}' to get started"})` without having to escape the quotes (and if somehow you had the `\'` sequence in your static string, the backslash won't suddenly disappear when you add an interpolation).

=> unless @omry wants to revisit this choice, I'd rather keep it that way :)

docs/source/grammar.rst
odelalleau: > Does [e71ac8d](https://github.com/omry/omegaconf/commit/e71ac8db4efb48913565fa198b99296083274d83) make it clearer?

@omry just checking if that commit looks good to you?

docs/source/grammar.rst
```diff
@@ -170,15 +170,21 @@ For instance:
...
 .. doctest::
 
     >>> c = OmegaConf.create({"path": r"C:\\${dir}", "dir": "tmp"})
     >>> print(c.path)  # *does* interpolate into the `dir` node
     C:\tmp
 
+Note that we use Python raw strings here to make examples
+more readable -- otherwise all ``\`` characters would need be duplicated due to how Python handles
+escaping in regular string literals. For instance, the above example would become ``"C:\\\\${dir}"``
+without using a raw string.
```
pixelb, line 185: This is good. Nits:

- I would avoid the last sentence as it adds noise without added understanding
- I would change "make examples more readable" to "make code more readable" to avoid any confusion that this isn't generally useful and might just be related to how the example docs are marked up or something

odelalleau: Thanks, done in 29cd492

docs/source/grammar.rst
odelalleau: > > Does [e71ac8d](https://github.com/omry/omegaconf/commit/e71ac8db4efb48913565fa198b99296083274d83) make it clearer?
> 
> @omry just checking if that commit looks good to you?

Actually check this diff which combines the two latest (related) commits: https://github.com/omry/omegaconf/pull/792/files/21f0217c51b562b335eb98af116690fb576bcc6e..29cd49235bfd4c88a733dd0416fb7f5e52176d5f

pixelb: This adjustment is much clearer to me. Thanks

PR #799, opened 2021-10-12
Interpolation grammar: allow pipe `|` in unquoted strings

Jasha10:
This PR adds support for the pipe character `|` to appear unquoted in OmegaConf's interpolation grammar.
Related to https://github.com/facebookresearch/hydra/pull/1850.

For good measure, I've also added a test for the pipe appearing in quoted strings.


Jasha10:
> Looks good, thanks! Just curious about the CI errors, that seem unrelated => I tried to restart the build but somehow it didn't like it...

Yes, I believe the CI errors are unrelated. Jieru and I were trying to fix them in PR #800. I'm not sure exactly what's going on yet -- it seems related to interaction between `nox` and `pip`.

PR #800, opened 2021-10-12
Fix CI

Jasha10:
There is a CI failure in #799.
I want to see if CI fails on the `master` branch.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -11,11 +11,7 @@ PYTHON_VERSIONS = os.environ.get(
 
 
 def deps(session, editable_install):
-    if editable_install:
-        # Editable installs not working in pip 21.3.0, https://github.com/pypa/pip/issues/10573
-        session.install("--upgrade", "setuptools", "pip!=21.3.0")
-    else:
-        session.install("--upgrade", "setuptools", "pip")
+    session.install("--upgrade", "setuptools", "pip")
     extra_flags = ["-e"] if editable_install else []
     session.install("-r", "requirements/dev.txt", *extra_flags, ".", silent=True)
 
--- a/setup.py
+++ b/setup.py
@@ -46,8 +46,6 @@ with open("README.md", "r") as fh:
         description="A flexible configuration library",
         long_description=LONG_DESC,
         long_description_content_type="text/markdown",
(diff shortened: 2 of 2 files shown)


jieru-hu:
it's probably still a good idea to upgrade pip. in Hydra's nox, `session.run` is used instead of session install, do you mind giving that a try instead? https://github.com/facebookresearch/hydra/blob/main/noxfile.py#L186-L188

layday:
The issue here is that the `CleanCommand` from `build_helpers` is being called while building omegaconf to install it in the nox session, which wipes the `.nox` folder clean.  You are only experiencing this in 21.3, [because](https://github.com/pypa/pip/blob/main/NEWS.rst):

> - In-tree builds are now the default. ``--use-feature=in-tree-build`` is now
  ignored. ``--use-deprecated=out-of-tree-build`` may be used temporarily to ease
  the transition.

... and is unrelated to the editable issue.

layday:
To clarify, previously, pip would make a copy of your entire project folder in $TMPDIR and build the package there. After 21.3, it runs the build from the project's root.

Jasha10:
Yes indeed, it seems that not only is `pytest-runner` deprecated, but also:
- `setup.py test` is [deprecated](https://github.com/pypa/setuptools/pull/1878)
- The `setuptools.setup` keyword argument `setup_requires` is [deprecated](https://setuptools.pypa.io/en/stable/userguide/dependency_management.html#package-requirement).
- The `setuptools.setup` keyword argument `tests_require` is [deprecated](https://setuptools.pypa.io/en/stable/userguide/dependency_management.html#optional-dependencies).

@omry, do you know if our downstream dependents are relying on omegaconf's `setup.py test` command? If not, we can remove `setup_requires` and `tests_require` from the `setup.py` file so that pip will not need to be pinned for editable installs in the noxfile.

Jasha10:
> I don't know. Maybe the conda package.

Looking at the [recipe](https://github.com/conda-forge/omegaconf-feedstock/blob/1d398211d3bf63699c12712d48e8dfb7beb50e4c/recipe/meta.yaml) in the omegaconf-feedstock repo, I'm confident that `setup.py test` is not being used there.

PR #804, opened 2021-10-14
document interaction between OmegaConf.merge and the "readonly" flag

### LINKED ISSUE #802: Read-only flag not working during `OmegaConf.merge`
## Summary
Even if we set Read-only flag, it do not prevent `OmegaConf.merge` (write operation).  

## To Reproduce
Work on value reassignment.  
```python

conf = OmegaConf.create({"field_a": "hello"})
OmegaConf.set_readonly(conf, True)

conf.field_a = "updated"
# ReadonlyConfigError: Cannot change read-only config container
```

Not work on config merge.  
```python

conf = OmegaConf.create({"field_a": "hello"})
OmegaConf.set_readonly(conf, True)

conf_update = OmegaConf.create({"field_a": "updated"})
conf_v2 = OmegaConf.merge(conf, conf_update) 
print(conf_v2.field_a)
# updated
```

## Expected behavior
`merge` will be block when read-only field will be replaced.  

`merge` is NOT literally *in-place write operation*, but it is practically **write operation on single true config**.  
In my opinion, Read-only flag should be useful when I want to protect some field from accidental override by merge.  

At least, this behavior should be explicitly documented.   

## Additional context
- [x] OmegaConf version: 2.1.1
- [x] Python version: Python 3.7.12
- [x] Operating system: Ubuntu 18.04.5 LTS

Jasha10:
Closes #802.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -515,19 +515,10 @@ You can temporarily remove the read only flag from a config object:
     >>> conf.a.b
     20
 
-Note that it is possible to call `OmegaConf.merge` on a read-only config:
-
-.. doctest:: loaded
-
-    >>> cfg1 = OmegaConf.create({"a": 123})
-    >>> cfg2 = OmegaConf.create({"a": 456})
-    >>> OmegaConf.set_readonly(cfg1, True)
-    >>> OmegaConf.set_readonly(cfg2, True)
-    >>> OmegaConf.merge(cfg1, cfg2)
-    {'a': 456}
-
-This merge operation creates a new config object as its output;
-the inputs (which are read-only) are not modified.
+Note that it is possible to use read-only config objects as inputs to
+`OmegaConf.merge`; the merge operation creates a new config object as its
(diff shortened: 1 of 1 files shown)


omry:
> I've realized that this update to the docs does not paint a full picture of the interaction between merge and readonly:
> 
> ```python
> cfg1 = OmegaConf.create({"a": {"b": 123}})
> cfg2 = OmegaConf.create({"a": {"b": 456}})
> OmegaConf.set_readonly(cfg1.a, True)
> OmegaConf.merge(cfg1, cfg2)
> # omegaconf.errors.ReadonlyConfigError: Cannot change read-only config container
> #     full_key: a.b
> #     object_type=dict
> ```
> 
> If `cfg2.a` is set as readonly instead of `cfg1.a` then no exception is raised.
> 
> I'm thinking it will be better if I write a general note on interaction between `OmegaConf.merge` and config flags, i.e. what flags are expected to be set on the merge operation's output.

This feels like a bug. I think there are some tests testing the combination of readonly and merge but something is obviously weird here.

Issue #793, opened 2021-10-01
hasattr raises KeyValidationError when querying key with wrong type


Jasha10:
Orthogonal to this fix, we should also probably fix OmegaConf to not fail when hasattr is called if the key is mismatching (it should return False).

_Originally posted by @omry in https://github.com/facebookresearch/hydra/issues/1839#issuecomment-931560166_

This is related to issue https://github.com/facebookresearch/hydra/issues/1834 and PR https://github.com/facebookresearch/hydra/pull/1839.

Jasha10:
Here is an example reproducing the issue:
```python
from dataclasses import dataclass, field
from enum import Enum
from typing import Any, Dict

from pytest import raises

from omegaconf import OmegaConf
from omegaconf.errors import KeyValidationError

class Category(Enum):
    X = 0
    Y = 1


@dataclass
class Conf:
    enum_dict: Dict[Category, str] = field(default_factory=dict)
    int_dict: Dict[int, str] = field(default_factory=dict)
    any_dict: Dict[Any, str] = field(default_factory=dict)


cfg = OmegaConf.structured(Conf)

with raises(KeyValidationError):
    hasattr(cfg.enum_dict, "foo")  # incorrectly raises KeyValidationError

with raises(KeyValidationError):
    hasattr(cfg.int_dict, "bar")  # incorrectly raises KeyValidationError

assert hasattr(cfg.any_dict, "baz") == False  # as expected
```

#### The expected behavior
The following assertions should not fail:
```python
hasattr(cfg.enum_dict, "foo")
hasattr(cfg.int_dict, "bar")
```

Jasha10:
From the python [docs](https://docs.python.org/3/library/functions.html#hasattr):

> hasattr(object, name)
>    The arguments are an object and a string. The result is True if the string is the name of one of the object’s attributes, False if not. (This is implemented by calling getattr(object, name) and seeing whether it raises an AttributeError or not.)

So we need to make our `__getattr__` implementation raise an `AttributeError`.
To maintain our record of only raising exceptions that derive from `OmegaConfBaseException`, we can define a new exception class that inherits both from `AttributeError` and from `OmegaConfBaseException` (maybe `OmegaConfAttributeError`?).

PR #816, opened 2021-11-04
ensure hasattr returns False for invalid key types

pixelb:
Note we diverge __getattr__ (cfg.foo) and __getitem__ (cfg["foo"])
implementations here as the former attribute is implicitly always
a string, so we should not be validating types on it.

Note we don't overload the existing "validate_access" flag on
these methods, as we do want to validate config traversal
for the specified node, (for example checking for RecursionError).

Therefore we introduce a new "validate_key" flag to
select the separate behavior for __getattr__().

* omegaconf/dictconfig.py (__getattr__): Pass a new 'validate_key'
flag through __get_impl__ and __get_node__, setting it to false.
* tests/test_basic_ops_dict.py: Add a test case, checking
explicit hasattr() and implicit attribute.access syntax.

The significant tests for this change are:

  pytest tests/test_basic_ops_dict.py -k test_typed_hasattr
  pytest tests/test_basic_ops_dict.py -k test_getitem_with_invalid_key
  pytest tests/test_errors.py -k test_get_full_key_failure_in_format_and_raise


Jasha10:
Closes #793.

Jasha10 (approved):
lgtm!

PR #819, opened 2021-11-13
Add grammar regex test for invalid escaped brace

odelalleau:
This new test is meant to answer the question "why doesn't `_arg`
include the backslash character while it is supported in the
`UNQUOTED_CHAR` parser rule?"

The answer to that question is that if we included the backslash
character, then this new test would fail because the
`SIMPLE_INTERPOLATION_PATTERN` regex would believe that `${foo:\}` is a
valid interpolation syntax, while it is not.

There is currently no other test that would fail if the backslash
character was added to `_arg`, motivating the need for such a test.

NB: this commit does *not* modify `_arg`, it just rewrites it in raw
string format to make it more readable.


pixelb (approved):
thanks

PR #842, opened 2021-12-14
Refactor test_structured_config.py: followup to #558

Jasha10:
PR #558 introduced the `module` fixture into test_structured_config.py.
This PR converts a straggler function that slipped through in #558 to use the `module` fixture as well.


jieru-hu (approved):
nice!

PR #856, opened 2022-01-30
remove unneeded noqa: F601 comments

Jasha10:
I think these `noqa` comments are not necessary.

Let's see if the CI passes...


Jasha10:
flake8's [docs](https://flake8.pycqa.org/en/latest/user/error-codes.html) say that F601 applies when "dictionary key name repeated with different values", which does not seem to apply here.

PR #860, opened 2022-02-01
support slice assignment in ListConfig (#736)

pixelb:

PR #861, opened 2022-02-07
fix issue with new pytest

pixelb:
Getting similar issue as with hydra:
https://github.com/facebookresearch/hydra/pull/2016


pixelb:
merged as part of #860

PR #865, opened 2022-02-26
Adds literal node

### LINKED ISSUE #422: Allow the use of typing.Literal as a pythonic alternative to Enums

Enums are a bit awkward to define in Python. On the other hand, using strings to represent a set of choices is very common in Python code.

```python
from typing import Literal  # before Python 3.8, import from `typing_extensions`

@dataclass
class SvmConfig:
  kernel: Literal["linear", "poly", "rbf"] = "linear"

config: SvmConfig = OmegaConf.structured(SvmConfig(kernel="rbf"))

config.kernel = "poly"  # ok!
config.kernel = "exponential"  # error!
```

Enums, see above.

**Additional context**

My main motivation is to use this with Hydra.

jordan-schneider:
Adds literal node, solving #422 

I tried to implement this as a wrapper around `EnumNode`, but this turned out to be impossible. `Literal` can accept `int, str, bool` and `Enum` values, but enum keys must be strings. So for example

```python

@dataclass
class Config:
  mode: Literal["5", 5]
```

there is no way to map both of these values to separate enum keys without also causing a collision somewhere else. 

This PR validates a `LiteralNode` value by insisting on both `=` equality and type equality. Both are needed, otherwise python's promiscuous primitive equalities cause problems e.g. `1.0` is valid under `Literal[True]`.

I tried to add tests wherever `EnumNode` had tests, but I'm unfamiliar with the codebase. Please let me know if additional tests are needed.

### DIFF
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -46,7 +46,7 @@ except ImportError:  # pragma: no cover
 if sys.version_info >= (3, 8):
     from typing import Literal
 else:
-    from typing_extensions import Literal
+    from typing_extensions import Literal  # pragma: no cover
 
 
 # Regexprs to match key paths like: a.b, a[b], ..a[c].d, etc.
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -44,7 +44,7 @@ except ImportError:  # pragma: no cover
     attr = None  # type: ignore # pragma: no cover
 
 if sys.version_info >= (3, 8):
-    from typing import Literal
+    from typing import Literal  # pragma: no cover
 else:
     from typing_extensions import Literal  # pragma: no cover
 
@@ -600,10 +600,14 @@ def is_tuple_annotation(type_: Any) -> bool:
 
 def is_literal_annotation(type_: Any) -> bool:
     origin = getattr(type_, "__origin__", None)
-    if sys.version_info < (3, 7, 0):
-        return origin is Literal or type_ is Literal  # pragma: no cover
-    else:
+    if sys.version_info >= (3, 8, 0):
         return origin is Literal  # pragma: no cover
+    else:
+        return (
+            origin is Literal
+            or type_ is Literal
+            or ("typing_extensions.Literal" in str(type_) and not isinstance(type, str))
+        )  # pragma: no cover
 
 
 def is_dict_subclass(type_: Any) -> bool:
@@ -841,14 +845,15 @@ def type_str(t: Any, include_module_name: bool = False) -> str:
         or isinstance(t, bytes)
         or isinstance(t, Enum)
     ):
-        return str(t)
+        # only occurs when using typing.Literal after 3.8
+        return str(t)  # pragma: no cover
 
     if sys.version_info < (3, 7, 0):  # pragma: no cover
         # Python 3.6
         if hasattr(t, "__name__"):
             name = str(t.__name__)
         else:
-            if t.__origin__ is not None:
+            if hasattr(t, "__origin__") and t.__origin__ is not None:
                 name = type_str(t.__origin__)
             else:
                 name = str(t)
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -855,7 +855,7 @@ def type_str(t: Any, include_module_name: bool = False) -> str:
         if hasattr(t, "__name__"):
             name = str(t.__name__)
         else:
-            if hasattr(t, "__origin__") and t.__origin__ is not None:
+            if getattr(t, "__origin__", None) is not None:
                 name = type_str(t.__origin__)
             else:
                 name = str(t)
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -600,7 +600,7 @@ def is_tuple_annotation(type_: Any) -> bool:
 
 def is_literal_annotation(type_: Any) -> bool:
     origin = getattr(type_, "__origin__", None)
-    if sys.version_info >= (3, 8, 0):
+    if sys.version_info >= (3, 8):
         return origin is Literal  # pragma: no cover
     else:
         return (
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -12,7 +12,6 @@ from typing import (
     Dict,
     Iterator,
     List,
-    Literal,
     Optional,
     Tuple,
     Type,
@@ -43,6 +42,11 @@ try:
 
 except ImportError:  # pragma: no cover
     attr = None  # type: ignore # pragma: no cover
+
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
 
 
 # Regexprs to match key paths like: a.b, a[b], ..a[c].d, etc.
@@ -831,6 +835,13 @@ def type_str(t: Any, include_module_name: bool = False) -> str:
         return "Any"
     if t is ...:
         return "..."
+    if (
+        isinstance(t, int)
+        or isinstance(t, str)
+        or isinstance(t, bytes)
+        or isinstance(t, Enum)
+    ):
+        return str(t)
 
     if sys.version_info < (3, 7, 0):  # pragma: no cover
         # Python 3.6
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -606,7 +606,9 @@ def is_literal_annotation(type_: Any) -> bool:
         return (
             origin is Literal
             or type_ is Literal
-            or ("typing_extensions.Literal" in str(type_) and not isinstance(type, str))
+            or (
+                "typing_extensions.Literal" in str(type_) and not isinstance(type_, str)
+            )
         )  # pragma: no cover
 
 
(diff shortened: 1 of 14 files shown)


jordan-schneider:
Okay I clearly missed a bunch of tests, working on that. Not sure what's up with mypy not thinking Literal exists.

jordan-schneider:
The literal thing was because setup.cfg sets the mypy version to 3.6 by default. I had a legitimate problem where I need to import from `typing_extensions` before 3.8, which I now do. I also added a dependency to the pre-commit hooks to fix a bunch of mypy errors locally, although this didn't appear to show up in the last round of circleci results.

jordan-schneider:
Looks like I've ironed out all the python version issues.

Jasha10:
Hi @jordan-schneider! Thanks for taking the time to create this PR.
At the moment, I can't commit to a timeline for merging this, as the OmegaConf team's hands are full.

Jasha10:
Thanks very much @jordan-schneider.
Unfortunately it may be some time before we are able to give this PR a serious review as the Hydra/OmegaConf team has its hands full at the moment.

omegaconf/_utils.py
```diff
@@ -593,6 +598,20 @@ def is_tuple_annotation(type_: Any) -> bool:
         return origin is tuple  # pragma: no cover
 
 
+def is_literal_annotation(type_: Any) -> bool:
+    origin = getattr(type_, "__origin__", None)
+    if sys.version_info >= (3, 8, 0):
```
thomkeh, line 603: to be consistent with the others
```suggestion
    if sys.version_info >= (3, 8):
```

omegaconf/_utils.py
```diff
@@ -822,13 +841,21 @@ def type_str(t: Any, include_module_name: bool = False) -> str:
...
+        # only occurs when using typing.Literal after 3.8
+        return str(t)  # pragma: no cover
 
     if sys.version_info < (3, 7, 0):  # pragma: no cover
         # Python 3.6
         if hasattr(t, "__name__"):
             name = str(t.__name__)
         else:
-            if t.__origin__ is not None:
+            if hasattr(t, "__origin__") and t.__origin__ is not None:
```
thomkeh, line 858: maybe
```suggestion
            if getattr(t, "__origin__", None) is not None:
```

omegaconf/_utils.py
```diff
@@ -593,6 +598,20 @@ def is_tuple_annotation(type_: Any) -> bool:
...
+    origin = getattr(type_, "__origin__", None)
+    if sys.version_info >= (3, 8, 0):
+        return origin is Literal  # pragma: no cover
+    else:
+        return (
+            origin is Literal
+            or type_ is Literal
+            or (
+                "typing_extensions.Literal" in str(type_) and not isinstance(type_, str)
+            )
```
thomkeh, line 611: Hmm, why is this needed? Isn't `Literal` always from `typing_extensions` in this `else` branch?

omegaconf/nodes.py
```diff
@@ -400,6 +401,62 @@ def __deepcopy__(self, memo: Dict[int, Any]) -> "EnumNode":
...
+    ):
+        if not is_literal_annotation(literal_type):
+            raise ValidationError(
+                f"LiteralNode can only operate on Literal annotation ({literal_type})"
+            )
+        self.literal_type = literal_type
+        if hasattr(self.literal_type, "__args__"):
+            self.fields = list(self.literal_type.__args__)  # pragma: no cover
+        elif hasattr(self.literal_type, "__values__"):  # pragma: no cover
+            self.fields = list(self.literal_type.__values__)  # pragma: no cover
```
thomkeh, line 422: Maybe make these branches explicitly correspond to python versions. So I think
```suggestion
        if sys.version_info >= (3, 7):
            self.fields = list(self.literal_type.__args__)  # pragma: no cover
        else:
            self.fields = list(self.literal_type.__values__)  # pragma: no cover
```
but I'm not really sure about this.

tmke8 (commented):
Hi, I'm the one who originally opened #422. So I'm very happy you are doing this!

I'm obviously not an official member of OmegaConf, but I thought I would try to help you with a code review anyway.

Some questions:

1. What happens if I annotate a node with just `Literal` without any type arguments for `Literal`?
2. Don't we have to add something like `typing_extensions;python_version<'3.8'` to `requirements/base.txt`? Or is `typing_extensions` a dependency anyway?

omegaconf/nodes.py
jordan-schneider: Why do you feel this is better? Philosophically I tend to think that directly checking is the right way to do things, in case e.g. the attribute name every switches back or something equally cursed. This also has the benefit that mypy/pylance/other syntax parsers can correctly infer that the attribute exists, which is left implicit if you're going by version.

If this is about readability, I'd be happy to add comments saying why the check is necessary.

omegaconf/_utils.py
jordan-schneider: So what's happening here is that:

python 3.8 and later has `typing.Literal` which has an `__origin__` attribute
python 3.7 `typing_extensions.Literal` has an `__origin__` attribute
python 3.6 and earlier `typing_extensions.Literal` does not have an `__origin__` attribute or any equivalent attribute

AND

```python
A = Literal[5]
A is Literal  # False
isinstance(A, Literal)  # TypeError: isinstance() arg 2 must be a type or tuple of types
issubclass(type(A), Literal)  # TypeError: issubclass() arg 2 must be a class or tuple of classes
```

so as far as I can tell, for 3.6 and earlier the only way to check for Literalness is using the string. This is gross but I don't know of a better way.

This logic is unnecessarily complicated though. How do you feel about

```python
def is_literal_annotation(type_: Any) -> bool:
    origin = getattr(type_, "__origin__", None)
    return origin is Literal or ("typing_extensions.Literal" in str(type_) and not isinstance(type_, str))
```

omegaconf/nodes.py
thomkeh: Right, I was just thinking that at some point in the future, omegaconf might want to drop support for Python 3.6 and then it would be useful that you can directly see that this if-else check has become unnecessary.

But, yes, a comment is likely already sufficient.

jordan-schneider:
> Hi, I'm the one who originally opened #422. So I'm very happy you are doing this!
> 
> I'm obviously not an official member of OmegaConf, but I thought I would try to help you with a code review anyway.
> 
> Some questions:
> 
>     1. What happens if I annotate a node with just `Literal` without any type arguments for `Literal`?
> 
>     2. Don't we have to add something like `typing_extensions;python_version<'3.8'` to `requirements/base.txt`? Or is `typing_extensions` a dependency anyway?

1.  I don't have time at the moment to directly check this but Literal.__values__/__args__ still exists but is empty, so the LiteralNode is constructed and whatever the value is it fails to validate. This seems like the correct behavior. I'll write a test for this when I have a moment.

2. I think it's a dependency of one of the other dependencies but it should be added explicitly anyway, in case that changes. Again I'll handle this when I have a moment.

omegaconf/_utils.py
thomkeh: There is an `is_literal_type()` function [here](https://github.com/ilevkivskyi/typing_inspect/blob/master/typing_inspect.py#L222) which also work with Python 3.6. The solution to the problem seems very stupid, you have to import a hidden class `_Literal`:

```
Python 3.6.13 |Anaconda, Inc.| (default, Jun  4 2021, 14:25:59)
[GCC 7.5.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from typing_extensions import Literal, _Literal
>>> a = Literal[0]
>>> type(a) is Literal
False
>>> type(a) is _Literal
True
```
I have no idea why.

Jasha10:
> If I remove that type: ignore comment locally I do in fact get an error. Anyone know what's going on here?

@jordan-schneider maybe upgrade the version of mypy that you have installed locally?

Jasha10:
> If I remove that type: ignore comment locally I do in fact get an error. Anyone know what's going on here?

The CI should pass if you rebase this branch against the `master` branch to incorporate changes from #878.

jordan-schneider:
I think it was mostly outdated local versions of tools. The pre-commit version of black and the current version used by CircleCI also disagreed, so I updated the pre-commit version as well.

codekansas:
Are there plans to merge this? It would be super useful!

Jasha10:
Thanks for the nudge, @codekansas!
> Are there plans to merge this?

Yes. We haven't been able to finalize this PR yet as the OmegaConf team's bandwidth is limited. It is on the roadmap.

indigoviolet:
@Jasha10 nudge again to enquire about plans/status since some time has passed

Queuecumber:
@omry @Jasha10 would appreciate some eyes on this, it's been in limbo for a year and a half and would be a nice way to produce clean structured configs

daturkel:
+1 this feature would be great and feels more pythonic than `Enum`s.

spfrommer:
+1

gsq7474741:
+1

jbohnslav:
+1, still relevant

SheaML:
+1

jkulhanek:
+1

omry:
Hi @jordan-schneider,
Thanks for your work on this.
I reworked it in #1272 and will integrate into 2.4.0

Issue #844, opened 2021-12-20
Have `bytes` as supported value type.


pedropgusmao:
**Is your feature request related to a problem? Please describe.**
I need to instantiate an object with another one which is basically a dataclass instance like:
```@dataclass
class Parameters:
    """Model parameters."""
    tensors: List[bytes] 
    tensor_type: str 
```
when trying to use this I get:
`omegaconf.errors.ValidationError: Unsupported value type: <class 'bytes'>`
Example:
```from dataclasses import dataclass
from typing import List

from omegaconf import OmegaConf

@dataclass
class Parameters:
    """Model parameters."""
    tensors: List[bytes] 
    tensor_type: str 

# omegaconf.errors.ValidationError: Unsupported value type: <class 'bytes'>
OmegaConf.structured(Parameters)
```

**Describe the solution you'd like**
Accepting `bytes`.

**Describe alternatives you've considered**
Internally interpreting `bytes` as `str` (when parsing) then converting it to `bytes`.

**Additional context**
...

Jasha10:
Thanks for suggesting this @pedropgusmao.
Supporting `bytes` seems reasonable to me since `bytes` is a built-in type and since yaml supports the `bytes` type natively.

pixelb:
Thanks for the suggestion and implementation!

I agree this has some merit, as binary is a yaml primitive: https://yaml.org/type/binary.html
Playing devil's advocate, some thoughts against implementing this:

- "binary" is just a base64 string, so why not do a little manual conversion in code to/from storing b64 in existing strings?
  - It would be good to have a concrete example of before/after code for this use case. BTW I vaguely remember a suggestion to support tensor as a primitive type, which had less merit, but some of the same considerations as this, as that's what's presented as the use case. Also slightly related to leaning on explicit conversion more is issue #805 discussing numpy float handling.
- We have to be careful to avoid becoming a data serialization mechanism. We're tuned for smaller amounts of config, rather than larger amounts of data

If we do support bytes, I'm not sure about the support for converting to/from strings. That would have to be restricted to utf-8, so I don't see the use case for this conversion support

Jasha10:
I've agree about the conversion support -- it's cleaner in my opinion if assigning `str` to a `bytes` field fails and assigning `bytes` to a `str` field fails.
cc @jieru-hu

Jasha10:
>     * "binary" is just a base64 string, so why not do a little manual conversion in code to/from storing b64 in existing strings?
>       
[...]

PR #845, opened 2021-12-21
Include support to `bytes`  built-in type.

pedropgusmao:
Implements support for `bytes` in response to #844 .

PR #872, opened 2022-03-04
Include support for bytes built-in type.

Jasha10:
This PR is the direct successor to the BytesNode PR #845 (which I closed accidentally).

### DIFF
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -284,7 +284,7 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
             #   assert hash(0) == hash(False)
             #   assert hash(1) == hash(True)
             return bool(key)
-        elif key_type in (str, int, float, bool):  # primitive type
+        elif key_type in (str, bytes, int, float, bool):  # primitive type
             if not isinstance(key, key_type):
                 raise KeyValidationError(
                     f"Key $KEY ($KEY_TYPE) is incompatible with ({key_type.__name__})"
@@ -476,7 +476,7 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
         value: Optional[Node] = self.__dict__["_content"].get(key)
         if value is None:
             if throw_on_missing_key:
-                raise ConfigKeyError(f"Missing key {key}")
+                raise ConfigKeyError(f"Missing key {key!s}")
         elif throw_on_missing_value and value._is_missing():
             raise MissingMandatoryValue("Missing mandatory value: $KEY")
         return value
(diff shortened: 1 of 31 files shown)


Jasha10:
Closes #844.

Jasha10:
The only differences between this PR and #845 are (1) I have rebased the diff against `master` and (2) I have squashed the commits.
I feel that the PR is in good shape, but I wanted to ask: @jieru-hu and @pixelb, are either of you interested in giving this a final review before I merge it?

omegaconf/dictconfig.py
```diff
@@ -507,9 +507,11 @@ def pop(self, key: DictKeyType, default: Any = _DEFAULT_MARKER_) -> Any:
                 else:
                     full = self._get_full_key(key=key)
                     if full != key:
-                        raise ConfigKeyError(f"Key not found: '{key}' (path: '{full}')")
+                        raise ConfigKeyError(
+                            f"Key not found: '{key!s}' (path: '{full}')"
+                        )
                     else:
-                        raise ConfigKeyError(f"Key not found: '{key}'")
+                        raise ConfigKeyError(f"Key not found: '{key!s}'")
```
jieru-hu, line 514: noob q:
what is `!s` and what is the benefit of doing this? a quick test does not seem to tell me the difference...

```
>>> key=b"foo"
>>> s=f"Key not found: '{key!s}'"
>>> s
"Key not found: 'b'foo''"
>>> no_s=f"Key not found: '{key}'"
>>> no_s
"Key not found: 'b'foo''"
>>> s == no_s
True
```

Jasha10: There is no difference from python3's perspective. The `!s` is to silence a mypy warning about differences between python2 and python3:
```python
# tmp.py
key=b"foo"
s=f"Key not found: '{key!s}'"
no_s=f"Key not found: '{key}'"
```
```text
$ mypy tmp.py
tmp.py:4: error: On Python 3 formatting "b'abc'" with "{}" produces "b'abc'", not "abc"; use "{!r}" if this is desired behavior
Found 1 error in 1 file (checked 1 source file)
```

Jasha10: See the python docs on [Format String Syntax](https://docs.python.org/3/library/string.html#format-string-syntax):

> Three conversion flags are currently supported: '!s' which calls [str()](https://docs.python.org/3/library/stdtypes.html#str) on the value, '!r' which calls [repr()](https://docs.python.org/3/library/functions.html#repr) and '!a' which calls [ascii()](https://docs.python.org/3/library/functions.html#ascii).
> 
> Some examples:
> ```
> "Harold's a clever {0!s}"        # Calls str() on the argument first`
> "Bring out the holy {name!r}"    # Calls repr() on the argument first
> "More {!a}"                      # Calls ascii() on the argument first
> ```

PR #873, opened 2022-03-07
add pathlib.Path support

### LINKED ISSUE #97: Add support for pathlib.Path type
Once stronger typing is implemented in the library, consider adding support for the `pathlib.Path` type.

Motivation: paths are often found in configs and `pathlib.Path` usage is spreading across the Python ecosystem.

gwenzek:
Add support for Path in Omegaconf.

Path type will never be automatically inferred when loading from YAML, so we will continue creating StringNode by default.
But if you use `structured` or `_promote` with a dataclass using Path some of `StringNode` can be converted to `PathNode` eg:

https://github.com/gwenzek/omegaconf/blob/1df8a05f9f740b7085fb65fe7a4d8b1e034f2ea6/tests/structured_conf/test_structured_config.py#L819-L833

This change is backward compatible, because code that currently use dataclass with Path fields currently see errors when running `structured`.

This should resolve #97 

cc: @odelalleau @Mortimerp9

### DIFF
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -1,5 +1,6 @@
 import copy
 import os
+import pathlib
 import re
 import string
 import sys
@@ -680,16 +681,19 @@ def _valid_dict_key_annotation_type(type_: Any) -> bool:
     return type_ is None or type_ is Any or issubclass(type_, DictKeyType.__args__)  # type: ignore
 
 
+BASE_TYPES = (
+    int,
+    float,
+    bool,
+    bytes,
+    str,
+    type(None),
+)
+
+
 def is_primitive_type_annotation(type_: Any) -> bool:
     type_ = get_type_of(type_)
-    return issubclass(type_, Enum) or type_ in (
-        int,
-        float,
-        bool,
-        str,
-        bytes,
-        type(None),
-    )
+    return issubclass(type_, (Enum, pathlib.Path)) or type_ in BASE_TYPES
 
 
 def _get_value(value: Any) -> Any:
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -152,6 +152,20 @@ def get_yaml_loader() -> Any:
         ]
         for key, resolvers in loader.yaml_implicit_resolvers.items()
     }
+
+    loader.add_constructor(
+        "tag:yaml.org,2002:python/object/apply:pathlib.Path",
+        lambda loader, node: pathlib.Path(*loader.construct_sequence(node)),
+    )
+    loader.add_constructor(
+        "tag:yaml.org,2002:python/object/apply:pathlib.PosixPath",
+        lambda loader, node: pathlib.PosixPath(*loader.construct_sequence(node)),
+    )
+    loader.add_constructor(
+        "tag:yaml.org,2002:python/object/apply:pathlib.WindowsPath",
+        lambda loader, node: pathlib.WindowsPath(*loader.construct_sequence(node)),
+    )
+
     return loader
 
 
--- a/docs/source/structured_config.rst
+++ b/docs/source/structured_config.rst
@@ -6,6 +6,7 @@
     from enum import Enum
     from dataclasses import dataclass, field
     import os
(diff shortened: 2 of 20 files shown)


Jasha10:
Closes #97.

Jasha10:
Thanks very much for taking the time to create this PR, @gwenzek.
At the moment, I can't commit to a timeline for merging this, as the OmegaConf team currently has it's hands full.

Jasha10 (changes requested):
```python 
>>> OmegaConf.create({"foo": Path("bar")})
...
omegaconf.errors.UnsupportedValueType: Value 'PosixPath' is not a supported primitive type
    full_key: foo
    object_type=dict
```
I think we should support the above use-case. This would probably require updating the [`is_primitive_type`](https://github.com/omry/omegaconf/blob/5e73ee6a1c0f3a63fc420bf856750b80001a5644/omegaconf/_utils.py#L651) function

Jasha10:
Thanks @gwenzek. The implementation looks good to me.
I am planning to push some commits to this PR branch adding more tests.

gwenzek:
I rebased to resolve the conflict with the new `bytes` node.

omegaconf/_utils.py
```diff
@@ -648,16 +649,19 @@ def _valid_dict_key_annotation_type(type_: Any) -> bool:
...
     type_ = get_type_of(type_)
-    return issubclass(type_, Enum) or type_ in (
-        int,
-        float,
-        bool,
-        str,
-        bytes,
-        type(None),
-    )
+    return issubclass(type_, (Enum, pathlib.Path)) or type_ in BASE_TYPES
```
Jasha10, line 664: I think checking `issubclass(type_, pathlib.Path)` here is not the best approach.
We don't want to support custom user-defined subclasses of `Path` (i.e. `class MyPath(pathlib.Path): ...` should not be supported).

Jasha10 (changes requested):

omegaconf/_utils.py
gwenzek: the problem is that `pathlib.Path` isn't an actual class. The classes are `PosixPath` and `WindowsPath`

Jasha10: Oh, I see. Thanks for the clarification.

Jasha10:
Demo of the API:
```python
>>> from omegaconf import OmegaConf; from pathlib import Path
>>> OmegaConf.create({"foo": Path("hello.txt")})
{'foo': PosixPath('hello.txt')}
```
`pathlib.Path` can be used as a type annotation with structured configs too.
Paths are preserved through a round-trip to yaml:
```python
>>> cfg = OmegaConf.create({"foo": Path("hello.txt")})
>>> print(OmegaConf.to_yaml(cfg))
foo: !!python/object/apply:pathlib.PosixPath
- hello.txt

>>> OmegaConf.create(OmegaConf.to_yaml(cfg))
{'foo': PosixPath('hello.txt')}
```

Notes:
- There is no automatic conversion from `Path` to `str`. This means assigning
  `Path("foo.txt")` to `MyStructuredConf.string_field` will fail with a
  ValidationError. Conversion from string to path still works (assigning the
  string `"foo.txt"` to `MyStructuredConf.path_field` succeeds).
- Using the yaml tag `!!python/object/apply:...` means that the output of `OmegaConf.to_yaml` can be deserialized using vanilla `pyyaml` (i.e. without using OmegaConf):
  ```python
  import yaml
  yaml_document = OmegaConf.to_yaml(OmegaConf.create({"foo": Path("hello.txt")}))
  assert OmegaConf.create(yaml_document) == yaml.unsafe_load(yaml_document)
  ```

cc @rsokl, @odelalleau

jieru-hu (approved):
looks great!

Jasha10 (approved):

Mortimerp9:
Awesome! Thanks for getting this in Hydra!

omry:
RFC on new type adapter design for OmegaConf:

https://github.com/omry/omegaconf/issues/1288

PR #913, opened 2022-05-05
preliminary UnionNode implementation

Jasha10:
This PR implements support for unions of primitive types (i.e. unions of `int`, `float`, `bool`, `str`, `bytes`, and `None`).

A few notes:
- `Union` can be freely mixed with `Optional` and `None`. For example, all of the following are equivalent:
  - `Optional[Union[int, str]]`
  - `Union[Optional[int], str]`
  - `Union[int, str, None]`
  - `Union[int, str, type(None)]`
- Containers of unions (e.g. `List[Union[int, str]]`, or `Dict[int, Union[str, bytes]]`) *are* supported.
- Unions of containers (e.g. `Union[List[str], Dict[str, str], MyStructuredConfig]`) *are not yet* supported. This will come in a followup PR.

Here is a demo of the API:
```python
from omegaconf import OmegaConf, ValidationError
from typing import Union
from pytest import raises
@dataclass
class Foo:
    u: Union[int, str] = 123

cfg = OmegaConf.create(Foo)
assert cfg.u == 123
cfg.u = "abc"  # ok

with raises(ValidationError):
    cfg.u = b"binary"  # bytes not compatible with union

with raises(ValidationError):
    OmegaConf.create(Foo(1.2))  # float not compatible
```

Implementation Notes:
- `UnionNode` is a "pass-through" node. That means most validation logic is delegated to a child node.
- I've introduced a new method `Container._get_child` which parallels `Container._get_node`. The difference between these methods is that `_get_child` automatically "passes over" Unions. Similarly, I've introduced `Node._get_parent_container()`, which is similar to `Node._get_parent()` except that it passes over union nodes in the parent heirarchy.
```python
from typing import Dict, Union
from omegaconf import MISSING, OmegaConf, StringNode, UnionNode

@dataclass
class Example:
    cfg: Dict[str, Union[str, int]] = MISSING

cfg = OmegaConf.create(Example({"foo": "bar"})).cfg

assert isinstance(cfg._content, dict)
assert isinstance(cfg._content["foo"], UnionNode)
assert isinstance(cfg._content["foo"]._content, StringNode)
assert isinstance(cfg._content["foo"]._content._val, str)
assert cfg._content["foo"]._content._val == "bar"

assert cfg._content["foo"]._metadata.ref_type == Union[int, str]
assert cfg._content["foo"]._content._metadata.ref_type == str

assert cfg._get_node("foo") is cfg._content["foo"]            # UnionNode
assert cfg._get_child("foo") is cfg._content["foo"]._content  # StringNode
```

TODO before merging this
- [x] docs
- [x] serialization test
- [x] Double-check usage of `_get_parent` vs `_get_parent_container` throughout the codebase.
- [x] Double-check usage of `_get_node` vs `_get_child` throughout the codebase.

### DIFF
--- a/docs/source/structured_config.rst
+++ b/docs/source/structured_config.rst
@@ -32,14 +32,16 @@ in the input class.
 Currently, type hints supported in OmegaConf’s structured configs include:
  - primitive types (``int``, ``float``, ``bool``, ``str``, ``Path``) and enum types
    (user-defined subclasses of ``enum.Enum``). See the :ref:`simple_types` section below.
+ - unions of primitive/enum types, e.g. ``Union[float, bool, MyEnum]``.
+   See :ref:`union_types` below.
  - structured config fields (i.e. MyConfig.x can have type hint MySubConfig).
    See the :ref:`nesting_structured_configs` section below.
- - optional types (any of the above can be wrapped in a ``typing.Optional[...]``
-   annotation). See :ref:`other_special_features` below.
  - dict and list types: ``typing.Dict[K, V]`` or ``typing.List[V]``, where K is
    primitive or enum, and where V is any of the above (including nested dicts
    or lists, e.g. ``Dict[str, List[int]]``).
    See the :ref:`lists` and :ref:`dictionaries` sections below.
+ - optional types (any of the above can be wrapped in a ``typing.Optional[...]``
+   annotation). See :ref:`other_special_features` below.
 
 .. _simple_types:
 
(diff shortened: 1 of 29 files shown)


pixelb (approved):
The Box abstraction is great, and make things very neat.
The implementation looks good to me.

In general splitting out the union of containers support is fine and good.

Pre-approving assuming the outstanding tasks are done.

Thanks!

Issue #821, opened 2021-11-15
Clean up omegaconf 2.1 deprecations


jieru-hu:
here is a not necessarily comprehensive [list](https://github.com/omry/omegaconf/search?q=%23+DEPRECATED%3A+remove+in+2.2)

PR #925, opened 2022-05-12
remove support for OmegaConf.is_optional()

pixelb:
Was slated for removal in 2.2

The deprecation was handled in issue #698
This addresses issue #821


pixelb:
A little reluctant with this one since no fallback

PR #944, opened 2022-05-24
Make `is_primitive_list` more precise

### LINKED ISSUE #939: BC breaking when for "allow_objects=True" flag
**Describe the bug**
In 2.2.1 version, "allow_objects=True" flag for `DictConfg` behavior changed. 

**To Reproduce**
```python
from omegaconf import OmegaConf, DictConfig
from collections import namedtuple
Shape=namedtuple("Shape", field_names=["channels", "height", "width"])
a = Shape(10, 2, 3)
cfg = DictConfig(dict(shape=a), flags={"allow_objects": True})
```

In 2.1.1:
cfg is `{'shape': Shape(channels=10, height=2, width=3)}`. 

In 2.2.1:
cfg is `[10, 2, 3]`.

Jasha10:
Closes #939.


omegaconf/_utils.py
```diff
@@ -605,9 +605,7 @@ def is_int(st: str) -> bool:
 
 
 def is_primitive_list(obj: Any) -> bool:
-    from .base import Container
-
-    return not isinstance(obj, Container) and isinstance(obj, (list, tuple))
+    return type(obj) in (list, tuple)
```
pixelb, line 608: Won't that break code like:

```
from omegaconf import OmegaConf, ListConfig
class CustomList(list):
    pass
cfg = ListConfig(CustomList((1,2,3)))
print(cfg)
```

Jasha10: Yes, you're right. I'll update the PR.

Jasha10: So it turns out that the behavior on `2.1_branch` is a bit idiosyncratic: in some cases `CustomList`/`Shape` are coerced to `ListConfig` and in other cases they either rejected or wrapped in an `AnyNode` (depending on whether `allow_objects` is set).
Anyway, the new `TestCreationWithCustomClass` captures and enforces that behavior.

pixelb (approved):
Excellent.
Thank you

Jasha10:
TODO:
- [ ] cherry-pick to `2.1_branch`.

Issue #569, opened 2021-02-25
Omegaconf from_argparse - Provide some interoperability with default argument parser


georgepar:
**Is your feature request related to a problem? Please describe.**

OmegaConf.from_cli is good, but most frameworks / individuals already use argparse for parsing command line arguments.
Specifically I'm thinking about the use case of having a config file with some model training configuration and being able to override some properties using command line args for fast experimentation.

**Describe the solution you'd like**

In my code base I've come up with this hack

```python
class OmegaConfExtended(OmegaConf):
    @staticmethod
    def from_argparse(parser: argparse.ArgumentParser):
        dest_to_arg = {v.dest: k for k, v in parser._option_string_actions.items()}

        all_args = vars(parser.parse_args())
        provided_args = {}
        default_args = {}
        for k, v in all_args.items():
            if dest_to_arg[k] in sys.argv:
                provided_args[k] = v
            else:
                default_args[k] = v

        provided = OmegaConf.create(_nest(provided_args))
        defaults = OmegaConf.create(_nest(default_args))

        return provided, defaults
```

Basically, it takes an argument parser, parses the arguments and splits them in default and user-provided arguments.
Then I can do something like `config = OmegaConf.merge(default_cli_args, config_from_yaml, user_provided_cli_args)`.
(Merge precedence: user provided CLI args > Config file > default values for CLI args)

**Describe alternatives you've considered**
The alternative is to use the default `from_cli`.

**Additional context**

Off-course this solution makes a lot of assumptions (especially about nesting).

Specifically I do something like this for nested args:

```python
parser.add_argument("--hidden", dest="model.hidden", default=1233, type=int)
```
and then have recursive function that converts `{"model.hidden": 1233}` to `{"model": {"hidden": 1233}}`

Has anybody tried something similar / toyed with this idea?

omry:
Hi,
I general OmegaConf offers really limited support for CLI integration and I am not interested in extending it.
Frameworks like Hydra are building much more powerful integrations.

from_cli is a very thin wrapper around [OmegaConf.from_dotlist](https://github.com/omry/omegaconf/blob/cab594e9b195019a2a5993ec77e755717f98d24a/omegaconf/omegaconf.py#L326):
```python
    @staticmethod
    def from_cli(args_list: Optional[List[str]] = None) -> DictConfig:
        if args_list is None:
            # Skip program name
            args_list = sys.argv[1:]
        return OmegaConf.from_dotlist(args_list)
```

[...]

PR #965, opened 2022-06-17
Add CLI param parsing for missed formats

yu-havrylko:
I encountered the problem when use `OmegaConf.from_cli()` for params in following format:
`python script.py --param1 value 1  --param2 value2`.

The purpose of this PR is to cover more variants of passed CLI arguments.

### DIFF
--- a/omegaconf/nodes.py
+++ b/omegaconf/nodes.py
@@ -543,3 +543,9 @@ class InterpolationResultNode(ValueNode):
     def _is_interpolation(self) -> bool:
         # The result of an interpolation cannot be itself an interpolation.
         return False
+
+
+class CliArgType(Enum):
+    DEFAULT = "DEFAULT"
+    POSIX_SPACE = "POSIX_SPACE"
+    POSIX_EQUAL = "POSIX_EQUAL"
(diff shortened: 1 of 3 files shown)


jieru-hu:
hi @yu-havrylko  - thanks for your contribution.

omegaconf chose to only support very limited CLI integration, pls check https://github.com/omry/omegaconf/issues/569 for more context.  in particular omry's comments [here](https://github.com/omry/omegaconf/issues/569#issuecomment-786164790)

yu-havrylko:
> hi @yu-havrylko - thanks for your contribution.
> 
> omegaconf chose to only support very limited CLI integration, pls check #569 for more context. in particular omry's comments [here](https://github.com/omry/omegaconf/issues/569#issuecomment-786164790)

Thank's for the feedback!
So there's no chance to have it in OmegaConf
It seems to me overhead to add hydra for my repo, just for parsing CLI params

Jasha10:
> Thank's for the feedback! So there's no chance to have it in OmegaConf

Yes. The problem is there's no CLI parser that will solve everybody's needs.

> It seems to me overhead to add hydra for my repo, just for parsing CLI params

That makes sense. You should use your posix solution if it works in your codebase :)
You can also publish a package on pypi (e.g. `omegaconf-posix-cli`) if it will be useful for other projects!

PR #970, opened 2022-06-28
Fix CI: set Sphinx language = en

Jasha10:


Jasha10 (commented):
Revert typo from previous commit

Issue #950, opened 2022-05-26
`ListConfig.__setitem__(slice, ...)` does not roll-back cfg state upon mutation failure


Jasha10:
Most mutations to OmegaConf containers are atomic in the sense that a failed mutation will result in an unmodified object. This is not the case, however, when `ListConfig.__setitem__` is called on a slice.

**To Reproduce**
The last line of the script below demonstrates inconsistency (failure to roll-back `ListConfig` state on failed mutation).
```python
from pytest import raises

from omegaconf import OmegaConf, UnsupportedValueType

illegal = object()

cfg = OmegaConf.create([1, 2, 3])

with raises(UnsupportedValueType):
    cfg.append(illegal)
assert cfg == [1, 2, 3]

with raises(UnsupportedValueType):
    cfg.extend([illegal])
assert cfg == [1, 2, 3]

with raises(UnsupportedValueType):
    cfg.insert(1, illegal)
assert cfg == [1, 2, 3]

with raises(UnsupportedValueType):
    cfg[1] = illegal
assert cfg == [1, 2, 3]

with raises(UnsupportedValueType):
    cfg[1:2] = [illegal]
assert cfg == [1, 3]  # Inconsistent
```

**Expected behavior**
The last line above should be `assert cfg == [1, 2, 3]`

PR #975, opened 2022-07-14
ensure atomic updates for ListConfig sliced assignment (#950)

pixelb:
* omegaconf/listconfig.py (__setitem__): Operate on a copy
so any changes due to validation errors etc. are discarded.
Fixes issue #950

### DIFF
--- a/news/950.bugfix
+++ b/news/950.bugfix
@@ -1,0 +1,1 @@
+ListConfig sliced assignment now avoids partial updates upon error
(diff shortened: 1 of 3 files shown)


pixelb:
This works but is not very efficient.
It would be good to have a replace() method I think, that would avoid all the assignment checks etc.
Also ideally we would shallow copy the work list in this case.

pixelb:
I'll push this now, and we can optimize later

Jasha10 (approved):
Implementation looks good to me! I think this diff should include a news fragment.

Issue #838, opened 2021-12-06
String to float conversion edge case


rsokl:
Hello! [Hypothesis](https://hypothesis.readthedocs.io/en/latest/) found an odd test case during one of hydra-zen's nightly jobs.

The yaml field-value `'0_e0'` gets converted to a float:

```python
>>> instantiate(OmegaConf.create('x: 0_e0'))
{'x': 0.0}
```
however this string is not actually a valid representation of `0.0`:

```python
>>> float('0_e0')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-16-12374ef4f47c> in <module>
----> 1 float('0_e0')
```

I do not think that the underscore can precede `e` directly.

rsokl:
Actually, this might be more appropriate for Hydra, since it involves instantiation. If so, I will be happy to move it.

Jasha10:
I think that the culprit is [this regex](https://github.com/omry/omegaconf/blob/9f3d98c23a8f547f7f0f9b3c68243f663b8c635b/omegaconf/_utils.py#L144) Looking at OmegaConf's custom yaml loader :)

rsokl:
Yup! That sure looks like it!

pixelb:
tl;dr we should probably restrict underscore handling to less cases.
At least we should disallow trailing underscores (as is the case in this bug).

That would match most underscore allowing languages including python.
See https://peps.python.org/pep-0515/

I see yaml 1.2 doesn't even specify "_" as allowed in numbers, as the 1.1 spec was too loosely defined wrt "_"
See: https://github.com/nodeca/js-yaml/issues/627

I see our existing RE probably originates in ruamel.
See https://stackoverflow.com/q/30458977/4421 and commit a50bdeea

It's interesting that yaml being a superset of json, diverges from json here ?

pixelb:
I also see some inconsistencies between the antlr (interpolation) and basic loader.
pr #995 does not address that, but at least documents it.

PR #995, opened 2022-08-10
Restrict underscore formats which are converted to float

pixelb:
Disallow numbers with trailing underscore from being converted to float.

The specific change here is to treat 0_e0 as a string.

Fixes issue #838


Jasha10 (approved):
Looks good to me!

PR #1007, opened 2022-09-20
docs: make index TOC shallower, add deep TOCs to pages

Jasha10:
- The TOC on the OmegaConf landing page is now shallower
- The individual html docs pages now each include their own (deep) table of contents


pixelb (approved):
Yes this is quite a bit cleaner and better supports future additions. Thanks

PR #1009, opened 2022-09-22
docs: use :local: directive for page TOCs

Jasha10:
This prevents the title of each page from appearing in its own table of
contents.


Jasha10:
### Before:
image
### After:
image

PR #1013, opened 2022-09-29
Update to antlr 4.11

xhochy:
Update to the latest version (helps in aligning with other packages that also depend on ANTLR).


xhochy:
The failure in CI look unrelated to my changes.

pixelb:
antlr is generally not backwards compat, so we have to be careful with such updates, juggling various dependency constraints.
We will consider this. thanks

xhochy:
> antlr is generally not backwards compat

That was my main motivation behind this PR. I have several other dependency in my tree that have mismatching ANTLR versions. To get that solved, I thought that updating all of them to the latest is probably the easiest to solve it for a moment. 

Builds are green now as I have rebased on latest master.

Jasha10:
Hi @xhochy,

Thank you for contributing this PR.

I'm sorry to say that we can't merge this due due to Meta's internal requirements r.e. antlr. Please see my comment on [hydra#2491](https://github.com/facebookresearch/hydra/issues/2491) for an explanation as well as a suggested workaround.

omry:
We need to find a solution for this problem.
One option is to vendor a version of the antlr runtime with OmegaConf in a different package and use that version - eliminating the runtime dependency from both OmegaConf and Hydra. (Hydra can depend on the version from OmegaConf).

Issue #794, opened 2021-10-05
OmegaConf is vulenrable to ddos via specially crafted yaml input (billion laughs attack)


omry:
OmegaConf is vulnerable to [Billion laughs attack](https://en.wikipedia.org/wiki/Billion_laughs_attack).
Specially crafted yaml input files can use yaml anchors to create a yaml input that is fast to load but very slow to traverse because it contains billions of nodes.

This is considered medium priority and will be addressed in a future major version of OmegaConf.
Most potential fixes are breaking.

### Details
pyyaml is creating references to the same object when it loads an anchor, which is why it can do load such files quickly.

Example input:
```yaml
lol1: &lol1 "lol"
lol2: &lol2 [*lol1,*lol1,*lol1,*lol1,*lol1,*lol1,*lol1,*lol1,*lol1]
lol3: &lol3 [*lol2,*lol2,*lol2,*lol2,*lol2,*lol2,*lol2,*lol2,*lol2]
lol4: &lol4 [*lol3,*lol3,*lol3,*lol3,*lol3,*lol3,*lol3,*lol3,*lol3]
lol5: &lol5 [*lol4,*lol4,*lol4,*lol4,*lol4,*lol4,*lol4,*lol4,*lol4]
lol6: &lol6 [*lol5,*lol5,*lol5,*lol5,*lol5,*lol5,*lol5,*lol5,*lol5]
lol7: &lol7 [*lol6,*lol6,*lol6,*lol6,*lol6,*lol6,*lol6,*lol6,*lol6]
lol8: &lol8 [*lol7,*lol7,*lol7,*lol7,*lol7,*lol7,*lol7,*lol7,*lol7]
lol9: &lol9 [*lol8,*lol8,*lol8,*lol8,*lol8,*lol8,*lol8,*lol8,*lol8]
lol10: &lol10 [*lol9,*lol9,*lol9,*lol9,*lol9,*lol9,*lol9,*lol9,*lol9]
```

Such input creates billions of yaml nodes. OmegaConf will take an unreasonably long time to instantiate an object from it.

#### 1. Use interpolations instead of yaml anchors
Avoid yaml anchors and use interpolations instead.

The following config uses interpolations to achieve a similar behavior:

```yaml
lol1: "lol"
lol2: ["${lol1}","${lol1}","${lol1}","${lol1}","${lol1}","${lol1}","${lol1}","${lol1}","${lol1}"]
[...]

PR #1053, opened 2023-01-17
Raise error if there are conflicting keys after merge

RaulWCosta:
## What?

Change the default behavior of merge to raise an error when there is a conflict in the key names after merge.
Right now, if there is a key conflict after merge, the last key specified silently overwrites the other. As shown in the test case below.

https://github.com/omry/omegaconf/blob/master/tests/test_create.py#L416
```
def test_yaml_merge() -> None:
    cfg = OmegaConf.create(
        dedent(
            """\
            a: &A
                x: 1
            b: &B
                y: 2
            c:
                <<: *A
                <<: *B
                x: 3
                z: 1
            """
        )
    )
    assert cfg == {"a": {"x": 1}, "b": {"y": 2}, "c": {"x": 3, "y": 2, "z": 1}}
```

## Why?

I think there are two main points to justify this change:
1. Duplicated keys are syntactically invalid for YAML files, as shown in the quote below (from https://yaml.org/spec/1.2-old/spec.html#id2764044)
> The content of a mapping node is an unordered set of key: value node pairs, with the restriction that each of the keys is unique.

2. This behavior can cause unexpected problems in running time for OmegaConf's users.

---

I executed the tests in a Windows 10 machine, using Python 3.10.8 running in a Conda env.

I hope this PR is useful :)

### DIFF
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -424,30 +424,26 @@ def test_yaml_merge() -> None:
             c:
                 <<: *A
                 <<: *B
-                x: 3
                 z: 1
             """
         )
     )
-    assert cfg == {"a": {"x": 1}, "b": {"y": 2}, "c": {"x": 3, "y": 2, "z": 1}}
+    assert cfg == {"a": {"x": 1}, "b": {"y": 2}, "c": {"x": 1, "y": 2, "z": 1}}
+
 
 def test_yaml_merge_with_conflict() -> None:
-    import pdb; pdb.set_trace()
-    cfg = OmegaConf.create(
-        dedent(
-            """\
-            a: &A
-                x: 1
-            b: &B
-                y: 2
-            c:
-                <<: *A
-                <<: *B
-                x: 3
-                z: 1
-            """
+    with raises(yaml.constructor.ConstructorError):
+        OmegaConf.create(
+            dedent(
+                """\
+                a: &A
+                    x: 1
+                c:
+                    <<: *A
+                    x: 3
+                """
+            )
         )
-    )
 
 @mark.parametrize(
     "path_type",
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -429,7 +429,7 @@ def test_yaml_merge() -> None:
             """
         )
     )
-    assert cfg == {"a": {"x": 1}, "b": {"y": 2}, "c": {"x": 1, "y": 2, "w": 3, "z": 1}}
+    assert cfg == {"a": {"x": 1}, "b": {"y": 2}, "c": {"w": 3, "y": 2, "z": 1}}
 
 
 def test_yaml_merge_with_conflict() -> None:
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -431,6 +431,23 @@ def test_yaml_merge() -> None:
     )
     assert cfg == {"a": {"x": 1}, "b": {"y": 2}, "c": {"x": 3, "y": 2, "z": 1}}
 
+def test_yaml_merge_with_conflict() -> None:
+    import pdb; pdb.set_trace()
+    cfg = OmegaConf.create(
+        dedent(
+            """\
+            a: &A
+                x: 1
+            b: &B
+                y: 2
+            c:
+                <<: *A
+                <<: *B
+                x: 3
+                z: 1
+            """
+        )
+    )
 
 @mark.parametrize(
     "path_type",
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -446,6 +446,7 @@ def test_yaml_merge_with_conflict() -> None:
             )
         )
 
+
 @mark.parametrize(
     "path_type",
     [
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -429,7 +429,7 @@ def test_yaml_merge() -> None:
             """
         )
     )
-    assert cfg == {"a": {"x": 1}, "b": {"y": 2}, "c": {"w": 3, "y": 2, "z": 1}}
+    assert cfg == {"a": {"x": 1}, "b": {"y": 2}, "c": {"x": 1, "y": 2, "w": 3, "z": 1}}
 
 
 def test_yaml_merge_with_conflict() -> None:
--- a/tests/test_create.py
+++ b/tests/test_create.py
@@ -424,11 +424,12 @@ def test_yaml_merge() -> None:
             c:
                 <<: *A
                 <<: *B
+                w: 3
                 z: 1
             """
         )
     )
-    assert cfg == {"a": {"x": 1}, "b": {"y": 2}, "c": {"x": 1, "y": 2, "z": 1}}
+    assert cfg == {"a": {"x": 1}, "b": {"y": 2}, "c": {"x": 1, "y": 2, "w": 3, "z": 1}}
 
 
 def test_yaml_merge_with_conflict() -> None:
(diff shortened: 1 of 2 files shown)


omry:
Can you reproduce the problem without using yaml anchors?

RaulWCosta:
> Can you reproduce the problem without using yaml anchors?

Without anchors, i.e. by loading a file with direct conflict in the keys (not 100% that was your question), it raises an error as expected. The behavior in this PR also raises an error for that case.

Do you think would be good to add a test case for that scenario?

tests/test_create.py
```diff
@@ -424,14 +424,28 @@ def test_yaml_merge() -> None:
...
                 <<: *A
                 <<: *B
-                x: 3
+                w: 3
                 z: 1
             """
         )
     )
-    assert cfg == {"a": {"x": 1}, "b": {"y": 2}, "c": {"x": 3, "y": 2, "z": 1}}
+    assert cfg == {"a": {"x": 1}, "b": {"y": 2}, "c": {"x": 1, "y": 2, "w": 3, "z": 1}}
```
omry, line 432: What is the reason for this change in behavior?

omry:
Generally speaking, anchors sucks, and you should use interpolations instead.
We are considering disabling anchors support by default to address #794.

tests/test_create.py
RaulWCosta: It was a typo, just fixed it

RaulWCosta: The previous test would raise an error considering the new behavior of anchor merge

RaulWCosta:
> Generally speaking, anchors sucks, and you should use interpolations instead. We are considering disabling anchors support by default to address #794.

Got it! I haven't tested that scenario with interpolation, but I can do it today :)

RaulWCosta:
> Generally speaking, anchors sucks, and you should use interpolations instead. We are considering disabling anchors support by default to address #794.

I didn't manage to make the example you added in #794 run with `OmegaConf.load` (maybe there is some aditional step that I'm not aware of?)

```
# test.yml
lol1: "lol"
lol2: ["${lol1}","${lol1}","${lol1}","${lol1}","${lol1}","${lol1}","${lol1}","${lol1}","${lol1}"]
```

Output I got:
```
{'lol1': 'lol', 'lol2': ['${lol1}', '${lol1}', '${lol1}', '${lol1}', '${lol1}', '${lol1}', '${lol1}', '${lol1}', '${lol1}']}
```

Besides that, it looks like this topic is an ongoing discussion in the PR. So maybe, while this is not defined, would be useful to fix this behavior in merge.

tests/test_create.py
omry: If there previous test would not raise an error, why is the fix for it to change the expected return value?

RaulWCosta: With the behavior proposed by this PR the previous test raises an error.

This PR intend to raise an error in the case there is conflict in the keys after resolving the anchors. In this test (before the change proposed) inside `c: {}` we have a conflict in the `x` key.

Jasha10:
> I didn't manage to make the example you added in #794 run with `OmegaConf.load` (maybe there is some aditional step that I'm not aware of?)

@RaulWCosta see the docs on [OmegaConf interpolations](https://omegaconf.readthedocs.io/en/2.3_branch/usage.html#variable-interpolation).

```python

yaml_data = """
# test.yml
lol1: "lol"
lol2: ["${lol1}","${lol1}","${lol1}","${lol1}","${lol1}","${lol1}","${lol1}","${lol1}","${lol1}"]
"""

config = OmegaConf.create(yaml_data)

assert config.lol2[0] == "lol"  # lazy dereferencing

# Try printing `config` before and after calling OmegaConf.resolve:

print(config)
# prints {'lol1': 'lol', 'lol2': ['${lol1}', '${lol1}', '${lol1}', '${lol1}', '${lol1}', '${lol1}', '${lol1}', '${lol1}', '${lol1}']}

# Calling resolve would likely fill up your ram if you have lol1,lol2,...lol10 defined
OmegaConf.resolve(config)

print(config)
# prints {'lol1': 'lol', 'lol2': ['lol', 'lol', 'lol', 'lol', 'lol', 'lol', 'lol', 'lol', 'lol']}
```

tests/test_create.py
omry: I am confused because your fix for the raised error is to change the expected output instead of expecting the error.

omry:
@Jasha10, I am actually not sure interpolation can be used here: OmegaConf does not provide a parallel to YAML merge (e.g. `<<: *A`).

tests/test_create.py
RaulWCosta: I changed the expected output to still have an "it works" case for the anchor resolution (I created another test case to check the raising of the error). Do you think that makes sense? Otherwise, I can surely modify it to expect an error.

RaulWCosta:
> > I didn't manage to make the example you added in #794 run with `OmegaConf.load` (maybe there is some aditional step that I'm not aware of?)
> 
> @RaulWCosta see the docs on [OmegaConf interpolations](https://omegaconf.readthedocs.io/en/2.3_branch/usage.html#variable-interpolation).
> 
> ```python
> 
> yaml_data = """
> # test.yml
> lol1: "lol"
> lol2: ["${lol1}","${lol1}","${lol1}","${lol1}","${lol1}","${lol1}","${lol1}","${lol1}","${lol1}"]
> """
> 
> config = OmegaConf.create(yaml_data)
> 
> assert config.lol2[0] == "lol"  # lazily access interpolated value ${lol1} -> "lol"
> 
> # Try printing `config` before and after calling OmegaConf.resolve:
> 
> print(config)
> # prints {'lol1': 'lol', 'lol2': ['${lol1}', '${lol1}', '${lol1}', '${lol1}', '${lol1}', '${lol1}', '${lol1}', '${lol1}', '${lol1}']}
> 
> # Calling resolve would likely fill up your ram if you have lol1,lol2,...lol10 defined
> OmegaConf.resolve(config)
> 
> print(config)
> # prints {'lol1': 'lol', 'lol2': ['lol', 'lol', 'lol', 'lol', 'lol', 'lol', 'lol', 'lol', 'lol']}
> ```

Appreciate!

RaulWCosta:
Hey @omry really appreciate your time! Do you think it would be best if I close this PR?

omry:
I would still like to get this reviewed.
I am not currently working on OmegaConf but there might be some new maintainers coming soon.

omry:
Thanks for the PR and the discussion here.

I am closing this because the duplicate-key/anchor-merge handling was superseded by #1245, which introduced be34d69c35f1.

The current behavior makes an important distinction:

- true duplicate explicit YAML keys are still rejected while parsing
- YAML merge keys keep YAML merge semantics, where explicit local keys override keys coming from merged mappings

For example, this remains valid and resolves to `c.x == 3`:

```yaml
base: &base
  x: 1
c:
  <<: *base
  x: 3
```

That behavior is preferred because OmegaConf ultimately represents mappings with `DictConfig`/mapping semantics and cannot represent the same key twice. The right place to reject duplicates is when the YAML document has duplicate explicit keys. Merge-key conflicts, on the other hand, are resolved during YAML merge flattening by allowing explicit keys to override merged keys, which matches useful YAML merge behavior and avoids rejecting valid merge patterns.

So #1053 would make OmegaConf stricter than the behavior we want here, while #1245 keeps duplicate explicit key protection and preserves YAML merge override semantics.

PR #1082, opened 2023-05-24
feat: add deep merging for lists

### LINKED ISSUE #1080: Support deep merging of configs
I don't know if it's a bug or the expected behavior in the project, but I have the following example:

I want to merge the following configurations:
```yaml
dict: 
  list:
    - entry_a1: val_a1
      entry_a2: val_a2
```
```yaml
dict: 
  list:
    - entry_b1: val_b1
      entry_b2: val_b2
```

If I merge them, I get: ( see full script and steps to reproduce below
```yaml
dict: 
  list:
    - entry_b1: val_b1
      entry_b2: val_b2
```

However I would expect, that the configs get deep merged, so that I get
```yaml
dict: 
  list:
    - entry_a1: val_a1
      entry_a2: val_a2
    - entry_b1: val_b1
      entry_b2: val_b2
```
as the merged result. The order of the elements doesn't matter.

Is that already possible and I'm missing something or is this functionality not supported? 

The main thing with this feature is, that we have to check, if two types are __mergable__, so that not something like
```yaml
dict: 
  string: val
```
and
```yaml
dict: 
  list:
      - entry1
```
get merged, because the types `string` and `list` would be incompatible.

**Additional context**
<details>
<summary>Here is the full python script for you to reproduce the examples:</summary>

```python

source_a = """
dict: 
    list:
    - entry_a1: val_a1
      entry_a2: val_a2
"""

source_b = """
dict: 
    list:
    - entry_b1: val_b1
      entry_b2: val_b2
"""

cfg_a = OmegaConf.create(source_a)
cfg_b = OmegaConf.create(source_b)

cfg_merged = OmegaConf.merge(cfg_a, cfg_b)

print(OmegaConf.to_yaml(cfg_merged))
```

</details>

MatteoVoges:
Fixes #1080 

## Proposed changes
* new flags in the `OmegaConf.merge()` function
  * `extend_lists`: this enables (deep) merge support for `ListConfigs` (default: False to ensure backwards compability)
  * `allow_duplicates`:  this controls, if two equal items can be in a list (default: False)

## Examples
If you have a config:
```yaml
list:
   - a
```
and 
```yaml
list:
  - b
```
you will get with the enabled flag `extend_lists` this result:
```yaml
list:
  - a
  - b
```
The functionality of `allow_duplicates` is self explanatory, i guess.

## Tests and Documentation
* Added some basic tests and tests for edge cases (nested objects)
* @omry Where should I add the documentation for this? In `docs/source/usage.rst` or in `docs/notebook/Tutorial.ipynb` or somewhere else?

--- 
### Contributed by  [](https://www.nexenio.com)

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -497,7 +497,7 @@ Use ``list_merge_mode`` to control the merge behavior for lists.
 This Enum is defined in ``omegaconf.ListMergeMode`` and defines the following modes:
 * ``REPLACE``: Replaces the target list with the new one (default)
 * ``EXTEND``: Extends the target list with the new one
-* ``APPEND_UNIQUE_VALUE``: Appends items that do not already exist in the target list
+* ``EXTEND_UNIQUE``: Extends the target list items with items not present in it
 
 **example2.yaml** file:
 
@@ -509,7 +509,7 @@ This Enum is defined in ``omegaconf.ListMergeMode`` and defines the following mo
 .. include:: example4.yaml
    :code: yaml
 
-If you load them and merge them with ``list_merge_mode=ListMergeMode.APPEND_UNIQUE_VALUE`` you will get this:
+If you load them and merge them with ``list_merge_mode=ListMergeMode.EXTEND_UNIQUE`` you will get this:
 
 .. doctest::
 
@@ -518,7 +518,7 @@ If you load them and merge them with ``list_merge_mode=ListMergeMode.APPEND_UNIQ
     >>> cfg_1 = OmegaConf.load('source/example2.yaml')
     >>> cfg_2 = OmegaConf.load('source/example4.yaml')
     >>> 
-    >>> mode = ListMergeMode.APPEND_UNIQUE_VALUE
+    >>> mode = ListMergeMode.EXTEND_UNIQUE
     >>> conf = OmegaConf.merge(cfg_1, cfg_2, list_merge_mode=mode)
     >>> print(OmegaConf.to_yaml(conf))
     server:
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -493,7 +493,8 @@ Note how the port changes to 82, and how the users lists are combined.
     <BLANKLINE>
 
 By default, ``merge()`` is replacing the target list with the source list.
-Use ``list_merge_mode`` to control the merge behavior for lists.
+Use ``list_merge_mode`` to control the merge behavior for lists. 
+You can access the regarding enum with ``from omegaconf import ListMergeMode``.
 Currently there are three different merge modes:
 * ``OVERRIDE``: content from newer list gets taken (default)
 * ``EXTEND``: lists get extended
@@ -509,16 +510,16 @@ Currently there are three different merge modes:
 .. include:: example4.yaml
    :code: yaml
 
-If you load them and merge them with ``list_merge_mode="EXTEND_IGNORE_DUPLICATES"`` you will get this:
+If you load them and merge them with ``list_merge_mode=ListMergeMode.EXTEND_IGNORE_DUPLICATES`` you will get this:
 
 .. doctest::
 
-    >>> from omegaconf import OmegaConf
+    >>> from omegaconf import OmegaConf, ListMergeMode
     >>>
     >>> cfg_example2 = OmegaConf.load('source/example2.yaml')
     >>> cfg_example4 = OmegaConf.load('source/example4.yaml')
     >>> 
-    >>> conf = OmegaConf.merge(cfg_example2, cfg_example4, list_merge_mode="EXTEND_IGNORE_DUPLICATES")
+    >>> conf = OmegaConf.merge(cfg_example2, cfg_example4, list_merge_mode=ListMergeMode.EXTEND_IGNORE_DUPLICATES)
     >>> print(OmegaConf.to_yaml(conf))
     server:
       port: 80
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -492,6 +492,41 @@ Note how the port changes to 82, and how the users lists are combined.
       file: log.txt
     <BLANKLINE>
 
+The default behavior for lists is that the list from the last config overwrites the others. 
+Use ``OmegaConf.merge(cfg1, cfg2, extend_lists=True)`` to merge the lists by extending them. 
+You can specify whether duplicate entries are allowed with the flag ``allow_duplicates``. 
+This is false, by default, but is only compatible with ``extend_lists`` activated.
+
+**example2.yaml** file:
+
+.. include:: example2.yaml
+   :code: yaml
+
+**example4.yaml** file:
+
+.. include:: example4.yaml
+   :code: yaml
+
+If you load them and merge them with both attributes set to ``True`` you will get this:
+
+.. doctest::
+
+    >>> from omegaconf import OmegaConf
+    >>>
+    >>> cfg_example2 = OmegaConf.load('source/example2.yaml')
+    >>> cfg_example4 = OmegaConf.load('source/example4.yaml')
+    >>> 
+    >>> conf = OmegaConf.merge(cfg_example2, cfg_example4, extend_lists=True, allow_duplicates=True)
+    >>> print(OmegaConf.to_yaml(conf))
+    server:
+      port: 80
+    users:
+    - user1
+    - user2
+    - user3
+    - user2
+    <BLANKLINE>
+
 OmegaConf.unsafe_merge()
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -492,10 +492,12 @@ Note how the port changes to 82, and how the users lists are combined.
       file: log.txt
     <BLANKLINE>
 
-By default, merge is replacing the target list with the source list.
-Use ``OmegaConf.merge(cfg1, cfg2, extend_lists=True)`` to merge the lists by extending them. 
-You can specify whether duplicate entries should be removed or not with the flag ``remove_duplicates``. 
-This is false, by default, and is ignored unless ``extend_lists=True`` .
+By default, ``merge()`` is replacing the target list with the source list.
+Use ``list_merge_mode`` to control the merge behavior for lists.
+Currently there are three different merge modes:
+* ``OVERRIDE``: content from newer list gets taken (default)
+* ``EXTEND``: lists get extended
+* ``EXTEND_IGNORE_DUPLICATES``: only new (unique) elements get extended
 
 **example2.yaml** file:
 
@@ -507,7 +509,7 @@ This is false, by default, and is ignored unless ``extend_lists=True`` .
 .. include:: example4.yaml
    :code: yaml
 
-If you load them and merge them with both attributes set to ``True`` you will get this:
+If you load them and merge them with ``list_merge_mode="EXTEND_IGNORE_DUPLICATES"`` you will get this:
 
 .. doctest::
 
@@ -516,7 +518,7 @@ If you load them and merge them with both attributes set to ``True`` you will ge
     >>> cfg_example2 = OmegaConf.load('source/example2.yaml')
     >>> cfg_example4 = OmegaConf.load('source/example4.yaml')
     >>> 
-    >>> conf = OmegaConf.merge(cfg_example2, cfg_example4, extend_lists=True, remove_duplicates=True)
+    >>> conf = OmegaConf.merge(cfg_example2, cfg_example4, list_merge_mode="EXTEND_IGNORE_DUPLICATES")
     >>> print(OmegaConf.to_yaml(conf))
     server:
       port: 80
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -494,7 +494,7 @@ Note how the port changes to 82, and how the users lists are combined.
 
 By default, merge is replacing the target list with the source list.
 Use ``OmegaConf.merge(cfg1, cfg2, extend_lists=True)`` to merge the lists by extending them. 
-You can specify whether duplicate entries are allowed with the flag ``allow_duplicates``. 
+You can specify whether duplicate entries should be removed or not with the flag ``remove_duplicates``. 
 This is false, by default, and is ignored unless ``extend_lists=True`` .
 
 **example2.yaml** file:
@@ -516,7 +516,7 @@ If you load them and merge them with both attributes set to ``True`` you will ge
     >>> cfg_example2 = OmegaConf.load('source/example2.yaml')
     >>> cfg_example4 = OmegaConf.load('source/example4.yaml')
     >>> 
-    >>> conf = OmegaConf.merge(cfg_example2, cfg_example4, extend_lists=True, allow_duplicates=True)
+    >>> conf = OmegaConf.merge(cfg_example2, cfg_example4, extend_lists=True, remove_duplicates=True)
     >>> print(OmegaConf.to_yaml(conf))
     server:
       port: 80
@@ -524,7 +524,6 @@ If you load them and merge them with both attributes set to ``True`` you will ge
     - user1
     - user2
     - user3
-    - user2
     <BLANKLINE>
 
 OmegaConf.unsafe_merge()
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -493,9 +493,8 @@ Note how the port changes to 82, and how the users lists are combined.
     <BLANKLINE>
 
 By default, ``merge()`` is replacing the target list with the source list.
-Use ``list_merge_mode`` to control the merge behavior for lists. 
-You can access the regarding enum with ``from omegaconf import ListMergeMode``.
-Currently there are three different merge modes:
+Use ``list_merge_mode`` to control the merge behavior for lists.
+This Enum is defined in ``omegaconf.ListMergeMode`` and defines the following modes:
 * ``OVERRIDE``: content from newer list gets taken (default)
 * ``EXTEND``: lists get extended
 * ``EXTEND_IGNORE_DUPLICATES``: only new (unique) elements get extended
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -492,10 +492,10 @@ Note how the port changes to 82, and how the users lists are combined.
       file: log.txt
     <BLANKLINE>
 
-The default behavior for lists is that the list from the last config overwrites the others. 
+By default, merge is replacing the target list with the source list.
 Use ``OmegaConf.merge(cfg1, cfg2, extend_lists=True)`` to merge the lists by extending them. 
 You can specify whether duplicate entries are allowed with the flag ``allow_duplicates``. 
-This is false, by default, but is only compatible with ``extend_lists`` activated.
+This is false, by default, and is ignored unless ``extend_lists=True`` .
 
 **example2.yaml** file:
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -495,9 +495,9 @@ Note how the port changes to 82, and how the users lists are combined.
 By default, ``merge()`` is replacing the target list with the source list.
 Use ``list_merge_mode`` to control the merge behavior for lists.
 This Enum is defined in ``omegaconf.ListMergeMode`` and defines the following modes:
-* ``OVERRIDE``: content from newer list gets taken (default)
-* ``EXTEND``: lists get extended
-* ``EXTEND_IGNORE_DUPLICATES``: only new (unique) elements get extended
+* ``REPLACE``: Replaces the target list with the new one (default)
+* ``EXTEND``: Extends the target list with the new one
+* ``APPEND_UNIQUE_VALUE``: Appends items that do not already exist in the target list
 
 **example2.yaml** file:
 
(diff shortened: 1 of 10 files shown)


MatteoVoges:
The tests show the following error in the `lint-3.10` job:
```
tests/test_basic_ops_dict.py:276: error: Keywords must be strings  [misc]
Installing missing stub packages:
/home/circleci/project/.nox/lint-3-10/bin/python -m pip install types-Pygments types-colorama types-setuptools


Found 1 error in 1 file (checked 81 source files)
```
I didn't edit this file, so I guess that this error is due some wrong configuration or setup in CircleCI. Is there anything I have to look into?

omry:
Thanks @MatteoVoges.

@Jasha10, @odelalleau, @shagunsodhani:
What do you think about this functionality? We can discuss here or on #1080.

At a glance it looks pretty nice, but I didn't review properly.
1. Add a news fragment (similar to what is described here https://hydra.cc/docs/development/documentation/).
2. Update the docs showing usage of OmegaConf merge. Also update the notebook with an example.
3. Try to fix the lint error even if you didn't create it (If it's involved it can be in a separate PR).

MatteoVoges:
@omry I updated the docs and added some examples. The linting error is fixed in #1084 . Is there anything left to do?

docs/source/usage.rst
```diff
@@ -492,6 +492,41 @@ Note how the port changes to 82, and how the users lists are combined.
       file: log.txt
     <BLANKLINE>
 
+The default behavior for lists is that the list from the last config overwrites the others. 
+Use ``OmegaConf.merge(cfg1, cfg2, extend_lists=True)`` to merge the lists by extending them. 
+You can specify whether duplicate entries are allowed with the flag ``allow_duplicates``. 
+This is false, by default, but is only compatible with ``extend_lists`` activated.
```
shagunsodhani, line 498:
```suggestion
This is false, by default, and is ignored unless ``extend_lists=True`` .
```

news/1082.feature
```diff
@@ -0,0 +1 @@
+This change enables deep merging for ListConfigs. Usage: `OmegaConf.merge(cfg1, cfg2, extend_lists=True)`
```
shagunsodhani, line 1:
```suggestion
This change enables deep merging (and optionally retaining duplicate values) for ListConfigs. Usage: `OmegaConf.merge(cfg1, cfg2, extend_lists=True)`
```

omegaconf/omegaconf.py
```diff
@@ -257,11 +257,15 @@ def merge(
...
         ],
+        extend_lists: bool = False,
+        allow_duplicates: bool = False,
     ) -> Union[ListConfig, DictConfig]:
         """
         Merge a list of previously created configs into a single one
 
         :param configs: Input configs
+        :param extend_lists: flag to deep merge ListConfigs
+        :param allow_duplicates: lists entries are unique
```
shagunsodhani, line 268:
```suggestion
        :param allow_duplicates: lists entries are unique. This flag is ignored unless `extend_lists=True`
```

shagunsodhani (commented):
This mostly looks good to me! @omry is there a specific piece that I should look deeper into ?

tests/test_merge.py
```diff
@@ -1190,6 +1190,50 @@ def test_merge_list_list() -> None:
...
+        # don't allow duplicate entries
+        ([1, 2], [1, 2], [1, 2]),
+        ([{"a": 1}], [{"a": 1}], [{"a": 1}]),
+    ],
+)
+def test_merge_nested_list_extend(c1: Any, c2: Any, expected: Any) -> None:
+    a = OmegaConf.create(c1)
+    b = OmegaConf.create(c2)
+    a.merge_with(b, extend_lists=True)
+    assert a == expected
```
Jasha10, line 1220:
```suggestion
    merged = OmegaConf.merge(a, b, extend_lists=True)
    assert merged == expected
```

tests/test_merge.py
```diff
@@ -1190,6 +1190,50 @@ def test_merge_list_list() -> None:
...
+    [
+        ([1, 2], [1, 2], [1, 2, 1, 2]),
+        ([{"a": 1}], [{"a": 1}], [{"a": 1}, {"a": 1}]),
+    ],
+)
+def test_merge_list_extend_allow_duplicates(c1: Any, c2: Any, expected: Any) -> None:
+    a = OmegaConf.create(c1)
+    b = OmegaConf.create(c2)
+    a.merge_with(b, extend_lists=True, allow_duplicates=True)
+    assert a == expected
```
Jasha10, line 1234:
```suggestion
    merged = OmegaConf.merge(a, b, extend_lists=True, allow_duplicates=True)
    assert merge == expected
```

Jasha10 (changes requested):
I'm happy with the proposed feature, and the implementation looks good to me.

One nit regarding the tests: the `BaseContainer.merge_with` method is not documented and is not really intended as a public API. It would be better to test the public-facing `OmegaConf.merge` facade instead of calling `BaseContainer.merge_with` in the tests:

docs/source/usage.rst
```diff
@@ -492,6 +492,41 @@ Note how the port changes to 82, and how the users lists are combined.
       file: log.txt
     <BLANKLINE>
 
+The default behavior for lists is that the list from the last config overwrites the others.
```
omry, line 495:
```suggestion
By default, merge is replacing the target list with the source list.
```

news/1082.feature
omry, line 1:
```suggestion
OmegaConf.merge() can now concat lists instread of replacing the target list
```

omegaconf/basecontainer.py
```diff
@@ -304,7 +304,12 @@ def get_node_value(key: Union[DictKeyType, int]) -> Any:
         assert False
 
     @staticmethod
-    def _map_merge(dest: "BaseContainer", src: "BaseContainer") -> None:
+    def _map_merge(
+        dest: "BaseContainer",
+        src: "BaseContainer",
+        extend_lists: bool = False,
+        allow_duplicates: bool = False,
```
omry, line 311: I think deduping should not happen by default.
```python
remove_duplicates: bool = False
```
Docs and examples should be updated of course.

omry (changes requested):

news/1082.feature
MatteoVoges: Updated in the latest commit

omegaconf/basecontainer.py
MatteoVoges: I'm ok with it and I updated the flag, behavior, docs, tests and examples

news/1082.feature
MatteoVoges: I decided to choose a shorter sentence and took omry's suggestion. I think for the news summary, it makes sense to announce the feature, but keep it short. Users, that have interest in this feature, read the docs anyway

Jasha10:
The `remove_duplicates` flag only makes sense if `extend_lists` is `True`, right?
```python
>>> OmegaConf.merge([1,2], [3,4], remove_duplicates=True)  # remove_duplicates has no effect here?
[3, 4]
```

Should we issue an error or warning in the case where `remove_duplicates` is passed and `extend_lists` is `False`?
Or should we adopt a similar `Enum`-based approach as is used for [the `structured_config_mode` argument to `OmegaConf.to_container`](https://github.com/omry/omegaconf/blob/85b3892a7e1b934e0df830f4520fbd9b9a14ebd9/omegaconf/omegaconf.py#L549)?

Something like this:
```python
class OmegaConf:
    ...
    def merge(
        ...,
        list_merge_mode: ListMergeMode = ListMergeMode.OVERWRITE,
    )

...

class ListMergeMode(Enum):
    """How `OmegaConf.merge` handles lists
    OVERWRITE = 1  # overwrite left-hand list with right-hand list
    EXTEND = 2  # extend left-hand list with right-hand list
    EXTEND_NO_DUPLICATES = 3  # ...
```

Jasha10:
Also, I think the naming of `remove_duplicates=True` is a little bit confusing.
The duplicates from `src` are removed but the duplicates from `dest` are not removed.
```python
>>> list1 = [1, 2, 3, 3]
>>> list2 = [3, 4, 5]
>>> OmegaConf.merge(list1, list2, extend_lists=True)
[1, 2, 3, 3, 3, 4, 5]
>>> OmegaConf.merge(list1, list2, extend_lists=True, remove_duplicates=True)  # duplicates from list1 not removed
[1, 2, 3, 3, 4, 5]
```

MatteoVoges:
I thought about this before as well and my first implementation to make sure, that there are no duplicate value was indeed using a set. But I decided to go for the currently implemented approach, because the set was scrambling the order of the lists and it was hard to test. 

> The duplicates from the right-hand merge argument are removed but the duplicates from the left-hand merge argument are not removed.

I actually like the current behavior a lot, because I think if you merge something the original object should not be affected, but only extended. And if you have some duplicate values in your `list1`, then it is for a reason, and they should appear in the merged object.

> Should we issue an error or warning in the case where `remove_duplicates` is passed and `extend_lists` is False?

I think a warning would be good here.

> Or should we adopt a similar `Enum`-based approach?

I don't think that this is neccessary, because we don't have many merge options yet and I think two merge modes do not require an enum.

Jasha10 (approved):
> I actually like the current behavior a lot

Fair enough. I still think the name of the `remove_duplicates` flag is potentially misleading, but I can't think of a better name for it.

> I don't think that [using an Enum] is necessary, because we don't have many merge options yet and I think two merge modes do not require an enum.

There are three list merge modes, not two:
- no list extension - overwrite the left with the right
- extend lists normally
- extend lists with no duplicates from the right

Let me lay out the pros and cons that I see for using an Enum:

Pros:
- Using an Enum would make it easier to add more merge modes in the future without overly complicating the signature of `OmegaConf.merge`. Meanwhile, switching to an enum *later* would be more difficult because it would break downstream consumers of the API.
- With an enum, there's no need to issue a warning for the case where `extend_lists is False and remove_duplicates is True`. Clients won't have to worry about calling `OmegaConf.merge` incorrectly.

Cons:
- Using an Enum might make `OmegaConf.merge` harder to call; clients would need to import the `omegaconf.ListMergeMode` class to select a non-default list merge option. We could potentially make things easier for clients by allowing them to pass a `typing.Literal` string instead of a `ListMergeMode` variant:
```python
class OmegaConf:
    ...
    def merge(
        ...,
        list_merge_mode: ListMergeMode | Literal["OVERWRITE", "EXTEND", "EXTEND_NO_DUPLICATES"] = "OVERWRITE",
    )
```
Clients could then call `OmegaConf.merge` as follows:
```python
cfg1 = OmegaConf.merge(a, b)  # default
cfg1 = OmegaConf.merge(a, b, list_merge_mode="EXTEND")
cfg2 = OmegaConf.merge(a, b, list_merge_mode="EXTEND_NO_DUPLICATES")
cfg2 = OmegaConf.merge(a, b, list_merge_mode=ListMergeMode.EXTEND_NO_DUPLICATES)  # alternate syntax
```

Anyway, I don't feel super strongly about this, and I might be bikeshedding here, so I'll give an approval and see if anyone else has a strong opinion on this matter.

odelalleau:
FWIW I do agree an Enum would be cleaner, for the reasons @Jasha10 mentioned. I think if you rename `remove_duplicates` into `ignore_duplicates` it might be less confusing: "remove" suggests post-processing the result (implying duplicates from both sides are deleted), while "ignore" suggests pre-processing (skip over items already in the list).

omegaconf/basecontainer.py
```diff
@@ -471,7 +482,14 @@ def _list_merge(dest: Any, src: Any) -> None:
                 for item in src._iter_ex(resolve=False):
                     temp_target.append(item)
 
-            dest.__dict__["_content"] = temp_target.__dict__["_content"]
+            if list_merge_mode == "EXTEND":
+                dest.__dict__["_content"].extend(temp_target.__dict__["_content"])
+            elif list_merge_mode == "EXTEND_IGNORE_DUPLICATES":
```
omry, line 487: An Enum would also allow you to document this in a discoverable place.

omry (changes requested):
Overall looks good. Please change to Enum.

omry:
Also, can you fix the lint issues? (If you already did, please rebase on top of your fix).

MatteoVoges:
> Overall looks good. Please change to Enum.

We then have the problem, that you have to import the enum, if you want to use this feature. So maybe there is a little hint in the merge-function-docstring. Or we enable both options, so Enum and String like @Jasha10 proposed?

Where would be a good place for the enum? `_utils` module?

omegaconf/basecontainer.py
MatteoVoges: Could you elaborate on that? Should I add another code comment?

omry:
> We then have the problem, that you have to import the enum, if you want to use this feature. So maybe there is a little hint in the merge-function-docstring. Or we enable both options, so Enum and String like @Jasha10 proposed?

I think using a string an an alternative is a bit unusual and I am not a fan of this duality.

> Where would be a good place for the enum? `_utils` module?

Given that the enum  is  becomeing a part of the public API, it should live in omegaconf/OmegaConf.py

```python
from omegaconf import OmegaConf, ListMergeMode

...
OmegaConf.merge(cfg1, cfg2, merge_mode = ListMergeMode.EXTEND_IGNORE_DUPLICATES)
```

A docstring update to OmegaConf.merge is fine. Don't forget to update docs to reflect this.

Jasha10:
> Given that the enum is becomeing a part of the public API, it should live in omegaconf/OmegaConf.py

Note that we have an [`SCMode`](https://github.com/omry/omegaconf/blob/85b3892a7e1b934e0df830f4520fbd9b9a14ebd9/omegaconf/base.py#L792) enum in `omegaconf/base.py` ([exported](https://github.com/omry/omegaconf/blob/85b3892a7e1b934e0df830f4520fbd9b9a14ebd9/omegaconf/__init__.py#L1) by `omegaconf/__init__.py`).

omegaconf/omegaconf.py
```diff
@@ -53,7 +53,7 @@
     split_key,
     type_str,
 )
-from .base import Box, Container, Node, SCMode, UnionNode
+from .base import Box, Container, ListMergeMode, Node, SCMode, UnionNode
```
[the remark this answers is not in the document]
Jasha10: I've dismissed the alert.

odelalleau:
I can’t check right now but I believe there’s a way to make an enum that can be compared to a string so that just passing the string would work too.

Jasha10:
> I believe there’s a way to make an enum that can be compared to a string so that just passing the string would work too.

I think you're referring to [`enum.StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum):
```python
from enum import StrEnum
class ListMergeModeString(StrEnum):
    OVERWRITE = "OVERWRITE"
    EXTEND = "EXTEND"
    EXTEND_NO_DUPLICATES = "EXTEND_NO_DUPLICATES"

assert ListMergeModeString.OVERWRITE == "OVERWRITE"
```
One issue with `StrEnum` is that it's only available in python 3.11+.

MatteoVoges:
> @omry note that the [SCMode](https://github.com/omry/omegaconf/blob/85b3892a7e1b934e0df830f4520fbd9b9a14ebd9/omegaconf/base.py#L792) enum currently lives in omegaconf/base.py (and is [exported](https://github.com/omry/omegaconf/blob/85b3892a7e1b934e0df830f4520fbd9b9a14ebd9/omegaconf/__init__.py#L1) by omegaconf/__init__.py).

I even followed the SCMode and the ListMergeMode is implemented in the same way.

> One issue with StrEnum is that it's only available in python 3.11+.

Unfortunately, this is then unusable for us.

odelalleau:
> One issue with `StrEnum` is that it's only available in python 3.11+.

Yeah but I’m pretty sure I already implemented something similar in older versions… unfortunately I don’t have access to a computer until next week so I can’t really be more specific.

docs/source/usage.rst
```diff
@@ -492,6 +492,43 @@ Note how the port changes to 82, and how the users lists are combined.
       file: log.txt
     <BLANKLINE>
 
+By default, ``merge()`` is replacing the target list with the source list.
+Use ``list_merge_mode`` to control the merge behavior for lists. 
+You can access the regarding enum with ``from omegaconf import ListMergeMode``.
+Currently there are three different merge modes:
```
omry, line 498:
```suggestion
By default, ``merge()`` is replacing the target list with the source list.
Use ``list_merge_mode`` to control the merge behavior for lists.
This Enum is defined in ``omegaconf.ListMergeMode`` and defines the following modes:
```

docs/source/usage.rst
```diff
@@ -492,6 +492,43 @@ Note how the port changes to 82, and how the users lists are combined.
...
+
+.. doctest::
+
+    >>> from omegaconf import OmegaConf, ListMergeMode
+    >>>
+    >>> cfg_example2 = OmegaConf.load('source/example2.yaml')
+    >>> cfg_example4 = OmegaConf.load('source/example4.yaml')
+    >>> 
+    >>> conf = OmegaConf.merge(cfg_example2, cfg_example4, list_merge_mode=ListMergeMode.EXTEND_IGNORE_DUPLICATES)
+    >>> print(OmegaConf.to_yaml(conf))
```
omry, line 523: Some lines here could be too long and cause the docs to have a horizontal scroll bar.
You can build the docs locally to check that. I suspect you will need to format the code a bit with some newlines.

news/1082.feature
```diff
@@ -0,0 +1 @@
+OmegaConf.merge() can now concat lists instead of replacing the target list
```
omry, line 1:
```suggestion
OmegaConf.merge() can now takes a list_merge_mode parameter that controls the strategy for merging lists (replace, extend and more).
```

omegaconf/base.py
```diff
@@ -795,6 +795,12 @@ class SCMode(Enum):
     INSTANTIATE = 3  # Create a dataclass or attrs class instance
 
 
+class ListMergeMode(Enum):
+    OVERRIDE = 1  # content from newer list gets taken (default)
+    EXTEND = 2  # lists get extended
+    EXTEND_IGNORE_DUPLICATES = 3  # only new (unique) elements get extended
```
omry, line 801: I know we have went though several iteration of bikeshedding here already, but any thoughts on using append instead of extend?
Also I think for the default REPALCE is more appropriate than override.
```suggestion
class ListMergeMode(Enum):
    REPALCE = 1  # Replaces the target list with the new one (default)
    APPEND= 2   # Appends the new list to the target list
    APPEND_UNIQUE_VALUE = 3  # Appends items that do not already exist in the target list
```

omry (commented):
Starting to look good.
Looking at the code, I have a suggestion for slightly better enum keys. Let me know what you think (sorry for the excessive bikeshedding).

omegaconf/base.py
Jasha10: I think `APPEND` is confusing due to the way that python's builtin `list.append` method works:
```python
>>> lst = [1,2,3]; lst.extend([4,5,6]); print(lst)  # expected
[1, 2, 3, 4, 5, 6]
>>> lst = [1,2,3]; lst.append([4,5,6]); print(lst)  # unexpected
[1, 2, 3, [4, 5, 6]]
```

Jasha10: > Also I think for the default REPALCE is more appropriate than override.

👍🏼

MatteoVoges: I like `REPLACE` and `APPEND_UNIQUE_VALUE` a lot. But I agree to Jasha, that we should name it after the python functions, so that `extend` is __extending__ the list. For `APPEND_UNIQUE_VALUE` I like the way, that we don't extend lists anymore, but rather looking at the actual values and if they are unique in the target list. @omry ,what do you think?

docs/source/usage.rst
MatteoVoges: Fixed in the latest iteration.

MatteoVoges:
The linting error is fixed in #1090

omegaconf/base.py
omry: Okay, from the perspective of the target list:
```python
# replace target list
REPLACE
# Extend target list with new list
EXTEND
# Extend target list with items not already present in it.
EXTEND_UNIQUE 
```

Jasha10 (approved):
Looks good!

news/1082.feature
```diff
@@ -0,0 +1 @@
+OmegaConf.merge() can now takes a list_merge_mode parameter that controls the strategy for merging lists (replace, extend and more).
```
MatteoVoges, line 1:
```suggestion
OmegaConf.merge() can now take a list_merge_mode parameter that controls the strategy for merging lists (replace, extend and more).
```

omry (approved):
Looks good, thanks!

MatteoVoges:
Thanks, are there any plans for a (pre-)release in the near future?

omry:
Maybe @shagunsodhani or @Jasha10  can do that.
In the mean time, you can depend on github repo revision from your build.

Jasha10:
@MatteoVoges fyi I've just uploaded `v2.4.0.dev0` to pypi.org:
https://pypi.org/project/omegaconf/2.4.0.dev0/

omry:
Hi @MatteoVoges,

I have resumed maintenance of OmegaConf, and I do not think `list_merge_mode` is currently in the right shape.

Are you depending on this feature? It has not shipped in a stable OmegaConf release, but it has been present in the 2.4.0 dev/pre-release packages. I am considering removing it before the stable 2.4 release and re-evaluating it in the context of some long-standing Hydra requests.

One concern is that the current option applies to the entire merge operation, while real-world use cases likely need list merge strategies to be selected more granularly, for example per field or per subtree. There may be other issues as well.

I would rather not release this as-is, and instead re-evaluate it after 2.4.0.

PR #1084, opened 2023-05-25
tests: fix lint error in tests

MatteoVoges:
There was a lint error in the tests:

```
tests/test_basic_ops_dict.py:276: error: Keywords must be strings  [misc]
Installing missing stub packages:
/home/circleci/project/.nox/lint-3-10/bin/python -m pip install types-Pygments types-colorama types-setuptools


Found 1 error in 1 file (checked 81 source files)
```
The problem is, that mypy detects the dict unpacking operator `**` in the function call as syntax error. So I adjusted the test, that the unpacking is still tested, but it don't happen in the function call anymore.

This is by the way a common requested feature / bugfix in mypy.
References:
* https://stackoverflow.com/questions/47632128/how-to-get-mypy-to-accept-an-unpacked-dict
* https://github.com/python/mypy/issues/5869
* https://github.com/python/mypy/issues/10023

### DIFF
--- a/tests/test_basic_ops_dict.py
+++ b/tests/test_basic_ops_dict.py
@@ -273,7 +273,7 @@ def test_map_expansion() -> None:
     def foo(a: int, b: int) -> int:
         return a + b
 
-    assert 12 == foo(**c)  # type: ignore[misc]
+    assert 12 == foo(**c)  # type: ignore[misc] # (known mypi error with dict unpacking)
 
 
 def test_items_iterator_behavior() -> None:
--- a/tests/test_basic_ops_dict.py
+++ b/tests/test_basic_ops_dict.py
@@ -267,17 +267,13 @@ class TestGetWithDefault:
 
 
 def test_map_expansion() -> None:
-    c = OmegaConf.create({"a": 2, "b": 10})
+    c = OmegaConf.create("{a: 2, b: 10}")
     assert isinstance(c, DictConfig)
 
     def foo(a: int, b: int) -> int:
         return a + b
 
-    a = {**c}["a"]
-    b = {**c}["b"]
-
-    # assert 12 == foo(**c) works as well, but gives lint error
-    assert 12 == foo(a, b)
+    assert 12 == foo(**c)  # type: ignore[misc]
 
 
 def test_items_iterator_behavior() -> None:
--- a/tests/test_basic_ops_dict.py
+++ b/tests/test_basic_ops_dict.py
@@ -267,13 +267,17 @@ class TestGetWithDefault:
 
 
 def test_map_expansion() -> None:
-    c = OmegaConf.create("{a: 2, b: 10}")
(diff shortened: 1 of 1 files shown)


tests/test_basic_ops_dict.py
```diff
@@ -267,13 +267,17 @@ def test_dict_get_with_default_errors(
...
 
     def foo(a: int, b: int) -> int:
         return a + b
 
-    assert 12 == foo(**c)
+    a = {**c}["a"]
+    b = {**c}["b"]
+
+    # assert 12 == foo(**c) works as well, but gives lint error
+    assert 12 == foo(a, b)
```
shagunsodhani, line 280: I am not sure if this is testing for expansion of args. This is more along the lines of `assert 12 == foo(c.a, c.b)`. Maybe it is better to silence the mypy error ?

shagunsodhani (commented):
Thanks for the PR! I left a comment and would love to learn more!

shagunsodhani (commented):
Thanks for the PR! I left a comment and would love to learn more!

tests/test_basic_ops_dict.py
MatteoVoges: Yeah, that is definitely possible and seems like a better solution. Thank you for the hint. I will fix it soon...

Jasha10 (changes requested):
The string `"{a: 2, b: 10}"` is a valid yaml document.
The line `c = OmegaConf.create("{a: 2, b: 10}")` is using `OmegaConf.create` to convert this yaml document into an OmegaConf DictConfig object.
The proposed diff changes what the test is testing.

Edit: I see now that the yaml document is not the point of the text. Please disregard my above comment.

tests/test_basic_ops_dict.py
Jasha10: > I am not sure if this is testing for expansion of args.

The name of the test is `test_map_expansion`. My guess is that it *is* testing expansion of args.

omry: I think we should surpress the lint error.
It's more important for me that the test intention is clear than to satisfy mypy.

MatteoVoges: The error is ignored now...

tests/test_basic_ops_dict.py
```diff
@@ -273,7 +273,7 @@ def test_map_expansion() -> None:
     def foo(a: int, b: int) -> int:
         return a + b
 
-    assert 12 == foo(**c)
+    assert 12 == foo(**c)  # type: ignore[misc]
```
omry, line 276: Please add a commend indicating this is known issue with mypy.

omry (changes requested):
Thanks.

tests/test_basic_ops_dict.py
MatteoVoges: Done.

Issue #1068, opened 2023-04-14
CI failure: DeprecationWarning: pkg_resources is deprecated as an API


Jasha10:
**Describe the bug**
CI is failing due to a DeprecationWarning being emitted by `setuptools`.

Here's an example failure on `master` branch:

<details>

```
$ nox -s omegaconf-3.11
nox > Running session omegaconf-3.11
nox > Creating virtual environment (virtualenv) using python3.11 in .nox/omegaconf-3-11
nox > python -m pip install --upgrade setuptools pip
nox > python -m pip install -r requirements/dev.txt .
nox > pytest 
============================================================= test session starts =============================================================
platform linux -- Python 3.11.0, pytest-7.3.0, pluggy-1.0.0
benchmark: 4.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /home/homestar/dev/omegaconf
configfile: pyproject.toml
plugins: benchmark-4.0.0, lazy-fixture-0.6.3, mock-3.10.0
collected 8120 items / 2 errors                                                                                                               

=================================================================== ERRORS ====================================================================
_______________________________________________ ERROR collecting build_helpers/test_helpers.py ________________________________________________
build_helpers/test_helpers.py:6: in <module>
    from build_helpers.build_helpers import find, find_version, matches
build_helpers/build_helpers.py:13: in <module>
    from setuptools.command import build_py, develop, sdist
.nox/omegaconf-3-11/lib/python3.11/site-packages/setuptools/command/develop.py:8: in <module>
    from setuptools.command.easy_install import easy_install
.nox/omegaconf-3-11/lib/python3.11/site-packages/setuptools/command/easy_install.py:50: in <module>
    from setuptools.sandbox import run_setup
.nox/omegaconf-3-11/lib/python3.11/site-packages/setuptools/sandbox.py:13: in <module>
    import pkg_resources
.nox/omegaconf-3-11/lib/python3.11/site-packages/pkg_resources/__init__.py:121: in <module>
    warnings.warn("pkg_resources is deprecated as an API", DeprecationWarning)
E   DeprecationWarning: pkg_resources is deprecated as an API
____________________________________________ ERROR collecting tests/test_pydev_resolver_plugin.py _____________________________________________
tests/test_pydev_resolver_plugin.py:23: in <module>
[...]

PR #1101, opened 2023-07-13
Fix pytest error due to deprecation warning

### LINKED ISSUE #1100: Tests fail due to `DeprecationWarning: pkg_resources is deprecated as an API`
**Describe the bug**

After a clean installation following the contribution guidelines, `pytest` fails with the error in the topic.

**To Reproduce**

Install as described in https://github.com/omry/omegaconf/blob/master/CONTRIBUTING.md and run `pytest`.

**Expected behavior**

Tests should run.

**Additional context**
- [X] OmegaConf version: `master`
- [X] Python version: 3.8
- [X] Operating system: Ubuntu

odelalleau:
Fixes #1100

The fix is similar to what was done in https://github.com/facebookresearch/hydra/pull/2696, instead of pinning a specific version of setuptools as in #1068

I'll mark this PR as draft since I still need to check if #1069 can be reverted now.
[Update: I double checked and we can indeed revert #1069 now, so adding it to this PR]

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ exclude = '''
 '''
 
 [tool.pytest.ini_options]
-addopts = "--import-mode=append -Werror"
+addopts = "--import-mode=append -Werror -W 'ignore:pkg_resources is deprecated as an API:DeprecationWarning' -W 'ignore:Deprecated call to `pkg_resources:DeprecationWarning'"
 pythonpath = ["."]
 
 [tool.towncrier]
--- a/noxfile.py
+++ b/noxfile.py
@@ -16,11 +16,7 @@ nox.options.error_on_missing_interpreters = True
 def deps(
     session: Session, editable_install: bool, requirements: str = "requirements/dev.txt"
 ) -> None:
-    session.install(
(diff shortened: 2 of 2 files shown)


odelalleau:
/poke @omry not sure if you've seen this

Jasha10 (commented):
Looks good to me! I like the idea of using `pyproject.toml` instead of `noxfile.py`. This way the pytest settings will take effect even when running pytest not via nox.

Jasha10 (approved):

omry:
lgtm

omry (approved):

pyproject.toml
```diff
@@ -13,7 +13,7 @@ exclude = '''
 '''
 
 [tool.pytest.ini_options]
-addopts = "--import-mode=append -Werror"
+addopts = "--import-mode=append -Werror -W 'ignore:pkg_resources is deprecated as an API:DeprecationWarning' -W 'ignore:Deprecated call to `pkg_resources:DeprecationWarning'"
```
winniehell, line 16: this fails now: https://github.com/omry/omegaconf/issues/1234

PR #1104, opened 2023-07-16
handle SCMode.INSTANTIATE with throw_on_missing=False

### LINKED ISSUE #1103: One should be allowed to re-instantiate a structured object with missing values
Defining dataclasses to introduce schema seems like a great idea. But, sometimes we have missing values, and there is `omegaconf.MISSING` to express this:

```python 
from omegaconf import OmegaConf, MISSING

@dataclass(frozen=True)
class A:
    x: int = MISSING

# If we are allowed to instantiate this
a = A()
print(a)
# ==> {'a': {'x': '???'}}

# Then we should be allowed to instantiate it again:
OmegaConf.to_container(
    a, 
    throw_on_missing=False, 
    structured_config_mode = omegaconf.SCMode.INSTANTIATE
)
# ...
# MissingMandatoryValue: Structured config of type `A` has missing mandatory value: x
#     full_key: x
#     object_type=A
```

The use case is we may have partially filled objects that we want to manipulate, construct new configs out of, etc.

The problem is that this line forgets that `.to_container`'s `throw_on_missing` argument:

https://github.com/omry/omegaconf/blob/7dae67e4a3869b584fd6d6408b2f916c176755db/omegaconf/basecontainer.py#L289

and inside this function, it could be allowed here:

https://github.com/omry/omegaconf/blob/7dae67e4a3869b584fd6d6408b2f916c176755db/omegaconf/dictconfig.py#L741

It's currently hard for me logistically to contribute the simple PR, but hopefully I have identified the changes necessary sufficiently if the maintainers agree this should be supported.

Jasha10:
This PR enables `OmegaConf.to_container(..., throw_on_missing=False)` to work for structured configs when `structured_config_mode=SCMode.INSTANTIATE`. Closes #1103.

In addition, this PR moves `MISSING` from `omegaconf.py` into `base.py` to avoid a circular import issue.

Here is an example of the behavior enabled by this PR:
```python
# repro.py
from enum import Enum
import omegaconf

@dataclass(frozen=True)
class A:
    x: int

a = OmegaConf.create(A)
container = OmegaConf.to_container(
    a,
    throw_on_missing=False,
    structured_config_mode=omegaconf.SCMode.INSTANTIATE,
)

assert isinstance(container, A)
print(container)
```
### BEFORE:
```
$ python repro.py
  File "/home/homestar/dev/omegaconf/repro.py", line 12, in <module>
    cfg = OmegaConf.to_container(
...
omegaconf.errors.MissingMandatoryValue: Structured config of type `A` has missing mandatory value: x
    full_key: x
    object_type=A
```
### AFTER:
```
$ python repro.py
A(x='???')
```

### DIFF
--- a/omegaconf/__init__.py
+++ b/omegaconf/__init__.py
@@ -1,4 +1,12 @@
-from .base import Container, DictKeyType, ListMergeMode, Node, SCMode, UnionNode
+from .base import (
+    MISSING,
+    Container,
+    DictKeyType,
+    ListMergeMode,
+    Node,
+    SCMode,
+    UnionNode,
+)
 from .dictconfig import DictConfig
 from .errors import (
     KeyValidationError,
@@ -19,16 +27,7 @@ from .nodes import (
     StringNode,
     ValueNode,
 )
-from .omegaconf import (
-    II,
-    MISSING,
-    SI,
-    OmegaConf,
-    Resolver,
-    flag_override,
-    open_dict,
-    read_write,
-)
+from .omegaconf import II, SI, OmegaConf, Resolver, flag_override, open_dict, read_write
 from .version import __version__
 
 __all__ = [
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -378,7 +378,8 @@ def get_dataclass_fields(obj: Any) -> List["dataclasses.Field[Any]"]:
 def get_dataclass_data(
     obj: Any, allow_objects: Optional[bool] = None
 ) -> Dict[str, Any]:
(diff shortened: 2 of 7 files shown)


omry:
Looks good overall and I agree that it's a better behavior.
I think there is still some circular dependency.

omry:
I want to understand the use case better before allowing this.

omry:
I am not convinced of the use case.
MISSING is, by definition, missing.
The object cannot represent missing values. Using OmegaConf's `???` sentinel for this is a bad idea.
Closing, feel free to follow up if you want to discuss.

PR #1109, opened 2023-07-28
Update pre-commit hooks and drop support for Python 3.6 & 3.7

odelalleau:
Fixes #1108

This PR updates pre-commit hooks to the latest current versions of isort / black / flake8 / mypy, and ensures we can run all these tools in the repository without any complaint.

PR #1113, opened 2023-08-08
Properly escape interpolation-like strings in resolved configs

### LINKED ISSUE #1081: Add `is_unresolved` to OmegaConf
I'd like to know when the value would be an indirect call and when it's a direct input

There are other helper functions like `is_missing` and `is_unresolved` would be able to allow custom logic around detecting the source of the value

None

### LINKED ISSUE #1112: Unexpected behavior for multiline-strings
**Describe the bug**
Hey, I'm trying to have a config, where some values are in an escaped interpolation, but in a multiline-string. I think this is a really specific use case, but I think that this applies to all escaped characters.

**To Reproduce**
```python
from omegaconf import OmegaConf as oc, Node

correct = {
    "config": {
        "subconfig": "MULTILINE \n    ${tag:ENV}",
    },
    "values": "${config}",
}

wrong = {
    "values": "${config}",
    "config": {
        "subconfig": "MULTILINE \n    ${tag:ENV}",
    },
}


def escape_tag(input: str, _node_: Node):
    """resolver function, that returns an escaped tag with the input"""
    if "\n" in str(_node_):  # multiline-strings
        escape = "\\\\\\"  # \\\
    else:
        escape = "\\"  # \

    return escape + "${" + input + "}"


oc.register_new_resolver("tag", escape_tag)


cfg_correct = oc.create(correct)
cfg_wrong = oc.create(wrong)

oc.resolve(cfg_correct)
oc.resolve(cfg_wrong)

print(cfg_correct["values"])
# >>> {'values': {'subconfig': 'MULTILINE \n    \\${ENV}'}}
print(cfg_wrong["values"])
# >>> {'values': {'subconfig': 'MULTILINE \n    \\\\\\${ENV}'}

### Another bug ?
print(cfg_correct["values"]["subconfig"])
# >>> [...] ${ENV}
print(cfg_wrong["values"]["subconfig"])
# >>> [...] \${ENV}
```

**Explanation**
Here I have two configs `correct` and `wrong` and the only difference between them is the key order in the python file.
Then I have a custom resolver that produces me an escaped interpolation. In multiline-strings `\` has to be escaped and because of the double escapes (https://github.com/omry/omegaconf/pull/695#discussion_r621335546) we need 3 `\`. Now the diff is that in the `wrong` config, each `\` is escaped again, and it produces 6 `\`.

Another bug (?) is, that when I print the actual strings, that the string gets dumped as multiline and not as quoted, but this could be intended.

**Expected behavior**
I expect, that the order in the config doesn't matter and both configs produce `\${ENV}`.

If there is any kind of workaround like sorting the keys in the config, I would be happy already. 

**Additional context**
- [X] OmegaConf version: latest / master / 2.4.0.dev0
- [X] Python version: 3.10.12
- [X] Operating system: Linux Ubuntu

odelalleau:
Fixes #1112
Fixes #1081

This is a draft PR for several reasons:
1. It needs tests + news fragment
2. It is built on top of #1109 that needs to be merged first
3. The behavior of `to_container()` and `to_object()` needs to be looked at (see #1120)

Note that this is a breaking change, but it can be argued that the previous behavior was a bug (all tests pass).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,24 +1,25 @@
+default_language_version:
+      python: python3.8
+
 repos:
-  - repo: https://github.com/timothycrosley/isort
-    rev: 5.0.9
+  - repo: https://github.com/PyCQA/isort
+    rev: 5.12.0
     hooks:
       - id: isort
 
   - repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 23.7.0
     hooks:
       - id: black
-        language_version: python3.8
 
-  - repo: https://gitlab.com/pycqa/flake8
-    rev: 3.7.9
+  - repo: https://github.com/PyCQA/flake8
+    rev: 6.0.0
     hooks:
       - id: flake8
-        additional_dependencies: [-e, 'git+https://github.com/pycqa/pyflakes.git@1911c20#egg=pyflakes']
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.931
+    rev: v1.4.1
     hooks:
       - id: mypy
         args: [--strict]
-        additional_dependencies: ['pytest']
+        additional_dependencies: ['attrs', 'pytest']
--- a/omegaconf/_impl.py
+++ b/omegaconf/_impl.py
@@ -10,6 +10,7 @@ from ._utils import (
     _get_value,
     is_primitive_container,
     is_structured_config,
+    maybe_escape,
 )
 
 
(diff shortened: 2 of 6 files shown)


Jasha10 (approved):
Looks good to me!

omry:
I think this needs tests to show what it's actually doing and to ensure it does not break in the future.

odelalleau:
> I think this needs tests to show what it's actually doing and to ensure it does not break in the future.

Oh yes (see point 1 above)

> Also, worth considering any performance impact.

I need to think about it a bit more, but I don't believe this is an issue. Will keep it in mind though.

MatteoVoges:
Hey I noticed a breaking change with this. See full description of it here: https://github.com/omry/omegaconf/issues/1112#issuecomment-1678573449

omry (changes requested):
Thanks for putting this together.

Closing this draft because both issues it was intended to address (#1112 and #1081) have been rejected / resolved without taking this implementation approach. This PR is therefore no longer planned for merge.

PR #912, opened 2022-05-05
update from antlr-4.8 to antlr-4.9

pixelb:
Use the latest 4.9 release (4.9.3)

Addresses issue #798

PR #1114, opened 2023-08-09
Upgrade Antlr to 4.11 and vendor the runtime

### LINKED ISSUE #1091: Request support for anltr 4.11.1
I realize that this has been brought up a few times before (e.g., #1000), but I currently have fugue installed in an environment with omegaconf, and I would like to be able to upgrade it, and it and dependent packages all have versions compatible with antlr-python-runtime 4.11.1.  So, I'd like to request a version of omegaconf that uses this version of antlr-python-runtime.

jlopezpena:
Bumping `antlr` binary and `antlr4-python3-runtime` in order to unlock doing the same thing in `hydra`, following [this discussion](https://github.com/facebookresearch/hydra/issues/2699)

### DIFF
--- a/omegaconf/version.py
+++ b/omegaconf/version.py
@@ -1,6 +1,6 @@
 import sys  # pragma: no cover
 
-__version__ = "2.4.0.dev1"
+__version__ = "2.4.0.dev2"
 
 msg = """OmegaConf 2.4 and above is compatible with Python 3.8 and newer.
 You have the following options:


Jasha10 (approved):
Thanks @jlopezpena!
This looks good to me.
I have confirmed that the sha256 of the antlr .jar file in this PR is the same as the sha256 of the .jar file available at [github.com/antlr/website-antlr4](https://github.com/antlr/website-antlr4/tree/gh-pages/download).

Jasha10:
(CI lint failures are unrelated to this PR)

omry:
@Jasha10, this is a breaking change (that would also require a parallel fix to Hydra.
1. News fragment.
2. I think it's really best if the effort is focused on fixing the root issue instead of bumping the version like this. In the mean time users with specific antlr version needs can build their own version of OmegaConf.

astrojuanlu:
> In the mean time users with specific antlr version needs can build their own version of OmegaConf.

How should users (like myself) do that, in a way that it works with various dependencies declaring `omegaconf` in its metadata?

(To keep the comment on-topic: from the peanut gallery, looks like this is effort well spent on a short term solution while folks address the root cause)

astrojuanlu:
> The problem with this approach is if you are working on a library that needs to be used by other people in their project, because then you are forcing THEM to also do the same thing

Yeah that's what I had in mind, ideally we should be able to do this in a way that transitive dependencies don't have to cooperate. Otherwise I don't see how users can cope with the current situation effectively.

shchur:
Hello, just wanted to clarify what the status of this PR is. Does the comment by @omry mean that this PR will be blocked and we will need to find another way to make `omegaconf` compatible with different `antlr4-python3-runtime` versions?

If this is the case, I would be happy to help with finding the root cause and working on a PR to address it.

Jasha10:
I think it would be good to at least discuss @omry's points before merging.

> @Jasha10, this is a breaking change

Yes, we are working on that: https://github.com/facebookresearch/hydra/pull/2733

> 1. News fragment.

Done! (commit https://github.com/omry/omegaconf/pull/1114/commits/bc4949b9a0995495a578c7712d24df764c4618f3)

> 2. I think it's really best if the effort is focused on fixing the root issue instead of bumping the version like this.

Any ideas for fixing the root issue? The antlr version is tightly coupled to the minor version of `antlr4-python3-runtime`.

> In the mean time users with specific antlr version needs can build their own version of OmegaConf.

Here is a recipe for users who need to build their own version of OmegaConf+Hydra:
https://github.com/facebookresearch/hydra/issues/2491#issuecomment-1344306280

My opinion is that we should go forward with this PR.
- Not clear how to fix the root cause.
- We've done PRs like this in the past (https://github.com/omry/omegaconf/pull/912 and https://github.com/facebookresearch/hydra/pull/2192)

odelalleau:
I haven't looked at this in details, but would it be possible by any chance to make Hydra / OmegaConf compatible with multiple versions of antlr? (compiling the grammar during `pip install` instead of shipping the pre-compiled files)

Jasha10:
> would it be possible by any chance to make Hydra / OmegaConf compatible with multiple versions of antlr? (compiling the grammar during `pip install` instead of shipping the pre-compiled files)

The current workflow for [compiling the grammar](https://github.com/omry/omegaconf/blob/f150237070a2571f8d54df67801c25f17b8bfddd/build_helpers/build_helpers.py#L30) invokes the following command:
```
java -jar build_helpers/bin/antlr-4.9.3-complete.jar -Dlanguage=Python3 -o omegaconf/grammar/gen -Xexact-output-dir -visitor omegaconf/grammar/grammar
```
This requires the antlr `.jar` to be present when the grammar is compiled.

If we compile the grammar during `pip install` then I think we would have to either (1) ship multiple antlr .jar files and select one of them at install time, or (2) expose an option for the user to provide their own antlr dependency.

Option (2) has been suggested before ([here](https://github.com/facebookresearch/hydra/issues/2491#issuecomment-1407480383)). I'm not sure specifically how we would expose such an option at `pip install` time -- maybe via an  [extras](https://peps.python.org/pep-0508/#extras) option?

natephysics:
It just so happens that I've run across a dependency problem involving this exact issue as well. Seems like a tricky issue to solve.

omry:
I think the best solution would be to "vendor" OmegaConf with a specific version of antlr, and have Hydra depend on it.
That would require some rewriting some imports in the generated code and would essentially eliminate the external dependency on antlr.

I think I outlined it at some point as one of the options for a permanent fix for this issue.

omry:
> Could an option be to replace the `anttrl4` java binary by [antlr-tools](https://github.com/antlr/antlr4-tools)?

I don't see how that would do anything.
The antlr java binary is just a matter of convenience during the build of OmegaConf.
The problem is that the antlr Python runtime only works with generated code that is of the exact same version used to generate the parser.

odelalleau:
I think @omry's suggestion of shipping a specific version of `antlr4-python3-runtime` within OmegaConf may be the most robust approach.

omry:
> I see, so the binary dependency is not runtime, but build-time only.

Not quite. The antlr runtime is checking that it's working with code generated by a generator from the exact same version.
This is a problematic design choice, as in reality antlr is very stable and they could be much laxer with this check - but I digress.

Other low-level libraries, like pip itself, solves the problem of dependency conflicts by optionally including their dependencies in under a different module. They call this vendoring ([pip page about it](https://pypi.org/project/vendoring)).
The idea is that by including your own version of the dependency, you isolate your dependencies from the dependencies of userland apps and eliminating these kinds of issues.

The complication is that Hydra should probably follow whatever OmegaConf is doing here.

astrojuanlu:
Is there a final decision on what the way forward should be? We are discovering more incompatibilities with other dependencies.

odelalleau:
> Is there a final decision on what the way forward should be? We are discovering more incompatibilities with other dependencies.

I think shipping our own version will be easier overall. We just need someone to actually implement it :)

astrojuanlu:
> I think shipping our own version will be easier overall.

I guess you mean @omry's vendoring idea from https://github.com/omry/omegaconf/pull/1114#issuecomment-1699656976 

> We just need someone to actually implement it :)

👍🏽 Thank you!

Jasha10:
> For whatever is worth, I don't think vendoring is the right solution. It is an anti-pattern

@jlopezpena, thank you for sharing your opinion about this. It seems you are advocating for the second approach from [your comment above](https://github.com/omry/omegaconf/pull/1114#issuecomment-1704768456): building "multiple versions of the grammars at build time (through multiple versions of the antlr binary), and select the grammar at runtime based on the installed version of antlr4-python3-runtime." I am concerned that this approach would make omegaconf's build and release process more difficult.

Currently, omegaconf's build & release process is very simple (here are the full instructions from the docstring of [`setup.py`](https://github.com/omry/omegaconf/blob/ceec6f45046d6d67b2e3faba32d8f2d3561ea81c/setup.py)):
```
OmegaConf setup
    Instructions:
    # Build:
    rm -rf dist/ omegaconf.egg-info/
    python -m build
    # Upload:
    twine upload dist/*
```
I worry that the proposal to build multiple versions of the grammar would significantly complicate this process. Meanwhile, the vendoring appraoch could be as "simple" as copying the `antlr4-python3-runtime` source into omegaconf's source tree, allowing us to keep the same build procedure. Does this make sense to you?

> by vendoring a third party pakage you are left with the burden of incorporating security updates and whatnot yourself - which might or might not be a big ordeal depending on how often there are changes that need to be incorporated.

Note that, with our current approach, we have not been updating our pin on the antlr4 runtime very frequently. Likewise, I do not anticipate that we would update a vendored version of the antlr4 runtime frequently.

odelalleau:
Thanks @jlopezpena, unfortunately no time for me to work on OmegaConf this week.

> To get CI to run properly we need to decide whether we do the vendoring as part of the build or as a manual step that gets committed.

I would just do a manual commit, seems simpler that way.

Jasha10:
> Decide if we want to commit the antlr4 source files to the repo or get them downloaded and modified through the script at build time
@Jasha10 what is your opinion on the last one?

Thanks again for all of the discussion and for working on the lead with implementation, @jlopezpena. I agree with @odelalleau about committing the antlr4 source files to the repo.

Jasha10:
> There are a bunch of flake8 errors in tests

Could you please rebase this PR against a recent version of the `master` branch?
Most of those flake8 errors should have been fixed by https://github.com/omry/omegaconf/pull/1109, which was merged into `master` after you opened this PR.

Jasha10:
Not sure -- I don't know a lot about how to configure CodeQL.
Anyway, I've gone through and manually dismissed the CodeQL alerts related to the vendor folder.

odelalleau:
> Because changes in antlr4 binary/runtime will be API breaking (-ish) for `hydra`, to ensure good semantic versioning `omegaconf` might want to adopt a policy requiring that any changes to the antlr binary/runtime must be released with a major version bump (i.e. the current PR should be released as `omegaconf` 3.0, not 2.4).
> 
> I believe hydra and omegaconf versions are heavily interlocked already, so hopefully this will not bring much of an additional maintenance burden (albeit whenever vendoring some pain is unavoidable).

Yeah that makes sense, and it should be fine. Note however that:
* Technically Hydra could ship its own vendored version of antlr runtime. But I don't think it's worth the extra trouble at first sight (we can still do it later if needed).
* Hydra / OmegaConf have never followed semantic versioning: breaking changes are typically introduced in X.Y+1 versions.
* I'm not sure bumping the antlr runtime version in OmegaConf should be considered a breaking change, even if it breaks Hydra: Hydra is a "special" case in the sense that it's tightly coupled with OmegaConf, but no other library should be relying on OmegaConf's internal antlr runtime version. As long as we properly set Hydra's dependencies, there's no strong reason to consider a new antlr runtime in OmegaConf "breaking" in terms of user facing release notes.

Jasha10 (approved):
This looks great. Thanks so much @jlopezpena.

omegaconf/base.py
```diff
@@ -629,7 +628,7 @@ def _validate_and_convert_interpolation_result(
             # If the converted value is of the same type, it means that no conversion
             # was actually needed. As a result, we can keep the original `resolved`
             # (and otherwise, the converted value must be wrapped into a new node).
-            if type(conv_value) != type(res_value):
+            if type(conv_value) is not type(res_value):
```
omry, line 631: Why this change?

omegaconf/typing.py
```diff
@@ -0,0 +1,10 @@
+import sys
+
+if sys.version_info >= (3, 10):
+    from typing import TypeAlias
+else:
+    from typing_extensions import TypeAlias
+
+from .vendor.antlr4.ParserRuleContext import ParserRuleContext
+
+Antlr4ParserRuleContext: TypeAlias = ParserRuleContext  # type: ignore[valid-type]
```
omry, line 10: What is the purpose of this file, and in particular why did you have to "rename" ParserRuleContext  to Antlr4ParserRuleContext?

omry (commented):
Thanks for working on it.
Overall looks good to me.
I think should probably trigger a major version bump.

I added a few inline questions.
Also, what is the plan to address this in Hydra?

Jasha10:
> what is the plan to address this in Hydra?

In Hydra we can change `import antlr4` to `from omegaconf.vendor import antlr4`.

omegaconf/typing.py
omry: Gotcha.
I wonder if MyPy has a way to globally exclude a specific type...

If not, 
Can you add a comment explaining the alias?
Also, I think you could keep the name the same, just import it from a different place.

omry:
> > > what is the plan to address this in Hydra?
> > 
> > 
> > In Hydra we can change `import antlr4` to `from omegaconf.vendor import antlr4`.
> 
> There is still a potential issue in that we need to make sure the binary antlr jar in hydra needs to be the same version as the vendored runtime in omegaconf (but AFAIK that is already the case anyway)

The build of Hydra would need to change similarly to the build of OmegaConf to rewrite the generated code as well.

omry:
Sounds good.
For simplicity, I think you can copy the code. No need to include the build code related code with OmegaConf for that purpose.
The way to control OmegaConf dependency in Hydra (and by extension, it's dependencies) is by pretty strict version dependency of OmegaConf from Hydra.

https://github.com/facebookresearch/hydra/blob/main/requirements/requirements.txt#L1

mawildoer:
Hey folks! Thanks for the upgrade and vendoring!
I'm running into a similar issue to hyrda on a seperate project with antlr dependencies.

I've been mostly successfully using this branch today, however I've hit a few snags trying to install it.

Mainly, if I use anything other than an editable installation, I get this import error:

```
ModuleNotFoundError: No module named 'omegaconf.vendor.antlr4.error'
```

I've tried:
- `python setup.py build` + `pip install .`
- `python setup.py sdist` + `pip install dist/omegaconf-2.4.0.dev1.tar.gz`
- `python setup.py antlr` + `pip install -e .` (with only this option succeeding)

I'm not sure I've successfully emulated what will happen during an actual CI build, but I have checked and confirms that the subdirectories of the `omegaconf/vendor` we not present in the install.

I didn't see them in the checks, but is there a package I could manually install from CI while still on this branch?

Thanks!

Matt

astrojuanlu:
I don't know if it's related but note that direct `setup.py` invokation on the CLI is deprecated, see https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#what-commands-should-be-used-instead

mawildoer:
> I don't know if it's related but note that direct `setup.py` invokation on the CLI is deprecated, see https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#what-commands-should-be-used-instead

TIL; those are the old-school alias for the `-e` we know and love! I've only ever used "speciality" `setup.py` commands once before, for something more analogous to the `antlr` case there (eg. doing something special). I just saw them this time in the `setup.py` and figured it was worth giving a whirl, since other things weren't working yet. Thanks for the comment!

> @mawildoer the non-editable installs should work now, could you give it a check?

Yessir! Tested both `pip install -e .` and `pip install .`. Both working 🎉 w/ full IDE/pyright/pylint etc... finding them too.

Thanks for the quick response! I'll have to package and distribute your branch if we can't get this back to mainline soon 😅

Thanks for all the work!

Jasha10:
@jlopezpena I've just uploaded omegaconf dev release `2.4.0.dev1` to pypi based on this PR branch. This should be useful so there's something to pin against when working on the Hydra PR.
https://pypi.org/project/omegaconf/2.4.0.dev1/
```
pip install omegaconf==2.4.0.dev1
```

shchur:
Hi @Jasha10, what are the remaining TODOs before this PR can be merged?

Currently, the antlr version incompatibility is major blocker for our project (because of the clash between Omegaconf & https://github.com/Nixtla/statsforecast), so I would be happy to help with any of the remaining TODOs, if possible.

Jasha10:
In my opinion this PR is in good shape.
I hoped to merge this at about the same time as it's sister PR https://github.com/facebookresearch/hydra/pull/2733 where unfortunately we have a backlog of CI issues that I believe are mostly unrelated to the PR itself (rather just technical debt on the Hydra project).

shchur:
@Jasha10 are there some open issues related to the CI failures on the hydra side that I could take a stab at?

odelalleau:
> 
@Jasha10, thank you for the quick response! Are there some open issues related to the CI failures on the hydra side that I could take a stab at?

There are no open issues but you can basically look at recent open PRs on Hydra to check what's failing.

Jasha10:
@shchur I have made a list:
https://github.com/facebookresearch/hydra/issues/2841

Jasha10 (approved):

Jasha10:
Thank you @jlopezpena for taking on this vendoring project.
The PR is merged and omegaconf dev release `2.4.0.dev2` is now available on pypi.
https://pypi.org/project/omegaconf/2.4.0.dev2/
```
pip install omegaconf==2.4.0.dev2
```

flange-ipb:
Hello, I'm wondering if [ANTRL's LICENSE.txt](https://github.com/antlr/antlr4/blob/dev/LICENSE.txt) needs to be shipped with the vendored distribution. This is clearly a condition of the 3-clause BSD license.

- The vendored source files contain only a short copyright notice and refer to "the LICENSE.txt file in the project root".
- The [wheel of antlr4-python3-runtime==4.11.1](https://files.pythonhosted.org/packages/b6/3a/4f2e9d67277cc255a17a34c1b54e02d939ca5991bc596dee3d63e587e512/antlr4_python3_runtime-4.11.1-py3-none-any.whl) does not contain a license file in its dist-info directory (this can be done, but this is up to its package maintainers).

omry:
Good catch. I will include it with the 2.4 release.

Issue #1102, opened 2023-07-14
Merging nested readonly structured configs doesn't work.


amatsukawa:
**Describe the bug**
Merging nested readonly structured configs doesn't work.

There seems to be code to temporarily unset the read-only property of the outer-most container prior to merge to make this work in `OmegaConf.merge`. I believe the issue is that this logic needs to be recursive.

**To Reproduce**
```python
from omegaconf import OmegaConf, MISSING
from dataclasses import dataclass

@dataclass(frozen=True)
class A:
    x: int = MISSING

@dataclass(frozen=True)
class B:
    a: A = MISSING
    b: int = 1

# This is the config
b = OmegaConf.structured(B(a=A()))
print(b)
# ==> {'a': {'x': '???'}, 'b': 1}

# Works
print(OmegaConf.merge(b, {"b": 3}))
# ==> {'a': {'x': '???'}, 'b': 3}


# Doesn't work
print(OmegaConf.merge(b, {"a": {"x": 3}}))
# ...
# ReadonlyConfigError: Cannot set value of read-only config node
#    full_key: a.x
#    reference_type=A
#    object_type=A
```

**Expected behavior**
Merging nested structured readonly configs should work.

Jasha10:
@amatsukawa this behavior is as expected.
See the `omegaconf.read_write` context manager: https://omegaconf.readthedocs.io/en/2.0_branch/usage.html#read-only-flag.
```
>>> conf = OmegaConf.create({"a": {"b": 10}})
>>> OmegaConf.set_readonly(conf, True)
>>> with read_write(conf):
...   conf.a.b = 20
>>> conf.a.b
20
```

Jasha10:
Thanks, for the report, @amatsukawa.

For now, you can work around this issue using the `read_write` context manager to temporarily turn off the frozen state of the config.

```python
from omegaconf import read_write

...

with read_write(b.a):
    print(OmegaConf.merge(b, {"a": {"x": 3}}))
```

amatsukawa:
Thanks for the response @Jasha10. 

For those who come later: indeed that was the workaround I had come to, but to use merge with (1) an arbitrarily nested structured template config and (2) unknown (ahead of time) updates in the second argument of the merge, I had settled on recursively setting the readonly flag to false on every `DictConfig` and `ListConfig` object.

PR #1124, opened 2023-09-20
feat: set flag `readonly` recursively

### LINKED ISSUE #1123: Inconsistencies in recursive behavior in `set_readonly()`
**Describe the bug**
There is unexpected behavior for different types of recursive flag passing when using `OmegaConf.set_readonly()`

**To Reproduce**
```python

config = OmegaConf.create({"a": {"b": 1}})

# (1)
OmegaConf.set_readonly(config, True)
print(OmegaConf.is_readonly(config.a))
# >>> True (EXPECTED)

OmegaConf.set_readonly(config, False)
print(OmegaConf.is_readonly(config.a))
# >>> False (EXPECTED)

# (2)
OmegaConf.set_readonly(config.a, True)
print(OmegaConf.is_readonly(config))
# >>> False (EXPECTED)

OmegaConf.set_readonly(config, False)
print(OmegaConf.is_readonly(config.a))
# >>> True (WRONG)
```

Here we see in the first example, that when applying `readonly` to a config, that subconfigs get recursively updated with the correct state.
In the second example we apply `readonly` to a subconfig, and the parent node doesn't get updated, but when setting `readonly` to false in the parent node, the children don't get recursively updated, which is wrong.

**Expected behavior**
The child nodes should get correctly updated.

**Additional context**
- [X] OmegaConf version: master (2.4.0.dev1)
- [X] Python version: 3.10.12
- [X] Operating system: linux

MatteoVoges:
## Motivation

I want a function that sets a value for a flag (atm `readonly`, but can be extended for all / generic flags) recursively.

Yes.

## Test Plan

Tests will be added if the basic structure is confirmed.

## Fixes

Fixes #1123

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -539,9 +539,10 @@ Configuration flags
 -------------------
 
 OmegaConf support several configuration flags.
-Configuration flags can be set on any configuration node (Sequence or Mapping). if a configuration flag is not set
+Configuration flags can be set on any configuration node (Sequence or Mapping). If a configuration flag is not set
 it inherits the value from the parent of the node.
 The default value inherited from the root node is always false.
+To avoid this inhertence and set a flag explicitly to all children nodes, specify the option `recursive=True`.
 
 .. _read-only-flag:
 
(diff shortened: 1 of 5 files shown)


MatteoVoges:
@odelalleau with this I get:

```python

config = OmegaConf.create({"a": {"b": 1}})

OmegaConf.set_readonly(config.a, True)
OmegaConf.set_readonly_recursively(config, False)
print(OmegaConf.is_readonly(config.a))
# >>> False
```

odelalleau:
Thanks! I would suggest to do it a bit differently to have a more generic implementation:
* Add a `recursive` boolean argument to `_set_flag()`
* Also add it to `set_struct()` and `set_readonly()` (instead of adding a different function specific to readonly)

odelalleau:
Thanks for the update -- overall looks good to me, but I won't have time to look at it in details until next week. We'll need to add a news fragment as well and update the documentation.

indigoviolet:
@omry @odelalleau this is a blocker for #1102, would be great to get it in

omry:
Hi @MatteoVoges.
Thanks for the PR.

I'm closing this as I disagree with the direction.
My position is explained in my comment on #1123.

PR #1135, opened 2023-10-24
fix: `SCMode.INSTANTIATE` for attrs fields with aliases/leading "_"

### LINKED ISSUE #1063: `SCMode.INSTANTIATE` for attrs classes breaks when an attribute name has a leading underscore
**Describe the bug**
According to the attrs API reference, attribute names are stripped of leading underscores in the `__init__()`: different behavior to dataclasses. But this quirk isn't handled by OmegaConf when instantiating the class. 

**To Reproduce**
```
import attr
from omegaconf import OmegaConf, SCMode

@attr.s
class Foo():
    _bar: int = attr.ib(default=0)

OmegaConf.to_container(OmegaConf.structured(Foo), structured_config_mode=SCMode.INSTANTIATE)
```
**Result**
```
omegaconf.errors.ConfigTypeError: Could not create instance of `Foo`: Foo.__init__() got an unexpected keyword argument '_bar'
    full_key: 
    object_type=Foo
```

**Expected result**
```
Foo(_bar=0)
```

**Workaround**
Set an alias in `attr.ib(...)` to the actual attribute name.
```
    _bar: int = attr.ib(default=0, alias='_bar')
```

**Additional context**
- [ ] OmegaConf version: <2.3.0>
- [ ] Python version: <3.10.10>
- [ ] Operating system: <Windows 10>
- [ ] Please provide a minimal repro

bzczb:
## Motivation
`attrs` classes automatically alias any fields with a leading underscore to a stripped version in that class's `__init__()`.
So, the class
```python
@attr.s
class Foo():
    _bar: int = attr.ib(default=0)
```
has an `__init__()` that looks like:
```python
def __init__(self, bar):
    self._bar = bar
```
---
This PR changes the structured config instantiation behavior to take aliases into account, when setting up the params for an `__init__()`.

## Test Plan
Test `test_leading_underscore_fields()` added.

## Fixes
Fixes #1063

### DIFF
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -468,15 +468,27 @@ def is_structured_config_frozen(obj: Any) -> bool:
     return False
 
 
-def get_structured_config_init_field_names(obj: Any) -> List[str]:
+def _find_attrs_init_field_alias(field: Any) -> str:
+    # New versions of attrs, after 22.2.0, have the alias explicitly defined.
+    # Previous versions implicitly strip the underscore in the init parameter.
+    if hasattr(field, "alias"):
+        assert isinstance(field.alias, str)
+        return field.alias
+    else:  # pragma: no cover
+        assert isinstance(field.name, str)
+        return field.name.lstrip("_")
+
+
+def get_structured_config_init_field_aliases(obj: Any) -> Dict[str, str]:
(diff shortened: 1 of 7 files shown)


Jasha10:
Hi @bzczb,
Thanks for submitting these PR's! I will review them shortly.

Jasha10 (approved):
Nice work on this PR @bzczb. Thanks very much!

PR #1159, opened 2024-02-27
Update linter config & reformat code with black

Jasha10:
This PR updates pins on linters so that the same tool versions can be used for developing Hydra and OmegaConf. Corresponds to https://github.com/facebookresearch/hydra/pull/2854.
The code has also been reformatted with the latest version of `black`.


Jasha10:
> Looks good but it seems we are getting hit by [TvoroG/pytest-lazy-fixture#65](https://github.com/TvoroG/pytest-lazy-fixture/issues/65)

Thanks for for the link -- I've implemented the suggested workaround from that thread.

PR #1199, opened 2024-12-17
Change deepcopy logic to be constrained on correct type

andywag:
The deep copy is always done when it is done with an instance of Box and the roots are the same. Box doesn't contain get_root so this condition fails causes an error.

## Motivation

Broken Case

Yes/No

## Test Plan

This issue is showing up in an internal Meta codebase

## Fixes

No Issue

## Related PRs

None


omry:
Please create an issue with a minimal repro of the problem this is solving.

PR #1211, opened 2025-03-12
Migrate dependencies to pyproject.toml

wouterzwerink:
## Motivation

Moving to a more modern and static pyproject.toml setup for dependencies. Nice to have everything in one place and easier to discover dependencies for installers.

## Test Plan

Still installs fine, though I still need to check that CI is still working.


omry:
Thanks for the PR. Unfortunately it's become outdated due to some project changes, and it looks like it was never fully complete to begin with.

I'll follow up with a similar update in #1255.

Issue #755, opened 2021-06-17
negative list indices in variable interpolation


Jasha10:
This issue is based on https://github.com/facebookresearch/hydra/issues/1674.

**Describe the bug**
Negative list indices do not work in interpolations

**To Reproduce**
```python
from omegaconf import OmegaConf
cfg = OmegaConf.create(
    {
        "list": [8, 16, 32],
        "interp": "${list[0]}",
        "interp_negative": "${list[-1]}",
    }
)
assert cfg.interp == 8
assert cfg.interp_negative == 32  # raises InterpolationKeyError
```

**Expected behavior**
Negative list indices work in interpolations.

**Additional context**
- OmegaConf version: master
- Python version: 3.9.5

omry:
This is a feature request, not a bug.
Interpolation into a list using negative index never worked (and this is the first time someone asks for it).
It only seems appropriate because of the new support for brackets in interpolation syntax.

Using the supported syntax in 2.0 will look like this:
```python
cfg = OmegaConf.create(
    {
        "list": [8, 16, 32],
        "interp": "${list.0}",
        "interp_negative": "${list.-1}",
    }
)
```
We can consider this for 2.2, but I don't think it's high pri.

Jasha10:
> Using the supported syntax in 2.0 will look like this:

FYI the `"${list.-1}"` syntax is not working for me. I tried it on the 2.0 branch and on 2.1/master.

omry:
> > Using the supported syntax in 2.0 will look like this:
> 
> FYI the `"${list.-1}"` syntax is not working for me.
> I tried it on the 2.0 branch (gives ConfigKeyError) and on 2.1/master (gives InterpolationKeyError).

This was never supported functionality.
The reason I mentioned `${list.-1}` is because it's equivalent to `${list[-1]}`. The fact that we added support for bracketed access in interpolation does not mean this is a real list. We also do not support slice syntax.

royvelich:
Just my own 2 cents, IMHO This is a very important feature for anything related with ML and Deep Learning model configurations. Many times I would like to be able to access the last element of feature layer lists - It makes everything so much easier.

odelalleau:
> Many times I would like to be able to access the last element of feature layer lists - It makes everything so much easier.

You can still do it with something like:
```python
OmegaConf.register_new_resolver("last", lambda lst: lst[-1])
```
```yaml
layers:
- input
- hidden
- output
last_layer: ${last:${layers}}
```

[...]

Issue #1086, opened 2023-06-02
ListConfig errors when is [:-1] applied to empty list


YodaEmbedding:
**Describe the bug**
ListConfig errors when is [:-1] applied to empty list.

**To Reproduce**

```python
>>> from omegaconf import OmegaConf
>>> conf = OmegaConf.create({"a": None, "b": "keyname", "${b}": "${b} wow dynamic", "xs": []})
>>> conf.xs[:-1]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.11/site-packages/omegaconf/listconfig.py", line 218, in __getitem__
    self._format_and_raise(key=index, value=None, cause=e)
  File "/usr/lib/python3.11/site-packages/omegaconf/base.py", line 231, in _format_and_raise
    format_and_raise(
  File "/usr/lib/python3.11/site-packages/omegaconf/_utils.py", line 899, in format_and_raise
    _raise(ex, cause)
  File "/usr/lib/python3.11/site-packages/omegaconf/_utils.py", line 797, in _raise
    raise ex.with_traceback(sys.exc_info()[2])  # set env var OC_CAUSE=1 for full trace
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/omegaconf/listconfig.py", line 203, in __getitem__
    for slice_idx in itertools.islice(
                     ^^^^^^^^^^^^^^^^^
ValueError: Stop argument for islice() must be None or an integer: 0 <= x <= sys.maxsize.
    full_key: xs[None:-1]
    object_type=list
```

**Expected behavior**
```python
>>> xs = []
>>> xs[:-1]
[]
```

**Additional context**
- [ ] OmegaConf version: 2.3.0
- [ ] Python version: 3.11
- [ ] Operating system: Arch Linux
- [ ] Please provide a minimal repro

Jasha10:
Thanks @YodaEmbedding.

For reference, here is comparison with the current behavior for python's built-in `list` type:
```python
>>> [][:-1]
[]
```

PR #1212, opened 2025-03-15
Interpolate for negative indices in list resolution

sean-roelofs-ai:
## Motivation

Supports Interpolation in lists with negative intergers

Yes

## Test Plan

pytest with additional cases for negative indices in `test_grammar.py`

## Issues

Feature is requested in https://github.com/omry/omegaconf/issues/755


omry:
Thanks for the contribution! The core logic in `_select_one` is correct and this is a worthwhile feature. Before we dig into the details, a couple of housekeeping requests:

1. **Please rebase on `main`** — the branch has fallen behind and there are conflicts.
2. **Squash your commits into a single commit** — the current stack has 5 commits (`make negative indice resolution work`, `news fragment`, `fix import order`, `black on omegaconf`, `fix test names`). One clean commit makes the diff much easier to review.

Once that's done, please also address the following:

**Drop the unrelated reformatting.** Several functions in `omegaconf.py` were reformatted (`create`, `merge`, `unsafe_merge`, `register_new_resolver`, `_get_resolver`, `_node_wrap`) with no relation to this feature. Please revert those — they inflate the diff and will conflict with other open PRs.

**Clarify the `listconfig.py` change.** The new bounds check in `__getitem__` normalizes negative indices before passing to `_resolve_with_default`. It's not clear this was broken before — Python list indexing already handles negative indices natively, so `_content[-1]` would have worked. Please add a test that demonstrates `cfg[-1]` / `cfg["list"][-1]` was broken before your change, or remove it if it isn't needed. Also note that after normalization `index` is mutated to a positive value, so if an exception is raised afterward, the error message on line 227 will report the normalized key rather than the original negative one the user passed.

**Fix inconsistent out-of-range behavior in `_select_one`.** Out-of-range positive indices silently return `None` regardless of `throw_on_missing`. The new negative out-of-range path raises `InterpolationKeyError` when `throw_on_missing=True`. These should behave the same way — match the existing behavior and set `val = None` for out-of-range negative indices too.

**Test coverage and naming.**
- No test covers direct negative index access (`cfg[-1]`). If you keep the `listconfig.py` change, add one.
- `("list_access", "${list.10}", 9)` reuses an existing test ID — give it a unique name (e.g. `"list_access_10"`).
- Rename `"list_access_good_negative_1"` / `"list_access_bad_negative_12"` to match the existing convention — drop the `good`/`bad` qualifiers (e.g. `"list_access_negative_1"`, `"list_access_negative_oob"`).

**News fragment.**
- Rename the file from `negative_interpolation.news` to `755.feature` — the convention is `{issue_number}.{type}` (see other files in `news/`).
- Reduce to a single line with no code block, e.g.: `Added support for negative list indices in interpolations (e.g. \`${list.-1}\`).`
- The current description "Allows negative list access" is misleading — direct access (`cfg[-1]`) already worked; what this PR adds is interpolation support.

**Comments.** The two new inline comments (`# Handle negative indices in interpolation`, `# Convert negative index to positive`) describe what the code does, which the code already makes clear. Leave comments for the *why*, not the *what* — these can be removed.

omry:
Thanks, @sean-roelofs-ai, for working on this and for the contribution here.

I’m going to close this PR for now. I think this area should be handled together with the related issues in #755 and #1086, and I’m planning to address that in a separate PR.

Thanks again for the work on this.

PR #1214, opened 2025-05-03
Raise ConfigTypeError when updating through a null structured Optional node (closes #1214)

wu0607:
OmegaConf.update() would hit an AssertionError with a confusing message when
the key path traversed a structured Optional field set to None.

The root cause: DictConfig(content=None) is a Container, so the traversal
loop did not replace it with {}, then root[key] dereferenced to Python None,
and the post-loop assert fired with an unhelpful message.

The fix detects a null Container during traversal and raises ConfigTypeError
immediately, naming the offending intermediate key via _get_full_key:

    ConfigTypeError: Cannot set 'a.b' because 'a' is None

The post-loop assert is kept as an internal invariant guard with a simple
message. Untyped None values (AnyNode) are unaffected — they are replaced
with an empty dict as before.

Closes #1214.

### DIFF
--- a/omegaconf/omegaconf.py
+++ b/omegaconf/omegaconf.py
@@ -731,9 +731,11 @@ class OmegaConf:
 
         last = split[-1]
 
-        assert isinstance(
-            root, Container
-        ), f"Unexpected type for root: {type(root).__name__}"
+        assert isinstance(root, Container), (
+            f"Unexpected type for root: {type(root).__name__}. "
+            f"We cannot set '{last}' to '{value}' when root is {root}. "
+            f"Please make sure the config value of {split} is valid."
+        )
 
         last_key: Union[str, int] = last
         if isinstance(root, ListConfig):


omry:
Thank you for the report — this is a real bug, tracked in #1280.

This PR does not meet the project's contribution requirements:

1. **No reproduction script.** Bug fix PRs must include a repro that demonstrates the issue before the fix.
2. **No tests.** The fix must be covered by tests, including the exact error message.
3. **No news fragment.** Changes must include a `news/<issue>.bugfix` (or other appropriate type) entry for the changelog.
4. **Wrong exception type.** `assert` is disabled with `python -O`. The fix should raise a proper `ConfigTypeError`, not patch the assert message.
5. **Misleading error message.** Including the value being set and the key list as a Python list does not help the user understand what went wrong.

A corrected fix has been applied and will ship in the next release.

PR #1225, opened 2025-10-23
Add support for Python 3.12-3.14 and drop support for Python 3.8

rsokl:
- Stop requiring 100% coverage across all versions. Just enforce for 3.10
- Run coverage, lint, and docs in separate jobs so that failures are easier to understand at a high level

### DIFF
--- a/tests/test_nodes.py
+++ b/tests/test_nodes.py
@@ -1021,12 +1021,13 @@ def test_path_str_hash_collision_handling() -> None:
     assert hash(path_node) == hash(path_str)
     assert path_node != path_str
 
-    list_cfg = OmegaConf.create([path_obj])
-    assert path_obj in list_cfg
-    assert path_str not in list_cfg
-    list_cfg.append(path_str)
-    assert path_obj in list_cfg
-    assert path_str in list_cfg
+    list_cfg1 = OmegaConf.create([path_obj])
+    assert path_obj in list_cfg1
+    assert path_str not in list_cfg1
+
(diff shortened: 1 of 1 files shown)


noxfile.py
```diff
@@ -63,8 +64,9 @@ def version_string_to_tuple(version: str) -> Tuple[int, ...]:
     return tuple(map(int, version.split(".")))
 
 
-@nox.session(python=PYTHON_VERSIONS)  # type: ignore
+@nox.session(python=["3.10"])  # type: ignore
 def lint(session: Session) -> None:
+    # Note: Linting only runs on Python 3.10 to avoid running it on every CI job
```
Jasha10: If we only run lints on 3.10, might we run into issues if e.g. we use some new python syntax that's supported in python3.8 but not supported in python3.10? The 3.10 linter would fail to warn us in such a situation.

requirements/dev.txt
```diff
@@ -10,7 +10,7 @@ mypy==1.8.0
 nox
 pre-commit
 pyflakes
-pytest
+pytest<8.4  # Pin to <8.4 due to removal of RaisesContext from private API
```
Jasha10: how unfortunate

rsokl: Yeah... I spent a while looking into supporting the new `RaisesExc` simultaneously and it is a big mess. I think a separate PR will be needed to switch to `pytest>=8.4` and switch to solely supporting `RaisesExc`. Refactoring the tests that use it to use neither seems pretty non-trivial

PR #1229, opened 2025-11-23
Add support for arithmetic operators in number interpolations (fixes #91)

### LINKED ISSUE #91: Consider supporting basic operators in interpolations of numbers
```yaml
a: 1
b: 2.0
d: ${a} - ${b}
e: ${a} + ${b}
d: ${a} * ${b}
e: ${a} / ${b}
```
desired behavior:
```python
conf = OmegaConf.load("file.yaml")
assert conf.d == -1.0
assert conf.d == 3.0
assert conf.d == 2.0
assert conf.d == 0.5
```
This should work only if both operands are numbers (int or float). result type should be int if both are ints or float otherwise.

#### ✨✨✨ See the OmegaConf docs on [How to Perform Arithmetic Using eval as a Resolver](https://omegaconf.readthedocs.io/en/2.3_branch/how_to_guides.html#id1) ✨✨✨

suneel944:
This PR implements support for basic arithmetic operators (+, -, *, /) in number interpolations as requested in issue #91.

## Changes
- Implement arithmetic operations for numeric interpolations in `grammar_visitor.py`
- Return int when both operands are int, float otherwise
- Fall back to string concatenation for non-numeric operands (backward compatible)
- Add comprehensive tests for all operations and edge cases
- Optimize implementation for performance with single-pass processing

## Example Usage
```python
cfg = OmegaConf.create({'a': 1, 'b': 2.0})
cfg.d = '${a} + ${b}'  # Results in 3.0 (float)
cfg.e = '${a} - ${b}'  # Results in -1 (int if both int)
```

## Testing
All tests pass, including:
- Addition, subtraction, multiplication, division
- Type checking (int vs float results)
- Whitespace handling
- Fallback to string concatenation for non-numeric operands

### DIFF
--- a/news/91.feature
+++ b/news/91.feature
@@ -1,0 +1,2 @@
+Support basic arithmetic operators (+, -, *, /) in number interpolations.
+
--- a/omegaconf/grammar_visitor.py
+++ b/omegaconf/grammar_visitor.py
@@ -295,7 +295,10 @@ class GrammarVisitor(OmegaConfGrammarParserVisitor):
             if isinstance(c, OmegaConfGrammarParser.InterpolationContext):
                 return self.visitInterpolation(c)
 
-        # Otherwise, concatenate string representations together.
+        result = self._try_arithmetic_expression(list(ctx.getChildren()))
+        if result is not None:
+            return result
+
         return self._unescape(list(ctx.getChildren()))
 
     def _createPrimitive(
@@ -333,7 +336,9 @@ class GrammarVisitor(OmegaConfGrammarParserVisitor):
                 # A single WS should have been "consumed" by another token.
                 raise AssertionError("WS should never be reached")
             assert False, symbol.type
-        # Concatenation of multiple items ==> un-escape the concatenation.
+        result = self._try_arithmetic_expression(list(ctx.getChildren()))
+        if result is not None:
+            return result
         return self._unescape(list(ctx.getChildren()))
 
     def _unescape(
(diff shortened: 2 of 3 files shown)


suneel944:
@omry  can I get a review on this?

suneel944:
@rsokl could you give me a review here please

omry:
> @omry can I get a review on this?

@suneel944, yes. I started to catch up on some issues here.
I will take some time though.
Thanks for your work on this.

omry:
@suneel944, for now I will defer this to after 2.4.0 lands. this is too big of an enhancement to include at this stage.

omry:
Thanks for working on this and for putting together such a thorough implementation and test suite.

I’m going to close this PR for now. Arithmetic support in interpolations is a much larger design surface than it first appears, and I don’t want to merge a grammar/semantics change of this size without a deliberate design pass.

Some of the open questions include:

- whether arithmetic should be part of normal interpolation syntax, a dedicated resolver, or a separate expression syntax;
- how to avoid surprising backwards compatibility changes for existing string interpolations like `${foo}+${bar}`;
- whether invalid/non-numeric operands should fall back to strings or raise;
- how operator precedence, literal expressions, whitespace, quoting, and resolver arguments should behave;
- how this should interact with Hydra’s CLI grammar and override parsing.

The underlying request in https://github.com/omry/omegaconf/issues/91 remains open, but it needs a deliberate design before we accept an implementation. Closing this PR does not mean rejecting the feature permanently; it means this particular implementation path is not something I’m comfortable merging as-is.

Thanks again for the contribution. The tests and implementation here are still useful context for a future design.

PR #1231, opened 2025-12-17
Support keys as lists or tuples to `OmegaConf.update()` (fixes #1230)

### LINKED ISSUE #1230: Allow `OmegaConf.update()` to Accept Literal Keypaths as Lists
Long time OmegaConf user, first time participating here. Thanks for the great project. :)

## The problem

OmegaConf keypaths always use one of `:`, `.`, or `[ ]` as delimiters when parsing nested updates.

Additionally, YAML allows these characters in keys. In some cases this is an obvious observation. In cases with `:` it's less obvious, but still possible. For a valid example:

```yaml
some_settings:
  "mm(1:3)": 42
```

I'm working in a use case where I need to load a large number of configuration files into OmegaConf, but many of these files contain keys that include special characters `:`, `.`, `[`, and `]`. Unfortunately, OmegaConf interprets these characters as structural markers for nested keypaths.

This creates ambiguity when trying to reference or update values using `OmegaConf.update()`. Any string passed to the update function is parsed by `split_key()`, which treats these characters as delimiters for nested keypaths -- even when they are meant to be literal characters within a single key.
 
```python
OmegaConf.update(cfg, "some_settings:mm(1:3)", ...)   # ambiguous if key is mm(1:3) <-- this is my current issue
OmegaConf.update(cfg, "some_settings[mm(1:3)]", ...)  # still parsed
OmegaConf.update(cfg, "some_settings[mm[1:3]]", ...)   # ambiguous if key is mm[1:3]
OmegaConf.update(cfg, "some_settings.mm(1:3)", ...)   # still parsed
OmegaConf.update(cfg, "some_settings.mm(1.3)", ...)   # ambiguous if key is mm(1.3)
```

## Proposal
Allow `OmegaConf.update()` to accept a list of literal key segments:

```python
OmegaConf.update(
    cfg,
    ["some_settings", "mm(1:3)"],
    99
)
```

This would bypass `split_key()` entirely and treat the path as explicit and literal. I think/hope this would also be fully backward-compatible, since existing `split_key()` already returns a list.

### Minimal code change

In [OmegaConf.update](https://github.com/omry/omegaconf/blob/master/omegaconf/omegaconf.py#L718) change the key handling from 

```python
split = split_key(key)
```

to:

```python
split = key if not isinstance(key, str) else split_key(key)
```

This retains all existing behavior for string keypaths, while allowing unambiguous literal keypaths when a list is passed. I'm uncertain if this would interfere with the actual updating though. 

## What I've already considered:
- Escaping characters inside the keypath string (`mm(1\:3)`) — does not work, because `split_key()` recognizes multiple delimiters.
- Using bracket notation -- also gets parsed before brackets are interpreted.
- Manually walking the tree and inserting values. functional but requires reimplementing OmegaConf’s merge semantics.

I'm happy to implement this and open a PR if devs agree/support this request!

nannau:
## Motivation

There are legitimate edge cases where users may want to call `OmegaConf.update()` with keys that **contain special characters** such as `:`, `.`, or `[]`. 

Previously, all keys were passed through `split_key`, which made it impossible to update values whose literal keys contained these characters.

This change allows users to **explicitly specify a key path** by passing a `list` or `tuple` to `OmegaConf.update()`. Doing so bypasses automatic key splitting and enables updates to keys containing special characters while preserving existing behavior for string-based keys.

## Contributing Guidelines

- [x] I have read the [Contributing Guidelines on pull requests](https://github.com/omry/omegaconf/blob/master/CONTRIBUTING.md)

## Test Plan

Added several new test cases to `test/test_update.py` covering:
- [x] when a `key` is a `list` or `tuple`, the value is correctly inserted/merged
- [x] when a special delimiter appears in a key, the key-strings (with the delimiters) in the key-path are preserved
- [x] verified that all existing and new tests pass across tested Python versions 3.9 -> 3.14

I tried building docs, but noticed a failure with `sphinx >= 9`. I temporarily limited sphinx to be `<9` and the docs built successfully. I suspect that might come up when the docs are built next, but it's not related to this PR.

## Fixes

Fixes #1230


nannau:
Hi @rsokl. I'm just tagging you based on recent activity, apologies if you're not the preferred contact.

I'm wondering what the review process is for this project and if changes like this are currently being accepted/encouraged? Or, is there anything more I can do to help out

omry:
Thanks for the issue report and the pull request.
This has been superseded by #1279

PR #1249, opened 2026-04-22
Preserve unresolved interpolations when merging missing structured nodes

### LINKED ISSUE #1205: When merging DictConfs and interpolation fails due to MISSING, the default dataclass is used
**Describe the bug**
If I merge two configs with the same schema but where one has a missing value that another entry uses for interpolation, the default dataclass for that variable type is used instead. This is dangerous since using the default is definitely not intended!
For instance, my use case merging resolved and unresolved configs to give a config which is "as resolved as possible"

**To Reproduce**
```
import copy

from dataclasses import *

import omegaconf
print(omegaconf.__version__)

@dataclass(unsafe_hash=True)
class WithDefaults:
	x: int = 0
	y: int = 0


@dataclass(unsafe_hash=True)
class B:
	b: WithDefaults = '${a}'  # this value relies on `a` in the top level


@dataclass(unsafe_hash=True)
class Parent:
	a: str = '???'  # deliberately leave missing
	b: B = B()


cfg = OmegaConf.structured(Parent())
cfg2 = OmegaConf.structured(Parent(b='???'))
cfg.merge_with(cfg2)  

print(cfg)

```

Output:
```
2.3.0
{'a': '???', 'b': {'b': {'x': 0, 'y': 0}}}

```

**Expected behavior**
cfg2 should not have the default `B` it should remain unresolved since I'm merging with a a config that contains `???`.
A merge of `'${interpolation}' <- '???'` should result in '${interpolation}' not the default dataclass. 
In a merge, MISSING must not be preferred and certainly a default dataclass should not be preferred
Output should read: `{'a': '???', 'b': {'b': '${a}'}}`

**Additional context**
- [x] OmegaConf version: 2.3.0
- [x] Python version: 3.11.10
- [x] Operating system: Mac OS 14.6 (23G80)

The undesired behaviour seems to occur in `expand` function that is inside BaseContainer._map_merge: `val = rt`.

omry:
When a merge combined a destination node holding an unresolved interpolation
with a source node that was MISSING but typed as a structured config,
OmegaConf synthesized a prototype for the source and then expanded the
destination interpolation. That replaced the interpolation with the
structured type's default value, so the merged config no longer preserved
the original interpolation semantics.

Fix this by tracking when the source started out as a missing structured
node and treating the synthesized prototype as type information only. In
that case, keep the destination interpolation intact, update its type
metadata, and skip the expansion path that would otherwise materialize the
default structured value. Add a regression test covering a structured child
whose interpolation target must remain unresolved after merge.

Closes #1205

### DIFF
--- a/news/1205.bugfix
+++ b/news/1205.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug where merging a missing structured config into an unresolved interpolation could replace the interpolation with the structured type's default value instead of preserving the interpolation.
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -358,11 +358,12 @@ class BaseContainer(Container, ABC):
 
             node._set_value(val)
 
-        if (
+        src_was_missing_structured = (
             src._is_missing()
             and not dest._is_missing()
             and is_structured_config(src_ref_type)
-        ):
+        )
+        if src_was_missing_structured:
             # Replace `src` with a prototype of its corresponding structured config
             # whose fields are all missing (to avoid overwriting fields in `dest`).
             assert src_type is None  # src missing, so src's object_type should be None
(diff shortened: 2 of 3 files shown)


omry:
@odelalleau, @Jasha10 - remember this one? :)

PR #1294, opened 2026-05-11
Break exception reference cycle

### LINKED ISSUE #1295: Omegaconf exception has reference cycle
**Describe the bug**

Omegaconf exceptions raised by `_format_and_raise()` and `_raise()` functions have the `ex -> traceback -> stack frame -> ex` reference cycle:

```
def _raise(ex: Exception, cause: Exception) -> None:
   ..
   raise ex.with_traceback(sys.exc_info()[2])
```

Concretely, the exception raised (i.e. `ex`) has the `_raise` frame in its traceback and the frame's `f_locals` have `ex` which is a cycle.

Exception reference cycle is particularly bad because:

1. They capture all local variables of all frames of the call stack (all the way up to `main()` via frame.f_back) which can be large objects or open resources like file.
2. Only Python GC can collect objects with cycle but Python GC can be much delayed so you can see resources leaked for a long time or OOM.

This is a known issue in Python (https://peps.python.org/pep-3134/#open-issue-garbage-collection) and Python has https://peps.python.org/pep-3110/#semantic-changes to help us but it won't save us in our case since we manually store exception to a local variable hence the cycle.

**To Reproduce**

```
def repro():
   x = "non_exist_key" in DictConfig # this will create a `ConfigAttributeError` exception with reference cycle
```

**Expected behavior**

Technically it's not a bug since there is no real memory leak but the practical outcome is bad enough so that we should avoid exception reference cycle if possible.

**Additional context**
- [ ] OmegaConf version: main
- [ ] Python version: All python version
- [ ] Operating system: <__FILL__>
- [ ] Please provide a minimal repro

jjyao:
## Motivation

Currently exceptions raised by omegaconf have reference cycle from `ex -> traceback -> stack frame -> ex` since we store `ex` in local variable which is bad (see https://peps.python.org/pep-3134/#open-issue-garbage-collection for more details). Python has a solution (https://peps.python.org/pep-3110/#semantic-changes) that we can manually do here to break the cycle.

Yes

## Test Plan

N/A

## Fixes

What issue does this PR fix? Use https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue to link this PR to a corresponding issue.

Fixes #1295

## Related PRs

(Is this PR part of a group of changes? Link the other relevant PRs.)

### DIFF
--- a/news/1295.bugfix
+++ b/news/1295.bugfix
@@ -1,0 +1,1 @@
+Fix OmegaConf exceptions retaining traceback frame locals through reference cycles.
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -966,8 +966,13 @@ def _raise(ex: Exception, cause: Exception) -> None:
     env_var = os.environ["OC_CAUSE"] if "OC_CAUSE" in os.environ else None
     debugging = sys.gettrace() is not None
     full_backtrace = (debugging and not env_var == "0") or (env_var == "1")
-    if full_backtrace and cause is not ex:
-        ex.__cause__ = cause
+    if full_backtrace:
+        # In the alias case (`cause is ex`), assigning `ex.__cause__ = cause`
+        # would create a self-reference cycle that pins `ex` (and transitively
+        # its traceback frames and locals) until cyclic GC runs. Use a shallow
+        # clone so `__cause__` is non-None for the OC_CAUSE=1 contract while
+        # remaining a distinct object that doesn't loop back to `ex`.
+        ex.__cause__ = copy.copy(cause) if cause is ex else cause
     else:
         ex.__cause__ = None
     try:
--- a/omegaconf/_utils.py
+++ b/omegaconf/_utils.py
@@ -970,7 +970,12 @@ def _raise(ex: Exception, cause: Exception) -> None:
         ex.__cause__ = cause
     else:
         ex.__cause__ = None
-    raise ex.with_traceback(sys.exc_info()[2])  # set env var OC_CAUSE=1 for full trace
+    try:
+        raise ex.with_traceback(sys.exc_info()[2])  # set env var OC_CAUSE=1 for full trace
+    finally:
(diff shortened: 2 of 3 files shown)


jjyao:
cc @omry for review. Let me know if you have any questions.

omry (changes requested):
Hi @jjyao,
Thanks.
My initial reaction is FFS. what are they thinking introducing a change to Python that requires userland code to jump through hoops like this.

Second reaction:
What is the actual problem you are seeing? 
What Python version?
What OmegaConf version?
Do you have a repro of the impact of this?
What OmegaConf and Python versions are affected?

This is a very low level "fix" without enough context or justification.
Please create a proper issue with full details and link this PR to it.
This needs also needs test coverage and a news fragment.

jjyao:
@omry 

I created https://github.com/omry/omegaconf/issues/1295, let me know if you need more details.

jjyao:
@omry can you take a look when you get a chance? Thanks!

omry:
Yeah, busy day.

omry:
Hi, I improved the test and added a news fragment.

omry:
Thanks for working on this!
By the way, are you related to Ray?
Are you aware of the Hydra plugin?

jjyao:
Yup: I worked on Ray.

omry:
Nice.
I initiated the [Hydra Ray Plugin](https://hydra.cc/docs/plugins/ray_launcher/). A team member implemented it.
These days I think it's probably the least maintained part of Hydra. it's very difficult to test (we used to have actual AWS testing there).

Anyway, I am starting to bring Hydra back into shape, I might eventually revive that launcher too (or kill it if a mercy kill is needed :) ).

jjyao:
Cool. Let me know when you need any help on the Ray front.

Also when will the next release be so I can pull in this fix :)

omry:
I am in bug crashing mode here and in Hydra.
I can throw in a dev release for you if you want to start playing with it.

PR #1311, opened 2026-06-10
Unwrap UnionNode when iterating a ListConfig (#1310)

### LINKED ISSUE #1310: ListConfig iteration leaks UnionNode for List[Union[...]]
## Summary

Iterating over a `List[Union[float, int]]` from a structured config yields
`UnionNode` objects instead of the selected concrete values.

This appears to be OmegaConf-specific and was originally reported downstream in
Hydra:

- https://github.com/facebookresearch/hydra/issues/3067

## Repro

```python
from typing import List, Union

@dataclass
class MainConfig:
    a: List[Union[float, int]] = field(default_factory=lambda: [0.8, 0.9])
    b: List[float] = field(default_factory=lambda: [0.8, 0.9])


cfg = OmegaConf.structured(MainConfig)

print(type(cfg.a[0]).__name__)
print(type(next(iter(cfg.a))).__name__)

print([cfg.a[i] * 10 for i in range(len(cfg.a))])
print([i * 10 for i in cfg.b])
print([i * 10 for i in cfg.a])
```

## Actual behavior

On OmegaConf `2.4.0.dev11`:

```text
float
UnionNode
[8.0, 9.0]
[8.0, 9.0]
TypeError: unsupported operand type(s) for *: 'UnionNode' and 'int'
```

I reproduced this both with the installed `2.4.0.dev11` package and with the
local OmegaConf source checkout. The current GitHub `main` branch is one commit
ahead of that local checkout, but that commit only adds Backlog Atlas workflow
files and does not touch OmegaConf runtime code.

## Expected behavior

Iteration should behave like indexing and yield the selected concrete values:

```text
[8.0, 9.0]
```

## Notes

This looks like a public API leak of `UnionNode`.

`UnionNode` is designed as a pass-through wrapper around exactly one selected
value node, or a special value such as `None`, missing, or interpolation. The
class docstring says user apps and downstream libraries should not need to know
about `UnionNode`.

Indexing works because `ListConfig._get_child()` unwraps `UnionNode` to the
concrete child node. Iteration goes through `ListConfig.ListIterator.__next__`,
which currently unwraps ordinary `ValueNode` instances but not `UnionNode`.

gaoflow:
Iterating a `List[Union[...]]` from a structured config yielded `UnionNode` wrappers instead of the selected concrete values, while indexing already unwrapped them:

```python
cfg = OmegaConf.structured(MainConfig)   # a: List[Union[float, int]] = [0.8, 0.9]
type(cfg.a[0]).__name__            # 'float'
type(next(iter(cfg.a))).__name__   # 'UnionNode'  ← leak
[i * 10 for i in cfg.a]            # TypeError: 'UnionNode' * 'int'
```

Indexing resolves elements through `_resolve_with_default` → `_get_value`, which unwraps `UnionNode`. `ListConfig.ListIterator.__next__` unwrapped `ValueNode` but treated a `UnionNode` as a plain container and returned it raw.

The fix unwraps `UnionNode` via `_get_value` on the **resolved** iteration path only, so iteration now mirrors indexing. The unresolved path (`_iter_ex(resolve=False)`, used internally when copying a `ListConfig`) keeps returning the node, so a copy still preserves its `UnionNode` typing — covered by a dedicated test.

Tests in `tests/test_basic_ops_list.py`: iteration yields concrete values with the int/float distinction preserved (fails without the change), and a copy keeps the `UnionNode` node type.

Fixes #1310.

### DIFF
--- a/tests/test_basic_ops_list.py
+++ b/tests/test_basic_ops_list.py
@@ -196,22 +196,6 @@ def test_iterate_list_of_union() -> None:
     assert [v * 10 for v in cfg.a] == [8.0, 90]
 
 
-def test_copy_list_of_union_preserves_node_type() -> None:
-    # The unresolved iteration path is used internally when copying a ListConfig;
-    # it must keep the UnionNode so the copy retains its union typing.
-    from dataclasses import dataclass, field
-
-    from omegaconf.base import UnionNode
-
-    @dataclass
-    class Cfg:
-        a: List[Union[float, int]] = field(default_factory=lambda: [0.8, 9])
-
-    cfg = OmegaConf.create(OmegaConf.structured(Cfg))
-    assert isinstance(cfg._get_node("a")._get_node(0), UnionNode)
-    assert list(cfg.a) == [0.8, 9]
-
-
 def test_items_with_interpolation() -> None:
     c = OmegaConf.create(["foo", "${0}"])
     assert c == ["foo", "foo"]
--- a/tests/test_basic_ops_list.py
+++ b/tests/test_basic_ops_list.py
@@ -176,6 +176,42 @@ def test_iterate_list_with_missing() -> None:
         next(itr)
 
 
+def test_iterate_list_of_union() -> None:
+    # Regression test for the UnionNode leak: iterating a List[Union[...]] used
+    # to yield UnionNode wrappers instead of the selected concrete values, while
+    # indexing already unwrapped them.
+    from dataclasses import dataclass, field
+
+    @dataclass
+    class Cfg:
+        a: List[Union[float, int]] = field(default_factory=lambda: [0.8, 9])
+
+    cfg = OmegaConf.structured(Cfg)
+
+    assert type(next(iter(cfg.a))) is type(cfg.a[0])
+    iterated = list(cfg.a)
+    assert iterated == [0.8, 9]
+    # The concrete int/float distinction selected by the union is preserved.
+    assert [type(v) for v in iterated] == [float, int]
+    assert [v * 10 for v in cfg.a] == [8.0, 90]
+
+
+def test_copy_list_of_union_preserves_node_type() -> None:
+    # The unresolved iteration path is used internally when copying a ListConfig;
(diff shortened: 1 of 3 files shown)


tests/test_basic_ops_list.py
```diff
@@ -176,6 +176,42 @@ def test_iterate_list_with_missing() -> None:
...
+    assert iterated == [0.8, 9]
+    # The concrete int/float distinction selected by the union is preserved.
+    assert [type(v) for v in iterated] == [float, int]
+    assert [v * 10 for v in cfg.a] == [8.0, 90]
+
+
+def test_copy_list_of_union_preserves_node_type() -> None:
+    # The unresolved iteration path is used internally when copying a ListConfig;
+    # it must keep the UnionNode so the copy retains its union typing.
+    from dataclasses import dataclass, field
```
omry: This is a higher level test that is potentially already covered. see existing copy tests in the test suite and consider adding this check there.

omry:
Thanks for your help @gaoflow.

binggao1230:
Thanks for the review! You're right — I removed the copy test.

I checked whether it was guarding anything: copying a `List[Union[...]]` (via `OmegaConf.create` or `copy.deepcopy`) preserves the `UnionNode` even with the iterator change reverted, because the elements get re-wrapped against the list's element type on the way back in. So the test wasn't catching a real failure mode, and the whole suite stays green either way.

I kept `self.resolve and …` on the iterator branch so the unresolved path (used internally by the copy in `_set_value_impl`) keeps returning the node rather than resolving it — that matches the existing intent of `resolve=False` even though the re-wrapping makes it harmless in practice. Happy to drop the gate too if you'd prefer the simpler unconditional form.

The remaining `test_iterate_list_of_union` is the actual regression and fails without the fix.

Disclosure: this PR was written by an AI coding agent (Claude) running under my account and direction.

omry:
No need for the disclosure, it was obvious by the speed of the PR :).
Curious though, what prompted you to pitch in on this one?

Issue #1058, opened 2023-02-19
structured config dataclasses are discarded on list merge


raphCode:
**Describe the bug**
`OmegaConf.merge(config, structured)` fails to merge dataclasses contained in a list. This includes type information as well as values.
Related:
https://github.com/facebookresearch/hydra/discussions/2586

**To Reproduce**
```python
from omegaconf import OmegaConf
from attrs import frozen

@frozen
class A:
    num: str = 13

@frozen
class C:
    a: A
    aa: list[A]
    num: int =  42

c = OmegaConf.structured(C)
c = OmegaConf.merge(dict(a=dict(num=1), aa=[dict(num=2)], num=1337), c)
assert OmegaConf.get_type(c.a) == A
assert c.num == 42
# these fail:
assert OmegaConf.get_type(c.aa[0]) == A
assert c.a.num == 13
assert c.aa[0].num == 13
```

**Expected behavior**
All asserts pass.

**Additional context**
- [ ] OmegaConf version: 2.3.0
- [ ] Python version: 3.11.1
- [ ] Operating system: Arch Linux

Jasha10:
Thanks, @raphCode. I agree that this looks like a bug.
A few notes:
- same symptoms with `dataclasses`; this is not an `attrs`-specific bug.
- there's a similar failure if we use `dict` instead of `list` in the type hint:
```python
from omegaconf import OmegaConf
from attrs import frozen

@frozen
class A:
    num: int

@frozen
class C:
    a: A
    aa: dict[int, A]

c = OmegaConf.structured(C)
c = OmegaConf.merge(dict(a=dict(num=1), aa={0: dict(num=2)}), c)
assert OmegaConf.get_type(c.a) == A
assert OmegaConf.get_type(c.aa[0]) == A  # fails
```

As far as fixing this:
There is a function called [`_deep_update_type_hint`](https://github.com/omry/omegaconf/blob/71f833b6046f0b514530ed791c57080287fa0bb0/omegaconf/basecontainer.py#LL821C5-L821C27) that is responsible for ensuring compatibility between the merge output and the given type hint (in this case, `list[A]` or `dict[int, A]`). I think the bug comes from bad behavior of that function.

PR #1323, opened 2026-06-24
fix: preserve structured element type when merging MISSING list into plain list

gaoflow:
## Problem

`OmegaConf.merge(plain_dict, structured_config)` silently drops the element type information for list fields when the structured config field is `MISSING` (has no default value).

**Reproduction (from #1058):**

```python
from typing import List
from omegaconf import OmegaConf, MISSING

@dataclass
class Item:
    num: int = MISSING

@dataclass
class Config:
    aa: List[Item] = MISSING

structured = OmegaConf.structured(Config)
result = OmegaConf.merge({"aa": [{"num": 1}]}, structured)

assert OmegaConf.get_type(result.aa[0]) == Item  # fails: returns dict
```

## Root cause

In `BaseContainer._list_merge`, when `src._is_missing()` is `True` (the structured list field has no default), the code already copies `element_type` from `src` to `dest` when `dest` has `Any` element type. However, it does **not** re-process the existing dest items against a typed prototype, leaving them as plain `dict`-backed `DictConfig` nodes without the correct `object_type`.

The non-missing path (lines 534–547) correctly builds a prototype and merges each item against it. The MISSING path was missing that step.

## Fix

After copying the element type, re-merge the existing destination items against a typed prototype when the source element type is a structured config. This brings the MISSING path in line with the non-missing path.

**After the fix:**
```python
assert OmegaConf.get_type(result.aa[0]) == Item  # passes
assert result.aa[0].num == 1                       # passes
```

## Tests

Added `test_merge_plain_dict_into_structured_config_preserves_list_element_type` to `tests/test_merge.py`, parametrized over both `OmegaConf.merge` and `OmegaConf.unsafe_merge`. All 7816 existing tests continue to pass.

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/news/1058.bugfix
+++ b/news/1058.bugfix
@@ -1,0 +1,1 @@
+Preserve structured list element types when merging a plain list into a missing structured config list field.
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -549,6 +549,18 @@ class BaseContainer(Container, ABC):
             # do not change dest if src is MISSING.
             if dest._metadata.element_type is Any:
                 dest._metadata.element_type = src._metadata.element_type
+                # Re-merge existing dest items against a prototype when the src
+                # element type is a structured config, so that items acquire the
+                # proper type information from the structured config definition.
+                src_is_optional, src_et = _resolve_optional(src._metadata.element_type)
+                if is_structured_config(src_et):
+                    prototype = DictConfig(
+                        src_et, ref_type=src_et, is_optional=src_is_optional
+                    )
+                    content = dest.__dict__["_content"]
+                    for idx, item in enumerate(content):
+                        if isinstance(item, DictConfig):
+                            content[idx] = OmegaConf.merge(prototype, item)
         elif src._is_interpolation():
             dest._set_value(src._value())
         else:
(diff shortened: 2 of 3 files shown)


binggao1230:
Addressed the Copilot review points in `d1113ac`:

- guarded the missing-source re-merge so it only enumerates existing destination content when `_content` is actually a list
- added coverage for `ListConfig(content=None)` and `ListConfig(content=MISSING)` with a missing structured list source
- changed the regression setup to use a raw plain dict and removed the line-number reference from the test docstring

Verification:

```bash
.venv/bin/python -m pytest tests/test_merge.py -q
.venv/bin/python -m flake8 omegaconf/basecontainer.py tests/test_merge.py
.venv/bin/pyrefly check omegaconf/basecontainer.py tests/test_merge.py
.venv/bin/isort --check-only omegaconf/basecontainer.py tests/test_merge.py
python -m black --check omegaconf/basecontainer.py tests/test_merge.py
git diff --check
```

omry:
Thanks for your help. Will review later.

omry:
Thanks for the update. I reproduced this locally against the PR branch: the original list case from #1058 is fixed, and the dict variant mentioned in the issue comments is also good. On current `main` the dict variant already passes, while the list case fails; with this PR both pass.

Please add a release fragment for this user-visible bug fix, e.g. `news/1058.bugfix`, with a short user-facing note about preserving structured list element types during merge.

omry (changes requested):
Bumping the remaining release-readiness request: please add a concise `news/1058.bugfix` fragment describing the user-visible fix for [#1058](https://github.com/omry/omegaconf/issues/1058), specifically that merges now preserve structured list element types.

The implementation changes and CI otherwise look ready for another review once the fragment is added.

omry (approved):
Thanks! I took this over, addressed the remaining review feedback, and verified the updated branch locally and in CI.

PR #1324, opened 2026-06-27
fix: preserve compatible values during DictConfig promotion

### LINKED ISSUE #898: _promote is erasing values from the config
**Describe the bug**
If you _promote a dict config to a dataclass with some default values, the default values from the dataclass will override the ones from the config.

**To Reproduce**

Modify `test_promote_to_dataclass` to add a default value to `bar` field.
And an assertion that `x.bar` isn't modified during promotion.

```py
    def test_promote_to_dataclass(self, module: Any) -> None:
        @dataclasses.dataclass
        class Foo:
            foo: pathlib.Path
            bar: str = "goodbye.err"
            qub: int = 5

        x = DictConfig({"foo": "hello.txt", "bar": "hello.txt"})
        assert isinstance(x.foo, str)
        assert isinstance(x.bar, str)

        x._promote(Foo)
        assert isinstance(x.foo, pathlib.Path)
        assert x.bar == "hello.txt"
        assert x.qub == 5
```

This tests currently fails with:

```
>       assert x.bar == "hello.txt"
E       AssertionError: assert 'goodbye.err' == 'hello.txt'
E         - hello.txt
E         + goodbye.err
```

This happens because in `_promote` the code is merging the `proto` with default values into the original config,
while it should be the other way around: merge the `config` into the `proto`.
This is a simple fix, but also a breaking change in a way.

**Additional context**
- OmegaConf version: 2.1.1.dev (pathlib branch)
- Python version: 3.9.5
- Operating system: linux

gaoflow:
Fixes #898.

`DictConfig._promote()` currently merges the structured prototype into the existing config. That correctly adds missing default fields and converts compatible values, but it also lets dataclass defaults overwrite existing compatible values.

This keeps the existing promotion behavior for incompatible values by starting with the current merge direction, then restores only previous top-level values that still pass validation against the promoted structured config. That preserves values like an existing `str` field while still falling back to the target default for values that cannot be converted to the promoted type.

Verification:

```text
uv run --with pytest --with attrs --with PyYAML --with antlr4-python3-runtime --with-editable . pytest tests/structured_conf/test_structured_config.py -q -k promote
uv run --with pytest --with attrs --with PyYAML --with antlr4-python3-runtime --with-editable . pytest tests/structured_conf/test_structured_config.py -q
uv run --with black black --check omegaconf/dictconfig.py tests/structured_conf/test_structured_config.py
uv run --with isort isort --check-only omegaconf/dictconfig.py tests/structured_conf/test_structured_config.py
git diff --check
```

### DIFF
--- a/news/898.bugfix
+++ b/news/898.bugfix
@@ -1,0 +1,1 @@
+Preserve existing compatible values when promoting a ``DictConfig`` to a structured config.
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -642,13 +642,23 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
 
         from omegaconf import OmegaConf
 
+        previous_nodes = {
+            key: copy.deepcopy(self._get_node(key, validate_access=False))
+            for key in self.keys()
+        }
         proto: DictConfig = OmegaConf.structured(type_or_prototype)
         object_type = proto._metadata.object_type
-        # remove the type to prevent assignment validation from rejecting the promotion.
+        # Remove the type to prevent assignment validation from rejecting the promotion.
         proto._metadata.object_type = None
         self.merge_with(proto)
-        # restore the type.
         self._metadata.object_type = object_type
+
+        for key, node in previous_nodes.items():
+            if key in self:
+                try:
+                    self[key] = node
+                except ValidationError:
+                    pass
 
     def _set_value(self, value: Any, flags: Optional[Dict[str, bool]] = None) -> None:
         previous_content = self.__dict__["_content"]
--- a/tests/structured_conf/test_structured_config.py
+++ b/tests/structured_conf/test_structured_config.py
@@ -982,7 +982,7 @@ class TestConfigs:
         @dataclasses.dataclass
         class Foo:
             foo: pathlib.Path
-            bar: str
+            bar: str = "goodbye.err"
             qub: int = 5
 
         x = DictConfig({"foo": "hello.txt", "bar": "hello.txt"})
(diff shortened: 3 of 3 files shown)


binggao1230:
CI triage note: the current red `process` check appears unrelated to this PR's code. The failure happens in the `Commit to Backlog Atlas branch` step when `github-actions[bot]` tries to push to `omry/omegaconf` and receives a 403:

```text
remote: Permission to omry/omegaconf.git denied to github-actions[bot].
fatal: unable to access 'https://github.com/omry/omegaconf/': The requested URL returned error: 403
```

Local verification for the changed area passed:

```text
pytest tests/structured_conf/test_structured_config.py -q -k promote  # 10 passed, 5 skipped
pytest tests/structured_conf/test_structured_config.py -q             # 655 passed, 323 skipped
black --check omegaconf/dictconfig.py tests/structured_conf/test_structured_config.py
isort --check-only omegaconf/dictconfig.py tests/structured_conf/test_structured_config.py
git diff --check
```

omry:
Thanks for working on this. The underlying issue (#898) is still untriaged, and the expected behavior/compatibility tradeoff is not yet settled. A change here may be breaking, so this PR is premature until there is clear maintainer buy-in on both the desired behavior and the implementation direction.

Please avoid sending PRs for untriaged behavior changes like this unless a maintainer has explicitly confirmed the approach first.

omry:
Thanks for working on this and for demonstrating the current behavior clearly.

We have now completed the triage of [issue #898](https://github.com/omry/omegaconf/issues/898), including reviewing the historical use of `_promote()`.

The method was used by Hydra’s original structured-config composition path, but Hydra removed that dependency when its composition logic was rewritten to use ordinary `OmegaConf.merge()`. There are no remaining call sites inside OmegaConf or current Hydra.

Rather than establish a new behavioral contract for an unused private API, we have decided to remove `_promote()` in OmegaConf 2.4. I am open to introducing `_promote()` or `_cast()` in the future if a compelling use case emerges.

Consequently, this PR is no longer the direction we want to take, and I’m going to close it. This is a decision about the future of the API rather than a problem with the contribution. Thanks again for investigating it.

PR #1331, opened 2026-07-17
Defer interpolation resolution when merging into a structured config (#1020)

### LINKED ISSUE #1020: Failure to merge interpolation into structured config
**Describe the bug**
As described in https://github.com/facebookresearch/hydra/issues/2416, merging a config with an interpolation key into a structured config can fail with an InterpolationKeyError or a ValidationError.

**To Reproduce**
There are two failure modes identified in https://github.com/facebookresearch/hydra/issues/2416.

##### Mode 1: merging an interpolation into an optional None value
```python
# omegaconf_failing.py
from typing import Optional

@dataclass
class TableConfig:
    rows: int = 1

@dataclass
class DatabaseConfig:
    table_cfg: TableConfig = TableConfig()

@dataclass
class ModelConfig:
    data_source: Optional[TableConfig] = None

@dataclass
class ServerConfig:
    db: DatabaseConfig
    model: ModelConfig

cfg = OmegaConf.merge(
    ServerConfig,
    {"db": DatabaseConfig},
    {"model": {"data_source": "${db.table_cfg}"}},
)  # raises "omegaconf.errors.InterpolationKeyError: Interpolation key 'db.table_config' not found"
```
```text
$ python omegaconf_failing.py
...
omegaconf.errors.InterpolationKeyError: Interpolation key 'db.table_cfg' not found
    full_key:
    object_type=ServerConfig
```

##### Mode 2: merging an interpolation into a structured value
```python
# omegaconf_failing_also.py

@dataclass
class TableConfig:
    rows: int = 1

@dataclass
class DatabaseConfig:
    table_cfg: TableConfig = TableConfig()

@dataclass
class ModelConfig:
    data_source: TableConfig = TableConfig()

@dataclass
class ServerConfig:
    db: DatabaseConfig
    model: ModelConfig

cfg = OmegaConf.merge(
    ServerConfig,
    {"db": DatabaseConfig},
    {"model": {"data_source": "${db.table_cfg}"}},
)  # raises "omegaconf.errors.ValidationError: Merge error: str is not a subclass of TableConfig. value: ${db.table_config}"
```
```text
$ python omegaconf_failing_also.py
...
omegaconf.errors.ValidationError: Merge error: str is not a subclass of TableConfig. value: ${db.table_cfg}
    full_key:
    object_type=ServerConfig
```

**Expected behavior**
I don't think either merge above should result in an error.

**Additional context**
OmegaConf version: HEAD (commit 8d110fc)

uttam12331:
Fixes #1020

### Problem

Merging a config containing an interpolation key into a structured config fails in the two ways described in the issue:

```python
cfg = OmegaConf.merge(
    ServerConfig,
    {"db": DatabaseConfig},
    {"model": {"data_source": "${db.table_cfg}"}},
)
```

- `Optional[TableConfig] = None` target → `InterpolationKeyError: Interpolation key 'db.table_cfg' not found`
- `TableConfig` target → `ValidationError: Merge error: str is not a subclass of TableConfig. value: ${db.table_cfg}`

### Cause

In `BaseContainer._map_merge`, for each source key the merge eagerly inspects the source node:

- `_validate_merge(value=src_node)` type-checks the raw `"${db.table_cfg}"` string (whose `get_type` is `str`) against the structured target type → subclass `ValidationError`.
- `_is_none(src_node, resolve=True)` **resolves** the interpolation to decide whether to `expand()` the destination — but it resolves in the *source's* context, which doesn't contain `db` yet → `InterpolationKeyError`.

An interpolation's type isn't known until resolution, and it must be resolved lazily against the *merged* result.

### Fix

Skip both eager steps when the source node is an interpolation (`get_value_kind(src_node) is ValueKind.INTERPOLATION`); it then falls through to normal assignment and resolves correctly afterwards.

### Verification

- Both failure modes now produce `{'model': {'data_source': {'rows': 1}}}` and resolve to the referenced value.
- Added `test_merge_interpolation_into_structured_config` (both target shapes × `merge`/`unsafe_merge`), which fails on the base branch and passes with this change.
- Full test suite: 8205 passed, zero regressions. `black`/`flake8` clean.

### DIFF
--- a/news/1020.bugfix
+++ b/news/1020.bugfix
@@ -1,0 +1,1 @@
+Fix merging an interpolation into a structured config field failing with InterpolationKeyError or ValidationError; the interpolation is now kept unresolved and resolved lazily against the merged result.
--- a/omegaconf/basecontainer.py
+++ b/omegaconf/basecontainer.py
@@ -434,14 +434,20 @@ class BaseContainer(Container, ABC):
 
             src_vk = get_value_kind(src_node)
             src_node_missing = src_vk is ValueKind.MANDATORY_MISSING
+            # An interpolation source must be merged as-is and resolved lazily:
+            # its type is not known until resolution, and resolving it here (in
+            # the source's context, which may lack the referenced keys) is wrong
+            # (issue #1020).
+            src_node_interpolation = src_vk is ValueKind.INTERPOLATION
 
-            if isinstance(dest_node, DictConfig):
+            if isinstance(dest_node, DictConfig) and not src_node_interpolation:
                 dest_node._validate_merge(value=src_node)
 
             if (
                 isinstance(dest_node, Container)
                 and dest_node._is_none()
                 and not src_node_missing
+                and not src_node_interpolation
                 and not _is_none(src_node, resolve=True)
             ):
                 expand(dest_node)
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -277,39 +277,6 @@ class InterpolationDict:
     dict: Dict[str, int] = II("optimization.lr")
 
 
-# Structures for merging an interpolation into a structured config (issue #1020).
-@dataclass
-class TableConfig:
-    rows: int = 1
-
-
-@dataclass
-class DatabaseConfig:
-    table_cfg: TableConfig = field(default_factory=TableConfig)
-
-
-@dataclass
-class ModelConfigOptional:
-    data_source: Optional[TableConfig] = None
-
-
-@dataclass
-class ModelConfigStructured:
-    data_source: TableConfig = field(default_factory=TableConfig)
-
-
-@dataclass
-class ServerConfigOptional:
-    db: DatabaseConfig = field(default_factory=DatabaseConfig)
-    model: ModelConfigOptional = field(default_factory=ModelConfigOptional)
-
-
-@dataclass
-class ServerConfigStructured:
-    db: DatabaseConfig = field(default_factory=DatabaseConfig)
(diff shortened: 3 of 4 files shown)


omry (changes requested):
Thanks for working on [#1020](https://github.com/omry/omegaconf/issues/1020). The implementation looks like the right direction, and the test matrix adequately covers both reported failure modes across `merge` and `unsafe_merge`.

A few changes before this is ready:

- Please fix the pyrefly failure at `tests/test_merge.py:1610`. `OmegaConf.to_container()` has a broad return type, so indexing it directly fails the Python 3.10 lint job.
- Instead of introducing six single-use dataclasses, please reuse the existing `A`, `B`, and `C` test fixtures. `C()` covers the optional-`None` target, while `B(x=A())` covers the populated structured target.
- Replace the `to_container()` assertion with a structured-type assertion such as `OmegaConf.get_type(res.model.x) is A`. This both avoids the lint problem and verifies that type information is preserved.
- Please add a concise `news/1020.bugfix` release fragment.

The failing Backlog Atlas check is unrelated to this patch; it is failing because the workflow cannot push with the current `github-actions[bot]` permissions.

uttam12331:
Reworked the test to reuse the existing A/B/C fixtures (C() for the optional-None target, B(x=A()) for the populated structured one) and replaced the to_container indexing with an OmegaConf.get_type(res.model.x) is A assertion, which clears the pyrefly failure. Also removed the six single-use dataclasses and added news/1020.bugfix.

omry (approved):
Thanks for the update. This addresses the review feedback, and the diff looks good to me.

PR #1347, opened 2026-07-22
Fix ListConfig insert index handling

### LINKED ISSUE #750: Followup: refactor ListConfig.{insert,append}
After some discussion (...), I became suspicious about why `append`'s code is so different from `insert`'s -- it seems that the two have diverged.

_Originally posted by @Jasha10 in https://github.com/omry/omegaconf/pull/749#discussion_r651428069_

omry:
Normalize ListConfig insertion indices with Python's integer-index protocol and list insertion rules.

Use _set_item_impl() for wrapping and validation, and roll back atomically on failure. Add regression coverage for negative, out-of-range, and __index__-compatible indices and failed validation.

Closes #750

### DIFF
--- a/news/750.bugfix
+++ b/news/750.bugfix
@@ -1,0 +1,1 @@
+`ListConfig.insert()` now follows Python list semantics for negative and out-of-range indices and leaves the list unchanged when validation fails.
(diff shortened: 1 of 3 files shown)


omry:
Closing this accidental duplicate; the update is tracked in #1346.

PR #1352, opened 2026-07-23
Follow interpolated containers during nested updates

### LINKED ISSUE #1329: OmegaConf.update() discards container-valued interpolation when updating a nested key
**Describe the bug**

`OmegaConf.update()` does not dereference intermediate interpolation nodes. When an intermediate interpolation resolves to a container, `update()` treats the raw interpolation node as a non-container, replaces it with an empty mapping, and applies the nested update to that replacement instead of to the resolved target.

**To Reproduce**

```python

cfg = OmegaConf.create(
    {
        "block1": {"arg1": 1, "arg2": 2},
        "block2": "${block1}",
    }
)

OmegaConf.update(cfg, "block2.arg3", 3)

print(OmegaConf.to_container(cfg, resolve=False))
print(OmegaConf.is_interpolation(cfg, "block2"))
```

Actual output:

```text
{'block1': {'arg1': 1, 'arg2': 2}, 'block2': {'arg3': 3}}
False
```

**Expected behavior**

When an intermediate interpolation resolves to a container, `OmegaConf.update()` should follow the interpolation and continue the update against the resolved target. The interpolation itself should remain unchanged:

```text
{'block1': {'arg1': 1, 'arg2': 2, 'arg3': 3}, 'block2': '${block1}'}
True
```

Consequently, resolving the complete config should produce:

```text
{'block1': {'arg1': 1, 'arg2': 2, 'arg3': 3}, 'block2': {'arg1': 1, 'arg2': 2, 'arg3': 3}}
```

Adding the nested field must still obey the resolved target's normal mutation rules. For example, adding `arg3` should fail when the target is in struct mode or is a Structured Config that does not permit the field, unless the caller explicitly enables the corresponding override.

Invalid interpolations, including interpolation cycles and ancestor dereferences, should continue to raise their normal resolution errors.

**Compatibility consideration**

Correcting this changes established `OmegaConf.update()` behavior and may therefore be breaking for callers that rely on an intermediate interpolation being replaced. However, the update tests cover replacement of literal primitive intermediate nodes but do not cover interpolation nodes, and the documentation does not define update-through-interpolation semantics.

**Additional context**

- [x] OmegaConf version: 2.4.0.dev12
- [x] Python version: 3.13.12
- [x] Operating system: Linux
- [x] Minimal repro included above
- Found while investigating facebookresearch/hydra#2305.

omry:
Follow intermediate node-interpolation chains that end at existing config containers while applying nested updates. Mutate the resolved target while preserving the interpolation, and avoid invoking custom resolvers during traversal.

Cover dict, list, tuple, typed, chained, relative, dynamic, invalid, and permission-controlled interpolation cases. Document the behavior and add a release fragment.

Fixes #1329

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -921,6 +921,20 @@ The ``force_add`` flag ensures that the path is created even if it will result i
     >>> OmegaConf.update(cfg, "a.b.c.d", 10, force_add=True)
     >>> assert cfg.a.b.c.d == 10
 
+When an intermediate path element is a node interpolation whose reference
+chain ends at an existing config container, ``OmegaConf.update()`` follows the
+interpolation and updates the referenced container. The interpolation itself
+remains unchanged.
+
+.. doctest::
+
+    >>> cfg = OmegaConf.create({"base": {"x": 1}, "alias": "${base}"})
+    >>> OmegaConf.update(cfg, "alias.y", 2)
+    >>> OmegaConf.to_container(cfg, resolve=False)
+    {'base': {'x': 1, 'y': 2}, 'alias': '${base}'}
+    >>> OmegaConf.is_interpolation(cfg, "alias")
+    True
+
 .. _keypath-escaping:
 
 Key path escaping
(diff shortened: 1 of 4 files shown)


omry:
@codex review again

PR #1356, opened 2026-07-25
fix: keep key context on OmegaConf.resolve() interpolation errors

### LINKED ISSUE #1330: `OmegaConf.resolve()` drops interpolation error context
**Describe the bug**

`OmegaConf.resolve()` drops the diagnostic context from interpolation errors. Direct node access reports both `full_key` and `object_type`, while resolving the same config through `OmegaConf.resolve()` raises the same exception type without either field.

This is visible through Hydra `instantiate()`, which resolves its copied config with `OmegaConf.resolve()`, but the behavior reproduces in OmegaConf alone.

**To Reproduce**

```python
from typing import Any

@dataclass
class MyClass:
    bad_interp: list[Any]


cfg = OmegaConf.structured(MyClass(bad_interp="${foo}"))

try:
    cfg.bad_interp
except Exception as e:
    print(str(e))
    print(e.full_key)
    print(e.object_type)

try:
    OmegaConf.resolve(cfg)
except Exception as e:
    print(str(e))
    print(e.full_key)
    print(e.object_type)
```

Direct access reports `full_key: bad_interp` and `object_type=MyClass`. The exception from `OmegaConf.resolve()` has `full_key=None` and `object_type=None`.

The same result occurs when resolving a deep copy of the config.

**Expected behavior**

Errors raised by `OmegaConf.resolve()` should retain the same `full_key` and `object_type` context as direct node access.

**Additional context**

- [x] OmegaConf version: 2.4.0.dev12
- [x] Python version: 3.14
- [x] Operating system: Linux
- [x] Minimal repro provided
- Hydra report: https://github.com/facebookresearch/hydra/issues/2235

Sanjays2402:
## Motivation

`OmegaConf.resolve()` dropped the diagnostic context from interpolation errors: the same failure reported `full_key` and `object_type` on direct node access, but arrived with both set to `None` when raised from `resolve()`. This is visible through Hydra's `instantiate()`, which resolves a copy of its config.

`_resolve_container_value` called `node._dereference_node()` outside the `_format_and_raise` wrapper that direct access (`DictConfig.__getattr__`) uses, so the exception escaped before the key/type context was attached. Dereference errors are now routed through `cfg._format_and_raise`.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/omry/omegaconf/blob/main/CONTRIBUTING.md)?

Yes

### For any non-trivial change: is there an issue or design discussion where maintainers agreed on the direction?

#1330 describes the bug and the expected behavior. The change is a one-branch fix that makes `resolve()` match the existing direct-access behavior; no new API or semantics.

## Test Plan

`tests/test_omegaconf.py::test_resolve_error_keeps_key_and_object_type` asserts that the error from `OmegaConf.resolve()` carries the same `full_key` and `object_type` as the error from direct access. It fails on `main` (`None != 'num'`) and passes with the fix. No special setup required.

Full suite: 8514 passed, 362 skipped (baseline 8513 passed, same 5 pre-existing errors from a missing `pytest-mock` in my environment).

## Fixes

Fixes #1330

## Related PRs

N/A

### DIFF
--- a/news/1330.bugfix
+++ b/news/1330.bugfix
@@ -1,0 +1,1 @@
+Keep the failing key and object type on interpolation errors raised through `OmegaConf.resolve()`, so they match the errors raised by direct node access.
--- a/omegaconf/_impl.py
+++ b/omegaconf/_impl.py
@@ -22,7 +22,13 @@ def _resolve_container_value(cfg: Container, key: Any) -> None:
     node = cfg._get_child(key)
     assert isinstance(node, Node)
     if node._is_interpolation():
-        resolved = node._dereference_node()
+        try:
+            resolved = node._dereference_node()
+        except Exception as e:
+            # Attach the same key/object_type context that direct node access
+            # reports, so errors surfaced through OmegaConf.resolve() stay
+            # diagnosable.
+            cfg._format_and_raise(key=key, value=None, cause=e)
         if isinstance(resolved, Container):
             _resolve(resolved)
         if isinstance(resolved, InterpolationResultNode):
--- a/tests/test_omegaconf.py
+++ b/tests/test_omegaconf.py
@@ -572,6 +572,20 @@ def test_resolve_raises_on_resolver_arg_to_missing(restore_resolvers: Any) -> No
         OmegaConf.resolve(cfg)
 
 
+def test_resolve_error_keeps_key_and_object_type() -> None:
+    """https://github.com/omry/omegaconf/issues/1330"""
+    cfg = OmegaConf.structured(StructuredWithMissing)
+    cfg.num = "${no_such_key}"
+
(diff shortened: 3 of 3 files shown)


Sanjays2402:
Both addressed in 0e0ab68.

The `object_type` assertion was comparing the two exceptions to each other, which would have passed if both had been `None`. It now asserts `StructuredWithMissing` explicitly on each path.

Also added `news/1330.bugfix` — I missed that CONTRIBUTING requires a fragment for user-visible fixes.

`pytest tests/test_omegaconf.py` is 194 passed. Reverting `_impl.py` to the parent commit fails the new test with `full_key` `None` vs `num`, so it is still catching the regression.

omry:
Thanks @Sanjays2402.
Looks good, merging.

PR #1359, opened 2026-08-07
Make OmegaConf containers unhashable

### LINKED ISSUE #1333: Consider making OmegaConf configs unhashable
## Summary

Consider making OmegaConf container objects unhashable so they cannot be used as dictionary keys or set members.

`DictConfig` and `ListConfig` are mutable, value-comparable containers, but both currently implement:

```python
def __hash__(self) -> int:
    return hash(str(self))
```

This makes it easy to use a mutable config as a dictionary key even though its hash changes when its contents change.

The question also affects the new `TupleConfig` being designed in #392. Although its outer structure would be immutable, it would remain a live OmegaConf config whose effective value can change through interpolations, resolver results, parent context, or mutable nested containers.

### Mutation changes the hash of an existing key

```python
cfg = OmegaConf.create([1])
mapping = {cfg: "value"}

cfg[0] = 2
mapping[cfg]  # may no longer find the existing entry
```

The test suite currently asserts that a `ListConfig` or `DictConfig` hash changes after mutation. That behavior is precisely what makes the object unsafe as a key in a hash table.

### Equality and hashing use different interpolation semantics

Container equality resolves interpolations, while container hashing uses the unresolved string representation:

```python
a = OmegaConf.create({"x": 1, "y": "${x}"})
b = OmegaConf.create({"x": 1, "y": 1})

assert a == b
assert hash(a) != hash(b)  # current behavior
```

This violates the Python invariant that equal hashable objects must have equal hashes.

Hashing resolved values instead is not a good general solution: it could execute custom resolvers, be expensive, fail on missing values or cycles, and still produce a hash that changes when an interpolation source or uncached resolver result changes.

## Proposed direction

Make all OmegaConf `Container` instances unhashable:

- `DictConfig`
- `ListConfig`
- `TupleConfig`

In practice this means setting `__hash__ = None`, causing `hash(cfg)`, dictionary-key insertion, and set insertion to raise `TypeError`.

This keeps the existing resolved-value equality behavior while preventing configs from entering hash tables with unstable hashes. Users who need a hashable key should first create an explicit immutable snapshot of the resolved data.

The `readonly` flag should not make a config hashable. It is reversible, and it does not guarantee that interpolation sources or resolver outputs remain stable.

## Python precedents

This follows Python's general treatment of mutable, value-comparable objects:

- The [Python data model](https://docs.python.org/3/reference/datamodel.html#object.__hash__) says that a class with mutable values and value equality should not implement `__hash__`; overriding `__eq__` without `__hash__` implicitly makes instances unhashable.
- [PEP 557](https://peps.python.org/pep-0557/#specification) makes non-frozen, equality-enabled dataclasses unhashable by default. Hashing such a class requires the explicitly named `unsafe_hash=True` escape hatch.
- [PEP 3106](https://peps.python.org/pep-3106/) made live dictionary views unhashable even though the views do not directly expose mutation methods, because their values can change when the underlying dictionary changes. This is similar to a structurally immutable `TupleConfig` whose effective values can change through its config context.
- Python's [built-in type documentation](https://docs.python.org/3/library/stdtypes.html#mapping-types-dict) excludes lists, dictionaries, and other mutable value-comparable objects from dictionary keys.

## Compatibility and migration options

This is a breaking change because existing code may hash configs or use them as dictionary keys.

Options:

1. Make existing containers unhashable immediately in OmegaConf 2.4.
2. In 2.4, emit a deprecation warning from the existing `__hash__` implementations and keep returning the current hash; set `__hash__ = None` in OmegaConf 3.0.

`TupleConfig` is new in 2.4 and can be unhashable from its introduction without a compatibility period.

## Questions

1. Should this apply only to public `Container` types, or also to directly exposed mutable `Node` classes?
2. Is a 2.4 warning followed by removal in 3.0 preferable to making the correctness fix immediately?
3. Do we know of legitimate current uses of `DictConfig` or `ListConfig` as dictionary keys or set members?
4. Should the migration documentation recommend a particular immutable snapshot representation, without adding a new API?

omry:
## Summary

- make `DictConfig` and `ListConfig` unhashable, matching `TupleConfig`
- replace the old mutable value-based hashes with Python's standard `__hash__ = None` behavior
- document the behavior, including read-only configs
- update regression coverage and add the 2.4 API-change fragment

## Why

`DictConfig` and `ListConfig` used `hash(str(self))` despite being mutable and value-comparable. Mutation could change the hash of an existing dictionary key, and interpolation-aware equality could consider two configs equal while their unresolved string representations produced different hashes. Both behaviors violate Python's hashing contract.

OmegaConf 2.4 already contains significant compatibility changes, so this correctness fix is applied directly rather than going through a deprecation period.

## Impact

Calling `hash()`, using a config as a dictionary key, or inserting one into a set now raises `TypeError`. This also applies to read-only configs because the flag is reversible and effective values can change through interpolations or resolvers. Scalar node hashing is unchanged.

## Validation

- full test suite: 8,561 passed, 362 skipped
- documentation build and 483 doctests
- Pyrefly
- Ruff lint and formatting

Closes #1333

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -309,6 +309,25 @@ Manipulation
     >>> conf.database = {'hostname': 'database01', 'port': 3306}
 
 
+Hashing
+^^^^^^^
+
+OmegaConf containers are unhashable because their values can change. They
+cannot be used as dictionary keys or set elements. If you currently use
+containers this way, migrate to an application-defined stable key. OmegaConf
+does not provide a hashable representation that preserves all config semantics.
+Read-only configs remain unhashable because the read-only flag is reversible
+and interpolations or resolvers can change their effective values.
+
+.. doctest::
+
+    >>> conf = OmegaConf.create({"key": "value"})
+    >>> hash(conf)
+    Traceback (most recent call last):
+    ...
+    TypeError: unhashable type: 'DictConfig'
+
+
 Serialization
 -------------
 OmegaConf objects can be saved and loaded with OmegaConf.save() and OmegaConf.load().
--- a/news/1333.api_change
+++ b/news/1333.api_change
@@ -1,0 +1,1 @@
+Make ``DictConfig`` and ``ListConfig`` unhashable, preventing their use as dictionary keys or set elements.
--- a/omegaconf/dictconfig.py
+++ b/omegaconf/dictconfig.py
@@ -70,6 +70,7 @@ def _make_key_suggestion(key: Any, available: Iterable) -> str:
 class DictConfig(BaseContainer, MutableMapping[Any, Any]):
     _metadata: ContainerMetadata
     _content: Union[Dict[DictKeyType, Node], None, str]
+    __hash__ = None  # type: ignore[assignment]
 
     def __init__(
         self,
@@ -660,9 +661,6 @@ class DictConfig(BaseContainer, MutableMapping[Any, Any]):
         if x is not NotImplemented:
             return not x
         return NotImplemented
-
-    def __hash__(self) -> int:
-        return hash(str(self))
 
     def _promote(self, type_or_prototype: Optional[Type[Any]]) -> None:
         """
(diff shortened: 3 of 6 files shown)


omry:
@codex review

omry:
@codex

Issue #1274, opened 2026-04-24
Python 3.10 minimum: remove version guards and modernize type annotations


omry:
Now that the minimum supported Python version is 3.10, we can remove dead version guards and adopt modern Python conventions throughout the codebase.

### `omegaconf/` source

- **`omegaconf/version.py:13`** — `if sys.version_info < (3, 9): raise ImportError` is always false; remove the block and its error message.
- **`omegaconf/_utils.py:334`** — `if sys.version_info >= (3, 10)` in `is_union_annotation`: always true. Inline the `isinstance(type_, types.UnionType)` check unconditionally and delete the `getattr` fallback path.

### `tests/` data and fixtures

- **`tests/structured_conf/data/dataclasses.py`** lines 12, 29, 384 — `if sys.version_info >= (3, 8)` guards around `TypedDict`/`TypedDictSubclass`: always true, unwrap.
- **`tests/structured_conf/data/dataclasses.py:831`** — `if sys.version_info >= (3, 10)` for `int | str` union syntax: always true, unwrap.
- **`tests/structured_conf/data/dataclasses.py:843`** — `if sys.version_info >= (3, 9)` for `SupportPEP585`: always true, unwrap.
- Same patterns in **`tests/structured_conf/data/dataclasses_pre_311.py`** (lines 12, 29, 382, 829, 841).
- Same patterns in **`tests/structured_conf/data/attr_classes.py`** (lines 11, 28, 379, 788, 800).

### `tests/` skip decorators and conditionals

- **`tests/structured_conf/test_structured_config.py:553`** — `@mark.skipif(sys.version_info < (3, 8), ...)`: always runs, remove the decorator.
- **`tests/test_utils.py`** lines 845, 1272, 1296, 1320, 1446 — multiple `@mark.skipif(sys.version_info < (3, 10))` and `if sys.version_info >= (3, 9/10)` guards: always true, remove skipifs and unwrap the true branches.

## Type annotation modernization

With 3.10 as the floor, we can adopt modern annotation syntax everywhere:

- Replace `Optional[X]` → `X | None` (~231 occurrences in source)
- Replace `Union[X, Y]` → `X | Y`
- Replace `List[X]`, `Dict[K, V]`, `Tuple[X, Y]`, `Set[X]` → `list[x]`, `dict[k, v]`, `tuple[x, y]`, `set[x]` (~152 occurrences; available since 3.9)
[...]

PR #1364, opened 2026-08-26
Modernize type annotations

jorenham:
## Motivation

Now that OmegaConf supports Python 3.10+, modern type annotation syntax can be used to improve readability and make the project future-proof. 

### Have you read the [Contributing Guidelines on pull requests](https://github.com/omry/omegaconf/blob/main/CONTRIBUTING.md)?

Aye!

### For any non-trivial change: is there an issue or design discussion where maintainers agreed on the direction?

It's pretty trivial, but #1274 I guess?

## Test Plan

No behavior changes, so the tests still pass. Pyrefly also seems happy.

Oh and FWIW, after making these changes myself (viz. a human) I had Claude Fable 5 go over the changes to see if missed anything or messed in some other way, and iterated until it had nothing left to complain.

## Fixes

Towards #1274 (i.e. the "Type annotaiton modernization" part)

## Related PRs

None that I'm aware of :shrug:

PR #1365, opened 2026-09-10
100% type coverage

jorenham:
This fills in the missing annotations so that `pyrefly coverage check` now reports 100% type coverage for the public API. (excluding `vendor/` and `grammar/gen`). 

My plan is to follow this up with a PR that enforces this 100% type coverage using `pyrefly coverage check` in a lightweight CI step (unless you object, of course).

## Motivation

I noticed on https://jorenham.github.io/typestats/dashboard/report/#omegaconf that omegaconf was close to 100% coverage, so I figured I might as well increase that to 100%.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/omry/omegaconf/blob/main/CONTRIBUTING.md)?

Aye!

### For any non-trivial change: is there an issue or design discussion where maintainers agreed on the direction?

Pretty trivial I think, but maybe I'm biased :shrug:

## Test Plan

No behavior changes, so the tests still pass. Pyrefly also seems happy.

## Fixes

#1274 comes closest, but there's no dedicated type coverage issue as far as I can tell.

## Related PRs

#1364, kinda


omry:
Thanks! This looks good. Getting the public API to 100% type coverage is worthwhile, and I agree that enforcing it with a lightweight `pyrefly coverage check` in a separate follow-up PR is the right next step.

